PR #11, opened 2015-02-19
Adding grumpy.py

JasYoung314:
Adding grumpy. This strategy chooses a strategy based on two thresholds which depend on the history of the opponent.

### DIFF
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -6,6 +6,7 @@ from titfortat import *
 from gobymajority import *
 from alternator import *
 from averagecopier import *
+from grumpy import *
 
 
 strategies = [
@@ -17,4 +18,5 @@ strategies = [
         Random,
         Alternator,
         AverageCopier,
+        Grumpy,
         ]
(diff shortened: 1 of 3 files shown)


drvinceknight:
Hi @JasYoung314 thanks a lot for the pull request: the grumpy strategy looks cool (thanks also for the tests!).

I note that the strategy keeps another attribute (the state: grumpy or not). This is something that will need to be reset at every repetition of a duel in the tournament. I've actually just changed how reset works with this commit: 3d061048ae30f1f836efe94201fe3326e2024a11

Could you add a specific reset method to the `Grumpy` class so that it not only resets `history` and `score` but also the state (you'll have to add the method so that it overwrites the `reset` method it inherits).

drvinceknight:
(this commit also: 376d9cf587b6adbe95b4c8fa5d0c630682653e4c)

JasYoung314:
Thanks @drvinceknight I have made the change to the grumpy player class you requested.

drvinceknight:
Nice: could you just add a test for the reset method in the grumpy class? Once you've done that I'll merge!

JasYoung314:
@drvinceknight the reset method is now tested for grumpy

PR #14, opened 2015-02-20
Added a new strategy

JasYoung314:
Added a new (slightly cheating strategy).

The strategy will simulate the next 50 rounds and decide whether or not to cooperate or defect based on the outcome of this simulation.

Also added in tests for the strategy and added it to axelrod/strategies/__init__.py


drvinceknight:
This 'cheating strategy' will need to be in it's list of strategies in __init__ which will now need to be re-written to spit out 3 plots.

Also, the strategy itself needs to be broken up in to other methods: which all need to be tested individually (this needs to be done for clarity). Let me now when that's done and I'll look a bit closer.

axelrod/strategies/mind_reader.py
```diff
@@ -0,0 +1,88 @@
...
+
+        results.append(self.calculate_score(self.history, opponent.history))
+
+        self.history = copy.copy(dummy_history_self)
+        opponent.history = copy.copy(dummy_history_opponent)
+
+        return results.index(min(results))
+
+    def calculate_score(self, history_1, history_2):
+        """
```
drvinceknight: Can you use the calculate score from the tournament itself? You're duplicating code here...

JasYoung314:
@drvinceknight I've made those changes, renamed mind_reader to mindreader and tested the new methods of my class.

Let me know if there are additional changes to be made

drvinceknight:
Can you modify `__init__.py` so that it's something like:

```
...
strategies = [

]

# Write a short comment here so that this won't confuse anyone
cheating_strategies = [
MindReader
]
```

and then modify the run_tournament.py so that it produces 3 plots:

- just with strategies (keep this as being called `results.png`
- just with cheaters
- with both

JasYoung314:
@drvinceknight Ive modified the run tournament to produce three graphs, the results.png is still the strategies that follow the rules of Axelrods tournament.

axelrod/strategies/__init__.py
```diff
@@ -5,10 +5,12 @@
 from titfortat import *
 from gobymajority import *
 from alternator import *
+from grumpy import *
+from mindreader import *
 from averagecopier import *
 from grumpy import *
 
-
+#These are strategies that follow the rules of Axelrods tournement, which can be found in the readme
```
drvinceknight: Remove this line.

PR #18, opened 2015-02-20
Adding a strategy that learns through reinforcement learning

geraintpalmer:


drvinceknight:
Fix merge conflict.

drvinceknight:
I had a go at fixing this myself @geraintpalmer : there are all sorts of syntax errors in the tests.

Please make sure you run:

`$ python -m unittest discover axelrod/tests/`

and that all tests pass.

drvinceknight:
I'd rather you wrote this in a similar way to the ['gobymajority' strategies](https://github.com/drvinceknight/Axelrod/blob/master/axelrod/strategies/gobymajority.py). So write one global class and have classes inherit from it.

It'll then be easier for you to write tests that check that all classes work (see the loop in the tests for gobymajority).

drvinceknight:
I had missed that you are indeed inheriting but I think it's done more elegantly in gobymajority (tests also). Let me know if I'm missing something.

PR #31, opened 2015-02-21
Appeaser strategy

jomuel:
added a simple strategy that changes its behaviour every time the opponent defects.


axelrod/tests/test_appeaser.py
```diff
@@ -0,0 +1,23 @@
...
+
+class TestAppeaser(unittest.TestCase):
+
+    def test_strategy(self):
+        P1 = axelrod.Appeaser()
+        P2 = axelrod.Player()
+	P1.str = 'C';
+        self.assertEqual(P1.strategy(P2), 'C')
+	P1.history = ['C']
+	P1.history = ['C']
```
drvinceknight: I think this should be `P2.history`? Could you confirm and perhaps include another test to make sure this particular case (for which this worked) is also caught?

drvinceknight:
(Also the latest pull request I threw in has thrown up a tiny merge conflict most probably in strategies/__init.py__, pull from upstream to fix if that's ok :smile: thanks!)

jomuel:
Yes, the P1.history should have been P2.history of course, i changed it and resolved the merge conflict due to strategies/__init__.py.

axelrod/strategies/__init__.py
```diff
@@ -32,12 +33,16 @@
         Grumpy,
         Inverse,
         AverageCopier,
+<<<<<<< HEAD
```
drvinceknight: Looks like you didn't quite catch the merge conflict.

drvinceknight:
Let me know if you'd rather I took care of fixing this: not also the 'X Failed - The Travis CI build failed': this means that the tests have not passed (mainly because of my comment above). Thanks a lot for the contribution: I look forward to merging it :) :+1: 

You might also like to add your name to the bottom of the README.md file (I've just added a list of contributors there).

jomuel:
Sorry for the oversight, I fixed it now. :smiley:

drvinceknight:
:fireworks: :tada: :+1: Thanks! Just about to run it now to update plot :)

Issue #22, opened 2015-02-20
Get mind reader back in


drvinceknight:
mindreader and geller are 2 'cheater' strategies but they cannot actually play against each other.

Would be cool to figure it out (even if perhaps we do this with error handling...)

@JasYoung314

langner:
The infinite recursion caused by two strategies calling each other could be dealt with by inspecting the caller name and bailing out if it is the same as the current function name. But that would mean a 'default' action (C/D) would need to be defined, in other words, do gellers like other gellers?

drvinceknight:
This wouldn't deal with the case where a geller plays mindreader
(mindreader is kind of like geller but on steroids). Would a simple
recursion error catcher followed by a default be a good idea? (I think the
nice with the default would be to pick randomly). Tricky with dealing with
who plays who first though (which would be caught just fine with your
suggestion in the case of geller v geller). Could add an attribute to mind
reading strategies...

langner:
I think it should work fine (see #34) for any Player if it looks at the caller name. Personally, I think an attribute is a better solution, since it is impler. But that would be your decision to make.

langner:
Acutally, it has occured to me that the attribute could be set automatically by the constructor of Player, by parsing the source code of the `strategy` method at construction time to check if it uses the oponnent strategy.

drvinceknight:
I like #34. I hear what you're saying about being able to put it directly in the `strategy` method. I would prefer (for accessibility, as I'm hoping that people with very little coding experience might still contribute) to leave that as simple as it is and leave things like this (for people who are going to mess with 'cheating strategies') up to them.

I'm going to accept #34 but let me know what you think...

langner:
I don't think it matter that much; #34 is fine.

JasYoung314:
I've opened a pull request https://github.com/drvinceknight/Axelrod/pull/55 with @langner suggested fix.

PR #34, opened 2015-02-22
Add a default strategy to Geller (three variants)

langner:


drvinceknight:
@JasYoung314 take a look at these commits: should help with #22

drvinceknight:
Might also be of interest @martinjc.

PR #35, opened 2015-02-22
Adding forgetful grudger strategy

geraintpalmer:
I've added a strategy that is grudger but forgets after 10 turns.


axelrod/strategies/forgetfulgrudger.py
```diff
@@ -0,0 +1,46 @@
+from axelrod import Player
+
+class ForgetfulGrudger(Player):
+    """
+    A player starts by cooperating however will defect if at any point the opponent has defected, but forgets after 20 matches
```
drvinceknight: Is this line right? Looks like 10 matches to me in the code.

PR #38, opened 2015-02-23
Results visualisation on github pages

martinjc:
Initial version of a results visualisation for Axelrod.

Modifies run_tournament.py to output the scores for each player in a .csv, which is then read by D3 for the visualisation


drvinceknight:
This looks great @martinjc, I know nothing about d3 (but need to learn about it) but just a question:

Instead of outputting a csv file is there any advantage to using the mpld3 library which would output json of the plots (which also have nice zoomy in things)?

Another advantage to this being that the matplotlib code need not be 'repeated' in the d3 voodoo.

Again, I know nothing about this, delighted to follow your lead. Just wondering :)

drvinceknight:
The colors look awesome by the way!

martinjc:
Thanks :-)

You could certainly use mpld3 to generate html+js versions of the matplotlib output, yes. I've only taken a brief look at mpld3, but I think there are a couple of issues to be considered.

Firstly, it seems that the output it produces is fixed in size to be the same as the matplotlib output, which isn't great for the web, which can of course be viewed at many sizes/resolutions. I may be wrong about that, but one of my favourite points about using SVG through D3 is the scalability it brings. Also the html+js mpld3 generates to display the plots (without wishing to invite too much judgement on my handwritten code!) is pretty ugly.

The zooming/panning etc is nice, but we can do that with vanilla d3 anyway (I can add that to the next version). Just using d3 also allows us to do further interactions in the future if we wish. Doing any kind of advanced interaction with mpld3 seems to require you to write javascript functions inside your python, which isn't very elegant.

There's also a usability consideration, in that d3 attaches the raw data to the elements used to visualise the data, so the presentation of data is accompanied by the data itself. We could then expose the data values within the SVG elements so that screen-readers and other machine/algorithm tech can read the data. (For example, each box+whisker could have data-attributes storing the value of the median, iqr etc, which could then be read by a screen-reader).

One final point is that using mpld3 generates output that can only be used specifically for showing the mpl plot on the web. Outputting data as .csv or .json is more usable for other (unthought of) things.

I think the main advantage of using mpld3 is that it is ideal in a situation when you just want a version of your mpl plot for the web. The main disadvantage seems to be that trying to do anything more than that is more complicated that just using plain D3.

That's my tuppence on the matter, but happy whichever way you want to go!

langner:
Although I am not going to contribute to the visualization, I would agree with @martinjs. I've use d3 and mpld3 in the past, as well as c3. They are all nice, but in the end you go back to d3 if you need something that goes beyond the basics.

martinjc:
i've added some zooming and panning etc so it's a bit more snazzy

drvinceknight:
That's an awesome explanation @martinjc! I had the issue with size of plots with some html presentation I did recently using d3. Mergin now and more and more convinced that I must learn d3 :)

Thanks a lot!

PR #39, opened 2015-02-23
Forgiving Tit For Tat

uglyfruitcake:
I've added a Forgiving Tit For Tat strategy and I've written unittests for it.


drvinceknight:
Nice! Keep them coming :) :+1:

PR #41, opened 2015-02-24
Implement basic caching and add tricky strategies

langner:
These commits introduce a basic caching mechanism and simplify the code a bit (__repr__ doesn't need to be defined for each strategy, just and attribute with the name). I also cleaned up some of the docstrings.


drvinceknight:
Ridiculous pull request! So awesome.

PR #48, opened 2015-02-24
K78 patch 1

Karlos78:
A slightly dull strategy but hopefully have indented correctly and have tried writing tests.


drvinceknight:
That's a pretty neat one! I refactors slightly, merged with upstream etc... :) Thanks for the contribution! :+1:

Issue #45, opened 2015-02-24
Strategy Design Pattern


meatballs:
Implement the strategy design pattern described here: https://en.wikipedia.org/wiki/Strategy_pattern

This would:
* Separate the behaviour of a Player from its instantiation and/or participation in a Tournament
* Simplify the definition of Strategies
* Allow multiple Players in a Tournament to use the same Strategy
* Enable dynamic allocation of Strategy to Player at Tournament runtime
* Enable the Django model layer (#44) to represent Players and Tournaments as simple objects with properties

Both a function and a class based example of how the pattern might be implemented in Python are described here: http://codereview.stackexchange.com/questions/20718/strategy-design-pattern-with-various-duck-type-classes

meatballs:
Do we want to bother with this? The structure we have in place now would enable this to be done (I think), but it's still a fair amount of work as all the existing strategies would have to change.

In the short term, the main advantage would be to allow tournaments with more than one player executing the same strategy. If that's of interest, then I'll have a crack at it.

drvinceknight:
> In the short term, the main advantage would be to allow tournaments with more than one player executing the same strategy. If that's of interest, then I'll have a crack at it.

But this can already be done right?

```
import axelrod
P1 = axelrod.Cooperator()
P2 = axelrod.Cooperator()
P3 = axelrod.Cooperator()
P4 = axelrod.Defector()
A = axelrod.Tournament([P1, P2, P3, P4])
A.play().generate_scores()
```

Am I missing something?

Nice to keep this as an optional thing for once we've packaged this so people can run their own tournaments...

meatballs:
True. Doesn't seem worth the effort at the moment, then.

Probably one to bear in mind in the run up to djangocon, if this gets in as a sprint.

PR #50, opened 2015-02-25
Adds use of abc module for Player class

meatballs:
Makes the Player class a true Abstract Base Class with the abc module
Adds an AxelrodTestCase class with a concrete test implementation of Player
Alters tests to use AxelrodTestCase and testPlayer


drvinceknight:
Thanks for this @meatballs , again very little that I know about. What is the main advantage of using this?

meatballs:
Well, that's a matter of some debate!! 

I did it simply because the class claimed to be abstract (in its comments) and this implements that claim for definite.

Personally, I think it's clearer if the intention (that a Player subclass must have a name and a strategy) is in the code, and therefore enforced, rather than simply in a comment.

Feel free to disregard if you're a pure duck typing man!

langner:
> I did it simply because the class claimed to be abstract (in its comments)

Who added that comment? That would me :) I didn't think much of it at the time and had no intention  to really make it abstract. But that would be the role of this class here, right? Namely, it's not meant to play at all.

meatballs:
I think so. Could be my age, therefore previous exposure to static typed languages, but it seems cleaner somehow

drvinceknight:
I'm very happy to add this in, I've just gotten back from a DjangoCon Europe organisation meeting @meatballs (very likely to be an Axelrod sprint :)) so haven't had time to look through this to carefully. There doesn't seem to be any negative side to adding this in though right? 

From the point of view of a contributor who does perhaps not know the 'ins and outs' of this abstraction (I'm putting my hand up here) it doesn't add any further confusion right?

drvinceknight:
Does this relate to #45 (which I still haven't read through carefully - apologies but want to make sure I don't rush it) or am I more confused than I thought I was?

meatballs:
I'm not sure it's a great idea to introduce stuff that's not well understood by all - especially when it's not adding any huge benefit. Let's dump it for now.

It does indeed relate to #45, but it might be worth doing the whole of #45 in one go rather than piecemeal like this. Sorry if I've ended up causing confusion!!

drvinceknight:
No apologies needed at all: this is awesome. I'll leave the pull request open and try over the next couple of days to wrap my head around it better: if nothing else I really appreciate the opportunity to learn something.

Will get back to you asap @meatballs :)

meatballs:
I closed this as I made the schoolboy error of creating it on my master branch which causes all sorts of problems if it's not merged in.

I'll create a separate 'abc' branch for this change.

PR #53, opened 2015-02-26
Adds use of editorconfig

meatballs:
A config file which controls text editor settings for those editors which support it. (e.g. indentation format, line ending characters)

More details at: http://editorconfig.org/

Plugins for supported editors here: http://editorconfig.org/#download


drvinceknight:
Never knew about this... Grabbing the vim config plugin now...

meatballs:
Useful tool - especially when there's a mixture of indentation settings in the same project.

PR #55, opened 2015-02-26
Added a fix to mind reader so it can play geller, and tested that it can

JasYoung314:
Adding suggested change from https://github.com/drvinceknight/Axelrod/pull/34 to address https://github.com/drvinceknight/Axelrod/issues/22

### DIFF
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -20,6 +20,7 @@ from oppositegrudger import *
 from oncebitten import *
 from golden import *
 from retaliate import *
+from mindreader import *
 
 strategies = [
         Defector,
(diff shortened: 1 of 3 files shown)


drvinceknight:
Can you add tests for the reset, init and str methods please. Take a look at some of the other strategies to see what is needed. I'm guessing this might be the case for #54 also.

Also, the dosctring for the strategy needs a better explanation: this does more than just what the one sentence you've put down says: 'pretends to play etc...'

drvinceknight:
Note that @meatballs in #56 has added command line arguments to the `run_tournament.py` script so it's easier to control cheaters.

JasYoung314:
Ive added in some tests for all the methods you have listed as well as updated the documentation in the method.

Like you said this was also the case in https://github.com/drvinceknight/Axelrod/pull/54 which has also been taken care of.

Issue #32, opened 2015-02-22
Write a manage.py script


drvinceknight:
As the number of strategies grows: the mathces (or just the random repetitions) can be parallelised (possibly even making use of caching for non stochastic strategies...).

drvinceknight:
I'm going to incorporate this with a new `manage.py` script which will run everything:

- options to run tests
- options to run in parallel
- etc?

langner:
At first glance, caching would be the first things to do to boost performance.

langner:
See #41 for caching for pairs of deterministic strategies.

drvinceknight:
Absolutely huge pull request! Amazing! Awesome @langner ! I'll leave this issue open as I still want to parallelise when I get a chance. That does need the overall better `manage.py` script which I'll get to.

drvinceknight:
(I am learning a lot of Python from your pull request @langner : very much appreciated).

meatballs:
Can this one be closed now?

drvinceknight:
I think I'd still like a manage.py script, which we don't have so that we could move the functionality of `run_tournament` and include options like:

```
manage.py test
```

actually that's the only extra thing I'd want in there that I can think off.. Perhaps not needed?

meatballs:
why not just add any additional functionality as methods in the TournamentManager class? Makes them much easier to test.

drvinceknight:
Does it make sense to have `run tests` in there? I think that's the only
extra thing really needed which I'm now not sure warrants it... Although I
guess could be nice...

meatballs:
Probably doesn't make much sense. Anyone who knows enough to want to run the tests is going to be perfectly capable of typing the command to do so. Running tests from within TournamentManager would actually require more typing!

drvinceknight:
Yeah... Cool. Let's scrap this.

PR #56, opened 2015-02-26
adds command line arguments

meatballs:
Uses the argsparse module to add command line arguments

use `python run_tournament.py -h` to view the options

running it without any options defaults to the current settings.

### DIFF
--- a/run_tournament.py
+++ b/run_tournament.py
@@ -1,28 +1,37 @@
 """
 A script to run the Axelrod tournament using all the strategies present in `axelrod/strategies`
 """
-
 from __future__ import division
 
+import argparse
 import time
 
-import axelrod
-import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


drvinceknight:
This is awesome @meatballs ! I expect it will be modified when I get around to #32 (I want to create a manage script that runs tests as well).

Particularly helpful with @JasYoung314 's latest pull requests which add some cheater strategies that take a while to run (#55).

Have you seen [docopt](https://github.com/docopt/docopt) it's my command line argument tool of choice: extremely elegant but sadly not in the base library (I'm glad that everything so far is). It's just a single python file though...

meatballs:
Ha! Yes, I've seen docopt, but didn't want to introduce something that nobody else had heard of. (I've enough history of that already)!!

PR #57, opened 2015-02-26
Pass mock player objects to strategy

christianp:
So players can't cheat by modifying their opponents.

Cheating by looking at what the opponent's strategy will do is still possible (should it be?)

### DIFF
--- a/axelrod/tournament.py
+++ b/axelrod/tournament.py
@@ -114,7 +114,10 @@ class Player(object):
         an error if no strategy method is defined (which are defined through
         class inheritance).
         """
-        s1, s2 = self.strategy(MockPlayer(opponent)), opponent.strategy(MockPlayer(self))
+        if not (s1 in axelrod.cheating_strategies or s2 in axelrod.cheating_strategies):    
+            s1, s2 = self.strategy(MockPlayer(opponent)), opponent.strategy(MockPlayer(self))
+        else:
+            s1, s2 = self.strategy(opponent), opponent.strategy(self)
         self.history.append(s1)
         opponent.history.append(s2)
 
--- a/axelrod/tournament.py
+++ b/axelrod/tournament.py
@@ -114,7 +114,7 @@ class Player(object):
         an error if no strategy method is defined (which are defined through
         class inheritance).
         """
-        s1, s2 = self.strategy(opponent), opponent.strategy(self)
+        s1, s2 = self.strategy(MockPlayer(opponent)), opponent.strategy(MockPlayer(self))
         self.history.append(s1)
         opponent.history.append(s2)
 
@@ -132,3 +132,11 @@ class Player(object):
     def __repr__(self):
         """The string method for the strategy."""
         return self.name
+
+class MockPlayer(object):
+    def __init__(self,player):
+        self.name = player.name
+        self.history = player.history[:]
+        self.score = player.score
+        self.stochastic = player.stochastic
+        self.strategy = player.strategy


drvinceknight:
There are quite a few strategies that are now built that actually 'take advantage' of this though. They're put in their own 'cheaters' list and a separate tournament is run for them (see second plot in the README). I think it's good fun to be able to allow that (there's just been a pull request that changes the opponent's strategy to 'always cooperate', I want to write one that has a separate response to all known strategies etc...).

I'm happy to keep those in and just keep track of non cheating strategies by investigating the pull requests.

drvinceknight:
I've reread the code and pull request properly and see that yours would indeed just stop 'mind control'... That's perhaps not a bad idea... Not sure, what do you think @JasYoung314 (Jason wrote the mind control one).

There's a bit of me that wants to let everyone cheat in anyway they want (again pointing to the fact that they are kept separate from the other strategies), I like the idea that 'cheating' is a 'no rules' kind of thing. People might keep on thinking of smarter coding tricks to beat each other (I actually have a plan for `mindreader`...).

JasYoung314:
@christianp I think that having a no holds barred cheating tournament is kind of fun (Personally I'm enjoying the challenge of using coding tricks to beat the opponent).

However, having these anti cheat failsafes in the main tournament might not be a bad idea. Is there a way to only use mock players in the main tournament?

drvinceknight:
There probably (inherit a `CheatingTournament` class) is but I'd still have to inspect things carefully at every pull request so probably easier to just leave it as it is?

JasYoung314:
I think leaving cheating in is a good idea (as two of my three strategies rely on cheating)

langner:
My opinion is that this is a question of focus. Cheating is simply "beside the point", because it reduces the contest to technical questions of coding. Sure, you can continue to find ways to change an opponent's strategy, history or the opponent itself! For example, here is a player that kicks MindControl's butt:
```python
class MindWarper(Player):
    name = 'Mind Warper'
    def __setattr__(self, name, val):
        if name == 'strategy':
            print 'Not so fast!'
        else:
            self.__dict__[name] = val
    def strategy(self, opponent):
        opponent.strategy = lambda opponent: 'C'
        return 'D'
```

We could now devise a way to defeat MindWarper with a MindBender player that modifies the internal __dict__ of MindWarper, or whatever. It might also make sense to harden the Player base class to disallow certain types of cheating as they are discovered. Python is flexible enough to allow for all that.

This is all very entertaining, but I don't think it's relevant to the goal of the project. Presumably it is about game theory and the evolution of cooperation. Not about the subtlteies of the Python language. Python is super flexible and that is why conventions are so important. A convention that clearly states the rules of the game is enough to not waste time on so much extra code.

Just my two cents.

drvinceknight:
Yeah I completely agree with you and it's great that that's how this has
been going (not many cheating strategies have been submitted at all). I
don't see the harm of having a side tournament which includes the cheaters
(simply for fun and just seeing the response dynamics you illustrated is in
itself interesting). I plan on having a thorough go at some documentation
and revamping the readme over the weekend so I might clarify all this
(giving a slightly less visible spot to the cheaters).

I'd almost not want to incorporate checks and limits on the cheaters
tournament as I don't want to take away from the actually interesting one
(without cheaters). By having a no holes barred tournament it will either
interest no one or be a cool to for some to learn certain python
subtleties.

Adding my own two cents :)

langner:
> I'd almost not want to incorporate checks and limits on the cheaters tournament as I don't want to take away from the actually interesting one (without cheaters). By having a no holes barred tournament it will either interest no one or be a cool to for some to learn certain python subtleties.

Sure, that's the simplest approach. I think it will lead to a side tournament that will be very far from the original one and focused on the tricks. Anyway, I'm all for the fun, so I guess I should put `MindWarper` in there :)

drvinceknight:
Yeah I expect the side tournament to be completely different. If at some point it did become too chaotic then perhaps we could bring in some restrictions. 

Let me know what you think @christianp if you're OK with it I'll close this pull request. Happy to discuss further though :)

christianp:
How about this - strategies in the cheating list get the raw object, but non-cheating strategies get the mock object.

drvinceknight:
I'm hesitant. That would work just fine but either way I'd still have to inspect the pull requests carefully (to make sure that a cheater is not put in the wrong list), not really sure it adds much apart from a _slightly_ more complicated overall class...

I don't think anyone is going to wilfully try and actually cheat (as opposed to creating 'cheater' strategies) anyway so I'm not sure this is solving a problem. Not at all meaning to be negative just hesitant to add code that I'm not sure has a purpose... 

Perhaps I'm too optimistic about people submissions and also about my ability to not be tricked :)

drvinceknight:
What does everyone think?

langner:
I tend to agree with Vince - it's unnecessary, assuming people don't really cheat.

Plus, a cleaner solution to block mind control and basically kill all current cheating would be to pass just a copy of the opponent's history to `strategy`, since we don't expect it to use anything else.

drvinceknight:
I'm going to go ahead and close this pull request but feel free to reopen if anyone wants to talk about it more. Really appreciate it @christianp, it's been good to discuss. Please throw a strategy in :)

PR #66, opened 2015-02-27
Added pi and e as mathematical constants similar to golden ratio

timothyf1:


drvinceknight:
Awesome strategy @timothyf1 : thanks!

PR #67, opened 2015-02-27
Adds limitedRetaliate to strategies

meatballs:
Not sure how this went AWOL in the merge revert, but here it is again anyway


drvinceknight:
My bad. Should have caught that.

PR #70, opened 2015-02-27
Adds randomised retaliate and random limited retaliate strategies

meatballs:


drvinceknight:
Is random retaliate doing what you want? This will set the limit once when the class is loaded, do you not want to have it re sample every time it plays? If this I'm wrong that's completely fine, just checking.

drvinceknight:
It would be nice to see both those have a couple more tests. You can set the seed in tests to get consistent output for stochastic strategies. :smile:

meatballs:
It's behaving as intended. I wanted to randomise its settings slightly so that, over many tournaments, I can see whether the settings for the main classes are optimal. They should play an entire tournament with the same settings.

I didn't bother with tests as neither class has any new methods - they both just inherit from classes which are already tested. If I fix the settings in a test, I'd just be repeating the test that are already in place for the parent classes.

meatballs:
Having typed the previous comment, I've just realised it causes a problem.

In any given tournament, the the randomised players are wholly deterministic, but will appear as stochastic and so not cached. That slows the tournament unnecessarily.

I'll add a mechanism whereby a class can declare itself to be stochastic or deterministic which will take precedence over the detection of the random module.

meatballs:
I don't believe there are any additional tests required. Shout if I've missed something!

drvinceknight:
Just confirming that this is the required behaviour (you won't get a different value over each repetition of the tournament as the class is set and the random number chosen for every Python instance):

```
Axelrod (meatballs-limitedRetaliate)$ python
Python 2.7.8 (default, Oct 19 2014, 16:02:00)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.54)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import axelrod
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.15/11)
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.15/11)
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.15/11)
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.15/11)
>>> import axelrod
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.15/11)
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.15/11)
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.15/11)
>>> exit()
Axelrod (meatballs-limitedRetaliate)$ python
Python 2.7.8 (default, Oct 19 2014, 16:02:00)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.54)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import axelrod
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.06/26)
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.06/26)
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.06/26)
>>> import axelrod
>>> import axelrod
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.06/26)
```

If that is the behaviour you won't (I'm not entirely sure it is) then I think it needs to be tested... (sorry for being a pain). Similarly as you're re-writing the init method and even the repr method I'd like to see a test for that, (I know that it's trivial but I'd feel better about it).

So something that tests:

```
>>> import axelrod
>>> import random
>>> random.seed(5)
>>> axelrod.RandomLimitedRetaliate()
Limited Retaliate (0.12/12)
```

Although note that that does not work because of where you are importing the `random` library so I'm not too sure what the solution is. There are a couple of random strategies that have tests with seeds, maybe importing the random library at the top of the python file and overriding the stochastic thing for each strategy would take care of it...

meatballs:
The idea is that each time run_tournament.py is executed, the randomised players get different settings which are then fixed for all tournaments within that run.

The reason behind it is that the two fixed settings for LimitedRetaliate are based on only two or three runs of run_tournament.py and it would be interesting to see how they perform against the randomised players.

There's no re-write of the repr method - the init method simply sets the stochastic attribute in a different fashion.

The init method on LimitedRetaliate and Retaliate uses the Player init and then additionally sets the name. That's already tested by test_representation on both classes.

On RandomLimitedRetaliate and RandomRetaliate, the init additionally sets the stochastic attribute which is then tested by test_stochastic on both classes.

As for the strategy, if I fix the settings, the tests become identical to those which are already there for LimitedRetaliate and Retaliate. If I use random.seed, all I'm doing in addition is testing whether the Python random module is working as expected, which seems over the top.

Sorry to be a pain too!! I'm a great believer in unit testing, but I also like all my code to be DRY and that includes tests. I honestly can't see anything I could usefully add which isn't already being tested.

meatballs:
It might be worth renaming some of the tests. I went with the convention that's grown up from other strategies (e.g. test_representation), but that's possibly a little misleading in this case. It's not really the repr method that we want to test as that's simply inherited from the parent. It's the effect on the name of the init that needs testing.

Same test code in the end, but it might make more sense if the test were called test_init instead.

drvinceknight:
> The idea is that each time run_tournament.py is executed, the randomised players get different settings which are then fixed for all tournaments within that run.

Cool, we're in agreement that that's what this currently does :+1: :smile: 

> The reason behind it is that the two fixed settings for LimitedRetaliate are based on only two or three runs of run_tournament.py and it would be interesting to see how they perform against the randomised players.

I'm still a bit confused by this: every time the results are updated a different `LimitedRetaliate` will be in place. I'm not sure you'd be able to 'see how they perform' because you won't have a history of how they do... Am I missing something? I'm not at all against this, am just not sure I understand the point. Would it not be better to have:

- a spread of `LimitedRetaliate` with a bunch of different parameters (like `GoByMajority`)
- another `LimitedRetaliate` which updates the parameters randomly at every run of the tournament (using a `reset` method).

I think that if the current behaviour is indeed what you/we want (I'm going to need to understand it a bit better though) than that behaviour needs to be tested (more or less copying the output of my Python shell above). I can imagine someone coming along and thinking that this isn't what we want the behaviour to be and put the random selection of the parameter in the init or the reset (so that the player changes behaviour slightly at every game). Without a test, that refactoring would break your wanted behaviour (although at the moment I think I understand the 'imaginarily refactored' strategy better)...

meatballs:
> Would it not be better to have:
>
> a spread of LimitedRetaliate with a bunch of different parameters (like GoByMajority)
> another version of LimitedRetaliate which updates the parameters randomly at every run of the tournament (using a reset method).

I considered using the GoByMajority approach, but the strategy seems very sensitive to small changes in its settings, so I'd end up creating a whole raft of them. Didn't want to swamp your tournament with twenty retaliate players!

On LimitedRetaliate, it would be even more so as it has two settings rather then one.

I'm not that bothered about the overall history, it's just interesting to see on each run whether a randomised player has come up with a better setting than the current fixed settings.

>  I can imagine someone coming along and thinking that this isn't what we want the behaviour to be and put the random selection of the parameter in the init or the reset (so that the player changes behaviour slightly at every game). 

Hadn't thought of that!!

I'll write something to test that the settings don't change between rounds

drvinceknight:
> I considered using the GoByMajority approach, but the strategy seems very sensitive to small changes in its settings, so I'd end up creating a whole raft of them. Didn't want to swamp your tournament with twenty retaliate players!

I think I'd prefer a range of strategies (perhaps not twenty as I'm sure 6 over the same range would give insight because of the linearity and continuity of the payoffs). Indeed after running 4 or 5 you might even be able to 'zoom' in on a particular parameter space area...

I've been thinking about this a bit more while walking the puppy and even from a 'scientific' point of view of reproducibility I would prefer to not have a 'randomly different' strategy at ever run of the tournament (*especially* if it is extremely parameter sensitive). All other stochastic strategies are to some extent 'determined' and will give quasi-reproducible results because of the 50 repetitions which aims to reduce effect of variability. 

What do you think?

> Hadn't thought of that!!

I get why you'd rather have DRY tests (and code) but I always like to have too many rather than too little just in case I haven't thought of something... 

I'm a pain with all my students as I more or less demand everything under the sun is tested. Twice. Even if it is something that is potentially 'tested' because of inheritance. You never know what someone is going to try and do through refactoring... I recently lost 3 hours of my life because I (being an idiot) refactored something that broke everything, it would have been immediate to notice if correct tests had been written... (I'm sure you know someone who could tell you how painful about it I can be :smile: )

drvinceknight:
(On a side note: I am thoroughly enjoying this conversation, I love this about open source software. I hope it's not too painful from your point of view. I am enjoying us working together to figure out the best way forward.)

meatballs:
> (On a side note: I am thoroughly enjoying this conversation, I love this about open source software. I hope it's not too painful from your point of view. I am enjoying us working together to figure out the best way forward.)

Not painful in the slightest!! It's always great to be both contributing and learning on the same project.

> I think I'd prefer a range of strategies (perhaps not twenty as I'm sure 6 over the same range would give insight because of the linearity and continuity of the payoffs). Indeed after running 4 or 5 you might even be able to 'zoom' in on a particular parameter space area...

OK. I'll get rid of the random stuff and do that instead.

> I'm a pain with all my students as I more or less demand everything under the sun is tested......

I shall bear that in mind!

> (I'm sure you know someone who could tell you how painful about it I can be :smile: )

Pretty sure he could talk about how painful I can be about all sorts of stuff too.....

Obviously, all of this will have to wait until the Wales match is over.

drvinceknight:
> Obviously, all of this will have to wait until the Wales match is over.

Of course: enjoy the game! (I've got one eye on Scotland - Italy before giving the big game my full attention)

PR #71, opened 2015-02-28
Use full payoff matrix, save scores to CSV file and add basic ecosystem simulation

langner:


drvinceknight:
This looks big (and awesome). Will read through asap. Love the idea of the tournament matrix being output. Could plot a heatmap maybe...

langner:
> This looks big (and awesome). Will read through asap. Love the idea of the tournament matrix being output. Could plot a heatmap maybe...

Possible. I tried drawing a table in matlplotlib containing payoffs per-turn, but it becomes really unreadable or too large when all strategies are considered.

P. S. Please wait some, I want to push something else to the PR.

axelrod/tournament.py
```diff
@@ -1,11 +1,12 @@
 """Recreate Axelrod's tournament."""
 
 import inspect
-import itertools
+import numpy
```
drvinceknight: I know this makes things a lot easier but I really would like do like that the base tournament only requires base python (I know the analysis uses numpy and matplotlib)... A quick scan shows that you're using this to create an empty matrix, perhaps you could throw in an exception and do that another way if need be?

drvinceknight:
> P. S. Please wait some, I want to push something else to the PR.

Cool: let me know :) There's a big rugby game here on TV later and I have a huge massive amount of papers that need grading so it'll probably wait till after/tomorrow (I'm just lying to myself: I'm too excited about this and will probably look at it sooner rather than later).

axelrod/tournament.py
langner: It's not just creating an empty matrix, it's creating an array filled with zeros. Plus, having that array allows indexing that replaces cumbersome nested loops. Nonetheless, it can be replaced easily (see below) and the performance gain is negligible at this point.

langner:
Something might be wrong in the simulation, because for the normal strategies there is a rather quick equilibrium. Or it just might be that stable.

Also, legend for stackplot is not streaightforward.

drvinceknight:
Any chance you could put the evolutionionary stuff in a separate pull request? Haven't look through it yet but if there's something wrong and it needs work I'd like to be able to discuss it separately. I'd it's a pain or makes sense to be in this for some reason just leave it here. :+1:

langner:
OK, I removed it from here.

langner:
Also added one more thing: a basic set of strategies (did not commit the plots, though).

run_tournament.py
```diff
@@ -1,60 +1,89 @@
...
+
+def run_tournament(turns, repetitions, exclude_basic, exclude_strategies, exclude_cheating, exclude_all):
+    """Main function for running Axelrod tournaments."""
 
-    strategies = []
     cheating_strategies = []
     all_strategies = []
     graphs_to_plot = {}
 
+    init_strategies = lambda S: [s() for s in S]
```
drvinceknight: Love this.

drvinceknight:
This pull request up skilled my python knowledge by a gazillion.

drvinceknight:
I tweaked the `strategies/__init__.py` file so that there was no repetition/overlap in the two lists of strategies: 2ccbe05a621640e521e9fabe8411057f4811eab0

PR #76, opened 2015-03-02
splits test_tournament.py into separate files for each class under test

meatballs:
...and tournament files


drvinceknight:
These greatly improve the organisation of everything. Thanks!

PR #78, opened 2015-03-03
creates ResultSet class

meatballs:
I've created a new ResultSet class, amended the Tournament class to use it and shifted the test for scores initialisation.

The intention is to move the functions that create the plots and csv files into that class from run_tournament.py, but I won't start that unless you're happy with what I've done thus far.

There's a dummy test in test_resultset.py which deliberately fails so that this PR doesn't yet look ready to merge.


drvinceknight:
No no, I like the idea of this. Very modular. Should make #79 even easier (very minor tweak and clean up).

meatballs:
Ok - I'll get on with it!

meatballs:
The day job is going to take precedence for the next few days. The csv method is done, tested and working. I've not used matplotlib before, but there's some work required on the plot method so that it returns a fig object.

drvinceknight:
Cool: no worries :)

If I get time will send some pull requests your way (I doubt it though the
day job is becoming the day and night job at the moment).

meatballs:
As I understand it so far, run_tournament.py uses matplotlib.pyplot at the moment, which, according to the docs:

> Pyplot provides the state-machine interface to the underlying object-oriented plotting library. The state-machine implicitly and automatically creates figures and axes to achieve the desired plot

I think what's required for my plot method is to create the fig and axes objects explicitly, but I've no idea how to do that and then to achieve the boxplot at the moment!

drvinceknight:
Are you 'just' trying to recreate the same plot as before or trying to
create a new plot? (Sorry for not following.)

meatballs:
The same plot - just returned as an object from the method rather than saved directly to a .png

run_tournament can handle saving the file. The ResultSet plot method should just return an object.

drvinceknight:
With you. Does this sort it out:
https://gist.github.com/drvinceknight/efea1afb243e8170c120

meatballs:
Good grief!! It's that simple?!!! That may well have saved me hours of hunting around the docs and examples!!

I'll give that a whirl on the train to London tomorrow, with any luck.

drvinceknight:
Yeah... I go through phases of loving matplotlib and wondering who the hell
came up with the syntax. I really should just look in to it and understand
it...

meatballs:
Seems to be working - just need to write a test for it now

drvinceknight:
Hehe... I thought this was waiting for the train ride tomorrow? :)

meatballs:
Couldn't leave that go once I didn't have to read the docs

meatballs:
Tests are all passing, but the plot isn't correct. Not sure where the error is at the moment.

meatballs:
Nope - can't see what's wrong here. It's definitely the plot method in ResultSet - if I pass the ResultSet object from the tournament into run_tournament.py and use the existing code to produce the plot there, it all works fine (so the methods to produce the numpy arrays etc. must be ok).

If I use the plot method, I get a figure with the correct ticks and titles, but the boxes and whiskers are nonsense.

meatballs:
I think this might need a separate class for the box plot - so that the functions which generate the plot and  the ticks can be refactored as methods and therefore tested.

How about I get rid of the plotting and we merge the work I've done so far to give us the csv output?

drvinceknight:
That sounds good to me: you'd leave the plots as is in the `run_tournament.py` script?

meatballs:
Mostly - It works when I leave the plotting in run_tournament.py but drive it from the new methods in the resultset (rankings, ranked_names, scores)

drvinceknight:
What would you prefer to do:

- Leave the plot in this pull request and I can perhaps take a look (no promises on when...)?
- Throw it out of this pull request and work on it separately (as long as `run_tournament.py` still creates the plots).

I'd be tempted to go with the second option... To be honest the plotting could even be done from the data output anyway (in a way the written file just takes precedence of the other object you're trying to pass).

Having said that, it might well just be a weird matplotlib thing that wouldn't take long to fix... Your call :+1:

meatballs:
I think the second option. I'll fix up this PR with the plot code left in run_tournament, but based on the resultset methods.

drvinceknight:
Yup, sounds great!

meatballs:
that should do it for the moment!

drvinceknight:
Weird... Travis is taking forever...

meatballs:
the yanks are out of bed

drvinceknight:
Ah. Tests failed on Travis because of matplotlib not being there. I might need to add something to the 'travis.yml' file. Could you also though perhaps put in an exception for when matplotlib isn't available? Perhaps users could just get the csv files in that case... 

Hold on (haven't had my coffee yet), I thought this PR wasn't doing anything with matplotlib anymore? (haven't looked at changes yet)

meatballs:
the import statement for matplotlib was still in there - sorry. That's why we CI, I suppose!!

axelrod/result_set.py
```diff
@@ -0,0 +1,54 @@
...
+        self.results = [
+            [[0 for irep in repetition_list]
+                for j in player_list] for i in player_list]
+        self.output_initialised = False
+
+    def generate_scores(self):
+        numpy_array_results = numpy.array(self.results)
+        return numpy_array_results.sum(axis=1)
+
+    def generate_ranking(self, scores):
```
drvinceknight: On all these simple statements, could you add a single line or two of docstrings? Hoping to package this at some point (so people could `pip install axelrod`) so it should throw something up if and when `help` would be run it. Sorry to be a pain. Otherwise: all looks really great.

axelrod/result_set.py
```diff
@@ -0,0 +1,54 @@
...
+        ranking = sorted(
+            range(self.nplayers),
+            key=lambda i: numpy.median(scores[i]))
+        return ranking
+
+    def generate_ranked_names(self, ranking):
```
drvinceknight: Same comment here (and throughout if possible). I'm going to need to find that 8th and 10th day in the week and at some point go through everything and add at least one line of docstring but I can at least start asking for it from now...

meatballs:
Hmmmmm..... thought you might say that, but was hoping to get away with it!!

They are exatly the same code that was previously in run_tournament.py - just moved to separate methods so they can be tested. I have to confess, I don't actually know what they do!! I just made sure the methods gave me the same results as the original code!!

meatballs:
Hmmmmm..... thought you might say that, but was hoping to get away with it!!

They are exatly the same code that was previously in run_tournament.py - just moved to separate methods so they can be tested. I have to confess, I don't actually know what they do!! I just made sure the methods gave me the same results as the original code!!

drvinceknight:
I very much appreciate it :)

> I have to confess, I don't actually know what they do!! I just made sure the methods gave me the same results as the original code!!

I think I'm just panicking at times (this has garnered far more interest than I could have hoped for) and keep on thinking: 'Oh I need to make sure I ask for that'... If I knew what I was doing from the start I guess I would probably be and have been doing a better job :) I've got some tests and exams to write but as soon as that's done I'm going to go through everything really carefully...

Issue #73, opened 2015-03-01
Go through all strategies and write more tests


drvinceknight:
I'd like every strategy to be extensively tested. Some don't have tests for basic methods (`repr` etc...), whilst this might seem like overkill, I would prefer this project to be over tested than under tested.

PR #80, opened 2015-03-03
Simplify strategy tests (and some minor things)

langner:
This addresses #73 and reduces the code necessary for testing strategies. For simple things, it should be enough for someone to set the attributes `name`, `player` and `stochastic` and the basic tests will be done by `TestPlayer` if the test class inherits from it.


drvinceknight:
This is awesome: thanks.

PR #95, opened 2015-03-07
Autogenerate strategies for docs

langner:


drvinceknight:
Am I understanding this right: that make file gets run by readthedocs so it will auto generate the strategies.rst file?

langner:
Yes, it should. If there is a problem, we can remove that dependency and build `strategies.rst` by running `make strategies` and commiting it whenever needed.

drvinceknight:
Or do I need to run `make strategies` to rebuild the `strategies.rst` file? and if so, does that take in to account dependencies as long as a strategy has been added to the `axelrod/strategies` directory?

Have merged in to master either way :)

:+1:

drvinceknight:
The next PR I get for a strategy: I won't modify the `strategies.rst` file and see what happens...

langner:
I'm not entirely sure how readthedocs works internally, but if it runs `make html` with unmodified modification times it should build `strategies.rst` as well, since I added it to the dependencies of the `html` target. Still, updating `strategies.rst` in the repo is not a bad idea.

drvinceknight:
Sweet sweet sweet :)

This is much awesome :)

PR #96, opened 2015-03-07
Un-numpification and introduce payoff matrix

langner:


drvinceknight:
This 'core dependency' of numpy, unless I've missed something along the way in the PRs, it's only needed to run results right? Would someone be able to use this library without numpy? (Not to run the tournament).

Not the end of the world if I've indeed missed that.

drvinceknight:
Probably worth adding a `requirements.txt` file and some info about `pip` grabbing the dependencies to the `README`... (Not terribly relevant to this PR, just thinking aloud).

langner:
I have not read all of the PRs along the way, and I don't understand what you mean exactly, but you do now import NumPy in `axelrod/result_set.py`:
https://github.com/drvinceknight/Axelrod/blob/master/axelrod/result_set.py#L1

And since result_set is used by `axelrod/tournament.py`, it is pretty much impossible to run a tournament without NumPy installed at this point.

drvinceknight:
Sorry for not being clear. At one point I really wanted to keep this written in pure Python except for things like plotting obviously. Do you think this is something worth aiming for? The stuff we're using `numpy` for could be relatively straightforwardly done in pure python at a low extra computation cost.

What do you think? Worth sticking trying to stick to pure Python or should we just embrace using appropriate libraries where they are appropriate. Hold on this isn't a conversation for here, here's an issue: #98

With regards to this actual PR: is this ready to be pulled in or is there more coming?

langner:
There is some more coming, but feel free to pull at will since I cannot predict when I will have a moment to sit down and write.

drvinceknight:
This looks like a big piece of work: thanks! Need a hand getting the tests right or are you ok to keep working on it?

langner:
The tests now pass and I resolved merge cobflicts with master.

run_tournament.py
```diff
@@ -41,14 +51,27 @@ def run_tournament(turns, repetitions, exclude_basic, exclude_strategies, exclud
...
-            else:
-                print ("The matplotlib library is not installed. "
-                       "Only .csv output will be produced.")
+            figure = boxplot.figure()
+            figure.savefig(plot, bbox_inches='tight')
+            figure.clf()
+            
+            # Save plot with average payoff matrix with winners at top.
+            pmatrix_ranked = [[results.payoff_matrix[r1][r2] for r2 in results.ranking] for r1 in results.ranking]
+            fig, ax = plt.subplots()
```
drvinceknight: Does this not need to be in the line `if boxplot.matplotlib_installed:` statement?

drvinceknight:
Love the heat maps.

run_tournament.py
langner: No, because there is a conditional `continue` two lines before that for the opposite case.

drvinceknight:
Missed that: awesome.

PR #103, opened 2015-03-08
Creates BoxPlot class and tests

### LINKED ISSUE #88: Create a BoxPlot class
This class should handle the production of a matlibplot boxplot based on a ResultSet instance passed to its init method.

The functions to generate the boxes/whiskers, xticks, and title should be separate methods with tests for each.

The method which would return an instance of matlibplot.Figure probably cannot be tested as Travis does not have matlibplot installed.

run_tournament.py would handle the creation of .png files based on the Figure object returned from this class.

meatballs:
Fixes #88


drvinceknight:
Sweet! Just checked the travis output and see that it skipped the matplotlib test. I'll see if I can sort that.

drvinceknight:
Latest commits now get the matplotlib tests passing on travis also. For the record I've had to use http://ryanvarley.uk/solutions/2014/05/14/how-to-solve-travis-ci-matplotlib-tclerror-no-display/ (writing this down so that when it breaks I can come back and see it: might throw it in an issue somewhere...). 

On another note it looks like matplotlib is being installed every time a commit is pushed to travis. This seems to be the way things are done...

Issue #99, opened 2015-03-08
Remove Numpy throughout


drvinceknight:
After discussion #98, I've decided that it'll be nice to have this written in as pure Python as possible.

@langner seems to already be working on this here #96 (Thanks!)

meatballs:
I think this is now done

PR #107, opened 2015-03-10
Renames and extends BoxPlot Class

### LINKED ISSUE #105: Rename and extend Boxplot class
Rename and extend the BoxPlot class so that it also processes the payoff plots thus:
 
* Simplifying run_tournament
* Allowing the elements of the payoff plot, currently in run_tournament, to become individual methods which can be included in the test suite
* Only checking for matplotlib installation once - in the class

meatballs:
Fixes #105 

Also renames all the files in the assets directory (and the corresponding entry in Readme.md).

Also removes numpy from test case for BoxPlot/Plot #99


meatballs:
I'm done messing about with this. It's ready to roll.

PR #110, opened 2015-03-10
Add the Darwin strategy

pmslavin:
A hopefully interesting, but not terribly competitive, strategy which accumulates a record (the 'genome') of what the most favourable response in the previous round should have been, and naively assumes that this will remain the correct response at the same round of future trials.

This genome is a class property and is not reset between rounds or opponents, so it eventually grows to represent the historic correct responses to an 'average' opponent.  It also includes a mechanism to thwart cheating opponents who attempt to simulate its responses.

Some logic and consistency tests are included.  I've checked the responses for this strategy and it seems distinct from the existing ones(!) but please feel free to check this.


drvinceknight:
This is really cool @pmslavin but I'm afraid that I think this would not qualify it as an 'honest strategy'. Those only get to know the history of any given round...

Correct me if I'm misunderstanding though...

drvinceknight:
(This could still be merged in but as a 'cheating' strategy. My currently top of the todo list is to write clear contribution guidelines...)

pmslavin:
There is certainly information passing between rounds, but the strategy doesn't record opponents' histories explicitly; rather it assumes that the 'correct' response for trial *i* of round *j* will remain the same for trial *i* of all future rounds.  This response is then modified whenever that assumption proves incorrect (like a gradual Win-Stick-Lose-Switch).
That may well be cheating according to the tournament rules, and if so the strategy can be reclassified.

axelrod/strategies/darwin.py
```diff
@@ -0,0 +1,63 @@
...
+        assumes that this will remain the correct response at the same round of
+        future trials.  This 'genome' is preserved between opponents.
+        If no record yet exists, the opponent's response from the previous round
+        is returned.
+    """
+
+    name = "Darwin"
+    genome = ['C']
+    valid_callers = ["play"]    # What functions may invoke our strategy.
+    outcomes = { ('C','C') : 1,
```
drvinceknight: You can read this directly from axelrod.game (helps minimise repetition).

axelrod/strategies/darwin.py
```diff
@@ -0,0 +1,63 @@
...
+    def strategy(self, opponent):
+        # Frustrate psychics and ensure that simulated rounds
+        # do not influence genome.
+        if inspect.stack()[1][3] not in self.__class__.valid_callers:
+            return 'C'
+
+        trial = len(self.history)
+
+        if trial > 0:
+            outcome = self.__class__.outcomes[(self.history[-1], opponent.history[-1])]
```
drvinceknight: I am perhaps missing something: why use `self.__class__` and not just `self.`?

axelrod/strategies/darwin.py
drvinceknight: Sorry, just realised you're not using the actual game matrix here.

axelrod/strategies/darwin.py
```diff
@@ -0,0 +1,63 @@
...
+            self.__class__.genome.append(opponent.history[-1])
+            current = opponent.history[-1]
+
+        return current
+
+
+    def reset(self):
+        """ Reset instance properties. """
+        self.history = []
+        self.__class__.genome[0] = 'C' # Ensure initial Cooperate
```
drvinceknight: I'm afraid that this is definitely a 'cheating' per say strategy. Some of the q-learners wanted to do this sort of thing.

drvinceknight:
This looks really neat @pmslavin, I'm going to check through the tests carefully and will add in (I have one question remaining above). Once I've done that I'll put it in the 'cheaters' classification but it'll still be an awesome strategy to have: we're working towards building this as a library so people could play with whatever strategies they want. First, I need to get the docs done though :D

pmslavin:
Last commit fixes a typo in the super class constructor call.  The use of __class__ to scope class attributes makes it explicit that these reside on the class, rather than any instance, and makes sure that assignments to these are modifying the persistent class attribute and not creating an instance attribute with the same name.  In terms of the 'Darwin' analogy, this emphasizes that the type itself is changing in response to selection.

drvinceknight:
> The use of __class__ to scope class attributes makes it explicit that these reside on the class, rather than any instance

Does this mean that in a hypothetical tournament with multiple Darwin players that they would each have the same `genome`? I'm not sure I'm following why you would not just change the instance attribute...

pmslavin:
Yep, exactly - there is a single copy of the genome across all instances, and this persists between rounds.  Having this as an instance attribute would lose the 'learned' responses between rounds; as it is, each `Darwin` is giving responses learned by other instances in earlier rounds.

drvinceknight:
Well, instances are not being regenerated for each round: which is why they have a `reset` method.

So in a tournament (with a single `Darwin` player) - as the one run by the `run_tournament` script) - this is completely equivalent to just setting the attribute class. The only place it becomes relevant is that if you have two (or more) Darwin players in a tournament (which is completely possible to do) then they are basically learning together. 

If this is the behaviour you would like then it would be good to include that in the initial docstrings and also make sure it's tested (I have not checked your tests yet so it might already be there).

Apologies if I'm misunderstanding...

pmslavin:
Yes - that really is the plan.  Every subsequent instantiation results in the most recent `Darwin` that has accumulated all of the changes since the class was first loaded.  So although `run_tournament.py, line 16` creates a list of player objects once that is subsequently used for every round and repetition, the instance created at `round_robin.py, line 46` should also share the unique `genome` that characterises the class; as should any future `Darwin` objects created under any other circumstances.  The idea is that the `type` itself is building a permanent set of responses.

I'll add some comments to make this less opaque and move it into the `cheating_strategies` list.

langner:
Telepathic memory in the type object is a cool idea. But if the rounds are long enough, the benefit of this memory is not too large. Still, it violates the implicit assumption that consecutive round robins are independent of each other.

drvinceknight:
Ok, sounds good. Let me know when those changes are done @pmslavin :)

pmslavin:
@langner Yes, the rankings show that this gives worse results than even some very basic strategies.  Hopefully this gives some insight into the time-value of agents' histories (i.e. it is limited) and the portability of winning strategies between opponents (i.e. they are broadly not).  It would be interesting to see whether this strategy has different outcomes in the 'ecological' variants where there are explicit generations.

I've reclassified this as a `cheating_strategy`, clarified the class docstring, and added a few more tests.

drvinceknight:
> 'ecological' variants where there are explicit generations.

I still am not sure I understand why all instances of the class should communicate (the one specific line you mention in round_robin is something that could be tweaked anyway, it's just the strange case when they play each other) but that's not at all a problem. Just me not quite anticipating that such a strategy would work this way (which is what is awesome about putting this on github :D, I would have never written that as you did :)).

> I've reclassified this as a cheating_strategy, clarified the class docstring, and added a few more tests.

I've left an inline comment about the docs: the fact that all instances communicate with each other has been left out (and that's the thing that's quite particular I think...).

drvinceknight:
Apologies, I had in fact missed something. Thanks for the patience and the edits: this is a great addition :+1:

pmslavin:
My pleasure; this tournament is a really interesting concept, so thanks for making it available.

Issue #68, opened 2015-02-27
Parallelise the tournament


drvinceknight:
To speed up the running of the tournament it would be nice to add the option to parallelise the various matches (or indeed more simply the repetitions).

This could then be run by using a command line argument...

I'm pretty sure I've never parallelised anything 'in the right way' so I'm throwing the help wanted label on this one... If you'd like to pick it up: let's chat :)

meatballs:
I've no experience of this, but I fancy learning!

Any preferences on how this should be done and what library to use? From the (little) research I've done, it looks like multiprocessing might the easiest option.

drvinceknight:
That would be awesome, I'm just struggling for time and really think that docs should be my priority... So it's awesome if you can pick this up.

The multiprocessing library is what I've always used (and again it's in the base library: :+1: ). I might be able to help along the way though so throw some questions my way (although as I said above: not sure I've every actually done this the right way - it has always felt hacky).

meatballs:
Right - I'll make a start. Branch 68 in my fork will contain anything I've done so far so anyone can take a look and contribute. I'll create a PR here when there's something worth merging.

drvinceknight:
Just taken a look.

Is the idea to have this as an option that gets passed to the `tournament_manager`?

Eventually this could be then passed to the `run_tournament` script through something like:

```
python run_tournament.py # Run on one core (status quo)
python run_tournament.py -p # Run with all available cores
python run_tournament.py -p 2 # Run with two cores
```

meatballs:
Yep, that's exactly the plan!

To start, I'm just going to run the individual tournaments in parallel (basic_strategies, cheating_strategies etc.) I'll look into the rounds and repetitions once I get that working.

drvinceknight:
I was wondering about this myself. What do you reckon it make sense to
parallelise? The tournament or repetitions of the tournament?

I don't think in the grand scheme of things that it matters (as long as all
workers are busy) but the second approach would be easier to implement...

meatballs:
No idea! I intend to hit the easiest option first and see what happens.

drvinceknight:
Cool :)

meatballs:
Progress so far......

Running the individual tournaments in parallel is simple, but yields little benefit since the longest running tournament is significantly longer than the quickest.

[...]

PR #111, opened 2015-03-11
Creates TournamentManager class

meatballs:
and uses it in run_tournament

ground work for #68


meatballs:
well, that broke my PR!! I'll get my coat.

PR #112, opened 2015-03-11
Introduce Ecosystem extension

langner:


drvinceknight:
These plots look ridiculously cool. Will get to reviewer this asap (because it's awesome). Will probably have questions.

axelrod/ecosystem.py
```diff
@@ -0,0 +1,38 @@
...
+            pops = self.populations[-1]
+
+            # The unit payoff for each player in this turn is a sum of the payoffs
+            # obtained from all other players scaled by the size of the opponent population.
+            payoffs = [sum([self.pmatrix[ip][jp] * pops[jp] for jp in plist]) for ip in plist]
+
+            # In our case lower payoffs are better, so fitness should increase when
+            # the payoff descreases. This is a simple linear function that fulfills
+            # this and still behaves nicely, but we could chooise more complicated forms.
+            pmax = 5 * self.nplayers
```
drvinceknight: Just skimming through for now: should this 5 be a variable?

axelrod/plot.py
```diff
@@ -59,3 +61,33 @@ def payoff(self):
...
+            pops = [[populations[iturn][ir] for iturn in turns] for ir in self.result_set.ranking]
+            ax.stackplot(turns, *pops)
+
+            ax.yaxis.tick_right()
+            ax.yaxis.set_label_position("right")
+            plt.ylim([0.0, 1.01])
+            plt.ylabel('Relative population size')
+            plt.xlabel('Turn')
+            plt.title("Strategy population dynamics based on average payoffs")
+            trans = transforms.blended_transform_factory(ax.transAxes, ax.transData)
```
drvinceknight: Never seen this before: what does it do?

run_tournament.py
```diff
@@ -55,7 +55,17 @@ def run_tournament(turns, repetitions, exclude_basic, exclude_strategies,
         # This is where the actual tournament takes place.
         results = tournament.play()
 
-        # # Save the scores from this tournament to a CSV file.
+        # This is the "ecological" variant that uses the results of the above.
```
drvinceknight: Again, just skimming. Is there a cli argeparse option for running this or not?

axelrod/ecosystem.py
```diff
@@ -0,0 +1,38 @@
...
+            fitness = [pmax - p + 1 for p in payoffs]
+
+            # The new populations should be multiplied by something that is proportional
+            # to the fitness, but we are normalizing anyway so just multiply times fitness.
+            newpops = [p * f for p, f in zip(pops, fitness)]
+            norm = sum(newpops)
+            newpops = [p / norm for p in newpops]
+
+            self.populations.append(newpops)
```
drvinceknight: Were you thinking of perhaps adding in some randomness (mutations) so that we could assess the stability of the equilibria?

This could be a further addition, later down the road.

axelrod/plot.py
```diff
@@ -59,3 +61,33 @@ def payoff(self):
             return figure
         else:
             return None
+
+    def stackplot(self, populations):
+        if self.matplotlib_installed:
+            figure, ax = plt.subplots()
+            turns = range(len(populations))
+            pops = [[populations[iturn][ir] for iturn in turns] for ir in self.result_set.ranking]
+            ax.stackplot(turns, *pops)
```
drvinceknight: Does stackplot automatically sort these by the final score or is this happening somewhere else? (The output looks sorted but perhaps that's not happening).

axelrod/ecosystem.py
```diff
@@ -0,0 +1,38 @@
+class Ecosystem(object):
+    """Create an ecosystem based on the payoff matrix from an Axelrod tournament."""
+
+    def __init__(self, pmatrix):
+
+        self.pmatrix = pmatrix
```
drvinceknight: What `pmatrix` is is not clear to me.

axelrod/ecosystem.py
```diff
@@ -0,0 +1,38 @@
...
+        self.nplayers = len(self.pmatrix)
+
+        # Population histories will be recorded in a nested list,
+        # with each internal list containing a list of populations
+        # for all strategies. The first list, representing the
+        # starting populations, should have all equal values.
+        self.populations = [[1.0 / self.nplayers for i in range(self.nplayers)]]
```
drvinceknight: I think I'd prefer this to be `self.population_sizes`.

axelrod/ecosystem.py
drvinceknight: Ah: you're not re-running the tournament each time... You're running it once and using the outcome... (Currently throwing that around in my head... Not what I would have done off the bat...)

drvinceknight: Could we pay for the couple of extra characters and call that `payoff_matrix`?

axelrod/ecosystem.py
drvinceknight: Why `pmax`? Variable name makes me immediately think of max payoff... Perhaps a bit more verbose?

axelrod/tournament.py
```diff
@@ -42,4 +43,6 @@ def play(self):
             for i in plist:
                 for j in plist:
                     self.result_set.results[i][j][irep] = payoffs[i][j]
+
+        self.result_set.init_output()
```
drvinceknight: What is this for?

axelrod/tests/test_ecosystem.py
```diff
@@ -0,0 +1,27 @@
...
+        self.assertEquals(len(eco.populations[0]), 10)
+        self.assertAlmostEqual(sum(eco.populations[0]), 1.0)
+
+    def test_reproduce(self):
+        """Does the reproduction mechanism work as expected?"""
+
+        eco = axelrod.Ecosystem(self.zeros)
+        eco.reproduce(100)
+        self.assertEquals(len(eco.populations), 101)
+
```
drvinceknight: Could you please add a test for a full output (of course for small number of strategies and turns). This is just to ensure that changes further down the line don't break things.

Am I not at all convinced that `test_reproduce` does this comprehensively.

axelrod/plot.py
```diff
@@ -59,3 +61,33 @@ def payoff(self):
             return figure
         else:
             return None
+
+    def stackplot(self, populations):
```
drvinceknight: Could these be tested please (realise you've written a lot and this is a truly awesome contribution so please let me know if you'd like me to take some of this on when I get a chance).

axelrod/ecosystem.py
langner: It could be, but it is the function used to get fitness which is the main unknown. So it would make most sense to make that a variable. I don't know what the "best" choice there is, but I think it won't really matter and will affect only kinetics. As long as the function is monotonic. I could be wrong!

drvinceknight:
A really awesome PR @langner , so much cool stuff in it. Have commented on various parts of the code where I would prefer a couple of things to be clarified.

This is an awesome contribution: can't wait to get it in.

axelrod/ecosystem.py
langner: > Why pmax? Variable name makes me immediately think of max payoff... Perhaps a bit more verbose?

Yes, the meaning is the maximum possible payoff. If a strategy looses with everyobdy else, it's payoff will be this number.

drvinceknight: Cool, let's keep it as it is for now, perhaps we could open an issue (more as a bookmark than anything else) to look in to this more? (Perhaps throw an inline comment in here also)

langner: > Cool, let's keep it as it is for now, perhaps we could open an issue (more as a bookmark than anything else) to look in to this more? (Perhaps throw an inline comment in here also)

No, I wast too hasty and this is in fact not what I wanted. I wanted the payoff from each round to be in in the range 0...pmax, but forgot the payoffs are already scaled by the opponents population, so the range will be pmax-5...pmax. I will simplify this.

drvinceknight: I was just reading your other comment again and I still don't think I have my head around this (even with pmax-5...pmax). I might need to scratch my head a bit.

langner: The point is to transform the score of a player into a number that would describe it's fitness, which in turn determines how fast it reproduces. I just wanted to determine the range of the score by calcualting the maximum payoff. It does not matter that much, thought, because the specific form of this function should not influence the population dynamics or the final equilibrium.

axelrod/plot.py
langner: Transforms in matplotlib allow one to do nonstandard operations on various coordinate systems and to combine them. A tutorial: http://matplotlib.org/users/transforms_tutorial.html. In this case I used it to create extra tick marks on the left side for the strategy names.

run_tournament.py
langner: Not currently, no.

axelrod/ecosystem.py
langner: Sure, noise of some sort could be added.

axelrod/plot.py
langner: No, the list comprehension is over `result_set.ranking` in order to get the familiar order.

axelrod/ecosystem.py
drvinceknight: AAAAAAAhhhhhhhh

I'm with you. The five comes from the payoff matrix of the PD. I understand (well now understand what pmax is at least).

axelrod/ecosystem.py
langner: > Ah: you're not re-running the tournament each time... You're running it once and using the outcome... (Currently throwing that around in my head... Not what I would have done off the bat...)

Exactly. There is no point in running it every time, since the average payoffs will be the same approximately. An efficient option would be to also calculate payoff variances and generate a payoff from a normal distribution at each turn.

axelrod/plot.py
drvinceknight: I think it could be nicer to sort over the final populations. Could be (especially when we include possibility of mutation) that this does not coincide with the same order.

axelrod/ecosystem.py
langner: Yeah! :) I thought that was obvious somehow.

axelrod/ecosystem.py
drvinceknight: Cool, once I merge this PR in, I'll add an issue and that could be added at another time (this is already very cool as it is and does loads!).

axelrod/ecosystem.py
drvinceknight: I think it is now (perhaps I'm hungry...). Throw in a short in line comment perhaps?

`pmax = 5 * self.nplayers  # maximum payoff against every player`

axelrod/ecosystem.py
drvinceknight: Yeah: this is an awesome way of doing this. This whole way of doing it (including the rest of it) is very elegant. Not sure my applause carries much weight but :clap:

axelrod/tournament.py
langner: I found it necessary to get the payoff matrix generated before plotting begins.

axelrod/ecosystem.py
langner: Thanks! But the fun I have is my real reward.

axelrod/tournament.py
drvinceknight: Hmmm, could this be called by the ecosystem perhaps?

run_tournament.py
drvinceknight: Could you throw that in to this PR?

langner:
I think I added everything you mentioned. If not, let me know. Sampling payoffs instead of taking the average does not change anything. Much more noise would be needed to destabilize the dynamics.

drvinceknight:
You're a machine :) Just checking it out now.

axelrod/ecosystem.py
```diff
@@ -0,0 +1,59 @@
+
+import random
+
+
+class Ecosystem(object):
+    """Create an ecosystem based on the payoff matrix from an Axelrod tournament."""
+
+    def __init__(self, results, fitness=None, ):
```
drvinceknight: Any reason there is a `, )` at the end here?

axelrod/ecosystem.py
```diff
@@ -0,0 +1,59 @@
...
+        # representing the starting populations, should have all equal values,
+        # and all population lists should be normalized to one.
+        self.population_sizes = [[1.0 / self.nplayers for i in range(self.nplayers)]]
+
+        # This function is quite arbitrary and probably only influence the kinetics
+        # for the current code.
+        if fitness:
+            self.fitness = fitness
+        else:
+            self.fitness = lambda p: 5.0 / (5.0 + p)
```
drvinceknight: Sweet.

axelrod/ecosystem.py
```diff
@@ -0,0 +1,59 @@
...
+            # obtained from playing with all other players, scaled by the size of the
+            # opponent's population. Note that we sample the normal distribution
+            # based on the payoff matrix and its standard deviations obtained from
+            # the iterated PD tournament run previously.
+            payoffs = [0 for ip in plist]
+            for ip in plist:
+                for jp in plist:
+                    avg = self.payoff_matrix[ip][jp]
+                    dev = self.payoff_stddevs[ip][jp]
+                    p = random.normalvariate(avg, dev)
```
drvinceknight: This wasn't what I meant by the noise/mutation. I'm happy for you to revert this back to where it is (does no harm though so we can leave it in I suppose... Your call.

(Happy to discuss what I meant but I think we can just leave this PR as deterministic).

langner: I think I know what you meant. But this is good to have anyway, I think. It's not much computation, and is more realistic for nondeterministic strategies.

drvinceknight:
Yeah, this looks brutally awesome. Could you remove the normal distribution? This wasn't what I meant by mutations (although might be equivalent to what I meant), perhaps easiest to just remove it and keep this deterministic.

We can open an issue about mutations/death rates etc and discuss that there.

This is such a cool addition. :+1: :clap: 

I'm working on docs and documenting this will be quite a nice job thing to do (I've flying to CA on Saturday so will make use of my plane time :D).

axelrod/ecosystem.py
drvinceknight: Cool, ignore my comment on the PR: let's keep it in.

langner:
> Could you remove the normal distribution? This wasn't what I meant by mutations (although might be equivalent to what I meant), perhaps easiest to just remove it and keep this deterministic.

The normal distribution has zero standard deviation for deterministic strategies. It is there only to simulate the randomness of stochastic strategies, by sampling a normal distirubtion defined by their average payoff and the corresponding stddev. I doesn't change much, but also doesn't cost anything. One could imagine a tournament where it does make a difference, though (for example one involving only very randomizing strategies).

drvinceknight:
Yup: happy with that - really great job with this :D

PR #113, opened 2015-03-13
Creates TournamentManager class

meatballs:
and uses it in run_tournament.py

Ground work for #68


axelrod/tournament.py
```diff
@@ -8,9 +8,9 @@ class Tournament(object):
 
     game = Game()
 
-    def __init__(self, players, game=None, turns=200, repetitions=10):
-        """Initiate a tournmanent of players"""
-
+    def __init__(self, name, players, game=None,
```
drvinceknight: I think a default is needed so that not all tournaments need to be named.

meatballs: Good call - on it's way

drvinceknight: If it could be tested also...

drvinceknight:
Could you throw in a test for  the name and default name if none is passed...

PR #116, opened 2015-03-19
Provides optional parallel running of repetitions

meatballs:
* Adds run_serial_repetitions and run_parallel_repetitions methods to Tournament
* Adds new argument to run_tournament (-p or --processes) to define number of parallel processes to run. The default is serial processing. Actual number used is limited to the cpu_count.
* Refactors playing of round robin and updating of result set out of Tournment.play and into separate methods
* The result set update from above now lives within ResultSet.finalise (renamed from ResultSet.init_output)
* Passes logger from TournamentManager to Tournament to log number of processes in use
* Updates test suite to match

And possibly other things I've forgotten.

And possibly not some things I've overlooked!


meatballs:
On my laptop, running all tournaments serially takes approx. 90s. Running with -p 0 takes this down to approx. 75 seconds

meatballs:
That's me done fiddling with this one. It's ready to roll.

meatballs:
I'm finished with this one. It's ready to roll.

drvinceknight:
@JasYoung314 can you take a look at this one.

meatballs:
According to various online sources (e.g. http://stackoverflow.com/questions/10037319/how-do-i-tell-if-pythons-multiprocessing-module-is-using-all-of-my-cores-for-ca) python's multiprocessing.cpu_count will return double the actual number of cores for Intel CPUs with hyper threading.

On my Macbook Air, 'About my Mac' shows it has an Intel I7 with 2 cores. cpu_count returns 4.

Running run_tournament with -p 0 launches 4 processes and completes in approx. 75 seconds (as quoted previously). Running with -p 2 reduces this further to 70 seconds.

JasYoung314:
I've just taken a look over it: Looks great!

It's very similar to a way I had previously implemented parallel processing.

Looking through the code, should line 74 in the tournament.py file be 

```python
if self.processes < 2 or self.processes > multiprocessing.cpu_count():
```

I've left a comment in the commit itself.

axelrod/tournament.py
```diff
@@ -18,31 +18,100 @@ def __init__(self, players, name='axelrod', game=None,
...
+
+    def run_parallel_repetitions(self, payoffs_list):
+        # At first sight, it might seem simpler to use the multiprocessing Pool
+        # Class rather than Processes and Queues. However, Pool can only accept
+        # target functions which can be pickled and instance methods cannot.
+        processes = []
+        work_queue = multiprocessing.Queue()
+        done_queue = multiprocessing.Queue()
+
+        if self.processes < 2 or self.processes > multiprocessing.cpu_count:
```
JasYoung314: Should this be 

```python
if self.processes < 2 or self.processes > multiprocessing.cpu_count():
```

meatballs:
> Looking through the code, should line 74 in the tournament.py file be

> `if self.processes < 2 or self.processes > multiprocessing.cpu_count():`

Yep -that's a typo! Presumably I got away with it simply because it was treated as a property accessor.

JasYoung314:
It's actually because in python, certain types have a notion of ordering. It's arbitrary but consistent, so you could compare the function `cpu_count` to the integer `self.processes` and have the following results

```python
>>> multiprocessing.cpu_count()
4
>>> 5 < multiprocessing.cpu_count
True
>>> 5 > multiprocessing.cpu_count
False
```

because the type integer is always considered to be less than the type function.

Pull request seems awesome!

meatballs:
Thank you @JasYoung314 for both the explanation and the compliment!! 

(Is now a good time to own up to the fact that I'm only really doing this because I've not used Python before and fancied an excuse to learn)?

JasYoung314:
(I think the time to own up was before you started messing with multiprocessing! Now you're one of us I'm afraid)

drvinceknight:
Just going through it now: this is very nice to see!

drvinceknight:
I added something to the README also. Thanks for this @meatballs : huge contribution.

drvinceknight:
Thanks a lot @JasYoung314 for reviewing this.

PR #120, opened 2015-03-24
Sorts out use of cache and adds tests

meatballs:
This doesn't yet include the sharing of the cache across tourmanents or the pickling.

It simply makes the use of the cache consistent by properly using the same object in all cases.


meatballs:
Something went wrong in Travis when it attempted to download matplotlib. I'll close this PR and create a new one to kick that off again.

PR #121, opened 2015-03-24
Introducing some superior strategies

langner:


drvinceknight:
These look very cool!

Issue #117, opened 2015-03-20
Refactor deterministic cache mechanism


meatballs:
* Ensure that the mechanism is used consistently across all Classes which utilise it
* Write some tests to ensure it works correctly
* Share it across multiple tournaments within the TournamentManager Class

Depends on #116

meatballs:
Another thought:

We could pickle the cache out to a file and load it into the TournamentManager each time. That way, we wouldn't have to run the first round robin before kicking off the parallel processes. We could just run all repetitions in parallel and even the first would have access to the cache.

We'd need to remember to rebuild and re-pickle each time a new strategy is added but, even if we forget, it's not a problem.

meatballs:
I'll pick this up. It will be on branch 117 in my fork.

meatballs:
I've hit a snag on this one and could do with some help:

I've refactored the mechanism to ensure that there is only ever one cache object which is then passed correctly to the various classes that use it. There are some temporary logging messages which show that the same object id is being used throughout.

That all works fine right up until the last commit (f5689b8) where the cache is shared across tournaments by the tournament manager.

As hoped, run_tournament executes slightly quicker than before the commit, but always crashes on me  at the very end. The results are produced correctly, but I always have to kill the process to make it exit. This is true regardless of whether I run in parallel or not.

The change is on line 47 of tournament_manager.py. Commenting it out makes run_tournament execute without crashing. 

The cache counts in the log messages show the expected numbers both with that line and without.

Can anyone else confirm that they see the same behaviour for me?
Anyone got a clue as to what is going on?!!

drvinceknight:
Will take a look this pm, seems weird though :)

langner:
I ran the code without problems.

meatballs:
> I ran the code without problems.

Thank you!! I think that's good news!! I did wonder if it's something obscure on my machine or my python installation

meatballs:
Further good news - pickling the cache to a file and loading it into the tournament manager has now reduced running time on my machine from 70s to 47s!!

meatballs:
And not bothering with the first round robin in serial if the cache is already loaded gets it down to 25s

drvinceknight:
Ran just fine for me also.

> Further good news - pickling the cache to a file and loading it into the tournament manager has now reduced running time on my machine from 70s to 47s!!

How does pickling the cache speed things up? Is everything not just in memory anyway? I think I don't understand some of this...

meatballs:
[...]

PR #122, opened 2015-03-25
Refactors deterministic cache mechanism and adds test

meatballs:
Work on #117 

This doesn't yet include pickling of the cache.

It makes the use of the cache consistent by properly using the same object in all cases and also adds an option to TournamentManager to share the cache between tournaments.


meatballs:
Good to go. I'm done fiddling with it for now.

Issue #125, opened 2015-03-25
Parallelisation hangs


drvinceknight:
As far as I can tell this is something to do with #122 but I have no idea. 

It hangs at this stage:

```
➜ Axelrod git:(master) ✗ python run_tournament.py -p 3
Starting basic_strategies tournament with 10 round robins of 200 turns per pair.
Passing cache with 0 entries to basic_strategies tournament
Running repetitions with 3 parallel processes
Finished basic_strategies tournament in 0.0s
Starting ecological variant of basic_strategies
Finished ecological variant of basic_strategies in 0.0s
Cache now has 10 entries
Finished all basic_strategies tasks in 1.3s

Starting strategies tournament with 10 round robins of 200 turns per pair.
Passing cache with 10 entries to strategies tournament
Running repetitions with 3 parallel processes
````

langner:
Yup, also hangs here. Only for standard strategies, cheaters and basic go through fine. I am not versed well enough with the code in #122 to tell what the matter is.

langner:
Now I see you are using `multiprocessing`. My first guess: deadlock caused by a large queue.

drvinceknight:
<scratches chin> let me tag @meatballs just in case he hasn't seen this...

meatballs:
That is precisely the problem!

[image: Empiria Ltd] <http://www.empiria.co.uk/>Owen Campbell

+44 (0)1928 238150 <+44%20(0)1928%20238150>
+44 (0)7799 111652 <+44%20(0)7799%20111652>

[image: LinkedIn] <http://www.linkedin.com/in/owencampbell> [image: Twitter]
<http://twitter.com/opcampbell/> [image: GooglePlus]
<http://plus.google.com/+OwenCampbell1/> [image: Bitbucket]
<http://bitbucket.org/meatballs> [image: Github]
<http://github.com/meatballs>Empiria Ltd is a Private Limited Company
Registered in England and Wales No: 4044182
Registered Office: C127, The Heath, Runcorn, Cheshire WA7 4QX UK

On 25 Mar 2015, at 19:23, Karol M. Langner <notifications@github.com> wrote:

Now I see you are using multiprocessing. My first guess: deadlock caused by
a large queue.

—
Reply to this email directly or view it on GitHub
<https://github.com/drvinceknight/Axelrod/issues/125#issuecomment-86181092>.

meatballs:
[...]

PR #126, opened 2015-03-25
Introducing meta players

langner:
This abstracts the core of MetaHunter into an abstract MetaPlayer and adds some more involved metaplayers. Depending on the interpretation, these meta players could be regarded as cheaters, since they load other strategies at runtime. They do not, however, modify strategies or run separate tournaments, just record the running scores these strategies would have had in the current game. In princple, the same could be written explicitely without importing the strategies (alhtough it would be much longer). Anyway, these players are not particularly good (I though MetaWinner would do somewhat better, though).


drvinceknight:
Yeah, I agree: I don't think these count as cheaters. They're just taking very detailed notes throughout the tournament.

langner:
They also boost the walltime of the tournament... creating more optimization challenges! :)

drvinceknight:
Have you seen #125 ? I seem to be having problems with the parallelised version of the tournament now...

drvinceknight:
(That's not directly relevant to this PR but just pinging you in case you have the same problem. Heading out the door now but will take a look at this shortly.)

drvinceknight:
This is looking nice.

How much longer is the run time?

Also, could you tweak the contribution guidelines in the docs for the change from `__init__` to  `_strategies`.

langner:
The whole thing now runs in about 100s on my machine.

Issue #130, opened 2015-03-29
Test (and improve) the packaging


drvinceknight:
So there is now a packaged version of `axelrod` available via `pip`. This is all from the `packaging` branch and this issue is to discuss any issues before merging in to master and making things 'official' :)

To test, I suggest creating a virtualenv as this will install 'system wide':

```
virtualenv tmp-axelrod2 # I had about 7 of these going as I as testing and am just about to tear them down
cd tmp-axelrod2 # Go to that directory
source bin/activate  # (depending on your shell this might need a tweak)
```

Type the following just to make sure that you have got a virtualenv:

```
which pip
```

Install through the magics of the internetz:

```
pip install axelrod
```

Note that that pulls down `matplotlib` also (which if you're in the virtualenv you will notice).

Now to run the tournament I have made some tweaks (see the `packaging` branch).
Firstly I've renamed the `run_tournament.py` script to `run_axelrod` as `pip` now installs this as an executable file system wide.

So below will run the very basic tournament:

```
run_axelrod --xs --xc --xa --ne
```

One thing that will need work on this branch before pulling in to `master` will be updating the documentation.

drvinceknight:
One of the other changes that I've made is that output is by default in the current directory.

(Owen: I think we need to have a command line argument that points at a `cache.txt` file.)

drvinceknight:
(I also don't like that the authors are just down as me, does anyone know what the done thing is here?)

meatballs:
http://stackoverflow.com/questions/9999829/how-to-specify-multiple-authors-emails-in-setup-py

meatballs:
> I suggest creating a virtualenv as this will install 'system wide':

In case it helps anyone else, this repo: https://github.com/meatballs/dotfiles contains my dotfiles setup which will download and configure the zsh shell with a heap of useful plugins - one of which will auto-activate a virtualenv when you cd into the directory (and then show the active environment in the prompt)

drvinceknight:
Have added @langner and @meatballs as authors: https://pypi.python.org/pypi/Axelrod

Will need to clean up the readme or write a `long_description.rst` file at some point...

drvinceknight:
@meatballs if you're happy (I just merged #134) with the general structure of this, I'll go ahead and start working on a proper PR with updated docs etc... (and we can announce it once that's in :)).

meatballs:
[...]

PR #134, opened 2015-03-31
adds additional email addresses

meatballs:
#130

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,10 +4,7 @@ setup(
     name='Axelrod',
     version='0.0.5',
     author='Vince Knight, Owen Campbell, Karol Langner',
-    author_email=(
-        'Vince Knight <vincent.knight@gmail.com>, '
-        'Owen Campbell <owen.campbell@tanti.org.uk>, '
-        'Karol Langner <karol.langner@gmail.com>'),
+    author_email=('axelrod-python@googlegroups.com'),
     packages=['axelrod', 'axelrod.strategies', 'axelrod.tests'],
     scripts=['run_axelrod'],
     url='http://axelrod.readthedocs.org/',
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,10 @@ setup(
     name='Axelrod',
     version='0.0.5',
     author='Vince Knight, Owen Campbell, Karol Langner',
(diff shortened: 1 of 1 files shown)


drvinceknight:
I read somewhere about the multiple email address thing and the suggestion was to just have one (to ensure there was no confusion with some messages being sent to some people and not all at times). I think this makes a lot of sense. 

The suggestion was perhaps to create an email list but I thought this was potentially overkill for this (modest) so I was happy to be the central email and just relay but perhaps I'm wrong and setting up a Google group is not a bad idea (costs nothing). 

I don't think having multiple email addresses there is a good thing though (not at all to not pass on credit or anything -  just to make sure it's clear what to do if someone wants to get in touch...).

meatballs:
No problem - I was really only looking for something simple to see if we could usefully use the idea of creating PRs against branches other than master. Feel free to dump it!

drvinceknight:
What do you think though? I like the idea of a Google group? No harm
right... (have setup quite a few, should have probably just done it for
this)...

I think the PR thing works well for this sort of stuff.

meatballs:
> I like the idea of a Google group?

Why not? We've used just about ever other tool out there so far, so why not a Google Group?!!! Seriously, it seems a perfectly good way to create a mailing list to me.

drvinceknight:
Cool, can you create it? (I'm on a train with rubbish signal but if awkward
I can set it up later :))

You'll be able to add Karol and I as managers or something like that but
who cares as long as it's public and anyone can email it and read it :)

drvinceknight:
Thanks for creating the group, you seem to have created a private one? Should it not be a public one that everyone can see? (You know for when millions of people use this... :confused: )

meatballs:
It's public in that anyone can write to it but private in that only the three of us are members.

I thought the intention was simply to provide an address to PyPi for the authors of the package, rather than to create an open community for discussion. Surely we already have github issues and gitter for that?

drvinceknight:
Ah fair dues: I didn't realise anyone could write to it. Let me attempt to
send an email to see what happens :)

Vince

drvinceknight:
Just seen that you've done that... Now to go and figure out why I did not
see the email...

PR #137, opened 2015-04-01
Adding to waffle

drvinceknight:
Just adding a badge and a graph to the readme file which points at https://waffle.io/Axelrod-Python/Axelrod

I don't know how useful this is but it also throws in a plot about PRs and issues.


meatballs:
It's very useful!

PR #139, opened 2015-04-01
Packaging drvinceknight/axelrod/#136

drvinceknight:
I think this is now ready.

Once this is merged we can create a version tag on github :+1: (I don't exactly know what that does).

Note that the email @meatballs put in isn't immediately visible to me on PyPi? Am I missing something?


meatballs:
README has disappeared from PyPi

is that because it's been renamend to .rst but not included in MANIFEST.in?

drvinceknight:
Yeah I pulled it out. The readme that was there didn't work well on pypi (it referenced images etc)...

I also took a look at some other projects (Django for example) and they don't have a long description on pypi. I thought it was simpler to just remove it.

Simply removing from `setup.py` didn't do the trick as pypi automatically reads `readme.txt` so I then renamed the file (to indeed try and keep things away from pypi). `MANIFEST` is an automatically generated file so I wouldn't touch it.

If we do want a long description I would suggest writing a 'for pypi' one (I'll do it) but I don't think it's necessary. I think keeping pypi short and to the point is fine (I can't say I had ever looked through those pages before: I just find on github and `pip install`).

The one thing I wasn't sure of though is whether to have url for website to go to readthedocs page or to github?

meatballs:
OK - I wasn't sure if it was deliberate or not. Seems fine to me to keep PyPi short and to the point.

I don't think it matters where the url points - they both reference each other so anyone who is interested can find what they need.

drvinceknight:
Cool. Any thoughts about the email address not appearing on PyPi?

meatballs:
> Any thoughts about the email address not appearing on PyPi?

When I look at PKG-INFO, I can see it fine (https://pypi.python.org/pypi?name=Axelrod&version=0.0.7&:action=display_pkginfo). Should I be expecting to see it somewhere else?

drvinceknight:
Ah perhaps not! My mistake I wasn't sure where this should be :)

drvinceknight:
How exciting that it is merged :) Should I set the version number on the git repository to have the 0.0.7 tag?

We could possibly open a milestone for 0.1.0?

meatballs:
already tagged it. There's now 1 release showing

drvinceknight:
Sweet :D

Issue #92, opened 2015-03-07
Improve how docstrings for strategies appear/are generated?


drvinceknight:
It must be possible to automate the generation of a sub header etc for each strategy?

Currently in the `strategies.rst` file I have manually listed all of them. I can't seem to be able to find how to automate this...

(Here's the currently rendered documentation: http://axelrod.readthedocs.org/en/latest/)

All files are written using sphinx in rst and can be found in the `docs` directory.

langner:
What do you mean by sub header?

drvinceknight:
Sorry, I didn't really explain this well. 

So at the moment: http://axelrod.readthedocs.org/en/latest/strategies.html shows every strategy's docstring listed out.

That is how an excerpt of how this is done in: https://raw.githubusercontent.com/drvinceknight/Axelrod/master/docs/strategies.rst

```
Here are some of the basic strategies:
--------------------------------------

.. currentmodule:: axelrod.strategies
.. autoclass:: Alternator
.. autoclass:: Cooperator
.. autoclass:: Defector
.. autoclass:: Random
.. autoclass:: TitForTat

A list of all further (honest) strategies:
------------------------------------------

.. autoclass:: AntiTitForTat
.. autoclass:: Appeaser
```

So I'm listing each strategy manually, I think I must be missing some sphinx but can't seem to find it: I'd like to just be able to say 'Ok list all the class from `axelrod.strategies`' here. 

*Also* if possible if they could each have their own subheader (`<h4>Alternator</h4>` or something like that) that is slightly nicer than what currently appears (`class axelrod.strategies.Alternator`) that would be cool although perhaps that's not needed... Just trying to think about readability for none coders but maybe that's not important and this isn't exactly 'not readable'. As I've written this I think that that is perhaps really not needed...

langner:
You can use `automodule`, but then you can't easily break the list into the section you have.

drvinceknight:
I tried `automodule` but without success. Let me try again...

langner:
You could have a Python script generate the `rst` file, then you can do whatever you want :D

drvinceknight:
Yeah I think `automodule` picks up doscrtrings for given submodule but not for classes perhaps? Unless I'm doing something wrong: e6d71cc

> You could have a Python script generate the rst file, then you can do whatever you want :D

[...]

PR #147, opened 2015-04-11
92 strategy docs

drvinceknight:
This is relevant to #92. 

I started wanting to change how things were done but ended up just doing some spring cleaning.

- I set the `Makefile` back to how it was before @langner's cool fix which sadly didn't work with readthedocs. I think we'll just have to run `python strategies.py > strategies.rst` when a new strategy gets included (this could also be added to the documentation...)
- I also added running the `strategies.py` script to travis (for example a change in the `_strategies.py` file actually broke this script). It doesn't do anything except check that it runs without any errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,4 @@ before_install:
 install: "pip install -r requirements.txt"
 script:
   - python -m unittest discover axelrod/tests/
+  - cd docs/; python strategies.py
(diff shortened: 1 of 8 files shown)


drvinceknight:
Whether or not this completely addresses #92 is an open question but either way this stuff should still be merged in (as it fixes things that were broken).

langner:
OK. I'm not familiar enough with readthedocs to know why the Python call does not work from the Makefile.

Issue #148, opened 2015-04-11
Major change suggested: minimisation -> maximisation


drvinceknight:
I think I made an initial mistake with the setup. I would like to make the default game be more in line with a lot of the literature I have been reading recently. This is one in which the players aim to maximise payoffs and not minimise (so the game values change in term).

The payoff matrix would become:

+-----+-----+
|(3,3) |(0,5)|
+-----+------+
|(5,0) |(1,1)|
+-----+------+

The main reason to do this change would be to be in line with the literature if and when we wanted to write an academic paper.

I do not feel that this would be a lot of work (as everything has been built in a modular way and nothing actually takes count of the score until the very end):

- Change payoff matrix in game
- Change mapping function in ecological variant
- (potentially change ordering of the strategies as displayed by plots in `run_axelrod`)

Let me know what everyone thinks.

langner:
This baffled me when I looked at the code, although in principle it makes no difference.

drvinceknight:
Yeah it would indeed make no difference. It's just a question of esthetics
and convention...

drvinceknight:
@meatballs can you let me know what you think about this? Hoping to find some time over the next couple of days to put this in place but just want to make sure that you don't have any big opposition to it.

meatballs:
I'm out and about today but should be able to have a look tomorrow

drvinceknight:
No problem :) (No code to look at, just your opinion on the idea of changing from min to max)

meatballs:
Can't see any problem with this at all

drvinceknight:
Cool: I'll get working on it over the next couple of days.

meatballs:
Might need to put a comment on Artem Kaznatcheev's blog once this goes in - his stuff explaining the differences will be out of date

drvinceknight:
Yeah, I was thinking that :) I've just opened the PR so as soon as we've
worked through it (I might have missed some stuff) one of us should do that
:)

Issue #102, opened 2015-03-08
Include the two missing memoryone strategies (already written: just commented out - need tests etc...)


drvinceknight:
@marcharper wrote a big pull request that took me a little while to merge as it needed tests etc...

I ended up commenting out two of the strategies: fa579594cf8e285a05a0cf4808d1133aaebdb771

It would be great to get them in, it just needs some tests and a PR that clarifies a couple of things.

marcharper:
They are both zero determinant strategies. ZDChi is correct as submitted but ZDGTFT2 needs to be fixed for your game matrix. I left it in since it did reasonably well.

See the following references:
http://www.pnas.org/content/109/26/10409.abstract
http://www.pnas.org/content/109/26/10134.short
http://www.nr.com/whp/StewartPlotkinExtortion2012.pdf

as well as this blog entry:

https://golem.ph.utexas.edu/category/2012/07/zerodeterminant_strategies_in.html

drvinceknight:
Thanks @marcharper, will get to it at some point. References are very helpful.

drvinceknight:
Hey @marcharper I've worked on this a bit (the first strategy): https://github.com/Axelrod-Python/Axelrod/tree/102-ZDChi

Note that reading through the original paper the assumption there does not hold here as the game here aims to minimize the utilities. So the condition T>R>P>S is actually inverted. Without spending a while looking at the paper do you think that this would have an effect on the values for p_1, p_2, p_3, p_4?

I note for example that the test I have added gives a value of four_vector the one for p_4 to be greater than 1.

Let me know what you think before I open a pull request to get this in the library, it wan't actually need much work now.

The second missing strategy though I have not gotten to at all yet.

drvinceknight:
- [File](https://github.com/Axelrod-Python/Axelrod/blob/102-ZDChi/axelrod/strategies/memoryone.py)
- [Test file (see line 200)](https://github.com/Axelrod-Python/Axelrod/blob/102-ZDChi/axelrod/tests/test_memoryone.py)

drvinceknight:
Have checked through paper (which is a really nice paper btw - I had not actually gotten around to reading it). This PR #144 incorporates the first strategy, I will make my way to the second when I get a chance.

marcharper:
I'm not sure that the conditions matter as long as the Press and Dyson determinant is used correctly. I can try to work out the correct values for ZDGTFT2, unfortunately that reference is a light on details.

[...]

PR #149, opened 2015-04-11
102 zdgtft2

drvinceknight:
This PR would address #102.

Both the strategies added relevant to #102 are designed for a minimisation version of the game. I started modifying the strategies so that they would work for this case BUT this I feel would mean they no longer correspond to the strategies they reference.

As such they don't perform too well but I've just raised this issue #148 which I think would make this all work (note that these two strategies is not the only reason for which I've opened #148 but more that after looking at all the relevant literature to these two strategies it feels like the right way to go).


langner:
I presume you want to merge this only after #148 is done.

drvinceknight:
Yeah I think it make sense to leave it. I probably shouldn't have opened
the PR...

drvinceknight:
I'm closing this.

PR #151, opened 2015-04-16
Some improvements to run_tournament.py

langner:


drvinceknight:
@langner , `run_tournament.py` has been replaced by `run_axelrod` (mainly for the packaging purposes). Perhaps that's where the impending merge conflict comes from? (I don't think you have pull from master...)

langner:
Exactly right. This should be better.

PR #152, opened 2015-04-16
148 min to max

drvinceknight:
This addresses #148 and should be the big change from minimisation to maximisation. I think I have caught everything:

- code
- tests
- docs
- assets


meatballs:
Looks good to me so far. Just having a poke around....

drvinceknight:
Please do! :)

PR #153, opened 2015-04-16
changes default repetitions to 10

meatballs:


drvinceknight:
Have you checked if this changes the results greatly (if 10 does not get rid of enough variation they might which would be a shame...)?

meatballs:
> Have you checked if this changes the results greatly (if 10 does not get rid of enough variation they might which would be a shame...)?

The results have always been run previously with 10 repetitions as that's the default on the tournament and tournament manager classes. this problem only came to light as a result of passing the repetitions argument from run_axelrod into those classes.

drvinceknight:
Ahhhhhhh. I remember the good old days, when I actually knew how this
package worked :)

Seriously though, this whole thing has been so much fun and ridiculously
cool to see it all grow so much and so fast. I really love open source... I
wish academic research would catch up faster...

PR #163, opened 2015-05-01
Docs update -- removed suspect Tit-for-Tat statement

marcharper:
Very simple change to README.rst as I'm almost certain that Tit-for-Tat should not be expected to "win" the tournament if only there were more strategies. I'm also highly sceptical of the claims in the literature of the superiority of e.g. ZD-strategies as they make rather questionable assumptions (weak selection, arbitrarily long games, using stationary scores instead of the real payoff matrices, etc.) My own work with Chris Lee and Dashiell Fryer shows that non-memory-one strategies can be very effective, and the MetaHunter strategy in this repository is another good example (and much more simple).

In fact, I encourage you all to publish these results, perhaps computing things like mean score per round as the number of rounds gets large and other quantities. The literature is in sore need of some empirical data.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,10 +54,6 @@ the results from the latest run of the tournament here:
 .. figure:: ./assets/strategies_boxplot.png
    :alt:
 
-As you can see: the 'tit for tat' strategy has not won in this instance,
-that is mainly because more strategies are needed to get anywhere near
-Axelrod's tournament.
-
 You can see the results from the latest run of the tournament here with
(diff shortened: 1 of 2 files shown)


langner:
I agree, some useful statistics could be extracted from running the tournaments.

langner:
> I encourage you all to publish these results

Well, I've created a repository to on a paper: https://github.com/Axelrod-Python/paper so we can get to it.

PR #164, opened 2015-05-01
Noisy tournaments

marcharper:
So lots of people have studied the effect of noise on the Prisoner's dilemma, in particular how it allows Tit-for-Tat to be overcome heads up. I added a variable to the Player class that randomly flips each play based on a background noise probability. You can see the difference it makes [here](http://imgur.com/sF921cX,zV44yuB), where first image is with 5% noise and the second is zero noise as usual.

There may be a better way to code this, say in the tournament class so that it's more controllable from the run scripts. This was just the easiest way to do it.


langner:
Interesting. I'm quite curious how this will influence the stability of the evolutionary trends.

marcharper:
You can see the difference it has on the pairwise payoffs [here](http://imgur.com/a/bVYnd), the top plot is with 5% error and the bottom with 0%.

meatballs:
Nice idea!

> There may be a better way to code this, say in the tournament class so that it's more controllable from the run scripts.

I think so. I'll sort that out, but it will probably be next week before I get to it.

drvinceknight:
Great addition. If @meatballs is happy to look through and tweak, I'm happy to review afterwards.

meatballs:
I've moved the work done so far into issue #166 (available on branch 166) so I'll delete this PR and create a new one once it's ready to roll.

PR #165, opened 2015-05-02
Zero Determinant Fixes and Overhaul

marcharper:
The formulas for ZDGTFT2 were not correct. Unfortunately the parametrization in the Stewart and Plotkin reference appears to be incorrect, however there is a newer reference [here](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0077886). I was able to track down what I believe S&P used in their tournament and updated the strategies accordingly (see [here](http://www.math.pitt.edu/~bard/classes/mth3380/hw3.pdf) ). Both of these are easily specified from the new parametrization.

Now that we have the correct parametrization of the ZD strategy vectors, it is only necessary to have one abstract ZeroDeterminant player class (MemoryOne subclassed) that can be further subclassed for ZD-Extort-2 and ZD-GTFT-2. I was also able to put in some parameter checks to make sure that the conditional probabilities actually lie between 0 and 1, which they were not guaranteed to before.

There was an error in the Game class in game.py -- the RPST function swapped two of the values.


marcharper:
The failed tests are "designed to fail" since they rely on events with probability not 0 or 1. I'm guessing that these were somehow whitelisted before and that I messed that up by changing the class names? Looks like the checks starting failing when I renamed Extort2 to ZDExtort2.

marcharper:
I'm going to break this into two requests.

PR #168, opened 2015-05-03
New Strategy Anti Tit For Tat

marcharper:
A strategy that always plays the opposite of the opponents previous move.

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -34,6 +34,7 @@ basic_strategies = [
 ordinary_strategies = [
     AlternatorHunter,
     Appeaser,
+    AntiTitForTat,
     ArrogantQLearner,
     AverageCopier,
     CautiousQLearner,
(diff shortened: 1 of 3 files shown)


langner:
Is the only different with SuspiciousTitForTat that the first move is C instead of D?

marcharper:
Actually suspicious Tit For Tat is incorrectly named. From [this reference](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0077886):

"Suspicious Tit-for-Tat (which starts with defection and then repeats the opponent’s previous move), Always Defect, and an Anti-Tit-for-Tat strategy (which always plays the opposite of the opponent’s previous move)"

So Suspicious TFT should be TFT but with a difference first move. But yes that is the only difference with the current STFT implementation. (It's an important difference though.)

marcharper:
I renamed the existing sTFT to Bully and added the version of sTFT as described in the reference (with tests).

marcharper:
If you run the tournament now you can see there is a not insignificant difference between Bully, sTFT, aTFT, and TFT

langner:
I think at some point the original ATFT got merged with BULLY, which were once separate. Anyway, this looks nice. I only did a small refactoring for these strategies, see marcharper/Axelrod#1

PR #169, opened 2015-05-03
Zd patch

marcharper:
The formulas for ZDGTFT2 were not correct. Unfortunately the parametrization in the Stewart and Plotkin reference appears to be incorrect, however there is a newer reference here. I was able to track down what I believe S&P used in their tournament and updated the strategies accordingly (see here ). Both of these are easily specified from the new parametrization.

Now that we have the correct parametrization of the ZD strategy vectors, it is only necessary to have one abstract ZeroDeterminant player class (MemoryOne subclassed) that can be further subclassed for ZD-Extort-2 and ZD-GTFT-2. I was also able to put in some parameter checks to make sure that the conditional probabilities actually lie between 0 and 1, which they were not guaranteed to before.

There was an error in the Game class in game.py -- the RPST function swapped two of the values.

The failed tests are "designed to fail" since they rely on events with probability not 0 or 1. I'm guessing that these were somehow whitelisted before and that I messed that up by changing the class names? Looks like the checks starting failing when I renamed Extort2 to ZDExtort2.

### DIFF
--- a/axelrod/game.py
+++ b/axelrod/game.py
@@ -9,13 +9,13 @@ class Game(object):
             ('D', 'C'): (t, s),
         }
 
-    def RPTS(self):
+    def RPST(self):
         """Return the values in the game matrix in the Press and Dyson notation."""
         R = self.scores[('C', 'C')][0]
         P = self.scores[('D', 'D')][0]
-        T = self.scores[('C', 'D')][0]
-        S = self.scores[('D', 'C')][0]
-        return (R, P, T, S)
+        S = self.scores[('C', 'D')][0]
+        T = self.scores[('D', 'C')][0]
+        return (R, P, S, T)
 
     def score(self, pair):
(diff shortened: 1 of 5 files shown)


drvinceknight:
> The failed tests are "designed to fail" since they rely on events with probability not 0 or 1. I'm guessing that these were somehow whitelisted before and that I messed that up by changing the class names? Looks like the checks starting failing when I renamed Extort2 to ZDExtort2.

What do you mean by whitelisted? I had written these tests as checks not that output was correct but that any changes would ping up a message. If you're happy with output you can simply put in the output results in to the tests (for each seed) and I can take a look at the output to make sure it fits with the required output?

Really awesome of you to work on this one :) :+1:

marcharper:
Oh I see now. I'll update this shortly.

drvinceknight:
I'll get to this today :+1:

PR #176, opened 2015-05-04
Tests refactor

marcharper:
There are some substantial patterns in the test suite so added three functions to TestPlayer that dramatically simplify the necessary test code, and to make it much easier to add new tests.

They are
* `first_play_test` which tests the first play (of course)
* `markov_test` which takes a list like [C,D,C,D] to test the four responses to CC, CD, DC, and DD
* `responses_test` which takes two histories e.g. [C,C] and [C,D] and tests the subsequent responses [C,D,...] however many you'd like.

I rewrote the tests for test_titfortat.py and test_memoryone.py and you can see it dramatically reduces the code involved. Incidentally, for many strategies like TFT, first_play_test and markov_test completely specify all the possible responses.


langner:
I thought about doing something like this before. Looks nice.

meatballs:
Nicely done!

meatballs:
Just having a look through all the changes....

meatballs:
I've tidied up some PEP8 errors in here. I'll close this PR and create a new one with the commits from here plus my own changes.

PR #181, opened 2015-05-06
Heads Up play tests

marcharper:
This PR adds a new class TestHeadsUp to test_player.py that takes two strategies and a list of expected rounds of play for testing. It's very similar to the last testing PR. Several examples are included in test_headsup.py and there's a short explanation of how to make more such tests on the contributions page.


meatballs:
Another very nice contribution. Just having a look through it...

Issue #170, opened 2015-05-03
Label Memory One players


marcharper:
Since whether a player has a long memory or not is such a big deal in the literature, it might be useful to label strategies as memory one (relying only on the previous round's plays) or not. Then we could compare the two classes of players in various circumstances, and this could be part of an eventual publication narrative.

I can go through and do this adding a member variable in the Player class like the current self.stochastic (I don't think that this can be easily determined directly from the code). But before I do so I'd like to know if this would likely be merged.

Alternatively we could group the strategies by class in "_strategies.py" but I'm thinking it may be better to have say a dictionary for each player (as a member variable) that has its classes, e.g. {"stochastic": False, "basic": True, "cheater": False, "memory-one": True}. Then the lists in "_strategies.py" can be automatically derived at runtime.

drvinceknight:
I like the idea of having a `memory_one` attribute so I'd be happy to merge it but I'm tagging the others to see what they think: @meatballs @langner

meatballs:
Sounds a nice idea. It could also from part of the selection of players in DjAxelrod.

drvinceknight:
Yeah: that would be very neat :+1:

langner:
Yeah, I also think it'd be useful. @marcharper can you give any literature reference that looks at this issue of memory one strategies in detail?

drvinceknight:
This is probably a good one (referenced in the doc string for the corresponding strategy in `memoryone.py`): http://www.pnas.org/content/109/26/10409.full.pdf

(In particular that paper has a great result in appendix showing that memory one strategies are as effective as others in a particular context.)

Was planning on putting a PR together later with a couple of references for https://github.com/Axelrod-Python/paper...

marcharper:
Just note that the literature makes some significant assumptions to reach the conclusion that long memory players are not as effective, and that in a "real" tournament we may not see such results.

I'll go through and tag the existing strategies and submit a patch.

langner:
Perhaps that is something to discuss in a paper. It seems our focus is more on the side of "real" tournaments...

PR #182, opened 2015-05-06
Memory-one labels Re: Issue #170

marcharper:
So I went through and tagged strategies as memory-one or not (with a boolean class member variable) giving either the four probabilities defining a strategy as memory-one (if it wasn't derived from the MemoryOnePlayer class explicitly) or an explanation of why the strategy is not memoryone.

The class var isn't used for anything yet.


meatballs:
Just looking through this at the mo....

PR #183, opened 2015-05-06
Noise argument to run_axelrod should be a float rather than an int

marcharper:


meatballs:
Ah, the perils of copy and paste! Thanks.

PR #184, opened 2015-05-06
Close matplotlib memory leak that results from running several tourna…

marcharper:
Close matplotlib memory leak that results from running several tournments and generating many figures. Otherwise eventually this happens:

/usr/lib/pymodules/python2.7/matplotlib/pyplot.py:412: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_num_figures`).


drvinceknight:
:+1:

Issue #187, opened 2015-05-07
Matplotlib formats and colorbars enhancement


marcharper:
The colorbars on the heatmaps can be shrunk to fit the vertical axis of the plot.

In TournamentManager, _save_plots now has a kwarg for the image format so that you can easily choose 'png' or 'svg'.

marcharper:
There are two commits on [branch 187](https://github.com/marcharper/Axelrod/tree/187) in my repository that handle each one above.

PR #189, opened 2015-05-07
Fixes #166

### LINKED ISSUE #166: Noisy Tournaments
Lots of people have studied the effect of noise on the Prisoner's dilemma, in particular how it allows Tit-for-Tat to be overcome heads up.

@marcharper  added a variable to the Player class that randomly flips each play based on a background noise probability.

You can see the difference it makes [here](http://imgur.com/sF921cX,zV44yuB), where first image is with 5% noise and the second is zero noise as usual.

You can see the difference it has on the pairwise payoffs [here](http://imgur.com/a/bVYnd), the top plot is with 5% error and the bottom with 0%.

This work now just needs to move into the Tournament class so it can be controlled from a calling program.

meatballs:
Noisy tournament stuff with docs and tests


meatballs:
Oh dear - looks like there's more to do here

meatballs:
anyone got any objections to this one? It would be good to finish it off so I can get on to #187

axelrod/tests/test_round_robin.py
```diff
@@ -46,6 +50,22 @@ def test_deterministic_cache(self):
         self.assertEqual(rr.deterministic_cache[(axelrod.Cooperator, axelrod.Defector)], (0, 100))
         self.assertFalse((axelrod.Random, axelrod.Random) in rr.deterministic_cache)
 
+    def test_noisy_cache(self):
```
drvinceknight: Could you walk me through what/how the `noisy_cache` is/works?

meatballs: the cache should not be updated if the tournament is noisy. this test checks that the cache is still empty after playing a noisy tournament even with otherwise deterministic players.

Issue #156, opened 2015-04-17
Tests for `run_axelrod`


drvinceknight:
We should have some unit tests for `run_axelrod`. Easy enough to test the function itself...

Issue #172, opened 2015-05-04
Tests needed for parallel running of the tournament


drvinceknight:
A bug crept past the tests (see #171) this is (apart from me missing it) due to the fact that the tests only check the running of the tournament in series.

PR #190, opened 2015-05-11
Fixes #186

### LINKED ISSUE #186: Rework run_axelrod
Rework run_axelrod so that run_tournaments can be imported and used in other scripts without needing to construct argparse Namespace objects for the arguments.

Rename run_axelrod to run_axelrod.py (otherwise python won't import it).

Add some long versions of argparse parameters (mainly for readability and the above remark).

Move logging outside of run_tournaments since otherwise the handlers accumulate.

originally submitted in PR #185

meatballs:


meatballs:
Fixes #156 too

drvinceknight:
Slowly working through this one.

- Why not have a `test_utils.py` set of tests as well? (A quick flick through tests and am not convinced this does indeed address #156)
- Are files (results, cache...) being created during the tests? This wasn't clear from my initial scan.

Back to my marking now but will look through things properly soon.

meatballs:
> Why not have a test_utils.py set of tests as well?

run_tournaments in utils.py is now simply calling methods that are tested elsewhere. I've concentrated on creating methods which can be unit tested in isolation rather than integration testing -  which is what I think testing of utils.py would entail.

I'm not saying we shouldn't do it, just that comprehensive unit testing is more important in my opinion.

> Are files (results, cache...) being created during the tests?

No

drvinceknight:
>> Why not have a test_utils.py set of tests as well?
> run_tournaments in utils.py is now simply calling methods that are tested elsewhere. I've concentrated on creating methods which can be unit tested in isolation rather than integration testing - which is what I think testing of utils.py would entail.

> I'm not saying we shouldn't do it, just that comprehensive unit testing is more important in my opinion.

Just after posting I realised that the tournament manager was being pulled in to utils and not the other way around... 

There have been some bugs introduced where things were changed and that broke things in `run_tournament` (when methods were changed to private) but they were not tested there. I was hoping we would avoid that happening again by improving the coverage, what do you think, is it overkill? I just imagine someone changing something in the tournament manager, tests passing and not realising that the utils file no longer works.

>> Are files (results, cache...) being created during the tests?
> No

Ok, re my previous point I wonder if it's worth having some tests that do create some output... (Obviously that gets removed)...

What do you think?

meatballs:
Most of the problems we've had have been to do with #172. I think that's the most important part of the test suite that we should work on. After that, we can look at some better integration testing - perhaps extending test_full_tournament

drvinceknight:
Is there any reason not to have some tests for the `utils.py` stuff? (Even if it's opened as an issue and not included in this PR).

drvinceknight:
I agree that the problems are due to #172 but them squeezing through the tests is because we did not have any integration tests for the `run_tournament`...

meatballs:
I don't agree - I think it was because we don't have unit testing coverage of the parallel processing methods

Again - I'm not saying we shouldn't do some integration testing but I think it's more important to increase the unit testing coverage.

I also think that's beyond the scope of this specific issue and ought to be another issue in its own right.

drvinceknight:
> I don't agree - I think it was because we don't have unit testing coverage of the parallel processing methods

Was it not that we changed a `method` to a private `_method` and tests didn't catch that that had changed or am I thinking of the wrong thing? Surely that could happen a bit too easily here...

> I also think that's beyond the scope of this specific issue and ought to be another issue in its own right.

Sure :+1: Once this is merged in, we'll just open an issue for that :)

I'll look through it properly in 4 scripts time (this is the official unit of measurement of time during marking season), I'm sue it won't take me long :)

drvinceknight:
All looks great to me. Nice big chunk of work :+1: 

Have opened #191.

PR #192, opened 2015-05-12
Fixes #172

### LINKED ISSUE #172: Tests needed for parallel running of the tournament

meatballs:


meatballs:
Hold off on this one. I've just had an idea of how to test the _start_workers method that only has a placeholder at present.

meatballs:
should be good to go.

axelrod/tests/test_tournament.py
```diff
@@ -29,28 +57,244 @@ def test_init(self):
...
+        tournament = axelrod.Tournament(
+            name=self.test_name,
+            players=self.players,
+            game=self.game,
+            turns=200,
+            repetitions=self.test_repetitions,
+            processes=2)
+        scores = tournament.play().scores
+        actual_outcome = sorted(zip(self.player_names, scores))
+        self.assertEqual(actual_outcome, self.expected_outcome)
```
drvinceknight: This doesn't actually check that 2 processes were spawned right? Just that the thing didn't crash? I'm not too sure how we'd test the former and the later is certainly needed, just checking.

meatballs: Correct. The closest we get is in test_start_workers. I'll add another comment there....

axelrod/tests/test_tournament.py
```diff
@@ -29,28 +57,244 @@ def test_init(self):
...
+            repetitions=self.test_repetitions)
+        tournament._start_workers(workers, work_queue, done_queue)
+
+        stops = 0
+        while stops < workers:
+            payoffs = done_queue.get()
+            if payoffs == 'STOP':
+                stops += 1
+        self.assertEqual(stops, workers)
+
```
meatballs: `test_start_workers` above doesn't count the number of processes spawned by `tournament.start_workers` as that's well nigh impossible.

However, `tournament.start_workers` puts one `'STOP'` on to the work queue each time it spawns a process. This test checks that we have the expected number of STOPS and assumes that mulitprocessing correctly spawned the processes to go with them.

axelrod/tests/test_tournament.py
meatballs: Actually - it's doing more than checking that it doesn't crash. It's checking that we get the expected result too - but that could have been produced by any number of processes.

drvinceknight: :+1:

PR #193, opened 2015-05-14
Adds coveralls service and badges

meatballs:


meatballs:
Don't merge this yet. I'm still mucking about with it and I can only get it run by committing changes to this branch.

meatballs:
Not working properly. I've logged an issue here:

https://github.com/coagulant/coveralls-python/issues/75

drvinceknight:
Awesome: I looked in to this briefly, it'll be a great addition when it's ready.

meatballs:
There's an alternative package for sending data to coveralls.io I'm going to set that up on a different branch and see if it works any better.

meatballs:
You can see the builds appearing here: https://coveralls.io/r/Axelrod-Python/Axelrod

drvinceknight:
That is cool. Can this now be merged?

drvinceknight:
https://coveralls.io/builds/2563671 gives details as to where we are missing tests. Very cool.

drvinceknight:
97% is pretty good though I reckon! :)

meatballs:
No - the only data that successfully got there so far is from me running it locally. Still having problems with Travis

drvinceknight:
Ah right: I understand. :+1:

meatballs:
Just waiting to see how the other package does on PR #194

meatballs:
This isn't going to work. unittest fails to create the .coverage file when run on Travis. All the stuff I can find online suggest using nose or py.test to run the tests instead but neither of those likes the TestPlayer Class that all our players testing inherits from.

Seems like too much work and not enough benefit to change all that.

drvinceknight:
That's a shame...

meatballs:
Yep. I thought it would be simple and quick to setup.

Nose and py.test are both very nice tools for running the tests, but they really don't like the TestPlayer Class. The way to fix that involves making TestPlayer inherit from object rather than TestCase but each individual player test then has to inherit from both TestPlayer and TestCase.

Doesn't seem worth it.

We could still run coverage locally and submit to coveralls.io We just can't automate that from Travis.

PR #194, opened 2015-05-14
adds python-coveralls package and commands

meatballs:
Don't merge this.


meatballs:
Same problem here. Back to the drawing board!

PR #202, opened 2015-06-04
[WIP] Ported library to Python 3

tricoder42:
:warning: **Work in progress, do not merge** :bomb: 

Compatability issues with Python 2/3:
- `random.choice` doesn't respect `random.seed`. `random.random` produces the same numbers for given seed for both Python 2 and 3, but `random.choice` doesn't select the same elements from list. Following snippet produces different output in Python 2 and 3:

```python
    import random
    random.seed(10)
    print([random.choice(['C', 'D']) for i in range(20)])
```

Some strategies relies on random choices. Right now the tests fail for `QLearner`


drvinceknight:
Given @marcharper's work on #264 can we close this PR?

tricoder42:
Yep :)

PR #207, opened 2015-06-09
Initial work for 199 - Listing of strategies from Axelrod's 1st trnmt

drvinceknight:
Some of the strategies are rather ambiguously described in the original paper but they sometimes reference other papers (where I assume they are described fully) so I could at least incorporate a link to them... (That's not currently in there)

To make the docs locally:

```
cd docs
make html
```


drvinceknight:
:warning: It is very possible that some strategies that I have marked as not being implemented in fact are.  Also, some could be implemented very easily by generalising some of the ones that are implemented (but that could happen through the PR for each particular implementation).

PR #208, opened 2015-06-17
Three new strategies for Axelrod's first tournament, Shubik, Joss, and Grofman, with docs and tests

marcharper:
Also there was a bug in one of the test functions (test_markov) in that it did not reset the state of the player for each of the four subtests (which is problematic if the strategy retains state).


marcharper:
Ok, that's it for now. The remaining strategies are a bit short on at least one details (e.g. exactly how a statistical test was done). I may try some tentative implementations for the others but I'll make a separate PR if I do so. If any one else wants to try please do, likely I'll move on to docs for the later tournaments.

I made an assumption on how Feld reduces it's cooperation probability (I used a linear dropoff, it wasn't clear from the paper what the function was).

drvinceknight:
That's awesome. I'll get to reviewing this today or tomorrow: very exciting to get more strategies in :+1:

drvinceknight:
Can you point out which commit addresses `test_markov`?

axelrod/strategies/grudger.py
```diff
@@ -62,3 +62,17 @@ def strategy(self, opponent):
...
+    """A player starts by cooperating for 10 rounds then plays Grudger, defecting if at any point the opponent has defected."""
+
+    name = 'Davis'
+    memoryone = False  # Long memory
+
+    def strategy(self, opponent):
+        """Begins by playing C, then plays D for the remaining rounds if the opponent ever plays D."""
+        if len(self.history) < 10:
+            return 'C'
+        if 'D' in opponent.history:
```
drvinceknight: I think that technically (although this could be about interpreting the paper) this should be:

```
if 'D' in opponent.history[10:]
```

So that you 'play grudger' after the first ten moves.

marcharper:
The test_markov change is in the first commit; I just moved

+            P1 = self.player()
+            P2 = axelrod.Player()

inside the loop a few lines below (in test_player.py).

I wasn't sure about Davis -- the paper says: "This rule cooperates on the first ten moves, and then if there is a defection it defects until the end of the game." I could see it either way, and I don't have a preference.

axelrod/strategies/rand.py
```diff
@@ -10,3 +12,51 @@ class Random(Player):
...
+    """
+    Defects when opponent defects. Cooperates with a probability that decreases
+    to 0.5 at round 200.
+    """
+
+    name = "Feld"
+    memoryone = False # Probabilities are not constant
+
+    def _cooperation_probability(self):
+        """It's not clear what the iterpolating function is, so we'll do
```
drvinceknight: typo: interpolating

axelrod/strategies/rand.py
```diff
@@ -10,3 +12,51 @@ class Random(Player):
...
+        something simple that decreases monotonically from 1.0 to 0.5 over
+        200 rounds."""
+        slope = 0.5/200
```
drvinceknight: I've been wondering about the '200'... There are other strategies that do things up until K rounds from the end. Could be worth having the number of rounds as property that the players can access... In this particular instance I think it should be a parameter.... The idea of the players being able to access the number of rounds could be another PR perhaps...

axelrod/strategies/titfortat.py
```diff
@@ -82,3 +82,59 @@ class AntiTitForTat(Player):
 
     def strategy(self, opponent):
         return 'D' if opponent.history[-1:] == ['C'] else 'C'
+
+
+class Shubik(Player):
+    """
+    Plays like Tit-For-Tat with the following modification. After 
+    each retaliation, the number of rounds that Shubik retaliates
+    increased by 1.
```
drvinceknight: Typo: increases

marcharper:
I thought about making 200 a parameter. I guess it makes sense to also make 0.5 a parameter?

drvinceknight:
Yeah good idea :) I'm literally looking through it all right now :+1: Will be done in a bit: all looking awesome :+1:

axelrod/strategies/titfortat.py
```diff
@@ -82,3 +82,59 @@ class AntiTitForTat(Player):
...
+
+    def __init__(self):
+        Player.__init__(self)
+        self.is_retaliating = False
+        self.retaliation_length = 0
+        self.retaliation_remaining = 0
+
+    def _refresh_retaliation(self, opponent):
+        """Reset the retaliation counter based on the history"""
+        self.retaliation_length = 0
```
drvinceknight: Could this be made a bit more succinct, something like:

```
self.retaliation_remaining = zip(self.history, opponent.history).count(('C', 'D'))
``` 

Not terribly important but the above is probably faster...

axelrod/strategies/rand.py
```diff
@@ -10,3 +12,51 @@ class Random(Player):
...
+    Cooperates for the first 11 rounds then randomly cooperates 10% less often
+    than the opponent has in previous rounds."""
+
+    name = "Tullock"
+    memoryone = False # memory-10
+
+    def strategy(self, opponent):
+        if len(self.history) < 11:
+            return 'C'
+        counter = Counter(opponent.history[-10:])
```
drvinceknight: Instead of importing the `Counter` (not terribly disruptive to do so but I think this is slightly cleaner):

```
opponent.history[-10:].count('C')
```

Another thing that could be worth doing here is making 10 a parameter?

axelrod/tests/test_rand.py
```diff
@@ -20,3 +20,61 @@ def test_strategy(self):
...
+
+    name = "Tullock"
+    player = axelrod.Tullock
+    stochastic = True
+
+    def test_strategy(self):
+        """Cooperates for first ten rounds"""
+        self.first_play_test(C)
+        for i in range(10):
+            history_1 = [C]*i
```
drvinceknight: Do you mind tweaking the PEP8 here:

```
[C] * i
```

(Sorry)

marcharper:
Thanks for looking so closely! There was another typo in Tullock.

drvinceknight:
For the strategies that are implemented, in the documentation I had an `Implementation` header. In the interest of keeping things the same could you add this in? (So something like [this](http://axelrod.readthedocs.org/en/latest/overview_of_strategies.html#implementation).)

A part from these minor tweaks, I think this is pretty much good to go.

(Note my one comment for PEP8 was for one particular line but if you could grab those throughout that would be appreciated,)

meatballs:
All looks great to me. Everyone happy for me to merge this?

marcharper:
I'm happy with it

PR #235, opened 2015-07-24
New strategies Tester and SoftJoss, and Unit Tests update

marcharper:
In writing unit tests for Tester (one of the strategies from Axelrod and @tscizzle's tournament), I had to modify test_responses in test_player.py to actually submit the plays in sequence (rather than simply overwriting the history), because the strategy needed to set internal state for the tests. Because of this change, I had to correct ~10 tests for other strategies.

Please take a look!


marcharper:
Oh I need to update the docs...

axelrod/tests/unit/test_player.py
```diff
@@ -43,7 +46,21 @@ def test_noisy_play(self):
         self.assertEqual(p1.history[0], 'D')
         self.assertEqual(p2.history[0], 'D')
 
-C, D = 'C', 'D'
+
+class MockPlayer(axelrod.Player):
```
drvinceknight: Nice!

axelrod/tests/unit/test_player.py
```diff
@@ -43,7 +46,21 @@ def test_noisy_play(self):
...
+    """Creates a mock player that enforces a particular next move for a given
+    player."""
+
+    def __init__(self, player, move):
+        # Need to retain history for opponents that examine opponents history
+        # Do a deep copy just to be safe
+        self.history = copy.copy(player.history)
```
drvinceknight: What's happening with these to lines? `self.history` is being assigned twice (once to a copy).

axelrod/tests/unit/test_titfortat.py
```diff
@@ -18,8 +18,8 @@ def test_strategy(self):
     def test_effect_of_strategy(self):
         """Repeats last action of opponent history."""
         self.markov_test([C, D, C, D])
-        self.responses_test([], [C, C, C, C], [C])
-        self.responses_test([], [C, C, C, C, D], [D])
+        self.responses_test([C]*4, [C, C, C, C], [C])
```
drvinceknight: PEP8 (not important)

docs/overview_of_strategies.rst
```diff
@@ -432,6 +432,36 @@ Axelrod's second tournament
...
+   import axelrod
+   p1 = axelrod.Tester()  # Create a Tester player
+   p2 = axelrod.Random()  # Create a player that plays randomly
+   for round in range(5):
+       p1.play(p2)
+
+   print p1.history
+   print p2.history
+
+which gives::
```
drvinceknight: Can you throw in a comment that this is just for the given seed (see other similar strategies).

axelrod/tests/unit/test_player.py
marcharper: The MockPlayer needs to have the right history in case the opponent inspects or manipulates it -- I made it a deep copy in case some of cheating strategies mess with the history.

drvinceknight: But just underneath you have `self.history = player.history`?

marcharper: Oh yeah that should not be there! I'll fix it...

drvinceknight: That was what confused me :)

drvinceknight:
A part from that minor one, could you update the results and cache please?

PR #236, opened 2015-07-24
Update results script

marcharper:
This PR adds a shell script "update_results.sh" that generates the noisy tournaments images, then runs an ordinary tournament that rebuilds the cache and the noiseless images, and copies all the CSV and PNG files to assets.

If someone can take a quick look at [usage.rst](https://github.com/marcharper/Axelrod/blob/update_script/docs/usage.rst) that would be appreciated -- I switched from SVG to PNG and now the images are rendering very large, but look fine when the docs are actually built.


marcharper:
This branch is based on my new_strategies branch from PR #235. I can rebase it, but then the images would have to be regenerated for the new strategies again.

drvinceknight:
I've always thought about writing this... Nice!

Worth including a command to update the `index.rst`?

> I switched from SVG to PNG and now the images are rendering very large, but look fine when the docs are actually built.

why this switch?

marcharper:
SVG to PNG because there doesn't appear to be a way to pass through a format from run_axelrod.

If it can be run automatically after new strategies are added that would be convenient but it's ok if we just run it manually every so often IMO.

drvinceknight:
Can you take a quick look at the script @meatball, I always had to think hard to make sure I wasn't forgetting a command.

( @marcharper, you def built the docs right? Just making sure as I seem to remember having an issue with the static files when I tried to pull them in from assets... I can't think for the life of me what that was...)

drvinceknight:
> SVG to PNG because there doesn't appear to be a way to pass through a format from run_axelrod.

Yes it's all coming back to me now. I was indeed doing it in a rundabout way. Ignore my other question.

> 
> If it can be run automatically after new strategies are added that would be convenient but it's ok if we just run it manually every so often IMO.

Fair enough.

marcharper:
Yeah I ran "make html" to rebuild the docs. I didn't check every docs page but usage.rst (the one I modified) looks fine.

drvinceknight:
> Yeah I ran "make html" to rebuild the docs. I didn't check every docs page but usage.rst (the one I modified) looks fine.

Cool :+1: 

I'm happy for this to be merged (it's super helpful!), just good if @meatballs could quickly check nothing is being forgotten in the script :)

marcharper:
@meatballs [here is the script](https://github.com/marcharper/Axelrod/blob/update_script/update_results.sh)

meatballs:
I'll take a look..

update_results.sh
```diff
@@ -0,0 +1,17 @@
...
+# Run the noisy tournament
+./run_axelrod --noise 0.05 --xc -p 0
+# Copy these figures over, renaming as appropriate
+cp strategies_payoff.png assets/strategies_payoff_noise_5.png
+cp strategies_boxplot.png assets/strategies_boxplot_noise_5.png
+
+# Running noiseless tournament
+echo "-- Running Axelrod, regenerating cache, results, and figures"
+./run_axelrod --rc -p 0
+# Copy the results and figures over to assets
```
meatballs: any specific reason for not using the -o option on run_axelrod?

marcharper: for the noisy tournament I need the files to be renamed, and I don't necessarily want to overwrite everything. I guess it would be fine for the regular tournament?

meatballs: Yep. No problem for the noisy tournament, but there's no need here.

Issue #37, opened 2015-02-23
Add probabilistic ending variant


drvinceknight:
Add a variant where each round ends the match with given probability.

langner:
I don't understand. A variable length tournament?

marcharper:
This is a thing because if you know the number of rounds in advance it's sometimes possible to "back-solve" more optimal strategies. So people often study games of finite but arbitrary length to eliminate this possibility.

langner:
Now I understand. I've read about this before, too.

What does "back-solving" mean?

marcharper:
I'm not sure if it's the right technical term, but it means that if you know the game is exactly 50 rounds then you can infer what the optimal sequence of plays would be. For example versus TFT you would cooperate until the last round and then defect (getting the sucker payoff but avoiding the punishment since the game ends). But if you didn't know the game was about to end you wouldn't do this because it's usually better to get two cooperation payouts rather than one sucker and one defection payout (depending on the game matrix, of course). 

So the number of rounds is a parameter that a strategy has to include in its optimization if its not arbitrary. If the game is indefinitely long then you might be better off trying to maximize your stationary score (for two memory one players) or your long term score in general.

marcharper:
Here's a flexible way to do this that aligns with some approaches in the literature. Play N rounds as tournaments already do, then have a probability p that the game ends after each subsequent round. Then the expected number of rounds past N is ~1/p (but the actual number of rounds is potentially unbounded). Some approaches in the literature do this with N=0 or larger.

Alternatively we could draw the number of rounds from a scipy/numpy random variable and play exactly that many rounds. However cheating strategies could try to discover the number of rounds and use it to their advantage (but that's not a big deal really).

Finally we could do both. Draw from a distribution to ensure that the number of rounds is bounded, e.g. start with a Bernoulli distribution of probability p, calculate the distribution of expected first success (i.e. the number) of additional rounds), and then pass N + (a draw) in as the number of rounds. Still allows for cheatering but at least the tournament runs for a fixed number of rounds. I think the distribution is just a negative binomial NB(1, 1-p), which [scipy](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.stats.nbinom.html) and [numpy](http://docs.scipy.org/doc/numpy/reference/generated/numpy.random.negative_binomial.html) both implement.

drvinceknight:
[...]

PR #237, opened 2015-07-24
Introducing BackStabber and Aggravater strategies

### LINKED ISSUE #221: Backstabber Strategy
New world-beating strategy.
Work visible on my fork at https://github.com/uglyfruitcake/Axelrod/tree/221

uglyfruitcake:
Fixes #221 

I haven't updated the graphs or the cache because I don't have matplotlib installed and I don't understand the cache!


axelrod/tests/unit/test_oncebitten.py
```diff
@@ -146,3 +146,41 @@ def test_reset(self):
...
+    stochastic = False
+
+    def test_initial(self):
+        P1 = axelrod.BackStabber()
+        P2 = axelrod.Player()
+        self.assertEqual(P1.strategy(P2), P1._initial)
+
+    def test_strategy(self):
+        """
+        If opponent defects at any point then the player will defect forever
```
drvinceknight: Add here that you're also checking the 196 thing.

axelrod/tests/unit/test_oncebitten.py
```diff
@@ -146,3 +146,41 @@ def test_reset(self):
...
+        """
+        P1 = axelrod.BackStabber()
+        P2 = axelrod.Defector()
+        P1.history = ['C']
+        P2.history = ['D']
+        self.assertEqual(P1.strategy(P2), 'C')
+        P1.history = ['C']
+        P2.history = ['D']
+        self.assertEqual(P1.strategy(P2), 'D')
+        P2.history = ['C']*197
```
drvinceknight: PEP8 (have a space either side of the `*`).

docs/overview_of_strategies.rst
```diff
@@ -882,4 +882,19 @@ This strategy is not unambiguously defined in [S&P, PNAS 2012].
...
+
+Fool Me Once
+^^^^^^^^^^^
+
+This strategy begins by cooperating but will defect if at any point the opponent has defected more than once.
+
+Backstabber
+^^^^^^^^^^^
+
+A variation of Fool Me Once, this strategy will also defect on the final 2 turns of the round unconditionally.
```
drvinceknight: This is the line I alluded to above.

axelrod/strategies/oncebitten.py
```diff
@@ -111,3 +111,14 @@ def strategy(self, opponent):
...
         self.D_count = 0
         self.history = []
+
+
+class BackStabber(FoolMeOnce):
+
+    name = 'BackStabber'
+
+    def strategy(self, opponent):
+        if len(opponent.history) > 196:
```
drvinceknight: You say in the documentation:

> A variation of Fool Me Once, this strategy will also defect on the final 2 turns of the round unconditionally.

This doesn't quite line up with what is here (final 4 turns). Also could you change 'round' to 'duel'.

The other issue is that using `196` assumes the tournament length is 200 which it does not necessarily have to be. It would be nicer if you coded something generic that always did 4 rounds less than the length of the tournament itself (although that might need some tweaking as I'm not sure the players have access to the tournament length).

(An easier/hackier fix would be to say in the description that the strategy defects after 196 rounds.)

drvinceknight:
> Fixes #221

Looks neat, am intrigued that it's such a simple change :)

> 
> I haven't updated the graphs or the cache because I don't have matplotlib installed and I don't understand the cache!

You should install `matplotlib` :+1:

uglyfruitcake:
I would, but I will be completely resetting my computer within the next few days. So I will do it after that.

drvinceknight:
> I would, but I will be completely resetting my computer within the next few days. So I will do it after that.

:+1:

uglyfruitcake:
Ok. Just to clarify, 196 was a mistake. It should be 197 so that it defects on the last 2 duels.

drvinceknight:
> Ok. Just to clarify, 196 was a mistake. It should be 197 so that it defects on the last 2 duels.

Yup. 199 is the 200th round (python 0-indexing) so `history>196` means `history in [197, 198, 199]`.

My bigger 'issue' is that 196/197 is counting from the start ad not the end as alluded to in the docs (and indeed what you want it do to do). If a tournament was generated that ran for 1000 turns in a duel then it wouldn't do as well.

I'm not sure if the tournament length is available to the players (we should make it so) but you could even make this 'backstab' on the last 10% of the turns.

On another note, I just saw your commit message for 8d42e6c, you might want to have a read of something like: http://chris.beams.io/posts/git-commit/ and/or https://github.com/erlang/otp/wiki/Writing-good-commit-messages

uglyfruitcake:
I didn't know if using the tournament length counted as "cheating". However if it is not then I would much rather use that than hard code it.

drvinceknight:
> I didn't know if using the tournament length counted as "cheating". However if it is not then I would much rather use that than hard code it.

Quite a few historical strategies have the knowledge so yeah that's not cheating :)

marcharper:
See also the discussion for issue #37 regarding making the tournament length available to players.

uglyfruitcake:
Yes. However, some of the suggestions in #37 would mean that there would not be a fixed number of turns meaning that the strategies wouldn't be able to use that to their advantage.

uglyfruitcake:
If I were to use the tournament length in the code. How would I go about doing that?

drvinceknight:
Yeah I think #37 is a bit different isn't it? Here we're talking about for a tournament of known length, letting the players know what that is.

> If I were to use the tournament length in the code. How would I go about doing that?

Just speaking to @meatballs briefly about that, it's not immediately straightforward.

marcharper:
Well that's the point actually -- all players knowing the number of rounds changes the viability of the strategy landscape. If I know you are going to start defecting on round 195 then I'll start defecting on round 194 and then you start defecting on round 193 and so on until we're all playing ALLD. That's why often people either assume that the number of rounds isn't fixed, or if it is, that the number of rounds is unknown to the players (the current situation). Anyway my point is knowing the number of rounds is actually a major variant of these tournaments (like adding in noise) -- it can dramatically alter the performance of many strategies, and players that do not use this additional information as part of their strategy are at a substantial disadvantage (depending on how other strategies are behaving).

I'm sure there's some pity statement about how people tend to cooperate with those they expect to interact with again (in-groupers) but not strangers (out-groupers). Of course there are nuances like  [reciprocal altruism](https://en.wikipedia.org/wiki/Reciprocal_altruism), maintaining a reputation, etc.

Another similar variation would be giving players access to the outcomes of the games played between other players (sort of like having a reputation variant).

drvinceknight:
For Axelrod's original tournament players knew the length of the tournament
right? That's why I just assumed it was fair game.

Of course everything you are saying is correct as to how strategies who use
available information have an advantage (similar to strategies who look at
history and don't).

marcharper:
I think you are right -- IRCC in Axelrod's first tournament the number of rounds was specified before strategies were submitted.

Some have claimed recently in prominent journals that looking at history doesn't matter. But that's another digression...

drvinceknight:
And the second tournament also right? I think that letting players know the
tournament length is acceptable (for the case of the results we're talking
about everyone knows them anyway, as for Axelrod's original tournaments and
as shown by this strategy).

People are welcome to create tournaments that don't include strategies that
know tournament length or indeed after #37 to create probabilistic ending
tournaments.

marcharper:
Of course :) I'm wasn't suggesting otherwise. There are lots of other things that would be useful for players to know, such how many other players there are and what strategies they use, [perhaps the source code of those strategies](http://arxiv.org/abs/1401.5577). Sorry if I derailed the discussion!

uglyfruitcake:
I have put in a new issue for making the tournament length available to the players. Can this PR go through for now and I'll change the hard coding in #240?

uglyfruitcake:
I just made a few changes to Backstabber and Aggravater. Backstabber will now wait for the 4th defection (this drastically improves it for some reason). Aggravater will now defect 3 times at the beginning. I have changed the testing and the documentation accordingly.

drvinceknight:
> I have put in a new issue for making the tournament length available to the players. Can this PR go through for now and I'll change the hard coding in #240?

Yup that's fine, could you just change the wording of the docstring and documentation (for now) so that it says 'defects after the 19*th' round? Just in case it takes a while to come back to it.

uglyfruitcake:
There you go.

drvinceknight:
:+1: I'll rerun the results now.

drvinceknight:
(Just waiting on travis.)

meatballs:
I'll regenerate the results and cache for this and push straight to master

drvinceknight:
> :+1: I'll rerun the results now.

I'm in the middle of doing that @meatballs

meatballs:
OK. I'll stop!!

PR #243, opened 2015-07-25
Adds Doublecrosser strategy

uglyfruitcake:
New strategy. Again, no graphs or cache. Sorry :(


axelrod/strategies/backstabber.py
```diff
@@ -33,3 +33,48 @@ def strategy(self, opponent):
...
+        cutoff = 6
+        if len(opponent.history) > 197:
+            return 'D'
+        if not opponent.history:
+            return self._initial
+        if opponent.history[-1] == 'D':
+            self.D_count += 1
+        if len(opponent.history) < 180:
+            if len(opponent.history) > (cutoff):
+                firstelements = []
```
drvinceknight: this can be written as `firstelements = opponent.history[:cutoff + 1].`

So in fact you could just write `if 'D' not in opponent.history[:cutoff] and opponent.history[-2:] != ['D', 'D']:`

but I'm not sure I understand why you have the `opponent.history[-2:] != ['D', 'D']:` there?

axelrod/tests/unit/test_backstabber.py
```diff
@@ -47,3 +47,51 @@ def test_reset(self):
...
+
+    def test_initial(self):
+        P1 = axelrod.DoubleCrosser()
+        P2 = axelrod.Player()
+        self.assertEqual(P1.strategy(P2), P1._initial)
+
+    def test_strategy(self):
+        """
+        Forgives the first 3 defections but on the fourth
+        will defect forever. If the opponent did not defect
```
drvinceknight: You're missing a couple of things in the test below:

- I don't see that the first 6 rounds effect is being tested.
- The first three defections is not being tested as you're reseting the history and not appending to it.

drvinceknight: I just checked #237 and I missed this on that PR. Could you include this fix in this PR for backstabber also please.

axelrod/strategies/backstabber.py
drvinceknight: Ignore my comment about `opponent.history[-2:] != ['D', 'D']:`.

axelrod/tests/unit/test_backstabber.py
```diff
@@ -47,3 +47,51 @@ def test_reset(self):
...
+        P1 = axelrod.DoubleCrosser()
+        P2 = axelrod.Defector()
+        P1.history = ['C']
+        P2.history = ['D']
+        self.assertEqual(P1.strategy(P2), 'C')
+        P1.history = ['C']
+        P2.history = ['D']
+        self.assertEqual(P1.strategy(P2), 'C')
+        P1.history = ['C']
+        P2.history = ['D']
```
drvinceknight: So this should be a list with 2 'D's in it (and possibly other 'C's to highlight the point).

axelrod/strategies/backstabber.py
uglyfruitcake: Cheers, that's a much nicer way of doing it. 5 lines down to 2!

axelrod/tests/unit/test_backstabber.py
uglyfruitcake: Yes, I've made a bit of a mistake there. I'll fix that for both backstabber and doublecrosser.

drvinceknight: :+1: (My mistake too, I should have caught it before :))

axelrod/tests/unit/test_backstabber.py
```diff
@@ -24,14 +24,64 @@ def test_strategy(self):
         P1.history = ['C']
         P2.history = ['D']
         self.assertEqual(P1.strategy(P2), 'C')
-        P1.history = ['C']
-        P2.history = ['D']
+        P1.history.append('C')
+        P1.history.append('D')
```
meatballs: Is this correct? Should it be P2.history.append ?

axelrod/tests/unit/test_backstabber.py
```diff
@@ -24,14 +24,64 @@ def test_strategy(self):
...
         self.assertEqual(P1.strategy(P2), 'C')
-        P1.history = ['C']
+        P1.history.append('C')
+        P1.history.append('D')
```
meatballs: This looks like a repeat of the previous typo

axelrod/tests/unit/test_backstabber.py
```diff
@@ -24,14 +24,64 @@ def test_strategy(self):
...
+        self.assertEqual(P1.strategy(P2), 'C')
+        P1.history.append('C')
+        P1.history.append('D')
+        self.assertEqual(P1.strategy(P2), 'D')
```
meatballs: this too

drvinceknight:
Will get the results running now but am heading to :zzz: now so might not wait till they're finished.

PR #245, opened 2015-07-27
240

### LINKED ISSUE #240: Make tournament length available to players
The Backstabber strategy currently has the number of turns hard coded as 200. I believe that in the original tournament the number of turns was available to the players.

uglyfruitcake:
Fixes #240 

Tournament length is now available to players as ```self.turns```

### DIFF
--- a/axelrod/tournament.py
+++ b/axelrod/tournament.py
@@ -29,6 +29,13 @@ class Tournament(object):
         self._processes = processes
         self._logger = logging.getLogger(__name__)
 
+    def _create_players_list(self, players):
+        newplayers = []
+        for player in players:
+            player.turns = self.turns
+            newplayers.append(player)
+        return newplayers
+
     def play(self):
         payoffs_list = []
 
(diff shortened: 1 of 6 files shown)


meatballs:
Nice effort!

I think that we should make ```tournament.turns``` a property rather than a simple attribute. Your ```_create_player_list``` method would then become the setter for the property.

marcharper:
Everyone happy with this? My latest PR includes a strategy that relies on the total number of rounds, and if this is how we will do it I'll make a quick change to that PR to use self.turns.

drvinceknight:
Just looking through it now :)

axelrod/tournament.py
```diff
@@ -25,10 +25,21 @@ def __init__(self, players, name='axelrod', game=None, turns=200,
...
 
+    @property
+    def players(self):
+        return self._players
+
+    @players.setter
+    def players(self, players):
+        newplayers = []
+        for player in players:
+            player.turns = self.turns
```
drvinceknight: Small suggestion: call this `player.tournament_length` or something like that? `player.turns` could be understood as a counter of how many turns have happened so far or something like that?

drvinceknight: A part from that small suggestion: this looks really great - great job.

uglyfruitcake:
Tournament length is now available to players as self.tournament_length.

drvinceknight:
:+1: (Waiting on travis, will merge asap)

PR #246, opened 2015-07-27
New strategies

### LINKED ISSUE #233: Add in remaining strategies from the morality metrics paper
Some of these are original Axelrod strategies as well.

marcharper:
Closes #233 by adding Eatherley and Champion strategies from Axelrod's second tournament and by parameterizing Random, Joss, and GoByMajority.

Since the named strategies for Axelrod's tournaments were distributed widely, I moved most of them to a new strategy file `axelrod_tournaments.py".

Also, it was necessary to update the results script to rebuild the cache on the first run (for the noisy tournament) because run_axelrod complains if a strategy is moved between files without updating the cache.


marcharper:
I believe that @tscizzle's tournament is now completed by the following strategies:

Cooperator()
Defector()
Eatherley()
Champion()
GTFT(ep=0.1)
GTFT(ep=0.3)
GoByMajority(soft=True)
GoByMajority(soft=False)
TitFor2Tats()
Random(0.8)
Random(0.2)
WSLS() # Pavlov
TitForTat()
TwoTitsForTat()
Grudger() # Friedman
Tester()
SuspiciousTitForTat()
Joss(0.1)
Joss(0.3)

drvinceknight:
> I moved most of them to a new strategy file `axelrod_tournaments.py"

This one made me make a double take but I think I like it. What do you think @meatballs?

I'll continue to look through this when I get time, although I made a comment on #245 that might be relevant.

meatballs:
I find myself torn on this one....

Firstly, let me say this is definitely an improvement on what we had previously. However, this feels like a situation where 'one class per file' might be our best solution.

The problem, as I see it, is that we struggle to find the code for a given strategy. This improves that situation but still requires me to know whether a strategy is one of Axelrod's originals or not. For me, looking for a file within the strategies directory with the same name as the strategy itself would be far simpler.

meatballs:
I suggest we let this one through on the basis that it improves what we currently have and then we open an issue to discuss whether 'one file per strategy class' is something we want to do.

drvinceknight:
Cool, will look through everything when I get a chance unless you wanted to review this @meatballs ?

marcharper:
I've been running

> grep <name> strategies/*.py

to find specific strategies (when I can't remember where they are). I left all the strategies that have acquired generic names (like TFT, Pavolv, etc.) where they are rather than moving them to axelrod_tournaments.py (which has just strategies that are named after their creators).

marcharper:
I rebased onto master to get the self.tournament_length parameterization into Champion. I also had to add a tournament_length kwarg to responses_test in test_player.py since the default value is -1 and is not set to something greater than zero until it interacts with a tournament class. Setting self.tournament_length = 200 in the class definition causes another test to fail (that the default isn't -1), specifically this test that was added to "test_initialization" in TestPlayer:

>         self.assertEqual(self.player().tournament_length, -1)

meatballs:
I'm happy with this. Anyone object to me merging it in?

marcharper:
I'm going to go ahead and merge this one. Once the metrics are in, I'll run the computations versus tscizzle's strategies for verification.

Issue #251, opened 2015-07-31
Project Objectives


meatballs:
I think it might be helpful for us to define the aims and objectives of this project and to have those recorded. Here's my shot at what I think those should be:

1. To enable the reproduction of previous IPD research as easily as possible.
1. To produce the de-facto tool for any future IPD research.
1. To provide as simple a means as possible for anyone to define and contribute new and original IPD strategies.

I also think we have a further objective, currently implemented within the library itself, but which ought to be moved to a separate repository:

* To regularly run a small number of interesting tournaments and to publish their results online.

meatballs:
I also think that, whilst we would encourage interesting work and analysis of tournaments which use our library, that analysis should not form part of the library itself. The library should produce the information which the research requires, but the analysis should be done separately by the individual researcher. We should, perhaps, consider listing work which has used the library as part of the github.io pages.

drvinceknight:
I agree with all of these and think it's important to have them written down as it helps to guide some of the discussion on previous/ongoing PRs/issues.

I particularly like '2.' being written clearly as it's something that I think can encourage the use of the library outside of the repository.

I also think having a 'front end' like repository where we keep track of interesting tournaments is a good thing to do. Basically we're talking about doing things in a very '*nix'y: this library does what this library does and does it well.

drvinceknight:
Assuming everyone's ok and on board with this I think this should be a PR to have it written in the README and then arguably have `run_axelrod` lightened...

drvinceknight:
> I also think that, whilst we would encourage interesting work and analysis of tournaments which use our library, that analysis should not form part of the library itself. The library should produce the information which the research requires, but the analysis should be done separately by the individual researcher. We should, perhaps, consider listing work which has used the library as part of the github.io pages.

:+1: YES

I need to work on the github.io pages some more... Looks like I will setup a jekyll site if that's not too painful re python/ruby?

marcharper:
:+1: 

What do you think about having a separate (optional) repository for analysis that people can contribute to? There is value to having things centralized, and blog posts can disappear without records. This makes it easier to combine and build upon other's analyses.

meatballs:
> What do you think about having a separate (optional) repository for analysis that people can contribute to?

Yes, I think we should have our own repo(s) for analysis work. The current tournaments which we run from run_axelrod are a good example, but I'm sure there will be others.

drvinceknight:
> but I'm sure there will be others.

[...]

PR #252, opened 2015-07-31
Adding objectives and list of projects addressing #251

drvinceknight:
It is possible that the list of projects could grow and we need to move
elsewhere. I suggest that if that's the case (a delightful problem to
have) we figure out the best place then.

Let me know if you think this is ok :)


marcharper:
:thumbsup:

PR #254, opened 2015-08-02
Sets tournament length on second player in self interactions

### LINKED ISSUE #253: BackStabber and DoubleCrosser self interactions error
There appears to be a problem with interactions between two DoubleCrosser or BackStabber players where the scores are being calculated incorrectly.

For example:
```
import axelrod

p1, p2 = axelrod.BackStabber(), axelrod.BackStabber()
turns = 10
p1.tournament_length = turns
p2.tournament_length = turns

r = axelrod.RoundRobin([p1, p2], axelrod.Game(), turns)

print r.play()
```

gives:
```[[16.0, 6], [26, 16.0]]```

when I'd expect to see:
```[[26, 26], [26, 26]]```

I get exactly the same result for DoubleCrosser.

I've tried several other deterministic strategies and found no problems with any of them.

meatballs:
Fixes #253


meatballs:
anyone got a chance to have a look at this? It's only one additional line of code, so shouldn't take long to review, and I don't like the fact that our library is currently returning incorrect results!

marcharper:
Looks good to me! :thumbsup:

Issue #255, opened 2015-08-05
Slight reorganization of _strategies.py or change to test_full_tournament?


marcharper:
I think that Travis is taking so long because of the time it takes to run a full tournament with all the non-cheating strategies (and it uses a lot of RAM).

This is of course a good problem to have, since we have a lot of strategies in the module now, but some of them are fairly odd or unpopular, and not in much use anymore (e.g. some of the strategies from the first two Axelrod tournaments).

One way to deal with this would be to move some of these strategies, such as Champion, Feld, Eatherly, Grofman, Shubik, Tullock, all the Probers, Tester, maybe some others -- out of `ordinary_strategies` in `_strategies.py` into another list, perhaps something like `historical_strategies`. We will still be testing the strategies individually so I don't think we are losing much on the testing front by doing so.

An alternative would be to reduce the number of strategies used in test_full_tournament -- maybe a random selection of 30 strategies?

Thoughts?

Issue #157, opened 2015-04-19
Bug in Q-learners


drvinceknight:
@geraintpalmer has told me that apparently the Q-learners were not written as well as they should have been so the recent change from min to max will have broken them.

meatballs:
Do we know what the nature of this bug it yet?

drvinceknight:
@geraintpalmer does... He did explain it to me but I forget the details...
Geraint can you go over it?

I know that for one the game scores are hard coded and don't use the game
object. Also I think they're still working on minimisation...

geraintpalmer:
Yes the payoff matrix is hard coded into the strategy.

At the time the payoff matrix was a min, but q-learners need to max, so I did something like 5-i for each i in payoff matrix.

drvinceknight:
Are you going to have any time to work on this at some point
@geraintpalmer? (No problem if not, not your priority)

marcharper:
I've updated the game matrix and modified the tests and a few other things. @geraintpalmer Will the strategy work properly with max and the right game matrix? If so, please take a quick look.

drvinceknight:
I'll try and get @geraintpalmer to take a look.

PR #256, opened 2015-08-06
Performance Enhancements

marcharper:
Many strategies were scanning either the opponents or their own histories each time `strategy()` was called, resulting in about `O(tournament_length^2)` unnecessary operations in every matchup versus simply tracking the total number of cooperations and defections. So I added automatic counting of cooperations and defections to the `Player` class and modified many strategies accordingly.

This reduces the time to run all the tests to 50 seconds on my machine (despite the fact that I added many new tests as well), down from 135 seconds. Along the way I made a number of improvements to the existing strategies and tests.

Also, since the Qlearners are bugged, I commented them out of `_strategies.py`. They use the wrong game matrix and all the tests fail when the correct game matrix is inserted.

Finally, the suggestion in #255 only slightly improves matters (50 seconds down to 40 seconds for all tests), so I closed it.


meatballs:
An excellent idea! I'm more than happy for this to go in.

drvinceknight:
> Also, since the Qlearners are bugged, I commented them out of _strategies.py. They use the wrong game matrix and all the tests fail when the correct game matrix is inserted.

:+1: Great call. Tagging @geraintpalmer so that he is aware: #157

PR #262, opened 2015-08-08
Update reset function in Darwin to prevent bug when interacting with ForgivingTitForTat

marcharper:
Just one line!

### DIFF
--- a/axelrod/strategies/darwin.py
+++ b/axelrod/strategies/darwin.py
@@ -64,7 +64,7 @@ class Darwin(Player):
 
     def reset(self):
         """ Reset instance properties. """
-        self.history = []
+        Player.reset(self)
         self.__class__.genome[0] = 'C' # Ensure initial Cooperate
 
 


marcharper:
This is the exception that is thrown:

Process Process-18:
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user/repos/Axelrod/axelrod/tournament.py", line 124, in _worker
    payoffs = self._play_round_robin(cache_mutable=False)
  File "/home/user/repos/Axelrod/axelrod/tournament.py", line 137, in _play_round_robin
    payoffs = round_robin.play()
  File "/home/user/repos/Axelrod/axelrod/round_robin.py", line 68, in play
    p1.play(p2, self._noise)
  File "/home/user/repos/Axelrod/axelrod/player.py", line 61, in play
    s1, s2 = self.strategy(opponent), opponent.strategy(self)
  File "/home/user/repos/Axelrod/axelrod/strategies/forgiver.py", line 43, in strategy
    return opponent.history[-1]
IndexError: list index out of range

The issue is that by not calling Player.reset(self), the cooperation / defection counts are not updated. I checked the other strategies and I think they all do the right thing now.

drvinceknight:
I'm going to go ahead and merge this (without a 3rd person looking through) as it's an obvious bug. Anyone/everyone please do say if they think that's no ok (@meatballs - I know you're busy Karol so not pinging you.)

langner:
Perhaps there should be an unwritten rule that bugfixes don't need to go through 2 people...

drvinceknight:
Yeah that seems sensible. I think the two people rule is rule is really about new stuff so that everyone is aware of new features/strategies etc...

Issue #261, opened 2015-08-08
Bug in eco plots?


drvinceknight:
Working on what is a trivial change for #258 and re running the results I noticed that the eco plots were not working for me. See #260.

I've run this on `master` as well as my `258` branch and have the same outcome:

![basic_strategies_reproduce](https://cloud.githubusercontent.com/assets/2131546/9149924/bac1b7e8-3db9-11e5-85d5-e14ec1a5a9c5.png)

I have tried this on two machines and with matplotlib installed in a virtualenv directly from `requirements.txt`. Can anyone else reproduce this?

(Here's the command to quickly grab the basic plot:

```
./run_axelrod -c cache.txt -p 0 --xs --xc --xa
```

)

marcharper:
The plot produced by that command (for this branch) looks fine on my machine

marcharper:
Ok I was able to reproduce the bug (I had to reinstall Axelrod).

drvinceknight:
#263 has not fixed the issue for me... I still get the following:

![basic_strategies_reproduce](https://cloud.githubusercontent.com/assets/2131546/9154380/462ee6e2-3e83-11e5-8dff-472f7ec989f9.png)

I had originally tried with something a bit like this but it did not make a difference to the heatplot itself. This is what confuses me, the heatplot is fine...

I am also getting this behaviour on `master` so I'm pretty sure the colormap does not have much to do with it.
 
- Can anyone reproduce this bug (see image below) on master?
- I am reverting #263 on my `258` branch, does that give the same for others?

I have run this on MacOSX as well as ubuntu 14.04. When you say @marcharper that you had to 'reinstall' Axelrod, what do you mean? Just a re-clone of the repo? I'll try that.

![basic_strategies_reproduce](https://cloud.githubusercontent.com/assets/2131546/9154385/bf050e20-3e83-11e5-99ac-6d40b5dc1714.png)

drvinceknight:
A clean clone of Axelrod gives the same outcome (on `master`)...

```
>>> import matplotlib
>>> matplotlib.__version__
'1.4.2'
```

How very strange...

marcharper:
```
user@tower:~/repos/Axelrod$ git checkout 258 # With my patch
Switched to branch '258'
Your branch is ahead of 'github-axelrod/258' by 1 commit.
  (use "git push" to publish your local commits)
user@tower:~/repos/Axelrod$ sudo paco -lD "python setup.py install"
[...]

PR #263, opened 2015-08-08
Fetch colormap object before passing to matshow

marcharper:


langner:
Related to #261

marcharper:
Yes, I believe that it fixes it, but someone else should verify that.

drvinceknight:
Working through now.

PR #260, opened 2015-08-08
258 - Changing heatmap colormap

drvinceknight:
This changes the default to `YlGnBu`, take a look at #258 for other example heatmaps...

PR #264, opened 2015-08-10
Python 3 Support

marcharper:
I finished @tricoder42's work on PR #202.


tricoder42:
:tada: :confetti_ball: :rocket: :sparkles: :+1:

marcharper:
Once #260 goes in, I'll rebase onto master and then we can merge :)

marcharper:
Also I think we can add the qlearners back in, I believe I fixed them. FYI, Python 3 randoms dictionary key ordering, so it's best to use OrderedDicts unless there is reason not to; tracking that issue down in the qlearners tests took some sleuthing!

drvinceknight:
Awesome to get this going again @marcharper!

> FYI, Python 3 randomizes dictionary key ordering, so it's best to use OrderedDicts unless there is reason not to; tracking that issue down in the qlearners tests took some sleuthing!

That's not just something python 3 does but also python 2. I would suggest not using `OrderedDicts` but just checking equality of the dictionaries no? That's how I get around this with all my testing in Sage.

drvinceknight:
Perhaps I'm misunderstanding the issue though (just looking through the commit messages and not the actual code yet).

drvinceknight:
@marcharper either way I think it would be more helpful to just keep this PR about python 3, could you remove the stuff about the Q-learners and put that in a separate PR? Or am I missing something?

axelrod/tournament_manager_factory.py
```diff
@@ -1,6 +1,10 @@
-from tournament_manager import *
+from __future__ import absolute_import
+from collections import OrderedDict
+
 import axelrod.strategies
 
+from .tournament_manager import *
```
drvinceknight: This is something I don't know about python3, why the `.tourna...`, why the `.`?

drvinceknight:
I've restarted the build to see if we can make coveralls happy...

marcharper:
The tests check for equality of dictionaries just fine. The issue is to make the tests pass for qlearner in both python2 and 3 without using ordered dicts, since there is some randomness in how the keys and values are chosen by qlearner. The problem is that qlearner choses a state by its value using the built-in `max` function, and python chooses the first (or last?) key with the value (if they are equal) when it iterates over the items of the dictionary. Since python 3 randomizes the order of iteration over the keys, it causes the tests to randomly fail 50% of the time.

I guess a manual `max` function that iterates over sorted keys would work.

marcharper:
Here's more info on the dictionary iteration issue and how it differs between python 2 and 3:

http://stackoverflow.com/questions/14956313/dictionary-ordering-non-deterministic-in-python3

axelrod/tournament_manager_factory.py
marcharper: It's absolute versus relative imports in Python 3, going all the way back to  PEP 0328. It's necessary for the imports to work, although something like `from axelrod.tournament_manager import *` would probably also work. You just can't leave off the dot altogether any longer.

drvinceknight:
> The tests check for equality of dictionaries just fine. The issue is to make the tests pass for qlearner in both python2 and 3 without using ordered dicts, since there is some randomness in how the keys and values are chosen by qlearner. The problem is that qlearner choses a state by its value using the built-in max function, and python chooses the first (or last?) key with the value (if they are equal) when it iterates over the items of the dictionary. Since python 3 randomizes the order of iteration over the keys, it causes the tests to randomly fail 50% of the time.
> 
> I guess a manual max function that iterates over sorted keys would work.

Cool.

I really should wait till my second cup of coffee before writing on github. I've just walked the puppy (she says hi) and was thinking that I was actually quite negative in pretty much all my comments above: so apologies :D Consider those just my initial thoughts which really should stay in my head before I think about things more.

- `OrderedDicts`: why the hell not? Don't touch those, they look great (let's ignore the idiot who had some reason for not wanting to use them)
- q-learners: would have been neater to keep them separate but let's just review it all in one go here :)

I'll take a look through but as all the tests pass I don't think there's much to consider and also get @geraintpalmer to look through the q-learners and comment here.

I wonder why overalls doesn't seem happy, any idea?

drvinceknight:
This is so cool to see :+1: Presenting this at [pyconuk](http://www.pyconuk.org/), being able to say that python 3 is supported will be really good :)

screenshot 2015-08-11 08 09 45

drvinceknight:
Ah actually, re coveralls, it looks like it's running but that it's returning a negative :x: because of the fact that the tests coverage as decreased... That's cool if so because at least everything is still working...

drvinceknight:
So I'm happy with everything here :+1: subject to @geraintpalmer taking a look at the qlearners.

axelrod/tournament_manager_factory.py
drvinceknight: Just saw this: thanks, will look up PEP 0328.

meatballs:
> What's the difference between AssertEqual and AssertEquals?

https://docs.python.org/2/library/unittest.html#deprecated-aliases

tricoder42:
It's just an alias...

11. 8. 2015 v 10:20, Owen Campbell <notifications@github.com>:

drvinceknight:
:+1:

Thanks both :)

marcharper:
Fixed the game matrix and tests; added some tests for random_choice. There's no difference between assertEquals and assertEqual but the former is deprecated.

marcharper:
It's good to merge for my part. FYI I also tested with python 3.5b4, and we should be all set when 3.5 lands.

drvinceknight:
Is this after the comment by @geraintpalmer @marcharper ? He said he commented on a commit somewhere?

marcharper:
It was the same comment as above, to fix the typo in the game matrix.

drvinceknight:
Ah right, no @geraintpalmer found something that wasn't quite right. Let me ask him to put it here and to help out with the conversation.

drvinceknight:
There's a comment at the bottom of https://github.com/marcharper/Axelrod/commit/27e1913bec578b5138fae5d4d7e1dcb548000bda

screenshot 2015-08-11 17 01 55

(I suggest you respond here though to keep the conversation in one spot)

drvinceknight:
Ah sorry, hold on, I've caught up and see that he did comment here (jeeez I'm slow today). My bad.

drvinceknight:
Looks like this is good to go @meatballs if you wanted to click :)

marcharper:
Just added a quick test for the payoff matrix in qlearner.

PR #275, opened 2015-08-17
adds documentation section for cooperation matrix

### LINKED ISSUE #273: Write documentation for the cooperation etc measures
Would be good to include reference to the paper and also a small example. (This is only relevant once #272 goes in)

meatballs:
fixes #273


marcharper:
:+1:

PR #276, opened 2015-08-17
Remove Assets

marcharper:
Removes:
* image files that are either no longer used
* image files that have moved to the tournament repository
* results *.csv files 
* cache.txt


meatballs:
:+1:

PR #278, opened 2015-08-17
New strategies

marcharper:
Some additional strategies that behave cyclically and non-cyclically. There are now multiple strategies that hunt for other strategies that use patterns or cycles, but not that many strategies that actually behave in such a way.

This is post-deletion of assets #276.

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
 from __future__ import absolute_import
 
 from .alternator import *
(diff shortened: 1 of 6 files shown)


meatballs:
Could we have something in the docs for the new strategies?

uglyfruitcake:
Sorry if I'm missing something, but is Fool Me Forever the same as Opposite Grudger?

marcharper:
@uglyfruitcake They are not quite the same. FMF doesn't start playing 'C' until the opponent plays 'D'; for opposite grudger it's after the opponent plays 'C'.

I'll add some info to the docs.

marcharper:
Rebased onto the new upstream and added some documentation.

meatballs:
:+1:

Issue #158, opened 2015-04-19
Numpy Revisited


meatballs:
Given that this is now published on PyPi with full dependency management, should we rethink the decision not to use Numpy?

It could easily be added as a dependency so that pip install axelrod will also ensure that numpy is available.

meatballs:
Also - since matplotlib is already a dependency at PyPi, numpy is already being installed. I'd say we might as well use it.

drvinceknight:
I'd still like to stay away from Numpy. The numerics we're doing don't really make this too heavy and at present a user can download the zip file and run everything from there with nothing else (the matplotlib stuff is all optional).

meatballs:
It's just a suspicion on my part, but I wonder if numpy's  array representation might give us a performance improvement. There's a cython tutorial here: https://python.g-node.org/python-summerschool-2011/_media/materials/cython/cython-slides.pdf which talks about the boxing/unboxing issue and specifically mentions numpy arrays.

I also suspect that users who aren't too comfortable are more likely to use pip install rather than downloading a zip file.

In the end for me it comes down to whether or not the performance improvement is there and  I don't have evidence either way. It possibly needs some work to compare the speeds with and without.

drvinceknight:
> I also suspect that users who aren't too comfortable are more likely to use pip install rather than downloading a zip file.

That's an excellent point...

drvinceknight:
Just to add to this discussion (although as I said above I think your latest point makes me much more in favour of the idea). This weekend at a hackathon I had a nightmare attempting to get numpy (more precisely scikit learn which depends on numpy) to run nice on travis. This might have just been that I didn't have time to think through it properly but just putting this out there for us to think about.

marcharper:
Note that eigen.py now imports numpy.

drvinceknight:
Good point, I'll add it to the requirements.txt

drvinceknight:
Actually, without it being in the requirements.txt how did tests run on travis? Or do we not have tests for it?

drvinceknight:
> Actually, without it being in the requirements.txt how did tests run on travis? Or do we not have tests for it?

Ah because numpy is a dependancy of matplotlib... Cool... Still adding it to the requirements.txt, but putting it in a different order.

PR #279, opened 2015-08-18
Adding numpy to the requirements.txt

drvinceknight:
As pointed out by #158 we now use numpy so am adding it to the requirements.txt. **Wait to make sure that travis passes** I know that numpy can make things temperamental sometimes...


marcharper:
:+1:

Issue #280, opened 2015-08-18
Segfaults with cheating strategies


marcharper:
There is some kind of deepcopy recursion loop (?) going on with mindreader and the latest changes in `round_robin.py` (from the 220 branch). Excluding the cheating strategies seems to eliminate the issue, as does simply removing MindReader and ProtectedMindReader from `_strategies.py`

I've tried to [refactor mind-reader](https://github.com/marcharper/Axelrod/tree/segfault) to handle this issue but I don't seem to have solved it.

marcharper:
Possible workarounds for now:
* Strategies can mark themselves as "don't copy" with a class variable, which should be fine for any strategy that uses deep copy and does not take parameters in their `__init__` method. In this case we can fall back to the former method of grabbing the class and instantiating a new copy
* Use inspect to identify strategies that use the copy module and don't copy them

I'm not sure what to do if there is eventually a strategy that uses copy and also needs parameters for it's init method, but I guess we can worry about that later.

meatballs:
Trying to use deepcopy on a strategy that also uses copy sounds like a nightmare to get right!

We probably ought to agree a default (either deepcopy or instantiation of a new object) and then use a class variable to mark those that need to use the other.

Alternatively, we could use deepcopy as the default but instantiation for cheating strategies as I can't think of how a strategy could be using copy and yet still be within the non-cheating bounds.

As @marcharper says, we'll have a problem with anything that uses copy and also wants __init__ parameters!

marcharper:
We could also take the copy out of MindReader. I put that in a while back because it was not otherwise possible to rollback the effect of play on the opponent (since internal variables could be set during simulated play and not reversed along with the history).

In that case MindReader would not work exactly as specified, but it's a cheating strategy anyway, so I guess it doesn't matter if it messes with the opponent in various ways.

marcharper:
So I went ahead and changed MindReader since it's the only strategy that uses copy. We can use `inspect.getargspec()` on `__init__` methods to determine if there is more than one argument, and if so, copy instead of creating a new instance. This is also more efficient. I pushed both changes to my branch listed above. There is a PR for that branch if we are happy with this.

PR #281, opened 2015-08-19
Segfault

marcharper:
Re: #280


drvinceknight:
This seems good to me and fixes a bug.

PR #283, opened 2015-08-19
adds tournament length to player2 on self-interactions

### LINKED ISSUE #253: BackStabber and DoubleCrosser self interactions error

meatballs:
fixes #253

### DIFF
--- a/axelrod/round_robin.py
+++ b/axelrod/round_robin.py
@@ -84,10 +84,10 @@ class RoundRobin(object):
             # deepcopy rather than instantiating the class variable
             argspec = inspect.getargspec(player1.__init__)
             if len(argspec.args) > 1:
-                player1.reset()
                 player2 = copy.deepcopy(player1)
             else:
                 player2 = class1()
+                player2.tournament_length = self.turns
             class2 = class1
         else:
             player2 = self.players[player2_index]


marcharper:
I'm fine with this. The reset before the deepcopy does increase the efficiency of the copy though, since it's not copying any unnecessary history or internal variables. (I know it gets reset later.) We could also reset after each head-to-head match is over.

meatballs:
At the point where the copy takes place, the object has only just been instantiated. It has no history or anything else, so I don't believe the reset is necessary.

We reset just before each interaction - lines 105 and 106 of RoundRobin - so I'm not sure what a reset afterwards would add.

marcharper:
Ok, thanks for the explanation. It seemed like the reset was making a difference in runtime on my machine, but it's slight and it could be do to other things (perhaps other loads on the machine).

This is on :+1: to merge IMO.

marcharper:
Merging under the unspoken "bug fix doesn't need two people" rule.

PR #288, opened 2015-08-26
Adding docs for non default arguments

drvinceknight:
- Adding an example and plot for different turns and repetitions
- Adding a different game example
- For both of those am also including a new plot


meatballs:
:+1:

marcharper:
Do we not require numpy now?

drvinceknight:
> Do we not require numpy now?

Do you mean re:

> The whole library can be used without any other non base Python libraries
> however if you have `matplotlib <http://matplotlib.org/>`_ installed on your
> system (this is installed automatically if you used :code:`pip install axelrod`)
> there are various graphical things coded in and ready to go.

That's old stuff and should have been changed with the PR bringing numpy in as a requirement. I'll fix it with this PR.

drvinceknight:
Done: good catch :)

PR #289, opened 2015-08-26
Creates tournament attributes dictionary for players

### LINKED ISSUE #286: Tournament to Player Dictionary
Pass a dictionary from a tournament to all its players containing entries for tournament length and game matrix.

Replace the current tournament length mechanism with this new dictionary.

meatballs:
Fixes #286

### DIFF
--- a/axelrod/tests/unit/test_tournament.py
+++ b/axelrod/tests/unit/test_tournament.py
@@ -49,7 +49,8 @@ class TestTournament(unittest.TestCase):
             processes=4,
             noise=0.2)
         self.assertEqual(len(tournament.players), len(self.players))
-        self.assertEqual(tournament.players[0].tournament_length, 200)
+        self.assertEqual(
+            tournament.players[0].tournament_attributes['length'], 200)
         self.assertEqual(tournament.game.score(('C', 'C')), (3, 3))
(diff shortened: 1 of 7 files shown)


axelrod/tests/unit/test_tournament.py
```diff
@@ -49,7 +49,8 @@ def test_init(self):
             processes=4,
             noise=0.2)
         self.assertEqual(len(tournament.players), len(self.players))
-        self.assertEqual(tournament.players[0].tournament_length, 200)
+        self.assertEqual(
+            tournament.players[0].tournament_attributes['length'], 200)
```
drvinceknight: Could we include a test here for the game attribute also?

And perhaps include tests for non default values?

meatballs:
I've added a test for the game value in the tournament attributes dictionary and I've also added a non-default 'turns' value.

I don't think non-default game values belong here. If we want them, they should be in test_game.py and that's not really part of this ticket.

drvinceknight:
> I've added a test for the game value in the tournament attributes dictionary and I've also added a non-default 'turns' value.
> 
> I don't think non-default game values belong here. If we want them, they should be in test_game.py and that's not really part of this ticket.

Yeah, sounds good to me. 

I'm happy with this :+1:

PR #292, opened 2015-09-05
Display names

marcharper:
Just changes the display names of a few strategies so that multiple instances do not have the same name. For example, Random is now named "Random 0.5", and the name changes if you instantiate with different parameters. This way if you have multiple Random players (with different internal probabilities), they don't show up with the same name.

### DIFF
--- a/axelrod/strategies/gobymajority.py
+++ b/axelrod/strategies/gobymajority.py
@@ -41,7 +41,12 @@ class GoByMajority(Player):
     def __repr__(self):
         """The string method for the strategy."""
         memory = self.memory_depth
-        return 'Go By Majority' + (memory > 0) * (":%i" % memory)
+        name = 'Go By Majority' + (memory > 0) * (":%i" % memory)
+        if self.soft:
+            name = "Soft " + name
+        else:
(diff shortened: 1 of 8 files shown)


axelrod/strategies/gobymajority.py
```diff
@@ -41,7 +41,12 @@ def strategy(self, opponent):
     def __repr__(self):
         """The string method for the strategy."""
         memory = self.memory_depth
-        return 'Go By Majority' + (memory > 0) * (":%i" % memory)
+        name = 'Go By Majority' + (memory > 0) * (": %i" % memory)
+        if self.soft:
+            name = "Soft " + name
+        else:
+            name = "Hard " + name
```
drvinceknight: Could you throw in some tests for `Hard`?

drvinceknight:
Can you throw something in the documentation somewhere about the repr? The docs are going to eventually need a massive overhaul but I think it would be nice to just included a quick description and example?

marcharper:
Will do.

drvinceknight:
I'm happy with this :+1:

PR #293, opened 2015-09-08
Adds behaviour dictionary to Player class

### LINKED ISSUE #291: Is a strategy that pays attention to which strategy it is playing against considered "honest"?
https://axelrod.readthedocs.org/en/latest/contributing.html#is-your-strategy-honest says:

"The rules for an ‘honest’ strategy are very simple:

* It does not change what it’s opponents do/know.
* It forgets everything every time it starts playing someone (this is implemented with the reset method)."

Consider a strategy that obtains the class name or code of its opponent (so that it is able to simulate, but not change, what its opponent will do in response to its own move). Is this honest? It seems to meet the definition.

meatballs:
Fixes #291

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -37,10 +37,12 @@ class Player(object):
             "random" in inspect.getsource(self.__class__))
         if self.name == "Player":
             self.behaviour['stochastic'] = False
+        if 'memory_depth' not in self.behaviour:
+            self.behaviour['memory_depth'] = None
         if 'inspects_opponent_source' not in self.behaviour:
             self.behaviour['inspects_opponent_source'] = None
-        if 'updates_opponent_souce' not in self.behaviour:
-            self.behaviour['updates_opponent_souce'] = None
+        if 'updates_opponent_source' not in self.behaviour:
+            self.behaviour['updates_opponent_source'] = None
         self.tournament_attributes = {'length': -1, 'game': None}
         self.cooperations = 0
         self.defections = 0
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -46,10 +46,7 @@ class Player(object):
         for dimension in self.default_behaviour:
             if dimension not in self.behaviour:
                 self.behaviour[dimension] = self.default_behaviour[dimension]
-        self.tournament_attributes = {
-            'length': None,
-            'game': None
-        }
+        self.tournament_attributes = {'length': -1, 'game': None}
         self.cooperations = 0
         self.defections = 0
 
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -28,20 +28,16 @@ class Player(object):
     """
 
     name = "Player"
-    behaviour = {
-        'stochastic': None,
-        'reads_source': None,
-        'modifies_source': None,
-        'memory_depth': None
-    }
+    behaviour = {}
 
     def __init__(self):
         """Initiates an empty history and 0 score for a player."""
         self.history = []
-        self.behaviour['stochastic'] = "random" in inspect.getsource(self.__class__)
-        self.tournament_attributes = {'length': -1, 'game': None}
+        self.behaviour['stochastic'] = (
+            "random" in inspect.getsource(self.__class__))
         if self.name == "Player":
             self.behaviour['stochastic'] = False
+        self.tournament_attributes = {'length': -1, 'game': None}
         self.cooperations = 0
         self.defections = 0
 
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -37,10 +37,12 @@ class Player(object):
             "random" in inspect.getsource(self.__class__))
         if self.name == "Player":
             self.behaviour['stochastic'] = False
+        if 'memory_depth' not in self.behaviour:
+            self.behaviour['memory_depth'] = float('inf')
         if 'inspects_opponent_source' not in self.behaviour:
             self.behaviour['inspects_opponent_source'] = None
-        if 'updates_opponent_souce' not in self.behaviour:
-            self.behaviour['updates_opponent_souce'] = None
+        if 'manipulates_opponent_state' not in self.behaviour:
+            self.behaviour['manipulates_opponent_state'] = None
         self.tournament_attributes = {'length': -1, 'game': None}
         self.cooperations = 0
         self.defections = 0
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -29,6 +29,12 @@ class Player(object):
 
     name = "Player"
     behaviour = {}
+    default_behaviour = {
+        'stochastic': False,
+        'memory_depth': float('inf'),
+        'inspects_opponent_source': None,
+        'manipulates_opponent_state': None
+    }
 
     def __init__(self):
         """Initiates an empty history and 0 score for a player."""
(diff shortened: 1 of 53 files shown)


axelrod/player.py
```diff
@@ -28,14 +28,22 @@ class Player(object):
...
         """Initiates an empty history and 0 score for a player."""
         self.history = []
-        self.stochastic = "random" in inspect.getsource(self.__class__)
-        self.tournament_attributes = {'length': -1, 'game': None}
+        self.behaviour['stochastic'] = (
+            "random" in inspect.getsource(self.__class__))
         if self.name == "Player":
-            self.stochastic = False
+            self.behaviour['stochastic'] = False
+        if 'memory_depth' not in self.behaviour:
```
drvinceknight: These could all be kept in a default_behaviour dictionary and this could be replaced with a loop over the keys of it yes?

Also, I se why you're doing this: I can create a new player and only add the attributes that are non default. If I added a new classifying dimension I would either need to add another line like the one you have here are simply add to the `default_behaviour` I'm suggesting right?

meatballs: > These could all be kept in a default_behaviour dictionary and this could be replaced with a loop over the keys of it yes?

Yes, that would be neater. I'll sort that.

meatballs: > Also, I se why you're doing this: I can create a new player and only add the attributes that are non default. If I added a new classifying dimension I would either need to add another line like the one you have here are simply add to the default_behaviour I'm suggesting right?

Yep - that's the idea!

drvinceknight: Cool, am I correct in saying that to add a new dimension, I would add it in two places:

- the specific player I'm adding
- this default dictionary

Just checking I understand (we'll want to document that at some point).

meatballs: Yes - but if you only added it to the default dictionary, it wouldn't matter. All the players would pick up that default.

drvinceknight: Yeah that's really nice. I really like the fact that you don't have to set everything for everybody.

axelrod/tests/unit/test_player.py
```diff
@@ -20,7 +20,9 @@ class TestPlayerClass(unittest.TestCase):
 
     name = "Player"
```
drvinceknight: Should we include some tests here for the default dictionary etc?

meatballs: We should - but that's one of the four additional tasks I described in the gitter channel.

I think each strategy test needs a ```expected_behaviour``` dictionary with all dimensons specified.

drvinceknight: That's true re other strategy test but don't we want to test the specific
initialisation of the global Player class? That feels like it should be
here and in this PR (I realise you've done loads on this and sound busy so
I'm happy to take that on).

meatballs: Tricky to do - anything you write needs to take into account the fact that the dictionary will be overridden by all the inherited classes and, unless they have all dimensions specified, you will hit missing key errors.

drvinceknight: I'm just suggesting you test the functionality for in instance of the
`Player` class. So that upon initialisation of the `Player` class it's
behaviour dictionary is the same as the default dictionary.

I think it's a pretty trivial test but at least it checks that the
initialisation dictionary doesn't do something weird for the `Player`
class. I think I'm suggesting something like:

```
self.assertEqual(player.behaviour, default_behaviour) # we'd just need to
grab/create default_behaviour
```

For each inherited class, as you said above I agree we want to test them
individually.

meatballs: but all the tests are also inherited - from TestPlayer. Anything you put in its ```test_initialisation``` method will also be used by all the tests for all the strategies.

drvinceknight: Aaaah... Yes. I suppose we could create another specific method for this class, that would then also be inherited but is never called? I'm just thinking that it would be possible for someone to change something in init that breaks things...

meatballs: I think the TestPlayer class needs to loop over an ```expected_behaviour``` dictionary and, for each key in that dictionary, check that it exists in the player under test and that the values match. That will cope with existing tests which don't have all the dimensions specified.

But I also think that should be a separate ticket.

drvinceknight: Yeah I've thought about it more during my walk and agree (on both counts).

meatballs: > Yeah I've thought about it more during my walk

No decent idea ever occurs to me whilst sat on my arse!

drvinceknight: That probably explains why I very rarely have good ideas :) ;)

I'll look through the rest of this in a couple of hours time but expect
it's a slam dunk. I'll go ahead and open the issues for the 4 remaining
tasks you mentioned on gitter afterwards?

drvinceknight:
Assuming the tests pass (I've just restarted the Travis build): I'm happy with this: :+1: - I'm just going to open the subsequent issue that will build on this work.

drvinceknight:
Having said that: my one slightly 'humph' with it is that I think I'd rather the dictionary be called `classifier` as opposed to behaviour... If someone agrees with me I'm happy to refactor that, if someone is happy with `behaviour` let's just ignore me :)

drvinceknight:
Looks like we have an actual failed test:

```
20.60s$ coverage run --source=axelrod -m unittest discover
...........E...............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
ERROR: test_full_tournament (axelrod.tests.integration.test_tournament.TestTournament)
A test to check that tournament runs with all non cheating strategies.
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tests/integration/test_tournament.py", line 32, in test_full_tournament
    output_of_tournament = tournament.play().results
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tournament.py", line 49, in play
    self._run_serial_repetitions(self._outcome)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tournament.py", line 82, in _run_serial_repetitions
    self._run_single_repetition(outcome)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tournament.py", line 75, in _run_single_repetition
    output = self._play_round_robin()
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tournament.py", line 146, in _play_round_robin
    return round_robin.play()
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/round_robin.py", line 40, in play
    player1_index, player2_index)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/round_robin.py", line 58, in _score_single_interaction
    player1, player2, classes)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/round_robin.py", line 109, in _play_single_interaction
    player1.play(player2, self._noise)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/player.py", line 75, in play
    s1, s2 = self.strategy(opponent), opponent.strategy(self)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/strategies/meta.py", line 34, in strategy
    results = [player.strategy(opponent) for player in self.team]
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/strategies/backstabber.py", line 18, in strategy
    if len(opponent.history) > (self.tournament_attributes['length'] - 3):
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

drvinceknight:
I've had to revert c9d011b - which set the tournament length to `None`, tests needed it to be numeric, I thought keeping it as `-1` was ok...

drvinceknight:
Closing this as it has been included in #295

PR #295, opened 2015-09-11
294 - adds classifier dictionary and applies to all strategies with tests

drvinceknight:
This applies the tests and classifiers for all strategies.

I have also included a test for the known dimension, this is mainly to stop a potential declassification of something... Also potentially if a new dimension is added _and applied_ to all the strategies it could be added to this test.


axelrod/strategies/backstabber.py
```diff
@@ -30,8 +34,12 @@ class DoubleCrosser(Player):
     """
 
     name = 'DoubleCrosser'
-    memory_depth = float('inf')  # Long memory
-    stochastic = False
+    behaviour = {
```
marcharper: Do we want `'stochastic': False` for the backstabber strategies as well?

marcharper:
I'm not sure it's necessary to have manipulates* = False for all these strategies since the default None should suffice, is there a reason to include it everywhere? For added clarity? (Just curious)

axelrod/strategies/calculator.py
```diff
@@ -11,7 +11,13 @@ class Calculator(Player):
     """
 
     name = "Calculator"
-    memory_depth = float('inf')
+    behaviour = {
+        'memory_depth': float('inf'),
+        'stochastic': False,
```
marcharper: This one actually is stochastic, since the first 20 rounds are Joss, and it is stochastic. Looks like it was misclassified before the PR.

axelrod/strategies/darwin.py
```diff
@@ -23,7 +23,14 @@ class Darwin(Player):
     """
 
     name = "Darwin"
-    memory_depth = float('inf')
+    behaviour = {
```
marcharper: Darwin "cheats" because it retains state between tournaments. I guess we don't have a dimension for that.

axelrod/strategies/forgiver.py
```diff
@@ -8,7 +8,12 @@ class Forgiver(Player):
     """
 
     name = 'Forgiver'
-    memory_depth = float('inf')  # Long memory
+    behaviour = {
```
marcharper: stochastic = False?

axelrod/strategies/forgiver.py
```diff
@@ -28,7 +33,12 @@ class ForgivingTitForTat(Player):
     """
 
     name = 'Forgiving Tit For Tat'
-    memory_depth = float('inf')  # Long memory
+    behaviour = {
```
marcharper: stochastic = False?

marcharper:
So behavior is a class variable? That could cause problems when a tournament has multiple instances of the same player that differ on some attribute. This can occur for `memory_depth` (GoByMajority) and `stochastic` (MemoryOne players, GTFT, etc.)

axelrod/strategies/grudger.py
```diff
@@ -16,10 +22,16 @@ def strategy(self, opponent):
 
 class ForgetfulGrudger(Player):
     """A player starts by cooperating however will defect if at any point the
-    opponent has defected, but forgets after meme_length matches."""
+    opponent has defected, but forgets after mem_length matches."""
 
     name = 'Forgetful Grudger'
-    memory_depth = float('inf')  # Long memory
+    behaviour = {
+        'memory_depth': float('inf'),  # Long memory
```
marcharper: This one might be memory_depth = 10

axelrod/strategies/grumpy.py
```diff
@@ -7,7 +7,12 @@ class Grumpy(Player):
     when the opponent co-operates."""
 
     name = 'Grumpy'
-    memory_depth = float('inf')  # Depends on internal memory.
+    behaviour = {
```
marcharper: stochastic = False?

axelrod/strategies/inverse.py
```diff
@@ -7,7 +7,13 @@ class Inverse(Player):
     long ago the opponent defected."""
 
     name = 'Inverse'
-    memory_depth = float('inf')  # Long memory
+    behaviour = {
+        'memory_depth': float('inf'),  # Long memory
+        'stochastic': False,
```
marcharper: This one uses random, so seems like stochastic = True (though Player might catch it...)

axelrod/strategies/memoryone.py
```diff
@@ -9,7 +9,13 @@ class WinStayLoseShift(Player):
     """Win-Stay Lose-Shift, also called Pavlov."""
 
     name = 'Win-Stay Lose-Shift'
-    memory_depth = 1  # Four-Vector = (1,0,0,1)
+    behaviour = {
+        'memory_depth': 1,  # Memory-one Four-Vector = (p, p, p, p)
```
marcharper: The four-vector is (1, 0, 0, 1) [maybe copypasta?]

axelrod/strategies/memoryone.py
```diff
@@ -39,16 +45,21 @@ class MemoryOnePlayer(Player):
     with a initializing four_vector."""
 
     name = 'Generic Memory One Player'
-    memory_depth = 1
+    behaviour = {
+        'memory_depth': 1,  # Memory-one Four-Vector = (p, p, p, p)
```
marcharper: The Four-vector here is unknown (it's an abstract base class essentially). Whether it is stochastic or not is also still unknown.

axelrod/strategies/mindreader.py
```diff
@@ -46,15 +46,25 @@ class MindReader(Player):
     """A player that looks ahead at what the opponent will do and decides what to do."""
 
     name = 'Mind Reader'
-    stochastic = True # Don't cache me
+    behaviour = {
+        'memory_depth': -10,
+        'stochastic': False,
```
marcharper: I think this one is stochastic, since the simulated play could be stochastic. MindReader alters opponent state rather than the source (so those two booleans are flipped).

I'm not sure if it actually inspects the source -- it doesn't explicitly read the source with `inspect`, but it does use the source to simulate. Let's be sure that the docs explain that both behaviors are included.

axelrod/strategies/oncebitten.py
```diff
@@ -8,7 +8,13 @@ class OnceBitten(Player):
     """
 
     name = 'Once Bitten'
-    memory_depth = float('inf')  # Long memory
+    behaviour = {
+        'memory_depth': float('inf'),  # Long memory
```
marcharper: Again perhaps this one is memory_depth =12 (10 rounds of altered behavior following two rounds of opponent defection)

axelrod/strategies/retaliate.py
```diff
@@ -8,8 +8,12 @@ class Retaliate(Player):
     A player starts by cooperating but will retaliate once the opponent
     has won more than 10 percent times the number of defections the player has.
     """
-
-    memory_depth = float('inf')  # Long memory
+    behaviour = {
```
marcharper: stochastic = False ?

axelrod/strategies/retaliate.py
```diff
@@ -70,7 +74,12 @@ class LimitedRetaliate(Player):
     retaliation limit (20 defections).
     """
 
-    memory_depth = float('inf')  # Long memory
+    behaviour = {
```
marcharper: stochastic = False ?

axelrod/strategies/titfortat.py
```diff
@@ -18,7 +23,12 @@ class TitFor2Tats(Player):
     """A player starts by cooperating and then defects only after two defects by opponent."""
 
     name = "Tit For 2 Tats"
-    memory_depth = 2  # Long memory, memory-2
+    behaviour = {
```
marcharper: stochastic = False

axelrod/strategies/titfortat.py
```diff
@@ -7,7 +7,12 @@ class TitForTat(Player):
     """A player starts by cooperating and then mimics previous move by opponent."""
 
     name = 'Tit For Tat'
-    memory_depth = 1  # Four-Vector = (1.,0.,1.,0.)
+    behaviour = {
```
marcharper: stochastic = False

drvinceknight:
> So behavior is a class variable? That could cause problems when a tournament has multiple instances of the same player that differ on some attribute. This can occur for memory_depth (GoByMajority) and stochastic (MemoryOne players, GTFT, etc.).
> 
> Edit: I see that MemoryOne updates the attributes on instantiation through super (e.g. GTFT) but this could still be an issue for e.g. Joss and others.

I think I've found an actual problem with MemoryOne:

```
>>> import axelrod
>>> c = axelrod.MemoryOnePlayer((1,0,0,1))
>>> d = axelrod.MemoryOnePlayer((1,.25,0,1))
>>> c.behaviour == d.behaviour
True
```

Where really `c.behaviour['axelrod']` should be `False`.

What do we think is the best way to deal with this? Change `behaviour` to not be a class attribute or just make sure `__super__` is being used in the right places?

drvinceknight:
> What do we think is the best way to deal with this? Change behaviour to not be a class attribute or just make sure __super__ is being used in the right places?

I think I think that `__super__` being used in the right place is the way to go, I would suggest that:

```
>>> import axelrod
>>> c = axelrod.Cooperator()
>>> d = axelrod.Cooperator()
>>> c.behaviour['stochastic'] = True  # Not that you would ever need to do that
>>> c.behaviour == d.behaviour
True
```

is the right behaviour...

axelrod/strategies/darwin.py
drvinceknight: Good point. We need a dimension for it... What about: `modifies_own_state` ?

axelrod/strategies/mindreader.py
drvinceknight: > I think this one is stochastic, since the simulated play could be stochastic. MindReader alters opponent state rather than the source (so those two booleans are flipped).

I thought the same, __but__ it's only actually stochastic if it plays against a stochastic player. MindReader itself is not stochastic so should be cached against another non stochastic player.

drvinceknight: > I'm not sure if it actually inspects the source -- it doesn't explicitly read the source with inspect, but it does use the source to simulate. Let's be sure that the docs explain that both behaviors are included.

I agree, wasn't too sure what to do without creating another dimension (which I would not like to do). I suggest leaving it as it is but documenting? Or is it just better to leave just having `manipulates_opponents_state` which arguably is a nicer way of putting it...

drvinceknight:
A part from the couple of places I've commented, I believe I've addressed all the mis categorisations.

Just going to see if I can go through and change the initialisation stuff for the relevant strategies...

meatballs:
Surely, the whole point of a class is to group objects which have similar behaviour. If we have classes whose instances differ fundamentally in their behaviour, then our definition of either those classes or the behaviour we're considering must be flawed.

drvinceknight:
> Surely, the whole point of a class is to group objects which have similar behaviour. If we have classes whose instances differ fundamentally in their behaviour, then our definition of either those classes or the behaviour we're considering must be flawed.

Strategies like `Joss` which depending on certain input parameters act stochastically or not, do make sense to be in the same class as they exhibit overall similarly behaviour. 

Going back to the idea of being able to pick strategies according to behaviour:

```
strategies = [s() for s in axelrod.strategies if s().behaviour['stochastic']]
```

I think I agree with you @meatballs that we need to imply that by the behaviour dictionary we imply the defaults. Going back to `Joss`, it wouldn't necessarily make sense to have:

```
strategies = [s() for s in axelrod.strategies if s(p=1).behaviour['stochastic']]
```

Have said that it _is_ possible to create `axelrod.Joss(p=1)` which is not stochastic but sadly then changes all `Joss`s to have `behaviour['stochastic']=False`. One option is to change the `__init__` to call the super which I don't think is currently working correctly:

```
>>> import axelrod
>>> a = axelrod.Joss()
>>> a.behaviour['stochastic']  # This is giving the correct output
True
>>> b = axelrod.Joss(1)
>>> b.behaviour['stochastic']  # This is giving the correct output
False 
>>> a.behaviour['stochastic']  # This has now changed
False
```

At the same time it is imaginable that we want the following to work:

```
>>> strategies = [axelrod.Joss(x / 10.0) for x in range(11)]
>>> [s for s in strategies if s.behaviour['stochastic']]
```

which would give the wrong output... I think that as long as we can get the `__super__` to work correctly in the init we should be ok no?

meatballs:
This won't work if there are two Joss instances in the same tournament where one is stochastic and the other isn't. Both instances will belong to the same class and will, therefore, have the same behaviour dictionary.

drvinceknight:
> This won't work if there are two Joss instances in the same tournament where one is stochastic and the other isn't. Both instances will belong to the same class and will, therefore, have the same behaviour dictionary.

Yeah I agree and that's not ok because of the use cases I've suggested above. Is the simplest to make the attribute no longer a class attribute? 

One other option (which is messy) is that for the strategies in question, to create a copy of the behaviour dictionary on init which is messy...:

```
b = axelrod.Joss(p=1)
b.behaviour = copy.copy(b.behaviour)
b.behaviour['stochastic'] = False
```

meatballs:
If behaviour is an instance attribute, how could you then generate a list of players to pass to a tournament? e.g. How would you derive a list of deterministic strategies and pass that list to a new tournament object?

This feels wrong to me. The idea of having instances of the same class with differing behaviour really suggests something's not right.

drvinceknight:
> If behaviour is an instance attribute, how could you then generate a list of players to pass to a tournament? e.g. How would you derive a list of deterministic strategies and pass that list to a new tournament object?

Well this would work no?:

```
strategies = [s() for s in axelrod.strategies if s().behaviour['stochastic']]
```

> 
> This feels wrong to me. The idea of having instances of the same class with differing behaviour really suggests something's not right.

It's just that some strategies are parametrised, and depending on the parameters different behaviour occurs - I don't think there's anything wrong with that. Anything with variable memory length for example, that surely depends in the same class no (assuming they do the same thing for different memory lengths)?

drvinceknight:
I think I like the copy idea but possibly for every strategy...

So each player when it initiates gets a copy of the class dictionary (which in essence corresponds to the default behaviour), this allows us to have a dictionary for a class but then also every instance has it's own dictionary that can be modified without affecting the overall class...

meatballs:
> So each player when it initiates gets a copy of the class dictionary (which in essence corresponds to the default behaviour), this allows us to have a dictionary for a class but then also every instance has it's own dictionary that can be modified without affecting the overall class...

I suppose that's about the best we can do. Still smells bad to me, though!

drvinceknight:
So I've just pushed two commits that try this out: 

- 63ff631 Adds a tests that checks things for Joss - not suggesting this stays where it is. Let me know what you think.
- 0d842b8 Adds in the copy: this passes all previous tests as well as the above test.

This now means we have the following behaviour:

```
>>> import axelrod
>>> a = axelrod.Joss()
>>> b = axelrod.Joss(p=1)
>>> a.behaviour['stochastic']
True
>>> b.behaviour['stochastic']
```

but that the same time we still have:

```
>>> axelrod.Joss.behaviour['stochastic']
False
```

I think this is actually quite nice as it sets up the class behaviour as the default behaviour but gives the flexibility to each instance.

Let me know what you think, if we're happy I'll try and find the class that need the __supr__ to be thrown in.

drvinceknight:
(I just edited that as there were a couple of errors in the actual code output.)

axelrod/strategies/mindreader.py
drvinceknight: What about just changing all of these to be:

```
manipulates_state
inspects_source
manipulates_source
```

?

Any thoughts on this @meatballs ?

meatballs:
> Let me know what you think

Go for it

axelrod/strategies/mindreader.py
drvinceknight: Woops that wasn't supposed to go here.

axelrod/strategies/darwin.py
drvinceknight: Reposting this from below comment which was in the wrong spot:

What about just changing all of these to be:

manipulates_state
inspects_source
manipulates_source
?

Any thoughts on this @meatballs ?

meatballs: how would that differentiate between those that modify their own state/source and those that modify their opponent's?

drvinceknight: That's the point: it would not. Not sure if that differentiation is needed? 

I'm suggesting that to deal with the issue raised by @marcharper with `Darwin` which is the only strategy that cheats by modifying it's own state (it doesn't reset properly)...

meatballs: Ah - just got that now. Seems ok for the moment, but I wonder how long it will be before we have extend the dimensions!

drvinceknight: > Ah - just got that now. Seems ok for the moment, but I wonder how long it will be before we have extend the dimensions!

Now to figure out the right sed-foo to automate that change (just managed to make a mess of my local repository :))...

drvinceknight: This has been addressed by fb079f9 (changing the names - if we're not happy can always change back) and e0d1ed5 (sort out Darwin).

marcharper:
I like this solution -- an instance member is accessed before a class member, so this allows the class to retain the "default" behaviors, and prevents corner cases of parameters from breaking things.

I liked "inspects_opponents_source" better since it's more specific. OTOH I guess "inspects_own_source" doesn't make much sense... "manipulates_state" could be ambiguous. For Darwin perhaps "retains state" or "improper reset" is more telling. However, we're getting into corner cases with single exception instances, so if everyone else is happy with the latest commits, I don't object. It's just the mathematician in me that likes as-precise-as-possible definitions. We can always clarify later if needed.

drvinceknight:
I think I've more or less finished now. I've included docs (I really think the docs need a full big section of work so am currently aiming for 'good enough' but if there's something that really needs to go in that I've missed out please let me know) and also changed how the strategies are amalgamated using a new `is_cheater` function (which is tested etc). All strategies can now be found in `axelrod.strategies` which is basically where anyone can go and use the classifiers as required.

One side effect of this is that I don't think `run_axelrod` has the correct behaviour. This is because `ordinary_strategies` now contains `basic_strategies`, I've done this because I thought it was perhaps a bit nicer (basics are ordinary) and I thought it would be an easy fix in `run_axelrod` but in all honesty I got a bit lost going through the tournament, tournament manager, utils etc... 

I wonder if someone else could take a look at that, if it's complicated and/or people don't have time I could always set the `ordinary_strategies` to not contain the `basic_strategies`. 💪👍🎆

marcharper:
`ordinary_strategies` should contain `basic_strategies`, so that seems fine. We could define basic_strategies with a function, but that can go in later.

drvinceknight:
> ordinary_strategies should contain basic_strategies, so that seems fine. We could define basic_strategies with a function, but that can go in later.

Oh really? Cool :) I thought it didn't for some reason. Awesome.

Yeah I agree re function, but suggest that's a further PR :)

langner:
Wow, this is a lot of stuff to go through.  All looks generally fine. My one comment is that my own feeling is that the name *behaviors* is somewhat misleading. This attribute is more like an *annotation* on our part where we try to describe and classify various strategies. Strategies themselves (and their authors at time of writing) don't respect these annotations so although they can be class attributes for convenience, I see no problem with overwriting them for instances as needed. Especially if some behaviors are conditional on a parameter value or whatever else.

drvinceknight:
> My one comment is that my own feeling is that the name behaviors is somewhat misleading. This attribute is more like an annotation on our part where we try to describe and classify various strategies

I tend to agree. My other suggested name was `classifier` but I'm also happy with `annotations`. Any preference (or indeed other suggestions):

```
player.classifier['stochastic']
# or
player.annotation['stochastic']
```

langner:
> Any preference (or indeed other suggestions):

No preference. My only other observation is that manual classification is tedious and most of these things could be programmatic.

drvinceknight:
> No preference.

I've gone with classifier:  f7890d8

> My only other observation is that manual classification is tedious

Haha: no objection from my part on that one! :)

> and most of these things could be programmatic.

Yeah, that would be awesome. Stochastic still is actually being set programmatically, I wrote them in directly to the dictionary as we had previously had `stochastic=True` written in. Then, when I saw that I thought that, I thought it was actually quite a nice thing to be have written in the source code for contributor to be able to directly see.

If someone wanted to automate it all: that would be great :+1:, I guess my only concern would be making sure that it worked for sure, even for dimensions that are not planned etc... so I would still encourage contributors to write out the dimensions anyway (and the automated thing would act like `stochastic` now acts). Note thought that when he started this @meatballs put in something really cool which means you _don't HAVE to_ classifier everything across anything: it automatically sets unclassified keys to `None`, which makes things _slightly_ less tedious :) 

I suggest this is a conversation for another issue at some point in time perhaps?

drvinceknight:
Preempting a conversation over classifier vs any other name: In case anyone wants to try anything else, here's a sed script (for my previous commit I used this for code but went through docs manually): https://gist.github.com/72f71ddfed1668c6e24f

drvinceknight:
Just a ping to check if people are ok for this to be merged in or if anyone would like more time to look through it?

meatballs:
Fine with me

Issue #303, opened 2015-09-20
Add reproduce method to Player class


marcharper:
Sometimes players need to be duplicated (e.g. when playing against themselves or in a dynamic like the Moran process). Right now we handle this with `copy` in round_robin.py:
```
def _pair_of_players(self, player1_index, player2_index):
        player1 = self.players[player1_index]
        class1 = player1.__class__
        if player1_index == player2_index:
            # Check for arguments in __init__. If more than one (self), then
            # deepcopy rather than instantiating the class variable
            argspec = inspect.getargspec(player1.__init__)
            if len(argspec.args) > 1:
                player2 = copy.deepcopy(player1)
            else:
                player2 = class1()
                player2.tournament_attributes = player1.tournament_attributes
            class2 = class1
        else:
            player2 = self.players[player2_index]
            class2 = player2.__class__
        return player1, player2, (class1, class2)
```

Another (potentially better) solution would be for Player to have a reproduce or duplicate method that
* reapplies any arguments to `__init__` and instantiates a new player
* does not copy the history

The generic function in Player for most strategies would be as follows:
```
def reproduce(self):
    cls = self.__class__
    new_player = cls()
    return new_player
```

In strategies that use parameters in `__init__`, the function can be overridden to pass on those arguments appropriately.

Advantages:
* No more `copy` in round robin, which is slow and prevents strategies from themselves using copy.
* Can update MindReader to not alter opponent state
* Implementing the Moran process will be easier
* Concerns are properly separated (instances of Player handle their own copy constructor)

drvinceknight:
Yeah this looks great :+1:

langner:
I don't understand how using this is better than copy.deepcopy with additional tweaks.

PR #304, opened 2015-09-20
Reproduce method for Player

marcharper:
Re: #303 

I added `self.init_args = ()` to `Player.__init__`. A child strategy that needs to pass through arguments need only set `self.init_args = (...)` (overriding `reproduce` is not necessary). I updated all the appropriate strategies and added tests.

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -28,7 +28,6 @@ def is_cheater(s):
            classifier['manipulates_source'] or\
            classifier['manipulates_state']
 
-
 def update_histories(player1, player2, move1, move2):
     """Updates histories and cooperation / defections counts following play."""
     # Update histories
@@ -73,6 +72,7 @@ class Player(object):
         self.tournament_attributes = {'length': -1, 'game': None}
         self.cooperations = 0
         self.defections = 0
+        self.init_args = ()
 
     def __repr__(self):
         """The string method for the strategy."""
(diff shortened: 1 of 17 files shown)


axelrod/strategies/axelrod_tournaments.py
```diff
@@ -29,6 +29,7 @@ class Davis(Player):
     def __init__(self, rounds_to_cooperate=10):
```
langner: Maybe explain what rounds_to_cooperate is about in the docstring?

axelrod/player.py
```diff
@@ -98,6 +98,15 @@ def play(self, opponent, noise=0):
             s1, s2 = self._add_noise(noise, s1, s2)
         update_histories(self, opponent, s1, s2)
 
+    def reproduce(self):
```
langner: Why not just use copy.deepcopy and reset the history?

marcharper: * Using copy in a strategy can lead to deep-copying loops, we had to alter MindReader when round_robin was initially changed to use copy
* deepcopy is very slow and unnecessary for this purpose, there's no reason to copy complex objects

axelrod/strategies/axelrod_tournaments.py
marcharper: It's the number of rounds_to_cooperate. We can update the docstring, but frankly few if any of the strategies explain their parameters, so I'd prefer to open up another issue for that.

axelrod/player.py
langner: Oh yes, now I remember. I would say generally that is MindReader's problem and he should have a custom __copy__ method to prevent that. But I do soo the point of this... maybe rename to `clone` though?

marcharper: Sure we could change the name... I also thought about "duplicate". I went with reproduce since it was sort of Darwinian (copy but don't inherit acquired state or history). But whatever everyone prefers is fine.

drvinceknight:
I think I would suggest that because it's darwinian clone is a better option than reproduce. Reproduce implies chance for mutation or change...

uglyfruitcake:
Ok I'm gonna change it to clone and submit a PR.

uglyfruitcake:
The PR is on your branch Marc

meatballs:
:+1:

PR #311, opened 2015-09-21
Fix #310, change titles of plots to say 'turns'

### LINKED ISSUE #310: Change titles of plots to say 'turns' (200).

bricef:


drvinceknight:
:+1: from @meatballs and me.

PR #319, opened 2015-09-21
Add the Mirror cheating strategy

bricef:


axelrod/strategies/mindreader.py
```diff
@@ -94,3 +94,33 @@ def __setattr__(self, name, val):
...
+        Also avoid infinite recursion when called by itself or another mind reader
+        or bender by cooperating.
+        """
+
+        curframe = inspect.currentframe()
+        calframe = inspect.getouterframes(curframe, 2)
+        calname = calframe[1][3]
+
+        if calname in ('strategy', 'simulate_match'):
+            return 'C'
```
marcharper: Can we add a test for this line to make coverage happy?

bricef:
@marcharper Was just doing that :) should now have new test to cover the play vs self case!

drvinceknight:
:+1:

Issue #305, opened 2015-09-21
Update strategies to use tournament attributes properly


marcharper:
We updated the library to provide the game matrix to players from the Tournament class with tournament attributes:
```
            player.tournament_attributes = {
                'length': self.turns,
                'game': self.game}
```

However multiple strategies still pull the game matrix manually from `game.py`, including:
* several in memoryone.py
* q_learners
* MetaWinner
* others?

drvinceknight:
:+1:

marcharper:
One issue that's come up -- strategies need to access the game in their `__init__` methods but don't receive `tournament_attributes` until `Player.__init__` runs.  Sometimes it's possible to run the super `__init__` first but not always...

meatballs:
> strategies need to access the game in their __init__

Do they really need to or is that just how they do things right now? Until they are added to a tournament, it seems right to me that they should have no knowledge of the game that will be played.

meatballs:
It doesn't seem right to me that we should set the tournament attributes to some default in the player __init__ method when that could bear no relation to what's actually set later on by the tournament itself.

marcharper:
Yes some Players need to know e.g. the game matrix to init, unless a lot of computations want to be repeated, or strategy setup left incomplete. An alternative is to have players deal with the tournament attributes in their strategy methods, but it's somewhat silly to have strategies do something like

```
def strategy(self, ...):
    if not self.is_setup:
        self.setup()
    # Strategy as usual
```

This is why I made the `receive_tournament_attributes` function as a callback for delivering the tournament_attributes, so this only need happen once (and not in `strategy`).

We're also duplicating code / having testing issues because only tournament delivers the attributes, making it hard to test a strategy that isn't initialized operated on by a tournament object, which is why I put in the default attributes in `Player.__init__`.

I'm open to other suggestions.

PR #323, opened 2015-09-21
Properly use tournament attributes

marcharper:
This addresses #305. Perhaps not the most elegant solution -- comments welcome!


marcharper:
I'll add some tests if you all are ok with this approach.

meatballs:
:+1:

Issue #318, opened 2015-09-21
Read the docs not automatically updating for docs changes


emmagordon:
e.g. 9 days ago we had commit id '93a243fbff227fd2c3813053ccf12f0085489762' ('Adding more details to the docs'). This included the removal of incorrect instructions (in contributing.rst) to add any new strategy to one of the basic_/original_/cheating_strategies lists, but on read the docs that is still present.

drvinceknight:
This seems to be a known issue caused by matplotlib/numpy. There are various solutions that include mock and a dummy requirements.txt file but I'm failing to get that to work. Working on it.

drvinceknight:
It seems like 89d77ee5a0285d75889d4ef9c8e18872e7d60677 is the first commit where this fails but I see no reason why.

drvinceknight:
```
OPTIONAL LATEX DEPENDENCIES
                dvipng: yes [version 1.14]
           ghostscript: yes [version 9.10]
                 latex: yes [version 3.1415926]
               pdftops: no

============================================================================
                        * The following required packages can not be built:
                        * freetype, png
```

Looks like we were looking for the wrong stuff. The above error message is also present on most successful builds.

drvinceknight:
I am making no progress on this at all, *but* I have set read the docs up so I it no longer builds from master but from the `318` branch. (I really should have done this straightaway).

drvinceknight:
I've opened https://github.com/rtfd/readthedocs.org/issues/1671 on rtds.

PR #326, opened 2015-09-23
docs/strategies overview update

meatballs:
Work from @lvisintini with my fix for the Random link

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -377,14 +377,6 @@ You can see the latest build status `here <https://travis-ci.org/Axelrod-Python/
 Adding the strategy to the documentation
 ''''''''''''''''''''''''''''''''''''''''
 
-To index all the strategies and make sure their docstrings get added to the
-documentation::
-
-    cd docs
-    python auto_generate_strategies_list.py > index_of_strategies.rst
-
-This will write the file that is automatically used by `<https://readthedocs.org/>`_ to generate this `list <http://axelrod.readthedocs.org/en/latest/strategies.html>`_ of strategies.
-
 If you would like to build the documentation locally use::
 
     make html
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -377,6 +377,14 @@ You can see the latest build status `here <https://travis-ci.org/Axelrod-Python/
 Adding the strategy to the documentation
 ''''''''''''''''''''''''''''''''''''''''
 
+To index all the strategies and make sure their docstrings get added to the
+documentation::
+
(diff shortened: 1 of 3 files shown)


lvisintini:
The preview that file on this branch is
https://github.com/Axelrod-Python/Axelrod/blob/lvisintini-docs/strategies-overview-update/docs/overview_of_strategies.rst

the latest revision is a step in the right direction but, it's not looking good right now :(

meatballs:
That is not how it looks on my local build! Not good.

meatballs:
@lvisintini What version of sphinx are you using? I built with 1.3.1 and all looks fine.

meatballs:
I wonder if that's just the github preview making a mess of it.

lvisintini:
Not really building it.
I'm just going by what the github preview looks like.
If the priority is to get it to work with sphinx and you confirm it looks good, then there is not problem.

drvinceknight:
I've checked this locally and all looks great. I noticed one problem though, when I ran locally the index of strategies created an empty list for this page: http://axelrod.readthedocs.org/en/latest/index_of_strategies.html

In 07edf59 and 7d15170 I simply remove the index (as we're about to revamp the docs I think this is ok for now?) but if that's just a local issue to me (I wonder if it's left over problems with #318 ?) then revert my commits and either way this is good to go :+1:

drvinceknight:
Only small thing is that for me `Random` is now in italics (the only strategy like that). I don't think this is an issue.

meatballs:
Odd. On my local build, the index is fine and Random is not in italics!

We need some form of CI for the docs.

meatballs:
@drvinceknight I'm on sphinx 1.3.1 You?

drvinceknight:
Ah that might be it (not sure what v right now away from machine). Revert
my commits (weird though cause never been a problem before). Based on that
this gets a :+1: from me. @marcharper :+1: the previous version of this so
I think you can  go ahead and merge :)

meatballs:
I've created a version of the docs on readthedocs that builds from this branch and I've reversed the two commits from @drvinceknight 

You can see the overview doc here: http://axelrod.readthedocs.org/en/lvisintini-docs-strategies-overview-update/overview_of_strategies.html

and the index document here: http://axelrod.readthedocs.org/en/lvisintini-docs-strategies-overview-update/index_of_strategies.html

drvinceknight:
Yeah all looks awesome. :+1:

drvinceknight:
Am merging as @marcharper had ok'd a previous version of this.

PR #327, opened 2015-09-25
Addresses 328: Adding the axelrod absolute path for autodoc.

drvinceknight:
I have checked readthedocs and this does not break that so this is a simple one.


drvinceknight:
(I actually opened the issue on a random other repo...)

PR #331, opened 2015-09-30
Adds wins to result set

### LINKED ISSUE #248: Plot for number of wins / distribution of wins
In S&P's tournament they give results on how many wins each strategy gets in the tournament (in addition to the average score). It would be nice if the library also generated a plot for the number of wins in addition to the boxplots for mean scores.

meatballs:
Fixes #248

### DIFF
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -46,6 +46,7 @@ class ResultSet(object):
         self.ranking = None
         self.ranked_names = None
         self.payoff_matrix = None
+        self.wins = None
         self.cooperation = None
         self.normalised_cooperation = None
         self.vengeful_cooperation = None
@@ -61,6 +62,7 @@ class ResultSet(object):
             self.ranked_names = self._ranked_names(self.ranking)
             self.payoff_matrix, self.payoff_stddevs = (
                 self._payoff_matrix(self.results['payoff']))
+            self.wins = self._wins(self.results['payoff'])
         if 'cooperation' in self.results and with_morality:
             self.cooperation = self._cooperation(self.results['cooperation'])
             self.normalised_cooperation = (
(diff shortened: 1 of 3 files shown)


docs/usage.rst
```diff
@@ -284,6 +283,18 @@ Recall the ordering of the players::
...
+
+  import axelrod
+  strategies = [s() for s in axelrod.demo_strategies]
+  tournament = axelrod.Tournament(strategies)
+  results = tournament.play()
+  results.wins
+
+The resulting wins list is::
+
+  [0, 80, 0, 20, 30]
```
drvinceknight: Could you add a quick sentence saying something like 'which shows that Cooperator had 0 wins and Defector 80 wins' (just to point at the order that we get out)?

axelrod/tests/unit/test_resultset.py
```diff
@@ -154,6 +155,24 @@ def test_payoff_matrix(self):
         self.assertEqual(self.round_matrix(
             rs.payoff_stddevs, 2), self.expected_stddevs)
 
+    def test_wins(self):
+        rs = axelrod.ResultSet(self.players, 5, 2, self.test_outcome)
+        wins = rs.wins
+        self.assertEqual(wins, self.expected_wins)
+
+    def test_winner(self):
```
drvinceknight: Would a docstring here be helpful? I'm just glancing and not immediately sure what's happening.

marcharper:
Cool! Does this aggregate the wins over tournament repetitions? Just checking if we can make a nice plot of the distribution of wins (like the mean payoff plot).

meatballs:
At the moment, it aggregates, but I can easily split it out and show the wins per repetition, if that's useful.

drvinceknight:
> At the moment, it aggregates, but I can easily split it out and show the wins per repetition, if that's useful.

I think in the interest of not losing data, this is a good shout. You could have a `get_mean_wins_per_repetition` method to complement that (or something less horribly verbose)?

axelrod/result_set.py
```diff
@@ -250,6 +252,62 @@ def _payoff_matrix(self, payoff):
...
+                ]
+
+            i.e. one row per player, containing one element per opponent (in
+            order of player index) which lists payoffs for each repetition.
+
+        Returns:
+            A wins matrix of the form:
+
+                [
+                    [player1 winsin repetition1, player1 wins in  repetition2],
```
marcharper: "winsin" -> "wins in"
"wins in " -> "wins in"

drvinceknight:
This looks great to me! :+1:

PR #332, opened 2015-10-01
simplify cooperations and defections

bootandy:
Turn them in to properties and read them from player.history
Makes code simpler


bootandy:
breaks some strategies fixing.

PR #333, opened 2015-10-01
Fix Darwin outcome attribute

marcharper:
#323 Introduced a bug into Darwin which this PR fixes.


drvinceknight:
What was the bug? Is it easy to add a test that now passes that ensures it's not reintroduced?

marcharper:
`outcomes` should have been `self.outcomes`. The test update should catch it -- it's invoked on the second round of play.

drvinceknight:
:+1:

PR #334, opened 2015-10-01
simplify cooperations and defections2

bootandy:
Turn them in to properties and read them from player.history
Makes code simpler

### DIFF
--- a/axelrod/mock_player.py
+++ b/axelrod/mock_player.py
@@ -16,8 +16,6 @@ class MockPlayer(Player):
         # Do a deep copy just to be safe
         axelrod.Player.__init__(self)
         self.history = copy.deepcopy(player.history)
-        self.cooperations = player.cooperations
-        self.defections = player.defections
         self.move = move
 
     def strategy(self, opponent):
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -35,15 +35,6 @@ def update_histories(player1, player2, move1, move2):
     # Update histories
     player1.history.append(move1)
     player2.history.append(move2)
-    # Update player counts of cooperation and defection
-    if move1 == C:
-        player1.cooperations += 1
-    elif move1 == D:
-        player1.defections += 1
-    if move2 == C:
-        player2.cooperations += 1
-    elif move2 == D:
-        player2.defections += 1
 
 
 class Player(object):
@@ -71,10 +62,16 @@ class Player(object):
         for dimension in self.default_classifier:
             if dimension not in self.classifier:
                 self.classifier[dimension] = self.default_classifier[dimension]
-        self.cooperations = 0
-        self.defections = 0
         self.init_args = ()
         self.set_tournament_attributes()
+
+    @property
+    def cooperations(self):
+        return self.history.count('C')
+
+    @property
+    def defections(self):
+        return self.history.count('D')
 
     def receive_tournament_attributes(self):
         # Overwrite this function if your strategy needs
@@ -132,5 +129,3 @@ class Player(object):
         rest all other attributes.
         """
         self.history = []
-        self.cooperations = 0
-        self.defections = 0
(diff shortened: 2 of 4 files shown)


drvinceknight:
Yeah this simplifies things: I like the look of it. 

Minor question: does this not though (arguably) slow things down a bit as we have to count the C's and D's every time as opposed to just keeping count of them as we go?

drvinceknight:
(Speed is not a terribly sensitive issue so it doesn't mean that it's a no go: just thinking about things)

meatballs:
:+1:

marcharper:
It's O(rounds) vs O(rounds ^ 2) for the strategies that use the properties every round. The impact will depend on what strategies are in use.

We could run `time ./run_axelrod` as a simple benchmark (with and without the patch).

bootandy:
Yes it would slow things down slightly if the other player accessed the property on their strategy call. (Most current strategies dont).

I realise I have added something that does turn O(rounds) into O(rounds^2) which from a pure computer science point of view is a bad thing. 

However this should be balanced against not having to update 4 more variables after each round. And the fact the number of rounds is often low.

As long as the number of rounds a pair of player plays is low (say ~100) I do not think it will have a negative performance impact. However I would like to see the results of the ./run_axelrod benchmark as I may be wrong.

meatballs:
@bootandy I think the simplified code outweighs the performance. If lots of strategies start using it, then that view may change but, at the moment, I like the look of what you've done.

Not sure if you know, but we have a chat channel on gitter which you're welcome to join: https://gitter.im/Axelrod-Python/Axelrod

bootandy:
This branch:
 Starting basic_strategies tournament with 10 round robins of 200 turns per pair.
 Finished all basic_strategies tasks in 1.7s
 Starting strategies tournament with 10 round robins of 200 turns per pair.
 Finished all strategies tasks in 166.5s
 Starting cheating_strategies tournament with 10 round robins of 200 turns per pair.
 Finished all cheating_strategies tasks in 6.5s
  
Old code:
git checkout 0a3a9aa
 
  ./run_axelrod
 Starting basic_strategies tournament with 10 round robins of 200 turns per pair.
 Finished all basic_strategies tasks in 1.7s
 Starting strategies tournament with 10 round robins of 200 turns per pair.
 Finished all strategies tasks in 150.2s
 Starting cheating_strategies tournament with 10 round robins of 200 turns per pair.
 Finished all cheating_strategies tasks in 6.4s

So 150s -> 166s ~ 10% performance decrease. 
I think it is worth it for simpler code and less redundant data.

drvinceknight:
I agree with Owen re readability over performance, and those timing impacts
are minor. :+1:

marcharper:
I'm not sure I agree that this is more simple. It's shorter, of course, but only because it's hiding the calculations in `history.count`. It's computationally more complex, as we all agree.

As for readability, we could alternatively have a History object that keeps the history and updates a `collections.Counter` as the history grows. This object could be instantiated by an existing history (for easy restoration in eg. MindReader, or just copied), and the two attributes `.cooperations` and `.defections` could just call `counter['C']` and `counter['D']`, keeping the property interface for `Player` (or moving it to `Player.history.cooperations`). This also preserves the performance advantage, and allows for future changes/additions to the history without messing with the `Player` class.

drvinceknight:
> Edit: Making the history a class isn't easy at this point -- there are a lot of places in the code that rely on some aspect of the current implementation (e.g. shared history for MetaPlayer). I left my progress here if anyone else wants to give it a go.

I still think that this is more readable (as you say it hides the complexity in `count` which is fairly verbose).  Perhaps and approach could be to merge this as is (I think we're all relatively comfortable with the increased run time?) and perhaps at a later point (once @meatballs has finished his big refactor) we could bring in the History class?

drvinceknight:
What do you think of the above suggestion @marcharper ?

drvinceknight:
On reflection perhaps this can be shelved for now as it would be nicer perhaps to just include whatever the final solution is.

Really appreciate the contribution @bootandy :+1: :)

PR #337, opened 2015-10-03
Add how to run tests to README

bootandy:
Add how to run tests

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,6 +90,11 @@ version of the results:
 
 .. image:: http://axelrod-python.github.io/tournament/assets/strategies_boxplot.svg
 
+Running Tests
+========
+ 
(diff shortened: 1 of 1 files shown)


drvinceknight:
I think something like this was actually in the readme before and we moved it out to the docs. 

Also, with your edit to run with coveralls, if coveralls isn't installed it would fail.

marcharper:
It's in `Axelrod/docs/contributing.rst` (and always has been, IIRC).

drvinceknight:
There was a time before we had docs, that's probably what I'm thinking off :)

bootandy:
You are right it is in the docs. And I shouldn't have used coverage - I just copy pasted what I saw Travis was doing.

PR #340, opened 2015-10-04
Update MetaPlayer reset to also reset team

marcharper:
The MetaPlayer derived strategies retain state between games. I believe this is unintentional.

### DIFF
--- a/axelrod/strategies/meta.py
+++ b/axelrod/strategies/meta.py
@@ -52,6 +52,7 @@ class MetaPlayer(Player):
 
     def reset(self):
         Player.reset(self)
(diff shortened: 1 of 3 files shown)


drvinceknight:
Yeah I think this counts as a bug. Could you add atest to verify the this fixes it (and that subsequent changes don't reintroduce it)?

marcharper:
I don't think this test covers all possibilities, but not knowing the team a priori (and the fact that the histories are shared), it may be the best we can do.

drvinceknight:
Thanks: I agree, this is probably the best we can do :)

:+1: from me :)

Issue #307, opened 2015-09-21
Better documentation for parameters in various strategies 


marcharper:
@langner commented on #304 that we could use more documentation for the arguments in the `__init__` methods of some of the strategies that take parameters.

PR #342, opened 2015-10-04
Add explanations of parameters to docs strings of various strategies __init__ methods

marcharper:
Addresses #307.


drvinceknight:
This is lovely. :+1:

PR #345, opened 2015-10-05
Fix implicit imports for tournament repo

marcharper:
The script update_results.sh in the tournament repository doesn't run after the switch to explicit imports. This fixes it.


drvinceknight:
:+1:

Issue #343, opened 2015-10-04
Classifying Meta Players


marcharper:
I think that the MetaPlayers could be considered cheaters by the standard Axelrod rules because they effectively observe the outcomes of third party plays.

I propose a new classifier dimension, something like "observes_others" or "uses_outside_observations".

drvinceknight:
Yeah, I'm happy with this. :+1: 

However, I have been thinking about cheaters etc for a while (since you, @meatballs and I chatted about this). I suggest we don't change the `is_cheater` definition and keep that strictly to catch stuff that 'messes with code' (as is current). 

Similarly I'd suggest the `persistant_memory` strategies (the ones that use a pesistant set of q values or neural network etc...) would not return `True` for `is_cheater`. 

Perhaps we could have another function that returns True if the strategy obeys the original Axelrod rules? `obey_axelrod`

drvinceknight:
(I mentioned this to @meatballs earlier, I think I'm coming around to what you were both saying, it's just taken me a little while :))

marcharper:
I like `obey_axelrod`. Generally the literature has not well-explored the possibilities of outside observations being incorporated into a strategy.

What the Meta strategies do is a bit different though since it is dynamic. I would not consider a pre-trained neural network to be a cheater, but if it trains on the fly on the output of other games in the same tournament it's not as straightforward IMO.

marcharper:
The Meta strategies could manipulate state -- it would be simple for a strategy to override `getattr` to count how often `strategy` is invoked by the opponent, and update a counter. Not sure where this sits on the traditional cheating spectrum.

They also come really close to "inspecting the source" -- while they don't actually read the opponent's source, they definitely use it for a simulation of sorts (against still other opponents). In other words, the Meta strategies sort of "access the source of opponents and third parties", or if you will, read/copy the opponent's "binary".

It's actually quite an impressive "bending but not quite breaking (?)" of a set of rules.

drvinceknight:
> I like obey_axelrod. Generally the literature has not well-explored the possibilities of outside observations being incorporated into a strategy.

Cool, `obey_axelrod`, should be simple enough to implement right?

>  I would not consider a pre-trained neural network to be a cheater, but if it trains on the fly on the output of other games in the same tournament it's not as straightforward IMO.

[...]

Issue #299, opened 2015-09-18
Documentation revamp


drvinceknight:
I think that the documentation needs a lot of work. It doesn't really serve it's purpose. This is a relic of my initial poor contribution and then a process that built on that by incrementally adding to it.

The only way to understand the library is to read (and remember) the documentation from start to finish. This is because the documentation is a mix/hybrid of a tutorial and a reference document.

My suggestion is along the lines of the following:

- tutorials should be short compartmentalised how to guides
- references should explain what things do but again be compartmentalised and follow the code base
- the final aspect of the documentation is the more 'historical' side: the history of the tournament, the basics of the game theory, the mathematics behind moran processes etc... This is relatively ok I think in the library at present (with some things not in yet).

So I would suggest the docs file structure would look something like:

```
|-- docs/
     intro.rst
     |-- reference
          |-- tournament.rst
          |-- player.rst
          |-- strategy.rst
          ...
     |-- tutorials
          |-- basic_tournament.rst
          |-- ecological_tournament.rst
          |-- noisy_tournament.rst
     |-- backround # for want of a better word? 
          |-- axelrod_first_tournament.rst
          |-- axelrod_second_tournament.rst
          |-- theory_of_repeated_games.rst
          |-- index_of_strategies.rst
```

This would have the advantage of people being able to write new tutorials/references when new things are included. Basically everything becomes modular.

Another important thing I think we should ensure is doctesting the docs. Then have a script that we run to run tests and finally ensure this is on travis. The main reasoning behind this is to ensure that any changes in the code are tested in the docs. 

If everyone thinks this is a long the right lines I'd like to pick an initial stab up :)

marcharper:
Sounds good to me, I presume this subsumes #140, #199, and perhaps #225?

drvinceknight:
It certainly does for #140 and #199 (have closed them). Will leave it up to you for #225, in a way that is kind of an 'ongoing mission'...

marcharper:
We could pull in some of the stuff from my AxelrodExamples repo, like the [example tournaments](https://github.com/marcharper/AxelrodExamples/blob/master/example_tournaments.py), with more explanation and a narrative of course. In particular, duplicating the S&P tournament (and perhaps Tyler's tournament) make sense in line with the goals of the project (IMO).

drvinceknight:
[...]

PR #346, opened 2015-10-05
343 - Change name of is_cheater function.

drvinceknight:
Following on from the discussion on #343 which ended up being another discussion about cheaters. Have changed the `is_cheater` function to be `obey_axelrod` which makes things a bit less subjective (users can then include whatever strategies they want). 

At present it makes no difference to the library (although the tournament repository) but hopefully will clarify all future discussions about the incoming strategies.

I have added something to the documentation but waiting on #299 this is very minor.

### DIFF
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -4,7 +4,7 @@ from __future__ import absolute_import
 from .random_ import random_choice
 from .plot import Plot
 from .game import DefaultGame, Game
-from .player import is_basic, is_cheater, update_histories, Player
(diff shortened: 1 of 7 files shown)


meatballs:
:+1:

marcharper:
:+1: Not sure what the issue with Travis is -- looks like only the 3.5 build did not finish.

drvinceknight:
I've restarted that build :)

drvinceknight:
A part from coveralls this is good to go: will let one of you do the honours :)

Issue #317, opened 2015-09-21
`test_strategy` test in axelrod.tests.unit.test_player.py makes unwarranted assumptions


bricef:
The `test_strategy` method in the `TestPlayer` class makes unwarranted assumptions about the tests for a given strategy. in particular, it assumes that the tests for a given strategy will override the method, but this is not explicitly documented anywhere.

```
    def test_strategy(self):
        """Test that strategy method."""
        self.assertEqual(self.player().strategy(self.player()), None)
```

It feels that the test is really testing that a strategy is a sublcass of `Player`, but that's not really a behavioural test. It would be better to raise a `NotImplmentedError` and check that calling `strategy()` does not return this for all registered strategies.

My suggestions are as follows:

1. Make `strategy()` raise a `NotImplementedError` in the `Player` base class and resolve the broken tests
2. Make `test_strategy()` raise a `NotImplementedError` in the `TestPlayer` class and resolve the broken tests.
3. Change the docs to indicate all test subclasses should implement `test_strategy()`

I'm very strongly in favour of 1, as that's the pythonic and least suprirsing thing to do. It's also the one that would take the most work...

bricef:
Hey @meatballs @drvinceknight @uglyfruitcake What do you guys think?

drvinceknight:
I think I would go for the first option.

meatballs:
Yep, I think option 1 is the way to go. (Sorry it's taken me so long to get around to saying so).

bootandy:
I'd vote for option 1

drvinceknight:
I think that's a consensus :)

Anyone want to pick this up?

marcharper:
FYI (1) breaks about 150 tests.

drvinceknight:
> FYI (1) breaks about 150 tests.

Yeah, we saw that at the pycon sprint... :( I still think it's the better option, what do you rekon @marcharper ?

marcharper:
I'm for it, just a fair warning to anyone that takes it on.

drvinceknight:
Hehe... Yeah...

bootandy:
I'll have a look at this.

PR #350, opened 2015-10-12
Change Player.strategy() to raise Exception instead of None …

bootandy:
By default Player.strategy now raises an exception instead of returning None. For https://github.com/Axelrod-Python/Axelrod/issues/317

Test framework TestPlayer updated to handle this change. Added TestOpponent player in test framework for players to use as an opponent in tests.

Updated TestPlayer framework to force subclasses to implement expected_classifier correctly. This revealed several tests which were not and were relying on the parent class TestPlayer

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -106,7 +106,7 @@ class Player(object):
 
     def strategy(self, opponent):
         """This is a placeholder strategy."""
-        return None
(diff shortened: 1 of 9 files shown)


marcharper:
Very nice! Thanks for taking on #317.

PR #351, opened 2015-10-12
Plot for distribution of head-to-head wins

marcharper:
Re: #248 and #331 (thanks to @meatballs for the heavy lifting!)

This essentially recreates the published list for the [S&P tournament](http://www.nr.com/whp/StewartPlotkinExtortion2012.pdf), which we can now consider "reproduced".

If we are happy with the plot, I'll update the docs and the tournament repository appropriately.

Most Strategies

S&P


meatballs:
:+1:

drvinceknight:
Could you add a quick snippet about generating this to the docs please?

marcharper:
Let me see if I can update the tournament repo so that we can pull the plots from there.

drvinceknight:
This is great, sorry to be a pain could you add some tests also: along the lines of https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_plot.py#L70

marcharper:
I missed one plot (needed to add the new noise plot to the update script). Otherwise the docs look fine when built locally, so this is good to merge once the last plot is added to tournament.

drvinceknight:
> so this is good to merge once the last plot is added to tournament.

I'm happy for you to merge once that's in @marcharper :)

Issue #284, opened 2015-08-21
Full Interaction History


meatballs:
At the moment, we score the interactions between players within the RoundRobin class and return the payoff and cooperation matrix. The detail of the interactions themselves is lost.

Should we have an option to included the full history as an additional entry in the dictionary produced by RoundRobin?

marcharper:
How much memory would that take up? I suppose it could be stored efficiently with a bitstring.

meatballs:
Not sure - but I'd want to use a flag, set with a default of False, to control whether it's returned or not

uglyfruitcake:
That would be really useful for analysing and improving strategies. Definitely like the sound of that.

drvinceknight:
I very much like this idea (obviously setting the flag by default to False).

marcharper:
If you all haven't looked at my [AxelrodExamples](https://github.com/marcharper/AxelrodExamples) repository lately, check it out. It basically reproduces the intent of this feature already, producing many plots that are useful for analyzing strategies, such as the following which shows how often Backstabber cooperations in each round with each opponent:

![](https://github.com/marcharper/AxelrodExamples/blob/master/assets/cooperation_heatmaps/BackStabber.png)

This one shows how often opponents cooperate with BackStabber:

![](https://github.com/marcharper/AxelrodExamples/blob/master/assets/opponent_cooperation_heatmaps/BackStabber.png)

There are also similar heatmaps for the mean score per round, mean score difference, and for noisy tournaments. I've been using these to develop some new strategies already (which I haven't submitted yet). I'm happy to write up some articles on all this (it's on my todo list, which has grown quickly recently).

Regardless, I'm not opposed to reproducing (or moving) any of this functionality inside Axelrod, but note that to generate these plots I did not run many tournaments over and over, rather I ran many direct play simulations. My vote would be to work on #248 first -- it's more closely attached to the tournament outcomes and would also be very useful for developing new strategies. It's also one of the few things that we need to get serious about writing some papers (the other big one being the rest of the strategies from the two Axelrod tournaments).

meatballs:
I agree that #248 and the remaining original strategies are higher priorities.

drvinceknight:
:+1:

meatballs:
I also think there's a strong case that the RoundRobin class should only return the interaction history. Computing the payoff and cooperation matrices fit more naturally with the ResultSet class - along with all the other computation methods.

marcharper:
[...]

PR #352, opened 2015-10-12
Pre-Work for #284

meatballs:
This is pre-work for #284. It takes the existing methods from the ResultSet class and creates new top-level functions instead.

It also includes a function to generate a payoff matrix from the interaction history. That function is, as yet, unused but shows how the RoundRobin will eventually return its results.

There should be no effect on the way the library currently works.


marcharper:
This branch rebases without conflicts on the soon-to-be new master.

PR #353, opened 2015-10-12
Trying to Fix Coverage ( DO NOT MERGE)

marcharper:


marcharper:
From the debug command, it looks like travis is running coverage without error, and that it attempts to report. However the coveralls.io website is not updated, so there is some communication issue.

PR #355, opened 2015-10-13
More plots and various plotting improvements

marcharper:
Two new plots:
* "*_sdvplot.svg" -- the score difference distributions in a violin plot
* "*_pdplot.svg" -- the score difference means in a heat map (like the payoff plot)

The new plots make it easier to see how a strategy is winning, either by mutual cooperation or successful defections. (Certainly the first one at least.)

Violin plots: these are a better representation of the actual distributions than boxplots in my opinion, so I updated the code to generate these instead. The downside is that matplotlib v1.4.3 doesn't handle them correctly in all cases, but there is a fix in for 1.5. So the code now catches an exception and falls back onto a boxplot if necessary.

Here's the noisy mean score distribution as a violin plot:

I also made various improvements to the plot sizes and spacing. The score difference plots use the same ordering as the mean score plots -- there are a few commented out lines that use another sorting.

Unfortunately I did have to touch results_set (don't hate me @meatballs!) but the changes are fairly minor (mostly additions). If you all like the new plots, I'll update the docs and the tournament repository.

### DIFF
--- a/axelrod/plot.py
+++ b/axelrod/plot.py
@@ -19,20 +19,13 @@ class Plot(object):
 
     ## Abstract Box and Violin plots
 
-    def _boxplot(self, data, names=None, title=None):
+    def _boxplot(self, data, names, title=None):
         """For making boxplots."""
         if not self.matplotlib_installed:
             return None
-
-        if not names:
-            names = self._boxplot_xticks_labels
-
         figure = plt.figure()
         plt.boxplot(data)
-        plt.xticks(
-            self._boxplot_xticks_locations,
-            names,
-            rotation=90)
+        plt.xticks(self._boxplot_xticks_locations, names, rotation=90)
         plt.tick_params(axis='both', which='both', labelsize=7)
         if title:
             plt.title(title)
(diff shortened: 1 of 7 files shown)


drvinceknight:
Ah awesome! I love violin plots. Will take a look through as soon as I get a chance (full day of teaching and first Cardiff python meetup this evening :dancer: :+1: ).

marcharper:
I think the coverage will go up once #357 is in. There are a few tests I can add in the meantime.

drvinceknight:
Could you add something to the docs?

axelrod/plot.py
```diff
@@ -1,4 +1,6 @@
-import numpy
+
+import numpy # for numpy.linalg.linalg.LinAlgError
```
CamDavidsonPilon: Why not be explicit and directly import LinAlgError
```
from numpy.linalg import LinAlgError
```
Then you can drop the comment and don't need to import all of numpy for a single reference.

marcharper: Good idea -- I'll change this.

marcharper:
Yes, I'll update the docs re: the new plots. @langner merged in the changes on the tournament repository, so the docs already have the violin plots.

drvinceknight:
> @langner merged in the changes on the tournament repository, so the docs already have the violin plots.

? Not sure I understand that, a merge on the tournament repo added something to the library docs?

Either way a part from what I think is a typo above I'm happy with this :+1:

marcharper:
I'll fix the typo. I pushed updated figures from this branch to the tournament repo that were merged, so for example there are already violin plots in the README for this repository.

drvinceknight:
> I pushed updated figures from this branch to the tournament repo that were merged, so for example there are already violin plots in the README for this repository.

Gotcha.

:+1:

PR #358, opened 2015-10-13
remove self.init_args

CamDavidsonPilon:
Correct me if I am wrong, but I don't think `self.init_args` is needed everywhere. `copy.copy` should transfer the properties over. This improves the maintainability of the code, and makes adding strategies easier. 

For review @meatballs et al.

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -73,7 +73,6 @@ class Player(object):
                 self.classifier[dimension] = self.default_classifier[dimension]
         self.cooperations = 0
         self.defections = 0
-        self.init_args = ()
         self.set_tournament_attributes()
 
     def receive_tournament_attributes(self):
@@ -119,9 +118,8 @@ class Player(object):
         """Clones the player without history, reapplying configuration
         parameters as necessary."""
 
-        cls = self.__class__
-        new_player = cls(*self.init_args)
-        new_player.tournament_attributes = copy.copy(self.tournament_attributes)
+        new_player = copy.copy(self)
+        new_player.reset()
         return new_player
 
     def reset(self):
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -119,7 +119,7 @@ class Player(object):
         parameters as necessary."""
 
         new_player = copy.copy(self)
-        new_player.reset()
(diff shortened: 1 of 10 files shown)


axelrod/player.py
```diff
@@ -119,9 +118,8 @@ def clone(self):
         """Clones the player without history, reapplying configuration
         parameters as necessary."""
 
-        cls = self.__class__
-        new_player = cls(*self.init_args)
-        new_player.tournament_attributes = copy.copy(self.tournament_attributes)
+        new_player = copy.copy(self)
```
CamDavidsonPilon: this may need to be `copy.deepcopy`

CamDavidsonPilon:
I imagine a few tests will fail, I'm just looking for feedback before I proceed.

marcharper:
Hi @CamDavidsonPilon ! Thanks for working on the library.

The presence of `init_args` is intentional. We have tried using copy before and there are some issues to consider:
* For some strategies one likely needs to use `copy.deepcopy` (like the meta strategies), since the internal variables can contain objects (like the `team` variable of a MetaPlayer)
* `deepcopy` can be slow
* There may be hard to track bugs due to copying that the tests do not cover (like retaining some kind of state between rounds)

That said, if you can make it work without a big performance hit, we'll certainly consider merging!

CamDavidsonPilon:
Hmm, ok, from there:

1. could the more complicated strategies reimplement their own `clone` function?
2. Agree - I bet `copy.copy` could be used for most cases that didn't require reimplementing `clone`.
3. Yes, I wonder what we would uncover. `copy.copy` should avoid these though.

I'll keep thinking about this.

marcharper:
It's a tricky part of the code. We are also having issues with the deepcopy in MindReader -- it's really slow when paired with some specific strategies. If you have any thoughts on that please do share :)

drvinceknight:
> The presence of init_args is intentional. We have tried using copy before and there are some issues to consider:

Is it worth having this in a docstring somewhere perhaps?

marcharper:
Yes, or a comment in the code I think, or in the docs...

CamDavidsonPilon:
I'll close this for now

PR #360, opened 2015-10-13
some small code cleanup

CamDavidsonPilon:
This is some small cleanup of code. The biggest change is moving `WinStayLoseShift` to be a `MemoryOnePlayer` subclass. This offers less code duplication. Thoughts?

### DIFF
--- a/axelrod/strategies/memoryone.py
+++ b/axelrod/strategies/memoryone.py
@@ -5,40 +5,6 @@ from axelrod import Player
 """IPD Strategies: http://www.prisoners-dilemma.com/strategies.html"""
 
 
-class WinStayLoseShift(Player):
-    """Win-Stay Lose-Shift, also called Pavlov."""
-
-    name = 'Win-Stay Lose-Shift'
-    classifier = {
-        'memory_depth': 1,  # Memory-one Four-Vector = (1, 0, 0, 1)
-        'stochastic': False,
-        'inspects_source': False,
-        'manipulates_source': False,
-        'manipulates_state': False
-    }
-
-    def __init__(self, initial='C'):
-        Player.__init__(self)
-        self.response_dict = {
-            ('C', 'C'): 'C',
-            ('C', 'D'): 'D',
-            ('D', 'C'): 'D',
-            ('D', 'D'): 'C',
-        }
-        self.classifier['stochastic'] = False
-        self._initial = initial
-        self.init_args = (initial,)
-
-    def strategy(self, opponent):
-        """Switches if it doesn't get the best payout, traditionally equivalent
-        to a Memory one strategy of [1,0,0,1], but this implementation does not
-        require random draws."""
-        if not opponent.history:
-            return self._initial
-        last_round = (self.history[-1], opponent.history[-1])
-        return self.response_dict[last_round]
-
-
 class MemoryOnePlayer(Player):
     """Uses a four-vector for strategies based on the last round of play,
     (P(C|CC), P(C|CD), P(C|DC), P(C|DD)), defaults to Win-Stay Lose-Shift.
@@ -71,21 +37,33 @@ class MemoryOnePlayer(Player):
         self.init_args = (four_vector, initial)
 
     def set_four_vector(self, four_vector):
+        if not all(0 <= p <= 1 for p in four_vector):
+            raise ValueError('An element in the probability vector, %s, is not between 0 and 1.' % four_vector)
+
         self._four_vector = dict(zip([('C', 'C'), ('C', 'D'), ('D', 'C'), ('D', 'D')], map(float, four_vector)))
         self.classifier['stochastic'] = any(0 < x < 1 for x in set(four_vector))
 
     def strategy(self, opponent):
         if not hasattr(self, "_four_vector"):
             raise ValueError("Fourvector not yet set")
-        if not len(opponent.history):
+        if len(opponent.history) == 0:
             return self._initial
         # Determine which probability to use
         p = self._four_vector[(self.history[-1], opponent.history[-1])]
         # Draw a random number in [0,1] to decide
-        r = random.random()
-        if r < p:
-            return 'C'
-        return 'D'
+        return 'C' if random.random() < p else 'D'
+
+
+class WinStayLoseShift(MemoryOnePlayer):
+    """Win-Stay Lose-Shift, also called Pavlov."""
+
+    name = 'Win-Stay Lose-Shift'
+
+    def __init__(self, initial='C'):
+        Player.__init__(self)
+        self.set_four_vector([1,0,0,1])
+        self._initial = initial
+        self.init_args = (initial,)
 
 
 class GTFT(MemoryOnePlayer):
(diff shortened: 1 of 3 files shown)


axelrod/strategies/memoryone.py
```diff
@@ -71,21 +37,33 @@ def __init__(self, four_vector=None, initial='C'):
...
+            raise ValueError('An element in the probability vector, %s, is not between 0 and 1.' % four_vector)
+
         self._four_vector = dict(zip([('C', 'C'), ('C', 'D'), ('D', 'C'), ('D', 'D')], map(float, four_vector)))
         self.classifier['stochastic'] = any(0 < x < 1 for x in set(four_vector))
 
     def strategy(self, opponent):
         if not hasattr(self, "_four_vector"):
             raise ValueError("Fourvector not yet set")
-        if not len(opponent.history):
+        if len(opponent.history) == 0:
```
CamDavidsonPilon: Safer to checkout explicitly against 0, rather than using `not`

axelrod/strategies/memoryone.py
```diff
@@ -106,7 +84,7 @@ def __init__(self, p=None):
 
     def receive_tournament_attributes(self):
         (R, P, S, T) = self.tournament_attributes["game"].RPST()
-        if not self.p:
+        if self.p is None:
```
CamDavidsonPilon: what if `p` is 0? This would previously fail the condition, but is a valid input. Instead I check against `None`

marcharper: Good catch!

axelrod/tests/unit/test_classification.py
```diff
@@ -38,7 +38,7 @@ def test_multiple_instances(self):
         self.assertNotEqual(P1.classifier, P2.classifier)
 
         P1 = axelrod.GoByMajority(5)
-        P2 = axelrod.StochasticWSLS(10)
+        P2 = axelrod.StochasticWSLS(.1)
```
CamDavidsonPilon: This was failing because `1 - 10 < 0 < 1` and was caught in the ValueError added

marcharper:
I'm fine with this. A few notes should you continue with this part of the library:

* There are several other memory one strategies not coded as special cases of `MemoryOnePlayer` -- some of the TFT players, of course Cooperator and Defector. We do like having some simple to follow strategies for newcomers, so let's not do the same to any of the MemoryZero strategies (Cooperator, Defector, and Random, at least).
* You could also make WSLS a special case of StochasticWinStayLoseShift. Just a thought. TFT could be a special case of GTFT as well and I am sure there are other examples in the library. It's ok to have multiple implementations I think -- it could be useful to test against each other when there is more than one implementation.
* I think that the previous code was (very) slightly more efficient (no random numbers). Perhaps we could modify `MemoryOnePlayer` to check for `p=0` or `p=1` and not draw a random in that case. I'm not sure if this would be a net gain or not, or if the difference is worth worrying about (probably not). Just a thought if you keep looking into this. Alternatively it might be better to leave the memory one players that have all probabilities equal to one or zero as is rather than as subclasses.

Anyone else have thoughts?

Some of the other maintainers are in Europe, so it may be a while before they can take a look (I think it's nighttime there now).

CamDavidsonPilon:
@marcharper, thanks for reviewing. I agree that leaving some more popular models out of this, it just felt odd to see duplication in a single python file. 

w.r.t to efficiency, the random() call is very efficient. In fact, the `hasattr` call in the same function takes 10x as long as a call to `random`. So any perf hit would be negligible. 

I'll add a few tests to have coveralls pass

axelrod/strategies/memoryone.py
drvinceknight: Can you break down why it's safer? What's the risk?

axelrod/strategies/memoryone.py
drvinceknight: Reading over this briefly, could you add a test that checks that this fixes that potential case?

CamDavidsonPilon: yeye just added

drvinceknight:
> There are several other memory one strategies not coded as special cases of MemoryOnePlayer -- some of the TFT players, of course Cooperator and Defector. We do like having some simple-to-follow strategies for newcomers, so let's not do the same to any of the MemoryZero strategies (Cooperator, Defector, and Random, at least).

I completely agree with this. Have just given a talk about the library and one of the things that was discussed was how nice it was that a lot of the strategies were very readable to non python experts.

> You could also make WSLS a special case of StochasticWinStayLoseShift. Just a thought. TFT could be a special case of GTFT as well and I am sure there are other examples in the library. It's ok to have multiple implementations I think -- it could be useful to test against each other when there is more than one implementation.

Not sure I'm a big fan of this suggestion as it's a contradiction with the previous one no? 

I'm also not sure about the multiple implementations idea (again: confusing for the non expert)... The fact that there is the generic memory one class means a user could code up other implementations if they wanted to but I'm not a fan of this being in the library... Perhaps I'm missing something?

Otherwise this all looks great, just the one suggestion about a test for what seems to be a bug that you caught?

axelrod/strategies/memoryone.py
drvinceknight: :+1:

CamDavidsonPilon:
Updated with more tests.

Conversely, one can see this refactor of `WinStayLoseShift` as an example implementation for the next non-expert to see how to implement a deterministic strategy using the `MemoryOnePlayer` class (which is super easy btw).

drvinceknight:
> Conversely, one can see this refactor of WinStayLoseShift as an example implementation for the next non-expert to see how to implement a deterministic strategy using the MemoryOnePlayer class (which is super easy btw).

Very true but I would suggest your non expert there is more familiar with Python than my non expert...

There are quite a few strategies already present that show how to implement a MemoryOnePlayer... 

TitForTat as a verbose one liner I would not like to be changed. It was something I genuinely spoke about for a while during my talk...

drvinceknight:
Thanks for the test: I'm happy with this PR as it is, perhaps we can open an issue to continue the discussion about TitForTat etc...?

CamDavidsonPilon:
Oh no need to the issue, I agree with you two. My comment was that there was no _deterministic_ strategy that used `MemoryOnePlayer` (that I can see at least)

marcharper:
I wasn't trying to suggest that TFT should be reimplemented as MemoryOne, I was just pointing out that there are a lot of cases where simply implemented strategies are special cases of parameterized strategies. (My wording was poor -- I meant the proverbial "you").

Maybe it's worth pointing these cases out in the docs for the relevant strategies, or in the docstring, and perhaps implementing tests for these corner cases.

drvinceknight:
> Maybe it's worth pointing these cases out in the docs for the relevant strategies, or in the docstring, and perhaps implementing tests for these corner cases.

I really like that idea :)

drvinceknight:
I think this can be merged or am I missing something?

marcharper:
Sure, I'll update the "duplicate strategies" issue with the suggestion above.

PR #361, opened 2015-10-13
Minor neatening of some strategies

bootandy:
Consistency of strategy method: strategy method is an instance method not a static method in most implementations.

Darwin strategy: neaten use class variable

Strategies that use opponent.history should check opponent.history exists not self.history

### DIFF
--- a/axelrod/strategies/appeaser.py
+++ b/axelrod/strategies/appeaser.py
@@ -18,7 +18,7 @@ class Appeaser(Player):
     }
 
     def strategy(self, opponent):
-        if not len(self.history):
+        if not len(opponent.history):
             self.move = 'C'
         else:
             if opponent.history[-1] == 'D':
--- a/axelrod/strategies/axelrod_tournaments.py
+++ b/axelrod/strategies/axelrod_tournaments.py
@@ -94,7 +94,7 @@ class Feld(Player):
                    self._end_coop_prob)
 
     def strategy(self, opponent):
-        if not self.history:
+        if not opponent.history:
             return 'C'
         if opponent.history[-1] == 'D':
             return 'D'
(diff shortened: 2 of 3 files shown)


axelrod/strategies/averagecopier.py
```diff
@@ -19,8 +19,7 @@ class AverageCopier(Player):
         'manipulates_state': False
     }
 
-    @staticmethod
-    def strategy(opponent):
+    def strategy(self, opponent):
```
drvinceknight: As was implemented here the strategy method was in fact static and not using self. By passing self you do make it a non static method but there is no advantage to doing so as the method does not use self.

drvinceknight:
A part from 20f728c I am happy with this (making the methods static when possible to do so was one of the improvements made at some point to the library). I suggest reverting 20f728c.

bootandy:
OK, static method commit reverted.

drvinceknight:
:+1:

marcharper:
Just FYI, we are working under the (unstated) assumption that len(self.history) == len(opponent.history), which allows more static methods. This looks good otherwise :+1:

drvinceknight:
> Just FYI, we are working under the (unstated) assumption that len(self.history) == len(opponent.history), which allows more static methods. This looks good otherwise

Actually the changes made by 794f597 make it possible to change those strategies to static methods right? Using `opponent.history` allows for more static methods.

@bootandy could you add to those methods the @static decorator as by changing them to use the opponent's length they are now in effect static methods. What do you think @marcharper?

drvinceknight:
Some other merges have caused a conflict (just raising in case this gets missed).

bootandy:
Will resolve conflicts and submit new pull request with static decorators

drvinceknight:
:+1:

Issue #359, opened 2015-10-13
Move all references of 'C' and 'D' to importable constants


CamDavidsonPilon:
The library is littered with `'C'` and `'D'`. I think it would be an improvement if these were constants

```
# axelrod/__init__.py

DEFECT = 'D'
COOPERATION = 'C'

# or maybe 
from enum import Enum

class Move(Enum):
  # this numbers don't really matter, they are just used for comparing equality. 
   DEFECT = 0
   COOPERATE = 1

```

Then in other code, you might have:
```
from axelrod import Move
...
# from rand.py

    def strategy(self, opponent):
        r = random.random()
        if r > self.p:
            return Move.DEFECT
        return Move.COOPERATE
```

Another example,  

```
from axelrod import Move
...
# from gobymajority.py

    def strategy(self, opponent):
        """This is affected by the history of the opponent.

        As long as the opponent cooperates at least as often as they defect then the player will cooperate.
        If at any point the opponent has more defections than cooperations in memory the player defects.
        """

        memory = self.classifier['memory_depth']
        history = opponent.history[-memory:]
        defections = sum([s == Move.DEFECT for s in history])
        cooperations = sum([s == Move.COOPERATE for s in history])
        if defections > cooperations:
            return Move.DEFECT
        if defections == cooperations:
            if self.soft:
                return Move.COOPERATE
            else:
                return Move.DEFECT
        return Move.COOPERATE
```

This reads a bit better, plus feels more maintainable - you won't accidently write `'c'` instead of `'C'` for example. Thoughts?

marcharper:
I think something can be done here. Several files have the following near the top (e.g lots of the test files):
`C, D = 'C', 'D'`

We could definitely be more consistent across the library.

bootandy:
I was considering suggesting this myself. 
Constants are less error prone (Have fun debugging the crash from a 'not C' that turned out to be a 'c' and not the letter 'D').

marcharper:
I like `C, D = 'C', 'D'` over `DEFECT` and `COOPERATE` just for brevity. I guess we could start files with `C, D = COOPERATE, DEFECT` or `C, D = Move.COOPERATE, Move.DEFECT`. We don't tend to use `C` or `D` as variables. 

However it shakes out I do think this would be a nice clean-up for the library, and as @bootandy says, reduce the chance of some unintentional mistakes.

drvinceknight:
[...]

PR #362, opened 2015-10-14
moving 'C' and 'D' to axelrod constants

CamDavidsonPilon:
w.r.t #359 , this moves 'C' and 'D' to an `actions.py` file
```
class Actions(object):
    C = 'C'
    D = 'D'
```
(`enum` is an external lib, and I didn't want to add dependency)

@marcharper @bootandy @drvinceknight, thoughts?

### DIFF
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 # The order of imports matters!
+from .actions import Actions
 from .random_ import random_choice
 from .plot import Plot
 from .game import DefaultGame, Game
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -1,0 +1,4 @@
+
+class Actions(object):
+    C = 'C'
+    D = 'D'
--- a/axelrod/game.py
+++ b/axelrod/game.py
@@ -1,21 +1,24 @@
+from axelrod import Actions
+
+C, D = Actions.C, Actions.D
 
 class Game(object):
     """A class to hold the game matrix and to score a game accordingly."""
 
     def __init__(self, r=3, s=0, t=5, p=1):
         self.scores = {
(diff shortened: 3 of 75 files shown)


marcharper:
Thanks for doing this. We have a 2 maintainers approval rule for merging, so hopefully one of the others will be able to look at it soon.

Any reason that a few of the files don't use `C, D = Actions.C, Actions.D`? I presume it was simply because the occurrences were few.

CamDavidsonPilon:
> Any reason that a few of the files don't use C, D = Actions.C, Actions.D? I presume it was simply because the occurrences were few.

Yea exactly, I used a rule of three approach: three or more occurrences => alias them.

marcharper:
:+1: For me. My only suggestion is to add something to the docs explaining the existence of Actions, C and D. There is not much we can do on the testing front other than C != D, or just a generic test to make sure no one changes the definition of the actions.

drvinceknight:
> There is not much we can do on the testing front other than C != D, or just a generic test to make sure no one changes the definition of the actions.

Sorry to delay this slighty, it all looks great. Could you just throw in a trival `test_actions.py`, just something that checks exactly what is suggested by @marcharper as well as perhaps `self.C,'C'` and `self.D,'D'` etc... Just to make sure that this doesn't break somewhere along the lines.

> My only suggestion is to add something to the docs explaining the existence of Actions, C and D.

Yes please :) (A sentence or two would do - the whole docs need work so at the moment it's really a question of getting something in there that at least acts as a placeholder).

drvinceknight:
Some other merges have caused a conflict.

CamDavidsonPilon:
Tests for `actions.py` have been added

drvinceknight:
Yeah this is good to go :+1: thanks for the great contribution :)

PR #363, opened 2015-10-14
Some simple additions to the documentation

marcharper:
Follow-ups from Tuesday's discussions.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -5,7 +5,7 @@ All contributions to this repository are welcome via pull request on the `github
 
 The project follows the following guidelines:
 
-1. Use the base Python library unless completely necessary. Numpy is ok.
+1. Use the base Python library unless completely necessary. A few external libraries (such as numpy) have been included in requirements.txt -- feel free to use these as needed.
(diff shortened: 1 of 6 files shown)


docs/contributing.rst
```diff
@@ -5,12 +5,16 @@ All contributions to this repository are welcome via pull request on the `github
 
 The project follows the following guidelines:
 
-1. Use the base Python library unless completely necessary.
+1. Use the base Python library unless completely necessary. Numpy is ok.
```
drvinceknight: Is it worth wording this more along the lines of: please use what is currently in `requirements.txt`, if you need to use an external library perhaps discuss this with us? (Or maybe get rid of this completely?)

drvinceknight:
This is great, just my one comment about the first guideline. Not sure it's still worth insisting on... Happy to go with whatever you feel is best.

marcharper:
Yeah I thought about that... how's this change? I just wanted to mention numpy specifically since we are starting to use it more.

drvinceknight:
> Yeah I thought about that... how's this change? I just wanted to mention numpy specifically since we are starting to use it more.

Yeah great! 

This is good to go (have already said so on the gitter channel) :+1:

PR #364, opened 2015-10-14
Add .mailmap file so that `git shortlog` groups names correctly

marcharper:
This allows us to use tools like gitinspector to analyze the repository:

[Python contributions] (http://marcharper.codes/axelrod/axelrod_py.html)
* Owen Campbell 39.77%
* Marc Harper 23.47%
* Vincent Knight	 15.48%
* Karol M. Langner 10.75%

[Documentation Contributions](http://marcharper.codes/axelrod/axelrod_rst.html)
* Vincent Knight	 71.46%
* Marc Harper 21.76%
* Owen Campbell 3.01%
* Luis Visintini 2.19%

PR #366, opened 2015-10-15
Revert "Add .mailmap file so that `git shortlog` groups names correctly"

meatballs:
Reverts Axelrod-Python/Axelrod#364


meatballs:
noticed, just too late, that my email address isn't correct

PR #368, opened 2015-10-16
Refactors payoff.py

meatballs:
1) Splits the payoff_matrix function into two, creating a new function: interaction_payoff which it then calls.
This gives us a function which can be used to obtain the scores for a single interaction of n turns between two players

2) Removes redundant parameters from various payoff functions where they can be derived from the payoff matrix.

### DIFF
--- a/axelrod/payoff.py
+++ b/axelrod/payoff.py
@@ -318,7 +318,7 @@ def wins(payoff):
     return wins
 
 
-def payoff_diffs_means(payoff, nplayers, repetitions, turns):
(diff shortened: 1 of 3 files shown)


drvinceknight:
This all looks good to me and a nice logical move after our conversation on Thursday.

drvinceknight:
:+1:

meatballs:
And our conversation led me to realised that the payoff_matrix function was double counting each interaction - hence the last commit on this PR!!

marcharper:
:+1:

PR #369, opened 2015-10-17
Minor neatening of some strategies

bootandy:
Follow up pull request after "Minor neatening of some strategies #361 " caused conflicts.

Darwin strategy: neaten use class variable

Strategies that use opponent.history should check opponent.history exists not self.history

Where possible @static decorator added to strategy method.


drvinceknight:
:+1: from me.

Issue #349, opened 2015-10-11
Default argument for stackplot?


drvinceknight:
I'm not sure I like that to get an evo plot we need to pass back the `eco.population_sizes` argument to the `stackplot` method as so:

```
import axelrod
strategies = [s() for s in axelrod.demo_strategies]
tournament = axelrod.Tournament(strategies)
results = tournament.play()
eco = axelrod.Ecosystem(results)
eco.reproduce(100) # Evolve the population over 100 time steps
plot = axelrod.Plot(results)
p = plot.stackplot(eco.population_sizes)
p.show()
```

In general (always?) this will always be the argument passed so I suggest we modify the method to have a default argument: this would not break any previous functionality for which the argument is passed (not that I can think of any).

Thus the above would become:

```
import axelrod
strategies = [s() for s in axelrod.demo_strategies]
tournament = axelrod.Tournament(strategies)
results = tournament.play()
eco = axelrod.Ecosystem(results)
eco.reproduce(100) # Evolve the population over 100 time steps
plot = axelrod.Plot(results)
p = plot.stackplot()  # This is now similar to plot.boxplot()
p.show()
```

Here's where the `stackplot` is: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/plot.py#L76

We could create a `_stackplot_dataset` method similar to `_payoff_dataset` which would be the default `population`.

PR #371, opened 2015-10-18
Changing stackplot input

drvinceknight:
- Adding depreciation warning in case of old system
- Change docs
- Adjusting tests.

This is my attempt at #349, having looked at things a bit closer getting rid of the argument all together implies passing the tournament object between all these things which is messy (I have a branch where that is implemented).

This seemed more elegant and just changes the input argument to be the whole ecosystem (which is slightly clearer from a ui pov). As people are beginning to use this library I've included a depreciation warning. :shipit:

### DIFF
--- a/axelrod/plot.py
+++ b/axelrod/plot.py
@@ -1,6 +1,7 @@
 
 from numpy.linalg import LinAlgError
 from numpy import arange, mean, median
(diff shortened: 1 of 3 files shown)


meatballs:
:+1:

PR #372, opened 2015-10-18
299 - Docs

drvinceknight:
I'm opening a PR as I feel like I'm getting somewhere with this and would like to have @marcharper @meatballs and (I know you're busy but tagging in case) @langner's opinion.

This achieves two main things:

- Doctests: everything (a part from stochastic stuff) is doctested. This is important, I caught a couple of things that aren't quite right in the current docs. Travis should be set up to run the doctests also. These are run from a './doctest.sh' script.
- Modularisation: There are two main parts to the docs, reference (I have basically not touched this apart from introducing doctests) and tutorials. The tutorials are hopefully a bit more streamlined and very much separated out. There are currently three types (outside of contributions) tutorials:
     - getting started: The basics
     - further topics: so far just classification and morality metrics
     - advanced: nothing yet but perhaps some neat tournaments etc... (like stuff that @marcharper has done?) could go here

There are still some things left to do (I haven't done them yet just because I'm tired and want to go to :zzz:):

- No tutorial for the command line tool
- A lot missing from contributions (but this can be lifted from status quo)
- The automatically generated docs for the strategies, are not in. I have no idea if this had something to do with breaking readthedocs a while back and will ease this back in at the last moment.

Let me know your thoughts. Also, if anything new comes in while this PR is still ongoing (which will no doubt be the case) please comment here just so that I know about it...

PR #373, opened 2015-10-19
Refactors payoff.py and cooperation.py

meatballs:
1. Refactors payoff_matrix to use interactions dictionary instead of matrix
2. Adds cooperation_matrix function to return cooperation matrix from interactions dictionary
3. Tidies up docstrings in both files

### DIFF
--- a/axelrod/cooperation.py
+++ b/axelrod/cooperation.py
@@ -48,6 +48,30 @@ def cooperation_matrix(interactions):
         and column (j) represents an individual player and the the value Cij
         is the number of times player i cooperated against opponent j.
     """
+
+    /*
+    The number of ways (c) to select groups of r members from a set of n
+    members is given by:
+
+        c = n! / r!(n - r)!
+
+    In this case, we are selecting pairs of players (p) and thus r = 2, giving:
+
+        p = n(n-1) / 2 or p = (n^2 - n) / 2
+
+    However, we also have the case where each player plays itself gving:
+
+        p = (n(n-1) / 2) + n or p = (n^2 + n) / 2
+
+    Using the quadratic equation to rearrange for n gives:
+
+        n = (-1 +- sqrt(1 + 8p)) / 2
+
+    Taking the real roots only allows us to derive the number of players given
+    the number of pairs:
+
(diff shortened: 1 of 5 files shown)


drvinceknight:
This looks good to me, could you make a note on #372 re the slight change that will be needed (nice to include the cooperation_matrix) to http://axelrod.readthedocs.org/en/299/tutorials/further_topics/morality_metrics.html

(Just trying to keep track).

:+1:

axelrod/cooperation.py
```diff
@@ -1,7 +1,66 @@
...
+
+        i.e. an n by n matrix where n is the number of players. Each row (i)
+        and column (j) represents an individual player and the the value Cij
+        is the number of times player i cooperated against opponent j.
     """
+    cooperation = [[0 for i in range(nplayers)] for j in range(nplayers)]
+    for players, actions in interactions.items():
+        p1_actions, p2_actions = zip(*actions)
+        p1_cooperation = p1_actions.count(Actions.C)
+        p2_cooperation = p2_actions.count(Actions.C)
```
CamDavidsonPilon: You can just use `C` and `D` here

axelrod/cooperation.py
```diff
@@ -1,7 +1,66 @@
 from . import eigen
+from axelrod import Actions
 
-def cooperation_matrix(results):
+C, D = Actions.C, Actions.D
+
+# As yet unused until RoundRobin returns interactions
+def cooperation_matrix(interactions, nplayers):
```
CamDavidsonPilon: given an `interactions` dict, couldn't the number of players be inferred? Basically solving for `len(dict.keys()) = n*(n-1)/2`

meatballs: Good call. That's me being lazy! I'll add it in.

axelrod/cooperation.py
meatballs: Ah yes! I added the variables in at the top of the file and then forgot to use them. Duh.

meatballs:
I'm done messing with this for the moment. Feel free to comment/merge/criticise!

marcharper:
:+1: to merge

PR #374, opened 2015-10-19
use random_choice over many calls of random.random

CamDavidsonPilon:
This PR settles on using random_choice over any custom random.random logic. 

- [ ] I may have reversed the logic, I would appreciate someone else to double check I didn't. 
- [x] remove `import random`s
- [x] fix any tests

### DIFF
--- a/axelrod/strategies/axelrod_tournaments.py
+++ b/axelrod/strategies/axelrod_tournaments.py
@@ -235,8 +235,7 @@ class Tullock(Player):
         cooperate_count = opponent.history[-rounds:].count(C)
         prop_cooperate = cooperate_count / float(rounds)
         prob_cooperate = max(0, prop_cooperate - 0.10)
-        r = random.random()
-        random_choice(prob_cooperate)
+        return random_choice(prob_cooperate)
 
 
 ## Second Tournament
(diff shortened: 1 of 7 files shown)


CamDavidsonPilon:
Ugh I think coverage is failing due to a rounding error!

CamDavidsonPilon:
Can you all merge with this red?

drvinceknight:
This looks like a nice refactoring. The coverage decrease doesn't bother me too much (it's minor) but will let another core decide if it's worth trying to address. :+1:

meatballs:
I can't get too excited about a 0.001% decrease!

drvinceknight:
> I can't get too excited about a 0.001% decrease!

Not excited as in you think this should be fixed or it can go through?

meatballs:
We should let it through!

drvinceknight:
Thanks for the contribution @CamDavidsonPilon :) :+1:

marcharper:
For posterity: `random_choice` was introduced to the library because `random.choice([C, D])` is not consistent with random seeds across python 2 and 3. However `random.random` is consistent. So it's ok if people want to use `random.random` for various purposes. This PR is of course an improvement -- I'm just noting that when I finished the python 3 compatibility work I only explicitly changed the instances of `random.choice`.

We should probably explain the `random.choice` and `random.random` situation somewhere in the documentation.

drvinceknight:
> We should probably explain the random.choice and random.random situation somewhere in the documentation.

I agree. I'll add a note to #372 as I continue to add to that.

PR #375, opened 2015-10-19
adding a reverse_action function

CamDavidsonPilon:
I saw a few instances where this function could be useful (I found it useful implementing my v1 of a person IPD project.). There are probably some more instances in other strategies. 

Couple things: 1) I'm so-so on the name `reverse_action` - I'm happy to hear other suggestions. 2) This changes the implementation of TitFotTat to something I think it more readable. 

Thoughts?

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -1,4 +1,4 @@
-from axelrod import Player, Actions, reverse_action
+from axelrod import Player, Actions, flip_action
 
 C, D = Actions.C, Actions.D
 
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -1,7 +1,6 @@
-from axelrod import Player, Actions
+from axelrod import Player, Actions, reverse_action
 
 C, D = Actions.C, Actions.D
-flip_dict = {C: D, D: C}
 
 
 class TitForTat(Player):
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 
 # The order of imports matters!
-from .actions import Actions
+from .actions import Actions, reverse_action
 from .random_ import random_choice
 from .plot import Plot
 from .game import DefaultGame, Game
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 
 # The order of imports matters!
-from .actions import Actions, reverse_action
+from .actions import Actions, flip_action
 from .random_ import random_choice
 from .plot import Plot
 from .game import DefaultGame, Game
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -2,3 +2,12 @@
 class Actions(object):
     C = 'C'
     D = 'D'
+
+
+def reverse_action(action):
+    if action == Actions.C:
+        return Actions.D
+    elif action == Actions.D:
+        return Actions.C
+    else:
+        raise ValueError("Encountered a invalid action")
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -4,7 +4,7 @@ class Actions(object):
     D = 'D'
 
 
-def reverse_action(action):
+def flip_action(action):
     if action == Actions.C:
         return Actions.D
     elif action == Actions.D:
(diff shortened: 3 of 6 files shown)


axelrod/strategies/titfortat.py
```diff
@@ -18,7 +17,9 @@ class TitForTat(Player):
 
     @staticmethod
     def strategy(opponent):
-        return D if opponent.history[-1:] == [D] else C
+        if len(opponent.history) == 0:
+            return C
+        return reverse_action(opponent.history[-1])
```
CamDavidsonPilon: hmm, something about this is causing lots of tests to fail - the logic looks identical to me however.

drvinceknight:
I haven't looked too closely. 

I'm hesitant (not against, just hesitant) about the readability claim of Tit for Tat. I think having it written in base python makes it an accessible example when describing the library to non python/coders.

drvinceknight:
Weird about the failing tests. On my phone, will take a closer look when in front of a machine.

meatballs:
> I think having it written in base python makes it an accessible example when describing the library to non python/coders.

I agree. We know there are some very inexperienced coders looking at this library and we need to ensure that we have some easy routes in - even if that sometimes leads to code that's not quite as DRY as it could be.

drvinceknight:
> I agree. We know there are some very inexperienced coders looking at this library and we need to ensure that we have some easy routes in - even if that sometimes leads to code that's not quite as DRY as it could be.

I would suggest (forgive me if I'm speaking incorrectly on your behalf @meatballs) that we both think this is a nice contribution and good to have but perhaps `TitForTat` should be left alone. @marcharper has suggested that we write `TFT` in an even more verbose basic python way than it is now which I don't think is a bad idea...

drvinceknight:
Not too sure what's happening with the tests, travis is saying something about numpy but running them locally I got other errors. For info this is how to run the unit tests @CamDavidsonPilon:

```
cd Axelrod
python -m unittest discover axelrod.tests.unit
```

meatballs:
> I would suggest (forgive me if I'm speaking incorrectly on your behalf @meatballs) that we both think this is a nice contribution and good to have but perhaps TitForTat should be left alone.

Exactly my thoughts.

marcharper:
While I agree re: TFT, it is somewhat silly to have `flip_dict` defined in several places, so I'm for the idea behind this PR. Perhaps `toggle_action` is a better name.

My suggestion for TFT rather than:

```
def strategy(opponent):
    return 'D' if opponent.history[-1:] == ['D'] else 'C'
```
We use something (rather verbose) like:
```
def strategy(opponent):
    # First move
    if len(self.history) == 0:
        return C
    # React to the opponent's last move
    if opponent.history[-1] == D:
        return D
    return C
```

The reason is that slicing is not something I would expect a python beginner to understand. In any case, perhaps we should put in a comment "Don't mess with TFT, thanks!" in the code.

drvinceknight:
> In any case, perhaps we should put in a comment "Don't mess with TFT, thanks!" in the code.

I like both of the above suggestion (the code for TitForTat) and the comment in TitForTat: "This strategy is written in a more verbose way than necessary however it is left that way as an example for contributors who are perhaps less familiar with Python" or something like that...

CamDavidsonPilon:
:ok_hand: I've removed changes to TFT, and tests pass locally. Waiting for :green_apple:

marcharper:
Any votes for `toggle_action`? `reverse_action` sounds more like rolling back than swapping. More suggestions: `swap_action`, `flip_action`, `transpose_actions`...

drvinceknight:
I think I prefer `flip_action`, but `toggle_action` would be my second favourite. I agree that `reverse_action` is not ideal, sounds like some sort of manipulation.

meatballs:
```flip_action``` for me

marcharper:
Looks like numpy is failing to install on Travis

drvinceknight:
This happened the other day. I'll restart it. Am I the only person who can
restart a travis run? We should fix that if so.

marcharper:
I thought so but it looks like I can do it (had to sign in through github).

drvinceknight:
Cool :)

drvinceknight:
Is this good to go now?

meatballs:
:+1:

drvinceknight:
Let's wait to make sure @marcharper is ok with `flip_action` as he's been involved with the review :tada: :shipit:

PR #376, opened 2015-10-20
Metafix

marcharper:
There is a bug in MetaWinner, a hold-over from before we switched the game matrix (!). (It was actually MetaLoser!)

Also the `memory_depth` calculation for the meta strategies was not correct. They are typically of depth `float('inf')`.


langner:
+1

Issue #348, opened 2015-10-10
History and noise


marcharper:
Currently the library records the post-noise play of each player in its history. By tracking its own plays a player can infer the noise rate by counting the number of flipped moves. (There is an example on my [noise_detector branch](https://github.com/marcharper/Axelrod/tree/noise_detector).) However it is not possible to tell if the opponent's moves are flipped.

In a related matter, we do not actually supply the noise parameter in the tournament arguments dictionary. If we did there would be no reason to infer the value but still reasons to track history and respond appropriately.

Any thoughts? I don't think that players usually know the noise level; but they don't usually know the number of rounds necessarily either.

drvinceknight:
Nice idea. I don't see why we wouldn't give the info to the players.

As always: up to the users to make use of strategies that use the information or not.

marcharper:
Agreed, but there is not an easy way of determining which strategies make use of any particular value in the tournament dictionary. So we might consider adding this to the classifier dictionary. Something like:
```
'makes_use_of': ("game_matrix", "tournament_length")
```

drvinceknight:
I really like that suggestion!

drvinceknight:
> I guess this could be automated by looking for e.g. "self.tournament_attributes["tournament_length"]" in the source.

I'm up for automating the process of writing the code but I'd still like to see it hard coded... (For readability...)

bootandy:
I think a player should just know the length of the tournament, it makes strategies more interesting if you know you are near the end.

bootandy:
I've just seen you can access self.tournament_attributes['length']

marcharper:
In any case I think it needs to be classified (as suggested above) for each player whether it makes use of any additional information other than the opponent's history (and its own state). Some researchers study tournaments without a defined endpoint (see the discussion #37). With the classification, the list of strategies can be filtered as needed for whatever conditions the user wants to allow for their tournaments.

drvinceknight:
> I think a player should just know the length of the tournament, it makes strategies more interesting if you know you are near the end.

I (humbly) think another way of saying that would perhaps be 'I think tournaments where strategies can know you are near the end are interesting' but of course as @marcharper points out other tournaments can be of interest to other people. The goal of the library is to facilitate **all those options** so as Marc says: it all needs to be classified and I **LOVE** the suggestion of all strategies having 'makes_use_of': ("game_matrix", "tournament_length")` which makes all these options really easy for the user.

[...]

PR #380, opened 2015-10-23
Strategy transformers

marcharper:
This PR introduces *strategy transformers*, which are ways of transforming the behavior of a strategy without rewriting the strategy's class. There is a generic transformer factory that will take a user-defined function and wrap the Player's strategy method.

For example, using `FlipTransformer`, we can turn `Cooperator` into `Defector`:
```
new_class = FlipTransformer(axelrod.Cooperator)
player = new_class() # It's the same as Defector now
```
Similarly we could turn AntiCycler into DefectingAntiCycler, playing
```
D DC DDC DDDC DDDDC ...
```
Instead of
```
C CD CCD CCCD CCCCD ...
```

Another transform adds TFT style retaliation to any other strategy:
```
RUA = RetailiateUntilApologyTransformer()
TFT = RUA(axelrod.Cooperator)
player = TFT()
```
This actually does transform `Cooperator` exactly into `TitForTat` (it passes all the tests for TFT, which specify TFT completely). For other strategies, it only affects what follows an opponent's defection, otherwise the player's desired plays are simply passed through.

There are additional transformers for the following modifications:
* Adding noise (similar to the noisy tournament but only to one of the players)
* Adding probabilistic defection forgiveness
* Defecting on the last N moves (only if the proper tournament attribute is available), or more generally finishing with any given sequence of moves
* Starting with any specific sequences of moves then playing as the strategy intended
* Internally tracking intended history (e.g. to infer ambient noise with)

If you've been wondering if MetaHunter would be better with a little TFT style retaliation, there's no need to write a new strategy! Moreover the transformations can be chained or composed, so you can add as many transformations to any strategy. The transformations are preserved when the player is cloned.

These transforms can also be used as class decorators. As an example in the library, I modified BackStabber's implementation:
```
@FinalTransformer([D, D, D]) # End with three defections
class BackStabber(Player):
...
    def strategy(self, opponent):
        if not opponent.history:
            return C
        if opponent.defections > 3:
            return D
        return C
```
while removing these two lines from the strategy method:
```
-        if len(opponent.history) > (self.tournament_attributes['length'] - 3):
-            return D
```
Now it works even if the tournament length is not known, and defects as intended if the length is available.

Some ideas for additional transforms:
* A grudger style never forgive
* Automatic noise detection
* Automatic memory_depth inference
* Adding hunting behavior to other strategies
* Add other types of retaliation (e.g. TitForTwoTats)

I only ran into one significant issue while testing -- the fact that the classifier dict is sometimes a class variable made it difficult to test, since it would overwrite the original class variable (in the super class). One workaround would be to set the classifiers in a method rather than have them as class variables.

Once #372  is merged I'll add a page of usage documentation.

### DIFF
--- a/axelrod/strategies/backstabber.py
+++ b/axelrod/strategies/backstabber.py
@@ -4,7 +4,7 @@ from .strategy_transformers import FinalTransformer
 
 C, D = Actions.C, Actions.D
 
-@FinalTransformer([D, D, D]) # End with three defections
+@FinalTransformer([D, D]) # End with three defections
 class BackStabber(Player):
     """
     Forgives the first 3 defections but on the fourth
--- a/axelrod/strategies/backstabber.py
+++ b/axelrod/strategies/backstabber.py
@@ -1,7 +1,10 @@
 from axelrod import Player, Actions
 
+from strategy_transformers import FinalTransformer
+
 C, D = Actions.C, Actions.D
 
+@FinalTransformer([D, D, D]) # End with three defections
 class BackStabber(Player):
     """
     Forgives the first 3 defections but on the fourth
@@ -20,8 +23,6 @@ class BackStabber(Player):
     def strategy(self, opponent):
         if not opponent.history:
             return C
-        if len(opponent.history) > (self.tournament_attributes['length'] - 3):
-            return D
         if opponent.defections > 3:
             return D
         return C
--- a/axelrod/strategies/backstabber.py
+++ b/axelrod/strategies/backstabber.py
@@ -1,6 +1,6 @@
 from axelrod import Player, Actions
 
-from strategy_transformers import FinalTransformer
+from .strategy_transformers import FinalTransformer
 
 C, D = Actions.C, Actions.D
 
--- a/axelrod/strategies/backstabber.py
+++ b/axelrod/strategies/backstabber.py
@@ -1,6 +1,6 @@
 from axelrod import Player, Actions
 
-from .strategy_transformers import FinalTransformer
+from axelrod.strategy_transformers import FinalTransformer
 
 C, D = Actions.C, Actions.D
 
--- a/axelrod/strategies/strategy_transformers.py
+++ b/axelrod/strategies/strategy_transformers.py
@@ -1,0 +1,272 @@
+
+"""
+Strategy Transformers -- class decorators that transform the behavior of any
+strategy.
+
+Run Axelrod one tournament
+
+Flip Action
+Forgiver
+Initial Sequence
+Final Sequence
+
+TFT -- force a repayment, return to other strategy
+
+Noisy
+RetaliateUntilApology
+
+Compose.
+
+Memory-depth inference
+
+As Decorators
+
+Meta Strategies
+
+"""
+
+import random
+from types import FunctionType
+
+import axelrod
+from axelrod import flip_action, random_choice, simulate_play, Actions
+C, D = Actions.C, Actions.D
+
+# Note: the history is overwritten with the modified history
+# Just like in the Noisy case
+# This can lead to unexpected behavior, such as when
+# Flip transform is applied to Alternator
+
+
+def generic_strategy_wrapper(player, opponent, proposed_action, *args, **kwargs):
+    """
+    Strategy wrapper functions should be of the following form.
+
+    Parameters
+    ----------
+    player: Player object or subclass (self)
+    opponent: Player object or subclass
+    proposed_action: an axelrod.Action, C or D
+        The proposed action by the wrapped strategy
+        proposed_action = Player.strategy(...)
+    args, kwargs:
+        Any additional arguments that you need.
+
+    Returns
+    -------
+    action: an axelrod.Action, C or D
+
+    """
+
+    # This example just passes through the proposed_action
+    return proposed_action
+
+def StrategyTransformerFactory(strategy_wrapper, wrapper_args=(), wrapper_kwargs={},
+                        name_prefix="Transformed "):
+    """Modify an existing strategy dynamically by wrapping the strategy
+    method with the argument `strategy_wrapper`.
+
+    Parameters
+    ----------
+    strategy_wrapper: function
+        A function of the form `strategy_wrapper(player, opponent, proposed_action, *args, **kwargs)`
+    wrapper_args: tuple
+        Any arguments to pass to the wrapper
+    wrapper_kwargs: dict
+        Any keyword arguments to pass to the wrapper
+    name_prefix: string, "Transformed "
+        A string to prepend to the strategy and class name
+    """
+
+    # Create a function that applies a wrapper function to the strategy method
+    # of a given class
+    def decorate(PlayerClass):
+        """
+        Parameters
+        ----------
+        PlayerClass: A subclass of axelrod.Player, e.g. Cooperator
+
+        Returns
+        -------
+        new_class, class object
+            A class object that can create instances of the modified PlayerClass
+        """
+
+        # Define the new strategy method, wrapping the existing method
+        # with `strategy_wrapper`
+        def strategy(self, opponent):
+            # Is the original strategy method a static method?
+            if isinstance(PlayerClass.strategy, FunctionType):
+                proposed_action = PlayerClass.strategy(opponent)
+            else:
+                proposed_action = PlayerClass.strategy(self, opponent)
+            # Apply the wrapper
+            return strategy_wrapper(self, opponent, proposed_action,
+                                    *wrapper_args, **wrapper_kwargs)
+
+        # Define a new class and wrap the strategy method
+        # Modify the PlayerClass name
+        new_class_name = name_prefix + PlayerClass.__name__
+        # Modify the Player name (class variable inherited from Player)
+        name = name_prefix + PlayerClass.name
+        # Dynamically create the new class
+        new_class = type(new_class_name, (PlayerClass,),
+                         {"name": name, "strategy": strategy})
+        return new_class
+    return decorate
+
+def flip_wrapper(player, opponent, action):
+    """Applies flip_action at the class level."""
+    return flip_action(action)
+
+FlipTransformer = StrategyTransformerFactory(flip_wrapper, name_prefix="Flipped ")
+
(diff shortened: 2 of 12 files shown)


drvinceknight:
This looks interesting and obviously fits the goal of the library of facilitating the study of IPDs. 

I need to get my head around it a bit more I think. I haven't looked at the code yet, so just thinking about the problem this solves: can you help me out a bit more than you already have. Are you thinking the (main) use case is for people using the library and wanting to create on the fly strategies? Or are these (mainly) to help create strategies?

Very much in favour, for example the `@FinalTransformer([D, D, D])` is a lovely decorator and it helps 'idiot proof' as you say:

> Now it works even if the tournament length is not known, and defects as intended if the length is available.

Just my usual initial slowness at getting my head around it :+1: :smile: 

> Once #372 is merged I'll add a page of usage documentation.

This is a perfect example of a 'further_topics' tutorial I think?
This is where the new modular tutorial docs I think come in to their own, instead of a big terrible monster being creating (as the previous docs were) we can make modular adjustments...

Also: anything we can do about 0.1% drop in coverage? (Not a disaster obviously)

marcharper:
I'm sure that I can get the coverage up.

This should prevent code duplication and expand the possible strategies the library can produce substantially. There are already many strategies that are e.g. TFT but defect on round one instead.

Plus you can, for example, run a tournament of many strategies and apply a transform to all of them. What happens if everyone defects on the first round? If everyone retaliates like TFT?

drvinceknight:
> This should prevent code duplication and expand the possible strategies the library can produce substantially. There are already many strategies that are e.g. TFT but defect on round one instead.
> 
> Plus you can, for example, run a tournament of many strategies and apply a transform to all of them. What happens if everyone defects on the first round? If everyone retaliates like TFT?

Have had more time to think about it: big big fan.

Will look through the code itself as soon as possible.

marcharper:
> Have had more time to think about it: big big fan.

Glad to hear about it, as I'm fairly certain these transformers are the key to studying the IPD with category theory. I'll write something up eventually...

drvinceknight:
Oh wow: that sounds awesome! Category theory is something I know very
little about.

meatballs:
> Category theory is something I know very
little about.

I can only aspire to attain such a level of knowledge

drvinceknight:
Be careful what you wish for :)

marcharper:
> I can only aspire to attain such a level of knowledge

It's nbd really, you just have to waste years of your life in mathematics graduate school, postponing other life goals and the development of a viable career in the meantime. Or learn a bit of Haskell I suppose.

drvinceknight:
Lol. Slightly related. I've seen a quote somewhere, something like:

'graduate school is reducing current income in order to reduce future
income'

axelrod/strategies/backstabber.py
```diff
@@ -1,7 +1,10 @@
 from axelrod import Player, Actions
 
+from .strategy_transformers import FinalTransformer
+
 C, D = Actions.C, Actions.D
 
+@FinalTransformer([D, D, D]) # End with three defections
```
meatballs: Shouldn't this be two defections? The orginal condition was for opponent history length to be greater than 197, so 198 is the first occurrence. If the the history is 198 long, then we are on turn 199.

meatballs:
This is an excellent piece of work and opens up all sorts of possibilities. Other than my comment on backstabber, I'm more than happy to see this go in.

axelrod/strategies/backstabber.py
marcharper: Thanks -- I fixed it, and updated the tests.

axelrod/strategies/strategy_transformers.py
```diff
@@ -0,0 +1,202 @@
...
+
+    if length < 0: # default is -1
+        return action
+
+    index = length - len(player.history)
+    if index <= len(seq):
+        return seq[-index]
+    return action
+
+# Defect on last N actions
```
drvinceknight: This comment isn't quite right? It's do seq on last len(seq) actions right? The default is default...

axelrod/strategies/strategy_transformers.py
```diff
@@ -0,0 +1,202 @@
...
+        if opponent.history[-1] == D:
+            self.is_retaliating = True
+        if self.is_retaliating:
+            if opponent.history[-1] == C:
+                self.is_retaliating = False
+                return C
+            return D
+        return action
+
+def RetailiateUntilApologyTransformer():
```
drvinceknight: Docstring?

axelrod/strategies/strategy_transformers.py
```diff
@@ -0,0 +1,202 @@
...
+    strategy_wrapper = RetaliationWrapper()
+    return StrategyTransformerFactory(strategy_wrapper, name_prefix="RUA")
+
+def history_track_wrapper(player, opponent, action):
```
drvinceknight: Docstring?

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -0,0 +1,184 @@
...
+        random.seed(12)
+        p1 = TFT()
+        p2 = axelrod.Random()
+        for _ in range(5):
+            p1.play(p2)
+        self.assertEqual(p1.history, [C, C, D, D, C])
+
+
+
+## Test that RUA(Cooperator) is the same as TitForTat
```
drvinceknight: What's going on below?

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -0,0 +1,184 @@
...
+    def test_flip_transformer(self):
+        # Cooperator to Defector
+        p1 = axelrod.Cooperator()
+        p2 = FlipTransformer(axelrod.Cooperator)() # Defector
+        self.assertEqual(simulate_play(p1, p2), (C, D))
+        self.assertEqual(simulate_play(p1, p2), (C, D))
+        self.assertEqual(simulate_play(p1, p2), (C, D))
+
+    def test_noisy_transformer(self):
+        # Test Noisy transformer
```
drvinceknight: Why go with `#` and not a Docstring? Also some missing below.

drvinceknight:
Just some minor things I've commented on. I think the main thing this is missing is an advanced tutorial or further topics. I'd also suggest that a pointer gets put in the contribution docs to that tutorial but as you say I guess that is easiest if you wait for #372?

I completely agree with @meatballs: this is great and leaves open the possibility of further transformations to be added :) (that could be explained in the tutorial perhaps?)

axelrod/strategies/strategy_transformers.py
marcharper: Good catch -- that's what it was initially but then I generalized to an arbitrary sequence.

drvinceknight: Thought as much :)

axelrod/tests/unit/test_strategy_transformers.py
marcharper: This is the "proof" that the transformed class passes all the TFT tests by reusing the `TestTitForTat` test class from `test_titfortat.py`. However it fails the memory_depth test (since the transformer doesn't modify it, though perhaps it could). If I set the memory_depth manually, it actually modifies the classifier of the originally transformed class, since the classifier dict is a class variable until `__init__` is called, and is inherited by the transformed class. In this case I think the issue is that we are abusing that fact that classifiers are generally instance specific but in some cases identical for all instances (like for `Cooperator`), using a class variable when we should be using an instance variable.

The test code is fine once we decide what, if anything, to do about this -- it's not necessary to run the test for coverage, but it would be really nice to be able to include more proofs that many strategies can be decomposed into a series of transformations. So I left it in as a commented-out example and will add an issue.

One solution would be to add a method `register_classifiers` to the to the `Player` class. We're already modifying classifiers in `Player.__init__` and some strategies are modifying these further based on their initializing parameters. It's probably cleaner and better to do this with a single function, and to call this function at the end of `__init__`.

marcharper:
I will certainly improve the documentation -- both the docstrings and an advanced tutorial, both of which are needed since we're metaprogramming now. But I do want #372 to hit first so I don't have merge conflicts on the docs (and I wanted to make sure that you all liked the idea).

drvinceknight:
Love the idea.

axelrod/tests/unit/test_strategy_transformers.py
marcharper: I'm willing to handle the above (soon but not today), and it's probably better done as a separate PR. This will impact #348. Just give me a thumbsup.

drvinceknight: :+1:

drvinceknight: Is it not (for this PR) worth removing this? Apologies if I'm misunderstanding things.

docs/tutorials/advanced/strategy_transformers.rst
```diff
@@ -0,0 +1,187 @@
...
+we applied the transformer. The transformer also changed the name of the
+class and player::
+
+    >>> player.name
+    'Flipped Cooperator'
+    >>> FlippedCooperator.name
+    'Flipped Cooperator'
+
+This behavor can be supressed by setting the :code:`name_prefix` argument::
+    FlipTransformer = StrategyTransformerFactory(flip_wrapper, name_prefix="")
```
drvinceknight: blank line needed here (before the code)

docs/tutorials/advanced/strategy_transformers.rst
```diff
@@ -0,0 +1,187 @@
...
+
+This defines a strategy that cooperates except on the first two and last two rounds.
+
+
+Included Transformers
+---------------------
+
+The library includes the following transformers:
+
+* :code:`FlipTransformer`: Flips all actions::
```
drvinceknight: Blank line between all of these?

drvinceknight:
Yeah this is brilliant. Just some blank lines to tidy things a bit.

I really like the tutorial.

Brilliant contribution, very exciting.

drvinceknight:
:+1:

langner:
This is very nice!

Issue #382, opened 2015-10-25
Implement verbose TitForTat


drvinceknight:
We have discussed on multiple tickets the fact that it's nice to have a very verbose strategy that is perhaps not coded as efficiently as possible.

This is now in the documentation http://axelrod.readthedocs.org/en/latest/tutorials/contributing/strategy/writing_the_new_strategy.html but has not actually gotten in to the library (oversight/confusion on my part). Fixing now.

PR #383, opened 2015-10-25
382 - Make TitForTat verbose

drvinceknight:
This addresses #382 and make TitForTat verbose.

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -4,8 +4,16 @@ C, D = Actions.C, Actions.D
 
 
 class TitForTat(Player):
-    """A player starts by cooperating and then mimics previous move by opponent."""
+    """
+    A player starts by cooperating and then mimics previous move by
+    opponent.
 
+    Note that the code for this strategy is written in a fairly verbose
+    way. This is done so that it can serve as an example strategy for
+    those who might be new to Python.
+    """
+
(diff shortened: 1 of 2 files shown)


axelrod/strategies/titfortat.py
```diff
@@ -4,8 +4,16 @@
 
 
 class TitForTat(Player):
-    """A player starts by cooperating and then mimics previous move by opponent."""
+    """
+    A player starts by cooperating and then mimics previous move by
```
marcharper: Can we say "mimics the previous action of the opponent." ?

marcharper:
:+1:

axelrod/strategies/titfortat.py
drvinceknight: Yes that's indeed what it should be. Away from a computer right now but
will make that change (and also change in the docs).

drvinceknight:
@langner jumped the gun a bit, I just need to push the slight doc change. Will throw that straight to master... (If this annoys anyone just let me now and I'll be sure to not do again.)

langner:
> @langner jumped the gun a bit, I just need to push the slight doc change. Will throw that straight to master... (If this annoys anyone just let me now and I'll be sure to not do again.)

Oops. Sorry about that.

drvinceknight:
No problemo :+1: :smile:

Issue #313, opened 2015-09-21
Inconsistent terminology used to describe rounds/turns/plays...


emmagordon:
This gets quite confusing!

Solution could be to clearly define some synonyms, or ideally make the docs/docstrings/comments/graphs etc.. all consistent in their use of terminology.

meatballs:
:+1:

drvinceknight:
I suggest we also add a `glossary.rst` and I suggest the following be the terminology:

- Turns are the 1 shot interactions between two players (throughout this has been confused with rounds).
- A Match is 200 turns
- A round robin is all the matches.

Note that this should be done in the documentation which is also being reworked #299

marcharper:
See also #359 and #362 for moves/actions

drvinceknight:
I've added a glossary #372 which is also starting to address #299.

drvinceknight:
I hope that #384 coupled with the already merged in #372 address this.

PR #384, opened 2015-10-25
313 - Address terms that are interchangeably used in documentation

drvinceknight:
This fixes some bugs in docs but also (mainly) aims to try and address #313:

- Have added player/strategy to glossary
- Have gone through and tried to tweak as necessary


marcharper:
:+1:

PR #386, opened 2015-10-26
Omega TFT

marcharper:
Adds the OmegaTFT strategy that won a 2004 tournament, described [here](http://arxiv.org/pdf/cs/0609017.pdf).

PR #388, opened 2015-10-26
Adaptive Pavlov, two versions

marcharper:
Two new strategies, both variants of APavlov that won a 2005 IPD tournament.

Also added a kwarg `attrs` to test_responses, which is a dictionary of attributes to check at the end of testing the sequence of plays. This simplifies the testing of internal variables of strategies in significant situations, and should useful in many other places in the library.

### DIFF
--- a/axelrod/strategies/apavlov.py
+++ b/axelrod/strategies/apavlov.py
@@ -1,0 +1,108 @@
+from axelrod import Player, Actions, flip_action
+
+C, D = Actions.C, Actions.D
+
+
+class APavlov2006(Player):
+    """
+    APavlov as defined in http://www.cs.nott.ac.uk/~pszjl/index_files/chapter4.pdf (pages 10-11).
+    """
+
+    name = "Adapative Pavlov 2006"
+    classifier = {
+        'memory_depth': float('inf'),
+        'stochastic': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self):
+        Player.__init__(self)
+        self.opponent_class = None
+
+    def strategy(self, opponent):
+        # TFT for six rounds
+        if len(self.history) < 6:
+            return D if opponent.history[-1:] == [D] else C
+        # Classify opponent
+        if len(self.history) % 6 == 0:
+            if opponent.history[-6:] == [C] * 6:
+                self.opponent_class = "Cooperative"
+            if opponent.history[-6:] == [D] * 6:
+                self.opponent_class = "ALLD"
+            if opponent.history[-6:] == [D, C, D, C, D, C]:
+                self.opponent_class = "STFT"
+            if opponent.history[-6:] == [D, D, C, D, D, C]:
+                self.opponent_class = "PavlovD"
+            if not self.opponent_class:
+                self.opponent_class = "Random"
+
+        # Play according to classification
+        if self.opponent_class in ["Random", "ALLD"]:
+            return D
+        if self.opponent_class == "STFT":
+            if len(self.history) % 6 in [0, 1]:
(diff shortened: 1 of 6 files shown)


langner:
+1

axelrod/strategies/apavlov.py
```diff
@@ -0,0 +1,108 @@
+from axelrod import Player, Actions, flip_action
+
+C, D = Actions.C, Actions.D
+
+
+class APavlov2006(Player):
+    """
+    APavlov as defined in http://www.cs.nott.ac.uk/~pszjl/index_files/chapter4.pdf (pages 10-11).
```
drvinceknight: Could you include a little description here also.

axelrod/strategies/apavlov.py
```diff
@@ -0,0 +1,108 @@
...
+            return D if opponent.history[-1:] == [D] else C
+
+    def reset(self):
+        Player.reset(self)
+        self.opponent_class = None
+
+
+class APavlov2011(Player):
+    """
+    APavlov as defined in http://www.graham-kendall.com/papers/lhk2011.pdf, as
```
drvinceknight: Same?

docs/tutorials/contributing/strategy/writing_the_new_strategy.rst
```diff
@@ -7,7 +7,7 @@ Identify a new strategy
 The library has a large number of strategies::
 
     >>> import axelrod as axl
-    >>> len(axl.strategies)
+    >>> len(axl.strategies) # doctest: +SKIP
     103
 
 If you're not sure if you have a strategy that has already been implemented
```
drvinceknight: Not sure I understand why we should skip this?

Is it not a nice thing to have and up to date? If it's already implemented it'll fall out of the review?

marcharper: Yes, but it creates a lot of merge conflicts. Not sure what the best solution is...

drvinceknight: Creates failures on travis? Yes indeed, this is non trivial...

Obviously one way is to ask for a rebase but that's not super accessible...

drvinceknight: I think that if we have the length of strategies in the docs then it should be tested (to make sure it's accurate). If we're going to ignore it then we should just remove it.

I would rather it was left in but yeah, not sure how to get around it.

marcharper: Well at least the diagrams will show how many strategies their are. We could also say that "the library has at least X many strategies" and periodically update that.

Technically the library can generate infinitely many strategies.

drvinceknight: > Could we have an update script that corrects the value somehow?

This can go wrong if we get lazy. I think that it's best to just remove it all together. As you say the diagrams show the number of strategies :) :tada:

drvinceknight: Same goes for #386.

docs/tutorials/contributing/strategy/writing_the_new_strategy.rst
```diff
@@ -7,8 +7,8 @@ Identify a new strategy
 The library has a large number of strategies::
 
     >>> import axelrod as axl
-    >>> len(axl.strategies)
-    104
+    >>> len(axl.strategies) # doctest: +SKIP
```
drvinceknight: Could you just remove this?

I think it's best to not be there if it's not being tested, as you say we know how many strategies there are from the diagrams.

PR #390, opened 2015-10-27
Rework strategy transformers and documentation.

marcharper:
This is an update to the strategy transformers to use a class to better handle arguments in transformers. 

Edit: removed discussion of #391 re: #392

### DIFF
--- a/docs/tutorials/advanced/strategy_transformers.rst
+++ b/docs/tutorials/advanced/strategy_transformers.rst
@@ -31,7 +31,7 @@ class and player::
 
 This behavor can be supressed by setting the :code:`name_prefix` argument::
 
-    FlipTransformer = StrategyTransformerFactory(flip_wrapper, name_prefix="")
+    FlipTransformer = StrategyTransformerFactory(flip_wrapper, name_prefix="")()
 
 Note carefully that the transformer returns a class, not an instance of a class.
 This means that you need to use the Transformed class as you would normally to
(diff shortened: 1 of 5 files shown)


docs/tutorials/advanced/strategy_transformers.rst
```diff
@@ -100,6 +96,26 @@ The library includes the following transformers:
...
+    cls1 = FinalTransformer([D,D])(InitialTransformer([D,D])(axelrod.Cooperator))
+    p1 = cls1()
+
+This defines a strategy that cooperates except on the first two and last two
+rounds. Alternatively, you can make a new class using
+:code:`compose_transformers`::
+
+    cls1 = compose_transformers(FinalTransformer([D, D]),
+                                InitialTransformer([D, D]))
+    p1 = cls1(axelrod.Cooperator)()
```
drvinceknight: Could these be doctested `>>>`, I know there is no output but say that at some point the input changes etc and typing this would give an error?

Same for quite a few here (sorry should have raised this in previous PR).

marcharper: Done. There was one that I specifically left untested, the one with `*args, **kwargs`, since they would have to be defined for the test to run properly.

drvinceknight:
:+1: from me.

drvinceknight:
I'm merging as this is related to the bug fixed by #392. 

Apologies for all the solo merges but they were bug fixes.

Issue #391, opened 2015-10-27
Strategy Transforming broke caching


marcharper:
The standard library pickle can't handle decorated classes. ~~The p~~ Possible fixes are:

* Use [cloudpickle](https://github.com/cloudpipe/cloudpickle) instead of pickle, which is as easy as `import cloudpickle as pickle`.
* Disable caching for transformed players.

Edit: the cloudpickle fix is in, alternate suggestions still wanted.

langner:
Another option would be to not pickle the classes themselves. Additionally, JSON might be a more natural, and human-readable format for storing the cache.

marcharper:
Yeah, there are many that could work. The cloudpickle approach is very simple and maybe be a good quickfix (I have a two-line commit to handle it). I'll open a PR for that in case we want to merge it while we discuss other solutions.

meatballs:
> Another option would be to not pickle the classes themselves. Additionally, JSON might be a more natural, and human-readable format for storing the cache.

I agree. The names of the classes could be used as the key rather than the classes themselves and JSON might be a much nicer way to represent the content - especially when we move to caching the interactions rather than the payoff and cooperation

drvinceknight:
Would JSON also be relevant if/when we build the web api?

meatballs:
> Would JSON also be relevant if/when we build the web api?

Doesn't make any difference as this would all be run within the celery task anyway

drvinceknight:
_Skuttles back to his cave to work on docs..._

marcharper:
The key should be strategy name + init_args to properly handle parameterized strategies.

drvinceknight:
I have merged #392 as it was a bug. 

When we decide what is the best way to fix it (ie this issue) we can go with that?

drvinceknight:
Do we want to discuss this further at some point (ie leave this issue open until people have the time) or are we happy with @marcharper's fix?

meatballs:
Let's go for the quick fix for now. #284 is going to to required some mucking about with the cache anyway

drvinceknight:
> Let's go for the quick fix for now.

Quick fix is already in (was a bug). Mainly just seeing if I can close this issue to keep things tidy. No harm in it staying open.

drvinceknight:
Have removed the bug label (no longer a bug).

marcharper:
So are we happy with cloudpickle for now? If so we can close this one as well.

drvinceknight:
Yup!

PR #392, opened 2015-10-27
Pickle fix

marcharper:
Quick fix for #391 using cloudpickle. Please see the discussion at #391.

### DIFF
--- a/axelrod/tournament_manager.py
+++ b/axelrod/tournament_manager.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, unicode_literals, print_function
 
 import os
-import pickle
+import cloudpickle as pickle
 
 from .tournament import *
 from .plot import *
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 numpy==1.9.2
 matplotlib==1.4.2
+cloudpickle==0.1.1


drvinceknight:
Travis errored out:

```
Collecting cloudpickle==0.11 (from -r requirements.txt (line 3))
  Could not find a version that satisfies the requirement cloudpickle==0.11 (from -r requirements.txt (line 3)) (from versions: 0.1.0, 0.1.1)
  Some externally hosted files were ignored as access to them may be unreliable (use --allow-external to allow).
  No distributions matching the version for cloudpickle==0.11 (from -r requirements.txt (line 3))
The command "pip install -r requirements.txt" failed and exited with 1 during .
Your build has been stopped.
```

Just reseting the build in case...

marcharper:
Shoudl be 0.1.1 I guess, let me fix...

drvinceknight:
What's the status on this? The discussion on #391 seems to suggest doing something different?

marcharper:
I think we should merge this now, or disable caching, until we decide on a permanent fix. Right now the tournaments run but they crash with an error when Backstabber is cached.

drvinceknight:
Merged as it's a bug.

marcharper:
Just note that this needs #390 to fix the bug still, as far as I can tell from my testing.

PR #394, opened 2015-10-27
Adding citation of library to docs.

drvinceknight:
- Thrown it at the root of the docs (wasn't sure where else to go).
- Still pointing at the project site (is this what we want?).


marcharper:
Looks good to me, we can update it later if needed. Thanks!

PR #397, opened 2015-10-28
377 -minor fix to docs

ABarriscale:


drvinceknight:
Worked with Anna on this one: :+1:

PR #399, opened 2015-10-28
New transformers: Apologizing, Retaliating variant, Deadlock Breaking, Grudging

marcharper:
`RetaliateTransformer(N)`: Retaliation N times after a defection:
```
    >>> import axelrod
    >>> from axelrod.strategy_transformers import RetaliationTransformer
    >>> TwoTitsForTat = RetaliationTransformer(2)(axelrod.Cooperator)
    >>> player = TwoTitsForTat()
```

`ApologizingTransformer`: Apologizes after a round of `(D, C)`:
```
    >>> import axelrod
    >>> from axelrod.strategy_transformers import ApologyTransformer
    >>> ApologizingDefector = ApologyTransformer([D], [C])(axelrod.Defector)
    >>> player = ApologizingDefector()
```
`DeadlockBreakingTransformer`: Attempts to break `(D, C) -> (C, D)` deadlocks by cooperating:
```
    >>> import axelrod
    >>> from axelrod.strategy_transformers import DeadlockBreakingTransformer
    >>> DeadlockBreakingTFT = DeadlockBreakingTransformer(axelrod.TitForTat)
    >>> player = DeadlockBreakingTFT()
```

PR #402, opened 2015-10-28
Some necessary changes for transformers to be applied to all strategies; additional test to catch implementation issues for new strategies.

marcharper:
We've been fast and loose with how we use `super` (myself included), and in some cases we hit the usual infinite recursion issue when a transformer is applied to a class that uses super in particular ways.

This PR corrects these issues, makes a change to the `staticmethod` testing of transformers to be more robust, and adds in a test to try to apply a transform to every strategy in the library to catch future strategy issues.

Needs to go in after #399 (for `IdentityTransformer`), tests should pass then. I did not add it to #399 since it's already got a :+1:

### DIFF
--- a/axelrod/strategies/axelrod_first.py
+++ b/axelrod/strategies/axelrod_first.py
@@ -223,7 +223,7 @@ class Joss(MemoryOnePlayer):
         """
         four_vector = (p, 0, p, 0)
         self.p = p
-        super(self.__class__, self).__init__(four_vector)
+        super(Joss, self).__init__(four_vector)
         self.init_args = (p,)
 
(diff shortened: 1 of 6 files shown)


marcharper:
Rebased onto latest master

marcharper:
This one is ready for review @drvinceknight @meatballs (should be easy)

drvinceknight:
:+1: Really like the test that checks that the transform works on all strategies.

marcharper:
I learned the hard way when running some tournaments applying transformers to every strategy. Hopefully this will keep anyone else from experiencing the same.

PR #405, opened 2015-10-29
Five New ZD strategies

marcharper:
This brings us to ~~100~~ 102 ordinary strategies!


axelrod/strategies/memoryone.py
```diff
@@ -146,7 +146,7 @@ def __init__(self, ep=0.05):
         self.set_four_vector(four_vector)
 
 
-class ZeroDeterminantPlayer(MemoryOnePlayer):
+class LRPlayer(MemoryOnePlayer):
```
drvinceknight: Why LRP? (I guess it's in the link, being lazy.)

drvinceknight:
My initial comment is that more or less all the docs strings are: "An Extortionate Zero Determinant Strategy.". Could you differentiate? Mention parameters?

axelrod/strategies/memoryone.py
marcharper: That's what they are called in the reference. It stands for Linear Relation -- do you prefer that?

marcharper:
Can do

axelrod/strategies/memoryone.py
drvinceknight: No no, this wasn't a review question. Was just curious and too lazy to click the link... :)

drvinceknight:
This looks good to me. :+1:

marcharper:
Ok, I'll stop for now.

drvinceknight:
Is `BiPolar` a name from the literature?

If it is: I'd like that to be referenced, if not, then I'd prefer to change the name. I realise I'm being a bit over cautious and I'm being overly inclusive here but I think that if we can it would be nice to avoid names that might make someone feel uncomfortable from a mental health point of view. I know that the Django project a couple of years ago made a move to change a word in there docs that at first seemed a bit over the top but quite simply 'did no harm to change' (I don't remember exactly). I realise we have names like `BackStabber` but for some reason that hasn't made me feel the same way. (Not trying to make this a big deal).

A part from that I think this is good to go and does a really great job of increasing the number of strategies.

marcharper:
I just made up a name -- I was thinking of a di-pole from physics rather than any kind of psychological condition. Do you have any suggestions? DiPolar? AllCorD?

drvinceknight:
> di-pole from physics

I thought that was probably the case. Not at all accusing anyone of being
insensitive. I'm being very very cautious here.

Whilst I like the sound of `Di/BiPolar` I think that `AllCorD` is best?

marcharper:
OK I'll make the change and rebase.

drvinceknight:
> OK I'll make the change and rebase.

I'm in front of my machine marking for a couple more hours so just ping me and I'll give the thumbs up :)

marcharper:
It will be much later / tomorrow, I did not bring a real computer with me.

drvinceknight:
> It will be much later / tomorrow, I did not bring a real computer with me.

No worries, thought you were back home or something :) Hope the Wedding went/goes well :)

docs/tutorials/further_topics/classification_of_strategies.rst
```diff
@@ -24,13 +24,13 @@ This allows us to, for example, quickly identify all the stochastic
 strategies::
 
     >>> len([s for s in axl.strategies if s().classifier['stochastic']])
-    31
+    36
 
 Or indeed find out how many strategy have only use 1 turn worth of memory to
```
drvinceknight: There's a typo here, should be `how many strategies only use 1 turn worth of memory`, nice if you can fix in this PR but no problem otherwise.

drvinceknight:
A part from that one typo (of mine) this is a :+1: from me.

meatballs:
I fixed the typo and merged this

drvinceknight:
:+1: Hooray! :)

PR #408, opened 2015-11-01
Fix deprecated stackplot calls

bootandy:
stackplot should be called with the ecosystem object and not its list
this commit fixes the test and some documentation.

Change made due to these commits:
c965e249ebb8fac64e23618e80ab19deb8a6dc30
a75c52f356b1dd50910974f1bbc52b3c429dbabf

### DIFF
--- a/docs/tutorials/getting_started/usage.old
+++ b/docs/tutorials/getting_started/usage.old
@@ -461,7 +461,7 @@ We can see how this differs when the initial population contains a large number 
     eco = axelrod.Ecosystem(results, population=[.1, .05, .7, .1, .05])
     eco.reproduce(50) # Evolve the population over 50 time steps
     plot = axelrod.Plot(results)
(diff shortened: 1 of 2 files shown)


docs/tutorials/getting_started/usage.old
```diff
@@ -479,7 +479,7 @@ Here is a with an even larger initial number of :code:`Defector` (note that it t
     eco = axelrod.Ecosystem(results, population=[.1, .05, 7, .1, .05])
     eco.reproduce(140) # Evolve the population over 140 time steps
     plot = axelrod.Plot(results)
-    p = plot.stackplot(eco.population_sizes)
+    p = plot.stackplot(eco)
```
drvinceknight: This doesn't matter terribly as really all the `.old` files should be deleted at some point *but* still good to not have any legacy code lying around.

drvinceknight:
Good PR: :+1:

PR #410, opened 2015-11-02
More refactoring of payoff.py and cooperation.py

meatballs:
Adds player_count function to derive number of players from interactions dictionary

### DIFF
--- a/axelrod/tests/unit/test_payoff.py
+++ b/axelrod/tests/unit/test_payoff.py
@@ -28,6 +28,12 @@ class TestPayoff(unittest.TestCase):
             [13, 11, 13]
         ]
 
+        cls.expected_scores_matrix = [
+            [26, 26],
+            [24, 24],
+            [24, 24]
+        ]
+
         cls.expected_payoff = [
             [[11.0, 11.0], [13, 13], [15, 12]],
             [[13, 13], [15.0, 15.0], [14, 7]],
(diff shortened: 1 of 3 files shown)


meatballs:
Don't merge this. I've changed my mind about the scores_matrix function. It means the payoff gets calculated twice and I don't like that! I'm going to get rid of it.

axelrod/tests/unit/test_payoff.py
```diff
@@ -74,14 +80,23 @@ def setUpClass(cls):
             [-1.0, 0.0, 0.0, 1.0, 0.0, 0.0]
         ]
 
+    def test_player_count(self):
+        nplayers = ap.player_count(self.interactions)
+        self.assertEqual(nplayers, 3)
```
drvinceknight: Worth including some edge cases? A single player etc...?

meatballs:
done!

axelrod/tests/unit/test_payoff.py
meatballs: Yeah, probably! I'll think of something to add

drvinceknight:
I'm happy with this. :+1:

meatballs:
No excuses left now. I'm going to have tackle the ripping apart of Round Robin etc. to get the interactions out!

drvinceknight:
:fearful: :muscle:

marcharper:
looks good to me, @meatballs just checking that the don't merge comment above no longer applies?

meatballs:
@marcharper correct. I removed my own code that was causing me offence!

PR #398, opened 2015-10-28
Added LookerUp strategy

mojones:

PR #411, opened 2015-11-03
Mojones-master this adds to #398

drvinceknight:
This PR simply adds a couple of commits to @mojones's #398 merging this would automatically merge in #398 (because github is cool).

This really only needs to be ok'd by @marcharper as #398 has gone through a detailed review by him and I. This is a minor review change:

- Adding an example to docstring
- Fixing a test (unless I'm mistaken the previous version was not running one set of tests)
- Changing something back to string notation for clarity (Marc: you might well have had a reason for changing this).


marcharper:
Looks fine to me!

meatballs:
and there we have it!

PR #414, opened 2015-11-05
Creates Match class

meatballs:
More work for #284 

This class isn't (yet) used by a tournament, but it's ready to roll and could be called directly.


marcharper:
Nice. This is going to make my code on in the AxelrodExamples repository a lot cleaner -- I've been not so elegantly simulating matches. So I'm a fan :+1: 

There's also a `simulate_match` in mindreader, and `simulate_play` in MockPlayer. At some point it may be worth trying to consolidate all these things.

marcharper:
Looks fine, can you add to the documentation regarding how to use the `Match` class? It looks like after initialization the intention is to access `match.results` rather than call `match.play()` (which would yield the same but ignore the cache, if present). I'm not sure that I would have guessed that -- maybe if `play()` were `_play()` I would not think to call it. Accessing `match.results` would yield another set of interactions right?

Also, is there a reason not to call `result` at the end of `__init__`? I get that you might want to repeat the match several times, but surely if one is building the object one intends to play at least one round. Or is it delayable for parallelization?

meatballs:
@marcharper I like both those suggestions. I'll rename play() to _play() and I'll create a setter method for results which gets called by __init__()

(Oh, and I'll write some docs too)!!

marcharper:
Just a thought -- you could combine both `_play` and `results` into just `play()`. I could see building the matchup and then wanting to call `play` repeatedly. I'm not sure exactly how you plan to use this class, so whatever you prefer probably has had more thought.

meatballs:
I'm having second thoughts about the setter and init stuff. I think we'll want match objects created once and the played repeatedly in a tournament. Setting the result at init makes that more tricky.

meatballs:
So I may go back to having results as a property with a getter whilst still leaving ```_play``` as an internal method.

So, any time ```match.results``` is accessed, it could replay the match depending on the cache and stochasticity.

meatballs:
> you could combine both _play and results into just play()

That might well be the way to go. It certainly makes it more obvious that the match will be replayed if you call it more than once.

meatballs:
it really depends on whether we want to hold the results of a match as an attribute or merely return them from a method.

marcharper:
I think you're right about `__init__` -- we probably don't want to run the match automatically.

I like `play()` better. It's fine to hold the results in an attribute and it since they are needed for sparklines that makes sense. Maybe `play()` can overwrite the stored results and also returns them?

drvinceknight:
Without looking to close in to the detail here @meatballs, could you include a **short** tutorial also? Perhaps in the 'Further Topics' section? (Create a match, see output, create a noisy match see output, cache...).

Something like `Creating your own matches`, just because I think some users could want/need to build their own tournaments etc and this sort of architectural stuff is nice to know about.

meatballs:
> Maybe play() can overwrite the stored results and also returns them?

Was pretty much the conclusion I came to last night whilst out watching the fireworks!

meatballs:
I'll sort out the docs once we're agreed the class and tests are what we want them to be.

drvinceknight:
> I'll sort out the docs once we're agreed the class and tests are what we want them to be.

Cool :+1: I like the conclusion so far between you and @marcharper :)

meatballs:
Here's another idea - match.results could hold all the results for each time the match is played - play() appends to the results rather than overwriting.

drvinceknight:
> Here's another idea - match.results could hold all the results for each time the match is played - play() appends to the results rather than overwriting.

I might be missing an underlying architectural reason but I think I would prefer if `play()` overwrote the results. 

If you and I were to play a match of tennis, each time we played it the outcome would be a result. Of course we might choose to keep track of things but I think that that should be something to the side. (So indeed a user could just keep track of results and/or the tournament does). To come back to the glossary: a match is a single thing, a matchup (doesn't exist) could potentially be something that keeps track of results (I am not suggesting we also create a matchup just mentioning that to highlight difference with how I'm thinking of a match).

marcharper:
Well I can see it both ways. If one wants to run a match several times, how does one preserve the history each time? I presume the results are a list that will be overwritten. So I do something like
```
results = [match.play() for _ in range(10)]
```
am I going to get the proper set of 10 runs or just 10 references to the last run?

If we follow at @meatballs suggestion, then match.results will be a list of lists of results, which is probably what I wanted. So how about this: `play()` returns the next result, but `match.results` holds a list of all the results?

We could also consider making `Match` iterable (calling `play` repeatedly) so that you can do something like this:
```
for results in match():
    for round in results:
        do_something();
```

`Match` would have the following API:
* `play` plays a match and returns the results of that match
* `match.results` holds all the matches, in a list
* `__iter__` goes through `match.results`, and then calls `play` as needed to get more results.

What do you think? Am I making this too complex?

drvinceknight:
> What do you think? 

I see your point and in essence you're wanting the behaviour that I think @meatballs said to start with.

Perhaps it's akin to the tennis court/club just keep a record of the match. So whilst Owen and I would forget what happened, the place where it happened would remember.

I'm down.

> Am I making this too complex?

Haha who knows :) I like the idea of the fancyness of making the match class iterable.

marcharper:
> Haha who knows :) I like the idea of the fancyness of making the match class iterable.

It's even better than that actually, because it's a lazy generator :)

marcharper:
@meatballs, just in case you've not written an iterable before, you just have to define `__iter__` and yield results:
```
def __iter__(self):
    self.play() # appends to self.results
    yield self.results[-1]
```

meatballs:
> Match would have the following API:

> ```play()``` plays a match and returns the results of that match

yes - that's exactly what I'm suggesting

> ```match.results``` holds all the match results, in a list

Again, yes, that's what I'm suggesting

> ```__iter__``` goes through match.results, and then calls play as needed to get more results.

I don't get the point of this. If I ask for 10 results and the match has already been played 5 times, how does it determine whether I want the 5 previous plus 5 new rather or 10 new results? Isn't it simply clearer to call ```play()``` and interrogate ```results``` implicitly?

marcharper:
It's just more pythonic IMO, perhaps not the strongest of reasons. Why save the results just to ignore them later? Maybe `match.reset()` clears results.

meatballs:
> Why save the results just to ignore them later? Maybe match.reset() clears results.

I still don't get this, I'm afraid. I'm not suggesting that the results are ignored, just that they can be accessed and iterated over explicitly as ```match.results``` I don't see what advantage the iterator for match itself provides.

marcharper:
Match does something repeatedly, and contains a list (of lists) that one would want to iterate over. So it makes sense to provide an iterator interface. There are at least two reasonable ways to do this:

Generate indefinitely:
```
for results in match:
    # results generated each loop
    do_something(results)
    # user-defined exit condition
    pass
```
or if one wants exactly 10 rounds:
```
from itertools import islice

match = Match(...)
for results in islice(match, 10):
    do_something(results)
```
or 
```
i = 0
match = Match(...)
for results in match:
    do_something(results)
    i += 1
    if i == 10:
         break
```

In this case I don't have to remember that `Match` stores its results in `.results`, nor what the nature of that storage is. I can just iterate over the match object. In fact in this case there's no reason to really store the results in `Match` at all (it's a generator).

Another option (store the results and iterate only over stored results):
```
match = Match(...)
match.play(10, rounds=100) # play 10 rounds
for results in match: # or for results in match.results:
    do_something(results)
    # StopIteration is raised automatically when the 10 results are exhausted and the loop breaks
    # whether it's a list (match.results) or match is an iterator
```

In the first case results are generated each time `__next__` is called, assuming that results was empty to begin with. In the second case, I had to generate the results first then iterate over them. My suggestion was to do both, in case something was stored in `match.results`: iterate over what was stored and then keep going. My personal preference would be not to store the results at all, because maybe I want to run 1000 matches of 1000 rounds and process them one at a time, so there's no need to keep them all in memory. So I'd have to define a new function that destroys the match object periodically, or an altogether new function duplicating the functionality.

But as I said above, I'm not sure how to intended to use this class in the library, maybe you have a good reason to save all the results, maybe you don't want to iterate over them, idk. I was trying to find a compromise that makes the class externally useful as well.

meatballs:
I don't have a need to store the results in order for them to be made available to the tournament. It can create the match objects, call ```play()``` once per repetition and then process the payoffs etc.

I suppose I was suggesting that it might be helpful to store the results if somebody were creating match objects outside a tournament - perhaps they would want to study the stochastic effects?

How about we ignore my suggestion entirely for the moment? We get #284 done and dusted using the class as it stands and we come back to this if and when we need it.

marcharper:
^ sounds good.

meatballs:
OK. I'll start looking at the docs for this one

meatballs:
I've included a doctest skip for the sparkline example as python2.7 really doesn't like it and I don't know how to include the Unicode within the .rst file in such a way as to to keep doctest happy on both 2.x and 3.x whilst also rendering properly via Sphinx.

drvinceknight:
> I've included a doctest skip for the sparkline example. Python 2.7 really doesn't like it and I don't know how to include the Unicode within the .rst file in such a way as to to keep doctest happy on both 2.x and 3.x whilst also rendering properly via Sphinx.

Once this is merged in can we open an issue for it? Not suggesting it's fixable but 'I know a guy' who might be able to take a look :)

drvinceknight:
The docs get a :+1: from me (still planning on looking at the rest at some point).

axelrod/match.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
+
+
+def sparkline(actions):
```
drvinceknight: Could we make sparklines take any character? Using `█` by default?

This would also allow us to keep the current doctest which is skipped and include another one where we use `X` (or whatever).

axelrod/match.py
```diff
@@ -0,0 +1,101 @@
...
+            if self._cache_update_required:
+                self._cache[self._classes] = result
+        else:
+            result = self._cache[self._classes]
+
+        self.result = result
+        return result
+
+    @property
+    def sparklines(self):
```
drvinceknight: So this would also take a default argument for `█`.

axelrod/tests/unit/test_match.py
```diff
@@ -0,0 +1,72 @@
...
+        self.assertEqual(
+            cache[(axelrod.Cooperator, axelrod.Defector)], expected_result)
+
+        # a deliberately incorrect result so we can tell it came from the cache
+        expected_result = [(C, C), (D, D), (D, C)]
+        cache = {(axelrod.Cooperator, axelrod.Defector): expected_result}
+        match = axelrod.Match(players, 3, cache)
+        self.assertEqual(match.play(), expected_result)
+
+    def test_sparklines(self):
```
drvinceknight: and we could have a test where we use a non default character?

meatballs:
I've no problem paramaterising the character, but I don't want to change the content of the doc just to keep doctests happy.

In the docs, I'd like the sparklines example to be as good as possible. The whole point of it is to make pattern recognition easy and that's best done with the solid block.

meatballs:
This seems a little OTT just to keep doctests happy - having to write another example in the docs adds very little new information.

If there's a good reason to use some other character, then fine, but I can't see one. The whole point of the sparklines is to make pattern recognition as easy as possible and that's best done with the solid block.

drvinceknight:
> This seems a little OTT just to keep doctests happy - having to write another example in the docs adds very little new information.

I'm not suggesting this is done to keep doctests happy. 

I think this should be parametrised no matter what (if you remember this was discussed at the sprints: someone suggested that a user could even use emoji if they wanted to :)). It's trivial to allow for it and nicer than to have the block hardcoded so to speak.

> The whole point of the sparklines is to make pattern recognition as easy as possible and that's best done with the solid block.

I'm not sure I agree that the solid block is universally the "best" but ultimately avoiding that discussion is the point of this suggestion.

Once we have it parametrised it's then easy enough to throw in an **extra** (not suggesting we replace) line to the docs:

```
Note that you can also create sparklines using a symbol of your choice::
>>> print(match.sparklines, symbol='X')
XXXXXXXXXXXXXXXXXXXXXXXXX
```

I'd certainly like this parametrised (I don't see why we wouldn't?), then to add a line or two to the docs showing it's possible to do that is a logical thing to do.

The side effect is that we then get a doctest but that's not at all the motivation.

meatballs:
OK. I'll sort it.

drvinceknight:
If you're busy, I've got 2 hours on a train and have pulled this locally already so I'm happy to throw it together?

meatballs:
already started :-)

drvinceknight:
:+1:

drvinceknight:
I had started as well so in case it's helpful (feel free to ignore) https://github.com/Axelrod-Python/Axelrod/pull/446

As I say over there, I made `sparklines` no longer be a property method... (There might be a better way)

meatballs:
Looks like we were writing pretty much the same code at the same time!

meatballs:
'symbol' is perhaps a better choice of name than 'character'

I'll change that and push another commit

drvinceknight:
Tests pass locally for me so not waiting on travis for my thumbs up: :+1: 

Thanks @meatballs for adding in the options :)

Issue #415, opened 2015-11-05
GoByMajority discuss documentation/implementation


drvinceknight:
So according to our documentation `Hard_Majo` is defined in the library as `GoByMajority`  - see screenshot - (it is also, in the documentation noted as not being present in the library: http://axelrod.readthedocs.org/en/latest/reference/overview_of_strategies.html#hard-majo) so this obviously needs to be fixed.

https://www.dropbox.com/s/f8vqlj66mtvms9x/Screenshot%202015-11-05%2013.46.39.png?dl=0

The problem is that by default `GoByMarjority` is implemented as `soft` and needs to be passed the keyword `GoByMajority(soft=False)` to create the `hard` Go by Majority. This is not good for newcomers to the library trying to reproduce Stewart and Plotkin's work (this was noted by @ABarriscale). Also, it means that the main tournament is not running the Hard variants of Go by Majority.

1. We should fix the documentation ⚠️
2. Do we want to have `GoByMajority` (or `SoftGoByMajority`?) **and** `HardGoByMajority` to the library as strategies of their own? (To be clear this is practically no work as they have already been implemented so to speak it's just a question of adding them to the name space).

marcharper:
I was about to mess with this to double check the suggestions for the paper.

Despite having only one line different the two variants do seem to behave differently. So whatever you feel is most clear is fine with me.

Another issue with GoByMajority -- the default case says memory zero, but it's really memory infinity, since `history = opponent.history[-memory:]` is the entire list when memory is zero. We should fix the classifier.

drvinceknight:
> Despite having only one line different the two variants do seem to behave differently. So whatever you feel is most clear is fine with me.

It was noted by Anna because the Soft did actually behave quite differently :)

I'll put something together now: probably going to go with Hard and Soft versions of everything implemented currently as well as doc fixes.

drvinceknight:
#416 addresses this.

PR #416, opened 2015-11-06
415 - Making HardGoByMajority explicit

drvinceknight:
This addresses #415:

- Adds `HardGoByMajority`
- Including docs
- Also fixes memory depth for `GoByMajority` base player (was 0, should be inf).


marcharper:
:+1:

PR #417, opened 2015-11-07
Add decorator for init_args

marcharper:
In attempt to make adding new strategies easier, I added a decorator so you can do this:

```
    @init_args
    def __init__(self, p=0.5):
        ...
```
and not have to put `self.init_args = (p,)` in the `__init__` method (or understand why).

I also did some various PEP8 while I was looking over all the strategies.


drvinceknight:
:+1:

PR #418, opened 2015-11-07
Refactor HeadsUpTest to be more intuitive, more flexible, more like test_responses

marcharper:
Also uncommented and modified a strategy transformer test.


marcharper:
I'm not sure how adding tests decreased coverage...

drvinceknight:
:+1:

PR #419, opened 2015-11-07
Remove update script

marcharper:
Just removes `update_results.sh`. which was moved to the tournament repository some time ago.


drvinceknight:
:+1:

PR #422, opened 2015-11-08
Plotting: New colormaps and image format passthrough

### LINKED ISSUE #420: Latest version of matplotlib released with new colormaps
http://matplotlib.org/devdocs/users/whats_new.html#new-colormaps

Are we going to change the defaults? This immediately implies that anyone using the library would need to update their matplotlib install.

Also, does this have an effect on violin plots when there is no variability? Can we use violin plots throught? I think you mentioned that at some point @marcharper ?

marcharper:
Closes #420.

This is a hard one to write tests for, so here are some images that I produced. For matplotlib < 1.5:

For 1.5+

You can also now pass `-i png` (or other format) to run_axelrod down through to TournamentManager.

### DIFF
--- a/axelrod/plot.py
+++ b/axelrod/plot.py
@@ -5,11 +5,21 @@ from warnings import warn
 
 matplotlib_installed = True
 try:
+    import matplotlib
     import matplotlib.pyplot as plt
     import matplotlib.transforms as transforms
     from mpl_toolkits.axes_grid1 import make_axes_locatable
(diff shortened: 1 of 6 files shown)


drvinceknight:
This looks good to me, why the `i` as the option for the format? (Just curious in case that's a thing)

marcharper:
'i' for image format

drvinceknight:
:+1:

PR #424, opened 2015-11-10
Improve efficiency of random hunter

marcharper:
Random hunter was doing a lot of history traversing that is slow. This also improves MetaHunter's runtime a bit.

I may have slightly altered the behavior of random hunter (it may start one round earlier?) but the tests still pass, so I think it's fine.


drvinceknight:
:+1:

PR #427, opened 2015-11-13
Update pattern string

mojones:
I re-ran the evolutionary algorithm to take account of the recently-added strategies.

### DIFF
--- a/axelrod/strategies/lookerup.py
+++ b/axelrod/strategies/lookerup.py
@@ -148,7 +148,7 @@ class EvolvedLookerUp(LookerUp):
                                          other_histories))
 
         # Pattern of values determed previously with an evolutionary algorithm.
-        pattern='CDCCDCCCDCDDDDDCCDCCCDDDCDDDDDDCDDDDCDDDDCCDDCDDCDDDCCCDCDCDDDDD'
+        pattern='CDCCDCCCDCDDDDDCCDCCDDDDDCDCDDDCDDDDCCCDDCCDDDDDCDCDDDCDCDDDDDDD'
         # Zip together the keys and the action pattern to get the lookup table.
         lookup_table = dict(zip(lookup_table_keys, pattern))
         LookerUp.__init__(self, lookup_table=lookup_table)
--- a/axelrod/tests/unit/test_lookerup.py
+++ b/axelrod/tests/unit/test_lookerup.py
@@ -111,8 +111,8 @@ class TestEvolvedLookerUp(TestPlayer):
 
     def test_init(self):
         # Check for a few known keys
-        known_pairs = {('DD', 'CC', 'CD'): 'D', ('DC', 'CD', 'CD'): 'D',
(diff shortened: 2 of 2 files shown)


drvinceknight:
:warning: do not merge: @mojones has asked on the gitter channel to check if this gives an improved performance. Once someone has the time/computer time to check this write it down here.

marcharper:
Do we want to run both at the same time?

drvinceknight:
> Do we want to run both at the same time?

I've set it running at home. Not both but with this as a replacement, does both make sense thought? We wouldn't have both in the library... (would we?)

marcharper:
There are several strategies that have many variants (5 of each of Hard and Soft GoByMajority). If these two behave significantly differently I don't see any reason not to have both. It would be nice to have some qualitatively different tables (e.g other than n=2=m). Even n=m=1 could be nice even if it's not the best, or n=2, m=1 (and of course one can hope for n, m > 2).

drvinceknight:
> If these two behave significantly differently I don't see any reason not to have both. It would be nice to have some qualitatively different tables (e.g other than n=2=m). Even n=m=1 could be nice even if it's not the best, or n=2, m=1 (and of course one can hope for n, m > 2).

Yeah I think I agree although I would suggest that this is an incremental change? And having each an every incremental change would not really be worth it? Having ones with different size tables would be great.

Just heading home so hopefully will be showing you the results on gitter shortly :) (will also throw them here)

marcharper:
No, I certainly wouldn't recommend adding every incremental change. Let's see how different the results are before getting too theoretical in this discussion.

drvinceknight:
Here's the updated version (pdf format):

[tuned.pdf](https://github.com/Axelrod-Python/Axelrod/files/34232/tuned.pdf)

Current master branch is here:

drvinceknight:
In terms of this strategy being an update of the originally evolved version I would suggest merging this in as is (ie that this is an incremental yet very effective improvement).

drvinceknight:
https://github.com/Axelrod-Python/tournament/pull/18 has all the new results ready in case we're happy.

mojones:
I propose merging as is to replace the current version. I think the stable of strategies is more "interesting" the more diverse it is i.e. having strategies that are fundamentally different. Maybe I could evolve an m=n=1 table using identical methods and add it as a separate strategy?

drvinceknight:
Then possibly some with m,n>2 ?(Which could be fine tuned even more once
there's enough computer power etc...)

marcharper:
Sounds good to me. Versions with n=m=1 would be great just as a point of comparison, also n=1, m=2 and n=2, m=1, if it's not too much trouble! I understand n, m > 2 is computationally more difficult.

drvinceknight:
Those sound like good suggestions to me. I assume this is good to merge
Marc?

marcharper:
I tagged it so

drvinceknight:
Sorry was just on my email :) I'll also merge the tournament update now.

Issue #403, opened 2015-10-29
Add definitions of play contexts


marcharper:
The documentation lacks names for the four possible rounds (C, C) (C, D), (D, C), (D, D) and the meanings of R, P, S, T in the game matrix.

drvinceknight:
I think this should go in an 'Description of Iterated Prisoner's Dilemma' file in the reference section of the docs.

That could then be nicely followed by the suggestion on #404

drvinceknight:
Have opened #428, not convinced it's as good as could be (I lazily copied the old background). Let me know what you think @marcharper :)

PR #428, opened 2015-11-13
403 Adding play contexts and RPST to docs

drvinceknight:
Addressing #403


marcharper:
It's fine I think -- nothing fancy required :+1:

PR #433, opened 2015-11-22
348 implementing `makes_use_of`

drvinceknight:
This is for #348 

Opening this to discuss and to have some help.

First of all I went with the `makes_use_of` value of the classifier dictionary to be a `set`, this seemed to make sense as we don't care about order.

I have 3 main remaining things (without the automatic aspects that @marcharper suggested on #348 which could or could not end up being in this PR):

- [x] The memory one strategies, I've managed to get these working but mainly had to do this by overwriting the inherited class `classifier`, I seemed unable to sort this without changing that (not sure if this is worth investigating).
- [x] Dealing with the strategy transforms (I need to make sure that tests update to take in to account when a transform changes what the strategy makes use of). Putting this here as a placeholder.
- [x] The meta strategies: **here be gremlins** :imp:. I am struggling to get the classifier dictionary to update properly with the team's `make_use_of`. You can see my attempt here: ab80a180d3ee7feebb2c1128a6abf489b2595fbd but tests still fail.

Halp :)

Edit:

I must make sure I write docs for all this:

- [x] Write docs.


marcharper:
So the reason that the classifier tests for the meta players is failing is because the `test_meta_classifiers` method of `TestMetaPlayer` doesn't overwrite the method `classifier_test` of `TestPlayer` and `classifier_test` doesn't build the correct expected classifiers dictionary. At the time that `classifier_test` is called by the `test_initialization` method of TestPlayer, the child classes of TestMetaPlayer don't have anything yet set for classifier["makes_use_of"].

This is fixed by renaming `test_meta_classifiers` to `classifier_test` to overwrite the parent method, which I pushed a commit for.

The strategy transformer issue is a little trickier. Everything is correct in the library but the two remaining tests are failing primarily because the classifier dictionary for many strategies are class variables rather than instance variables. This means that the transformer overwrites the base class's classifier dictionary because of how python handles class variables with inheritance.

So the issue with with `FinalTransformer` combined with a Player that has class-wide classifiers. The tests would pass if run in the proper order, but unittest loads all the tests first rather than running them sequentially. At the moment there are no issues in the library itself (but there could be if the transformers are applied more liberally). Specifically, when one invokes:
```
FinalTransformer([D, D, D])(axelrod.Cooperator)()
```
the transformer sets Cooperator's classifier dictionary to indicate use of tournament length because
Cooperator's classifier dictionary is a class variable and is inherited into the transformed class. For now I push a quick fix, having `Player.__init__` do a deepcopy of the classifier dictionary (which creates new references at the instance level).

We might consider (should this continue to be an issue going forward) making the classifier dictionary an instance-level feature throughout, perhaps with a method that sets the classifiers to prevent the need to overwrite `__init__` in every strategy, something like:

```
class MyStrategy(Player):
    ...
    def report_classifiers(self):
        ... # build classifier dictionary based on internal variables
        return classifiers
```

marcharper:
I think @drvinceknight intends to write some docs, otherwise this is :+1:  from me.

drvinceknight:
This should have triggered a travis failure. The doctests failed for me locally... Inspecting...

drvinceknight:
This is now documented. Note that I also changed how travis runs the doctests, this is following #439 (which if you look at the commits I merged and then reverted).

drvinceknight:
@marcharper I have labelled this with a positive review but if you want to see the changes to the tests:

- 87abb46
- 951c5a7

and if you want to see the docs:

- 7d4c5b5

Issue #434, opened 2015-11-23
Wins are being counted twice


drvinceknight:
The wins are not quite counted correctly:

```
>>> tnmt = axelrod.Tournament([axl.Cooperator(),axl.Defector(), axl.Alternator()])
>>> results = tnmt.play()
>>> results.wins
[[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
 [4, 4, 4, 4, 4, 4, 4, 4, 4, 4],
 [2, 2, 2, 2, 2, 2, 2, 2, 2, 2]]
```

Defector gets 4 wins but only has 2 opponents.

This happens because here https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/payoff.py#L371 a player gets a win when it beats an opponent but also when it is the opponent that gets beaten.

The simplest fix would be to change:

```
                if winner is not None:
                    wins[winner][repetition] += 1
    return wins
```

to

```
                if winner is not None:
                    wins[winner][repetition] += 1
    wins = {key:wins[key]/2 for key in wins}  # Simply divide the number by 2 (the number will always be even).
    return wins
```

I don't know if this has downstream implications that would need to be considered.

PR #436, opened 2015-11-23
Fixing number of wins.

drvinceknight:
Addresses #434


drvinceknight:
#437 is a better solution.

PR #439, opened 2015-11-24
Ensure `doctest` runner doesn't always exit 0

doismellburning:
Note that this does mean they're run serially; there's probably a better way to do this, but this should be an improvement over "always succeed"!

### DIFF
--- a/doctest
+++ b/doctest
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+set -e
+
 python -m doctest docs/tutorials/getting_started/*rst
 python -m doctest docs/tutorials/further_topics/*rst
 python -m doctest docs/reference/*rst


doismellburning:
@drvinceknight This should solve your doctest woe of #433

drvinceknight:
Super helpful, we've got a 2 core review merge rule (even though this is technically a bug and trivial). Have merged it in to #433 but in case another core sees this: :+1: :)

drvinceknight:
This isn't working quite right.

Running `./doctest`:

```
Axelrod(348) : ./doctest
**********************************************************************
File "docs/tutorials/further_topics/classification_of_strategies.rst", line 14, in classification_of_strategies.rst
Failed example:
    axl.Cooperator.classifier == expected_dictionary
Expected:
    True
Got:
    False
**********************************************************************
File "docs/tutorials/further_topics/classification_of_strategies.rst", line 20, in classification_of_strategies.rst
Failed example:
    s.classifier == expected_dictionary
Expected:
    True
Got:
    False
**********************************************************************
1 items had failures:
   2 of  11 in classification_of_strategies.rst
***Test Failed*** 2 failures.
```

but this is in fact missing a test:

```
Axelrod(348): python -m doctest docs/tutorials/contributing/strategy/*rst  # from doctest
**********************************************************************
File "docs/tutorials/contributing/strategy/classifying_the_new_strategy.rst", line 11, in classifying_the_new_strategy.rst
Failed example:
    for key in sorted(classifier.keys()):
       print("{}: {}".format(key, classifier[key]))
Expected:
    inspects_source: False
    manipulates_source: False
    manipulates_state: False
    memory_depth: 1
    stochastic: False
Got:
    inspects_source: False
    makes_use_of: set([])
    manipulates_source: False
    manipulates_state: False
    memory_depth: 1
    stochastic: False
**********************************************************************
1 items had failures:
   1 of   6 in classifying_the_new_strategy.rst
***Test Failed*** 1 failures.
```

drvinceknight:
So as far as I can tell that fix is not actually a solution to the problem as it makes some failures silent. On #433 I've reverted it (so that the `-e` is not in doctest) and then included the actual commands right in the `.travis.yml`... Which is causing all the failures we expect on travis and ensures that the `doctest` script gives the expected outputs when run locally.

doismellburning:
Yes, sorry, that's what I meant by "serially"; it runs through each command, and if one fails, the script exits with a failure code. Ideally there'd be a better test runner that would run everything and THEN aggregate failures, but I was going for minimal solution as I can't immediately see a nice way to do that.

drvinceknight:
> Yes, sorry, that's what I meant by "serially"; it runs through each command, and if one fails, the script exits with a failure code. Ideally there'd be a better test runner that would run everything and THEN aggregate failures, but I was going for minimal solution as I can't immediately see a nice way to do that.

Ah: I see :+1: I pretended I knew what 'serially' meant and just carried on :)

So I've got it setup as: https://github.com/Axelrod-Python/Axelrod/commit/87abb464152223e565ed80b5fc10aa772d402283

drvinceknight:
I'm going to close this @doismellburning, I merged it then reverted on #433. Thanks :+1: :muscle:

PR #441, opened 2015-11-24
Use Travis container infrastructure for faster builds

doismellburning:


doismellburning:
(This may not work, given the `xvfb` stuff, but figured it was worth a try)

doismellburning:
My bad, sorry, didn't realise the defaults had now changed

Issue #412, opened 2015-11-04
Create DeterministicCache class


meatballs:
Creating a class for the cache would give us some advantages:

1. Encapuslation - the cache has state (e.g. mutability or the number of turns for which it is valid) and it has methods (e.g. store, load, add). A class would allow us to encapsulate that behaviour instead of it existing several times around other classes.

1. Implementation - having a class with 'add' and 'retrieve' methods would allow us to change the underlying implementation (e.g. a database, a redis store) without affecting any of the classes which use it.

This should wait until #284 is complete

marcharper:
:+1:

drvinceknight:
:+1:

drvinceknight:
Closed by #527.

PR #443, opened 2015-11-24
Pin dependency versions

doismellburning:
Because explicit version pins mean consistently reproducible builds, and consistently reproducible builds are A Good Thing, and "hm suddenly the build is breaking, oh it's because a new backwards-incompatible version of numpy was released" is A Bad Thing

I'd suggest also pointing requires.io at this, and then you'll get notified of dependency updates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,7 @@ before_install:
   - sh -e /etc/init.d/xvfb start
 install:
   - pip install -r requirements.txt
-  - pip install sphinx
-  - pip install coverage
-  - pip install coveralls
+  - pip install -r requirements-test.txt
 script:
   # Build the documentation
   - cd docs; make html
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,0 +1,3 @@
+Sphinx==1.3.1
+coverage==4.0.3
+coveralls==1.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-numpy>=1.9.2
-matplotlib>=1.4.2
+numpy==1.10.1
+matplotlib==1.5.0
 cloudpickle==0.1.1


marcharper:
While I agree in principle, this hasn't been a problem in practice, and we've tried to make the library available to people on systems that they do not fully control. Moreover, Ubuntu frequently lags on new versions of python packages for some time, so we would be requiring people to build the newest versions from scratch (with pip this is ok but it's still added complexity to use the library). Since the library doesn't need the latest versions of these libraries it may be better to leave the versions with >= or to pin older versions.

doismellburning:
Fair. You may at least want an upper bound on some of those dependencies though...

marcharper:
For matplotlib and numpy I think we need to be flexible, but for sphinx and coveralls pinning specific versions are probably fine. We can always move them up later if we need a specific feature or bugfix, and they are necessary just to build the docs and test, not to actually run the library.

Our reliance on numpy is fairly simple, so unless they really screw something up with basic matrix operations we should be fine. We have some code in `plot.py` that deals with whether matplotlib is greater than 1.5.0 or not (there is an issue with violin plots that was not fixed until 1.5.0). There is also a similar check to set the default colormap, since that changed in matplotlib 1.5.

The cloudpickle dependency is contingent on a rework of the caching features #412 (the standard library pickle modules can't handle strategies that have had a transformation applied). We may be able to remove it completely.

I'd support pinning numpy and matplotlib to at least the Ubuntu LTS package versions; I'm not sure what the maximums would be. Anyone have suggestions?

drvinceknight:
I can't say that I have much to add of value here and don't know enough about the packages to be able to indicate an idea about the upper bound...

Just to add my :+1: to this:

> we've tried to make the library available to people on systems that they do not fully control

drvinceknight:
Any thoughts more thoughts on this @marcharper ?

marcharper:
Nothing definitive. I agree that pinning dependencies in principle is better for reproducibility, but in practice I'm not sure that it matters for this project. Any publication could note the exact versions (and requirements.txt could be modified to install the exact list for reproducibility). OTOH, having everyone on the exact same versions is good for reproductions... MPL is an issue since it's just moved to 1.5 and we want some of the changes (new heatmaps, violin plots).

drvinceknight:
Should we close the PR as is and open an issue to discuss?

marcharper:
Sure, we can always re-open later

PR #444, opened 2015-11-24
Improved doctest discovery / runner

doismellburning:
@drvinceknight Think #439 but better - NOW it aggregates failures

This isn't _much_ more than a cosmetic fix, but it does mean there's One True Command to run doctests, AND it just finds all the `rst` files, rather than a dir-by-dir enumeration


marcharper:
This is nice and future-proof as we expand the docs. What do you think @drvinceknight ?

drvinceknight:
:+1:

PR #449, opened 2015-12-07
447 - Adding a Mixed decorator **and** metaplayer

drvinceknight:
This adds the ability to make random mutations on a player (transform) **or** create a metaplayer that randomly switches between a team of players.

### DIFF
--- a/axelrod/strategies/meta.py
+++ b/axelrod/strategies/meta.py
@@ -269,10 +269,12 @@ class MetaWinnerLongMemory(MetaWinner):
         self.init_args = ()
 
 
-class MetaMutater(MetaPlayer):
-    """A player who randomly switch between a team of players.
+class MetaMixer(MetaPlayer):
+    """A player who randomly switches between a team of players.
     If no distribution is passed then the player will uniformly choose between
     sub players.
+
+    In essence this is creating a Mixed strategy.
 
     Parameters
     ----------
@@ -284,7 +286,7 @@ class MetaMutater(MetaPlayer):
         If none is passed will select uniformly.
     """
 
-    name = "Meta Mutater"
+    name = "Meta Mixer"
 
     def __init__(self, team=None, distribution=None):
 
@@ -298,7 +300,7 @@ class MetaMutater(MetaPlayer):
 
         self.distribution = distribution
 
-        super(MetaMutater, self).__init__()
+        super(MetaMixer, self).__init__()
         self.init_args = (team, distribution)
 
     def meta_strategy(self, results, opponent):
--- a/axelrod/strategies/meta.py
+++ b/axelrod/strategies/meta.py
@@ -2,6 +2,7 @@ from axelrod import Actions, Player, obey_axelrod
 from ._strategies import strategies
 from .hunter import DefectorHunter, AlternatorHunter, RandomHunter, MathConstantHunter, CycleHunter, EventualCycleHunter
 from .cooperator import Cooperator
+from numpy.random import choice
 
 # Needs to be computed manually to prevent circular dependency
 ordinary_strategies = [s for s in strategies if obey_axelrod(s)]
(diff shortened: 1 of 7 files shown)


axelrod/strategies/meta.py
```diff
@@ -266,3 +267,40 @@ def __init__(self):
                 == float('inf')]
         super(MetaWinnerLongMemory, self).__init__(team=team)
         self.init_args = ()
+
+
+class MetaMutater(MetaPlayer):
+    """A player who randomly switch between a team of players.
```
marcharper: switches

axelrod/strategy_transformers.py
```diff
@@ -239,6 +240,32 @@ def apology_wrapper(player, opponent, action, myseq, opseq):
 ApologyTransformer = StrategyTransformerFactory(apology_wrapper,
                                                 name_prefix="Apologizing")
 
+
+def mutate_wrapper(player, opponent, action, probability, m_player):
+    """Randomly picks a strategy to play, either from a distribution on a list
+    of players or a single player."""
+
+    # If a single probability, player is passed
+    if type(probability) in [float, int]:
```
marcharper: perhaps `isinstance` is more reliable here? (not sure myself)

axelrod/strategy_transformers.py
```diff
@@ -239,6 +240,32 @@ def apology_wrapper(player, opponent, action, myseq, opseq):
...
+        m_player = [m_player]
+        probability = [probability]
+
+    # If a probability distribution, players is passed
+    if type(probability) == type(m_player) is list:
```
marcharper: Same comment here

axelrod/strategy_transformers.py
```diff
@@ -239,6 +240,32 @@ def apology_wrapper(player, opponent, action, myseq, opseq):
...
+            normalised_prob = [prob / float(mutate_prob)
+                               for prob in probability]
+            if random.random() < mutate_prob:
+                p = choice(m_player, p=normalised_prob)()
+                p.history = player.history
+                return p.strategy(opponent)
+
+    return action
+
+MutateTransformer = StrategyTransformerFactory(mutate_wrapper, name_prefix="Mutated")
```
marcharper: So is the underlying player overwritten here in favor of the m_player list? I.e. it doesn't seem like it matters which player is transformed.

marcharper:
Just a general comment -- I'm not sure "mutator" is the best name, at least in the biological sense; game-theoretically I think what we technically have here are mixed strategies (or meta-mixed strategies). Whereas a mutation usually means that the strategy completely changes into another one.

Perhaps the difference is just semantic (mutation is more general than its usage in biology after all), I'm not saying it has to be changed, just a thought.

drvinceknight:
> Looks good overall. Just a general comment -- I'm not sure "mutator" is the best name, at least in the biological sense; game-theoretically I think what we technically have here are mixed strategies (or meta-mixed strategies). Whereas a mutation usually means that the strategy completely changes into another one.

Completely agree. I wasn't satisfied with the name. `MixedTransformer` and `MetaMixed` is what I'll go with but by all means suggest something if you can think of something that rolls off the tongue a bit nicer (should definitely be `mixed`-based though).

drvinceknight:
> So is the underlying player overwritten here in favor of the m_player list? I.e. it doesn't seem like it matters which player is transformed.

The underlying player never really gets touched... It's just that one of the other players gets tagged in randomly (after having it's history set to the underlying player's history). I'm very open to chat/input on this though: do you think this is the right way to go?

As is, it doesn't matter which player is transformed, I thought that from a UI point of view it might be a nice decorator to change all strategies in a given tournament.

Very open to suggestion: kind of feeling around in the dark with this one.

drvinceknight:
4a9d524 changes use of `type` to use of `isinstance` (coverage decreased): I have no idea if this is indeed better or not, kept the commit separate so that it can be easily reverted...

marcharper:
It's fine to ignore the underlying player, I just wanted to make sure I understood the code. Some of the other transformers can do this, e.g. `InitialTransformer` with a very long list effectively ignores the underlying player.

`isinstance` is more flexible -- it allows inherited types. But I'm not sure if that's relevant here except for other iterators than `list`. So either way is probably fine, but for the list case it may be better to do this:
```
import collections
isinstance('ciao', collections.Iterable)
```
in case someone wants to use a tuple or set instead of a list.

drvinceknight:
ab71ca0 addresses that @marcharper :)

PR #454, opened 2015-12-14
198 - Implements some hypothesis testing

drvinceknight:
This puts in place property based testing using https://github.com/DRMacIver/hypothesis

Have implemented this for:

- The game tests (trivial)
- The tournament tests (less trivial; hypothesis is used to here create 50 tournaments)

This would be a nice thing to put throughout the library if possible...

When this gets in, we should update: https://hypothesis.readthedocs.org/en/latest/usage.html


drvinceknight:
Investigating what this error means:

```
Timeout: Ran out of time before finding a satisfying example for test_property_serial_play. Only found 4 examples (4 satisfying assumptions) in 67.74s.
```

drvinceknight:
Once https://github.com/DRMacIver/hypothesis/commit/f0919fdae5e912e6b388631bf3f89cbcea6284c3 is in the latest release of Hypothesis (shouldn't be long) will need to modify df726bd

marcharper:
Looks fine to me. We could also consider (at some future time) using the newer [type hints](https://docs.python.org/3/library/typing.html) in Python 3.5, but I don't know if it will be an issue for earlier versions, which might be a deal breaker.

drvinceknight:
I would be up for this if only to be able to understand type hints :)

marcharper:
It's just a syntax for specifying types of arguments and return values that currently does not affect runtime at all AFAIK, but allows static checkers to verify various things during testing. This puts the type information somewhere that can be parsed and used rather than in the docstring (which could be in any arbitrary format or not at all present).

There are some nice generics like `Callable` and `Sequence` if you want to indicate that the argument is e.g. an iterable but not specify precisely that it is a list. To me this produces nicer code than using `if`/`try` statements with `isinstance(obj, list)` or `hasattr("__next__")` while still maintaining the benefits of dynamic typing.

It's sort of on the other side of the fence as using the `auto` keyword in C++ to achieve a degree of dynamic (but still deterministic at compile time) typing rather than rigidly specifying the type of everything. There was talk of adding [concepts](https://en.wikipedia.org/wiki/Concepts_%28C%2B%2B%29) to C++ that would allow one to specify an interface rather than a type (e.g. this argument is an unspecified type that implements a `__next__` method), but that didn't make it officially in. Hopefully it makes it in to C++17.

drvinceknight:
I listened to a podcast episode about it and mainly thought that it sounded pretty 'un pythonic' but certainly sounds worth trying out. I'm curious how it would (or wouldn't) be backwards compatible with earlier pythons though...

drvinceknight:
I'll open an issue for that :)

PR #455, opened 2015-12-29
Use RTD sphinx theme

meatballs:
Code from http://read-the-docs.readthedocs.org/en/latest/theme.html to use their default sphinx theme


drvinceknight:
:+1:

PR #458, opened 2016-01-15
New Strategy: FirmButFair

marcharper:
There's a simple memory-one strategy in the literature called FirmButFair that the library doesn't have yet.


drvinceknight:
:+1:

PR #460, opened 2016-02-15
Adding ThueMorse and ThueMorseInverse strategies

geraintpalmer:


drvinceknight:
Can you give some context about this one: why the T-M sequence?

Putting link from your documentation here in case others want to find it quicker: http://mathworld.wolfram.com/Thue-MorseSequence.html

axelrod/strategies/thuemorse.py
```diff
@@ -0,0 +1,60 @@
...
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def thuemorse_sequence(self, x):
+	    if x == 0:
+	        return 0
+	    if x%2==0:
```
drvinceknight: PEP8

axelrod/strategies/thuemorse.py
```diff
@@ -0,0 +1,60 @@
...
+	        return self.thuemorse_sequence(x/2)
+	    if x%2==1:
+	        f = x-1
```
drvinceknight: If x is indeed the round, replace f with `previous_round` or similar (again: make the variables verbose: it helps readability).

axelrod/strategies/thuemorse.py
```diff
@@ -0,0 +1,60 @@
...
+    classifier = {
+        'memory_depth': float('Inf'),
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def thuemorse_sequence(self, x):
```
drvinceknight: What is x? Is it the round? Could you make the variable name more verbose?

drvinceknight: Could you add a docstring here?

geraintpalmer:
Heard about it the other day. There are no cubes (certain strings of 3 never repeat n certain places) etc. Thought it'd be difficult to learn against.

axelrod/strategies/thuemorse.py
```diff
@@ -0,0 +1,60 @@
...
+	    if x == 0:
+	        return 0
+	    if x%2==0:
```
drvinceknight: PEP8

axelrod/tests/unit/test_thuemorse.py
```diff
@@ -0,0 +1,60 @@
...
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_strategy(self):
+        """Test that strategy is randomly picked (not affected by history)."""
+        self.first_play_test(C)
```
drvinceknight: Why are you repeating these?

docs/reference/overview_of_strategies.rst
```diff
@@ -1123,3 +1123,13 @@ LookerUp
 
 LookerUp uses a lookup table to decide what to play based on the last few rounds plus
 the first few plays made by the opponent at the start of the match.
+
+ThueMorse
+^^^^^^^^^
+
+This strategy uses the [Thue-Morse sequence](http://mathworld.wolfram.com/Thue-MorseSequence.html) in order to determine its actions. It will Cooperate on a 1 and Defect on a 0.
```
drvinceknight: So is this in fact independent of the other players' history?

axelrod/strategies/thuemorse.py
drvinceknight: I would suggest you make this just `n` but include a quick line of docstring.

axelrod/strategies/thuemorse.py
drvinceknight: I would get rid of the `f` and just have `(n - 1)/2`.

drvinceknight:
> Heard about it the other day. There are no cubes (certain strings of 3 never repeat n certain places) etc. Thought it'd be difficult to learn against.

Sounds cool! @geraintpalmer 

Made some stylistic suggestions and also wouldn't mind the tests being clarified a bit if that's not too much trouble.

axelrod/tests/unit/test_thuemorse.py
```diff
@@ -0,0 +1,60 @@
...
+        'memory_depth': float('Inf'),
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_strategy(self):
+        """Test that strategy is randomly picked (not affected by history)."""
```
drvinceknight: Is this random?

geraintpalmer: Sorry, copied test from another strategy as template. Will fix.

axelrod/tests/unit/test_thuemorse.py
```diff
@@ -0,0 +1,60 @@
...
+        self.first_play_test(C)
+        self.first_play_test(C)
+
+    def test_effect_of_strategy(self):
+        self.markov_test(['D', 'D', 'D', 'D'])
+        self.responses_test([], [], [C, D, D, C, D, C, C, D, D, C, C, D, C, D, D, C])
+        self.responses_test([C], [C], [D, D, C, D, C, C, D, D, C, C, D, C, D, D, C])
```
drvinceknight: Could you perhaps as a docstring include the first couple of terms of the sequence just so that it's evident what the strategy is doing.

axelrod/tests/unit/test_thuemorse.py
marcharper: It's still infinite memory I think even though it doesn't use the other players history. Unless the sequence repeats eventually or has a finite order recurrence relation (doesn't seem so) then the generation of the sequence depends on the round_number. So it depends on the history of play (just the length, not the content).

drvinceknight: Cool: thanks for clarifying. :)

marcharper:
My only suggestion, and this is probably overkill, would be to define a `SequencePlayer` that takes a generator function and a list of default arguments as `__init__` parameters. Separate the Thue-Morse sequence generation into a generic python generator and define players by inheriting from the abstract base class. Then if anyone else wants to then add a player that iterates through another sequence they need only define the sequence function and inherit from `SequencePlayer`.

This is similar to how the `Cycler` player works -- it generically iterates cyclically over a given input sequence.

drvinceknight:
I really like that idea!

@geraintpalmer if you fancy tackling that that's awesome but if you would
prefer not to, we could open an issue for that generic player (which would
eventually be how these two players would be generated).

geraintpalmer:
Yeah I'm happy to look into that

marcharper:
I'm happy to help if you go that route. You'll need to make sure to define `player.reset()` appropriately to renew the generator when the player is reset or cloned (that's why you need the generator function and the arguments passed to the `SequencePlayer` class). I think that these should be saved appropriately with the `init_args` decorator already in the library. Here's a guess as to what `SequencePlayer` would look like:

```
class SequencePlayer(object):

    @init_args
    def __init__(self, generator_function, generator_args, ...):
        # Initialize the sequence generator
        self.sequence_generator = generator_function(*generator_args)
        ...

    def strategy(self, opponent):
        for s in self.sequence_generator:
            yield s

    def reset(self):
        Player.reset(self)
        self.sequence_generator = generator_function(*generator_args)
        # Check to make sure that self.strategy starts over at the beginning of the sequence
        # I think it will but be sure to include a test
```

docs/reference/overview_of_strategies.rst
marcharper: That's right, but it's not independent of the full history of play. In other words it doesn't matter what the last `n` opponent plays were but it does depend on the length.

axelrod/tests/unit/test_thuemorse.py
marcharper: And maybe a link to the sequence at http://oeis.org/A010059 ^ :+1:

Here's another idea I had some time ago but never followed up on: use a python wrapper to OEIS to pull down sequences on the fly. The drawback is that OEIS doesn't list all the terms of the sequence (nor provides an API to generate more terms).

marcharper:
This is a neat sequence! I wonder if it being self-similar makes it easier or harder to infer.

axelrod/tests/unit/test_thuemorse.py
drvinceknight: This looks like it might be relevant: https://pypi.python.org/pypi/oeis/

drvinceknight: Although that doesn't look very maintained...

marcharper:
There's a PR on @geraintpalmer master branch for a general sequence player.

drvinceknight:
> There's a PR on @geraintpalmer master branch for a general sequence player.

Left one question over there. Looks great.

geraintpalmer:
Sequence player looks good thanks @marcharper

drvinceknight:
On https://github.com/geraintpalmer/Axelrod/pull/1#event-562192520 @marcharper wrote:

> Edit: With the reset/clone tests just added in the other PR, I think the two instance players do a fairly good job of testing the base class.

Perhaps @geraintpalmer you could add a simple test or two that just checks basic creation of an instance of the parent class? Perhaps Marc is correct and this is overkill but nice to have.

marcharper:
I'm on it.

drvinceknight:
Closing this as all apart from 4def36b was in #464. Thanks for the strategies @geraintpalmer :) :+1:

PR #461, opened 2016-02-18
37 - Adding probabilistic ending to the match class

drvinceknight:
This is with a view to addressing 37 (this in no way fixes it). This is just a small step but
going in the correct direction.

- Matches can be given a prob end argument which is the probability
  that a match ends on a given round.
- This is randomly sampled using inverse sampling so that we're not
  querying the need for it and the random module more than once.
- At present this can be used 'in conjunction' with turns so that a
  match stops either before some number of turns or probabilistically.
- Docs have been updated.

Note that this does not break any other functionality, not passing a
prob end means the match just behaves as normal.

I have also taken the opportunity to add hypothesis tests to the match.

I know that this will need the work that @meatballs is doing to be in for this to work but I've given up trying to do it in one go. No doubt this will change as and when I go down the rabbit whole (I need to look carefully at the results and how that will work).


axelrod/match.py
```diff
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from random import random
```
marcharper: I usually prefer to import the module in the case that you need a function of the same name to avoid later confusion:

```
import random
r = random.random()
# This still works elsewhere in the file
ri = random.randint()
```

```
from randomimport random
r = random()
# This doesn't work anymore
ri = random.randint()
```

Same for the `datetime` module. That's just my preference, I find it keeps me from making little mistakes later when I try to modify a larger file. Also note that some projects like numpy avoid this and have names like `numpy.random.rand` instead of `numpy.random.random`.

drvinceknight: Yeah good shout. I'll change that.

axelrod/match.py
```diff
@@ -77,11 +82,28 @@ def play(self):
 
         i.e. One entry per turn containing a pair of actions.
         """
+        if self._prob_end is not None:
```
marcharper: It feels like the statistical details here are not really the concern of the `Match` class -- it just needs to know the value of `end_turn`, not how to generate it. Perhaps it's better to factor this part out (which also has the advantage of allowing other ending schemes), and simply pass `end_turn` to `Match.__init__`.

OTOH perhaps it's better for the calling code to change the value of `turns` (handling the randomization outside the `Match` class). I guess there are several ways one might want to do this...

drvinceknight: This sampling is statistically equivalent to a flip of a coin at the end of
each pair of actions. Doing it this way just saves from repeatedly
'flipping a coin'.

I think it's better for the randomization to happen at a match level. It
makes things easier further down the line (I think) and also is closer to
the idea of my interpretation of a stochastic length game. I spent some
time thinking about this and I think that once Owen has finished what he's
working on things will be easier to modularize.

I'm mainly opening this PR to get people's thoughts as I'm not convinced
mine are correct so I'm not at all immovable here.

I don't think we want to change the value of turns. Mainly because of some
cheating strategies... I thought about that for a bit but felt it was
better to keep things separate?

marcharper: Ok, you've obviously thought about this quite a bit, so I'm fine with this approach.

drvinceknight: Well as I go up through the classes I'll keep pushing to this branch. It might well be that it's not quite right just yet...

Issue #465, opened 2016-02-23
Tests currently fail on py2


drvinceknight:
#464 has introduced a bug, tests fail on py2:

```
........................................Falsifying example: test_property_serial_play(self=<axelrod.tests.unit.test_tournament.TestTournament testMethod=test_property_serial_play>, s=[<class 'axelrod.strategies.memoryone.ALLCorALLD'>, <class 'axelrod.strategies.grudger.Aggravater'>, <class 'axelrod.strategy_transformers.BackStabber'>, <class 'axelrod.strategies.alternator.Alternator'>, <class 'axelrod.strategies.mindreader.MindReader'>], turns=2, repetitions=2, rm=random.seed(0))
E..............
======================================================================
ERROR: test_property_serial_play (axelrod.tests.unit.test_tournament.TestTournament)
Test serial play using hypothesis
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tests/unit/test_tournament.py", line 110, in test_property_serial_play
    min_size=2,  # Errors are returned if less than 2 strategies
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/hypothesis/core.py", line 502, in wrapped_test
    print_example=True, is_final=True
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/hypothesis/executors.py", line 57, in default_new_style_executor
    return function(data)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/hypothesis/core.py", line 103, in run
    return test(*args, **kwargs)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tests/unit/test_tournament.py", line 129, in test_property_serial_play
    results = tournament.play()
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tournament.py", line 49, in play
    self._run_serial_repetitions(self._outcome)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tournament.py", line 82, in _run_serial_repetitions
    self._run_single_repetition(outcome)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tournament.py", line 75, in _run_single_repetition
    output = self._play_round_robin()
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tournament.py", line 146, in _play_round_robin
    return round_robin.play()
[...]

PR #467, opened 2016-02-23
Adding in some examples for particular failures.

drvinceknight:
This is relevant to #465.

Seems to be an issue with MindReader + Backstabber and ThueMorse...

PR #468, opened 2016-02-23
Eliminate use of copy.deepcopy in mindreader

marcharper:
This is from an experimental branch a few months back to remove the deepcopy in mindreader. It fixes one of the bugs in #465 but is also much slower overall (running a few tournaments to verify timing, will edit with results). I simply rebased onto the latest master and squashed in a few fixes (e.g. `update_history` versus `update_histories`). The other bug is fixed in #469.

One possible downside is that this method doesn't necessarily copy stochastic players exactly because the history of play might change when the history is replayed. OTOH that may be a feature -- an advanced mindreader might run its experiments many times to determine the next best move.

### DIFF
--- a/axelrod/strategies/mindreader.py
+++ b/axelrod/strategies/mindreader.py
@@ -2,25 +2,23 @@ import copy
 import inspect
 
 from axelrod import Actions, Player, RoundRobin, update_history
+from .cycler import Cycler
 
 C, D = Actions.C, Actions.D
+
+def limited_simulate_play(player_1, player_2, h1):
+    """Here we want to replay player_1's history to player_2, allowing
+    player_2's strategy method to set any internal variables as needed. If you
+    need a more complete simulation, see `simulate_play` in player.py. This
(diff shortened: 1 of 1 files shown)


marcharper:
So the reason that this approach is slower is simply that the cloning and replaying of players is `O(turns^2)`. It's not terrible -- about an hour-ish for the combined tournament with the standard options. We could mark some strategies as "do not copy" and only use cloning on those. On the other hand, using `clone` instead of `copy` is the morally right thing to do IMO.

drvinceknight:
I don't have a problem with an extra hour of run time.

drvinceknight:
I have merged this in to #467. Assuming travis passes over there I'll merge which will close this PR. I will then open an issue to clearly describe everything that's been put in place to fix #465 in case @meatballs and @langner would like time to go over it all. We moved quickly here because this was a bug (I doubt anyone was in a rush to create a tournament with `MindReader` and `BackStabber` but who knows :)). :+1:

PR #469, opened 2016-02-24
Strategy transformer fix

marcharper:
This fixes the issue with mindreader and the final transformer from #465. The problem was that mindreader was reading beyond the end of the game length which causes an index error in FinalTransformer. The transformer now simply passes through the action if the tournament length is exceeded.


drvinceknight:
Merging as this is a bug fix.

Issue #404, opened 2015-10-29
Reorganize overview of strategies


marcharper:
I think we should de-emphasize the strategies from Axelrod's original tournaments unless they are particularly interesting or well-known now. Additionally, there are **many** strategies in the library that do not appear in the overview at all. I suggest we split the file into:
* One overview with all strategies, in alphabetical order
* Auxiliary files for various tournaments, if desired.

drvinceknight:
Yeah, this is a good suggestion. This further modularises the docs.

drvinceknight:
> One overview with all strategies, in alphabetical order

This could/should be automated from the Sphinx docstrings. We have a script that did that... I removed it during the re write but would be easy to put back in.

PR #473, opened 2016-02-28
404 - Cleaning up documentation for strategies

drvinceknight:
This addresses #404 and also is related to a conversation @meatballs and I had about apidoc.

I'm not actually using apidoc, using my editor it took me two seconds to create the file, as I wrote in the docs I suggest we just add in a new module when we need to but of course we could also automate this (if someone wanted to take that on: this would in fact just be going back to something we had in a script before but apidoc sound cool!).

You can see how this looks here: http://axelrod.readthedocs.org/en/404/reference/all_strategies.html


marcharper:
Nice! So it grabs all functions and classes in `axelrod.strategies`, yes? If so we might consider moving a few non-strategy classes to other places (like the `Memoized` decorator just added with the sequence players), and functions like `axelrod.strategies.hunter.detect_cycle`.

Alternatively we could just slightly update the page to say "Docstrings for all strategies and their auxiliary functions" or something similar.

drvinceknight:
_Technically_ it grabs all functions and classes from `axelrod.strategies.<file>`.  Each `<file>` is hard code at present.

But yeah, at present the  `Memoized` decorator is documented. I'm happy either way, perhaps I can include the line you suggest in the docs and open an issue so that we can move it (and similar) out in another PR? If you think it's nicer to do that all here I'm happy as well. 

Heading to bed now so will catch up with this in the am.

marcharper:
Yeah that's fine, it's a minor comment. Looks good to me.

drvinceknight:
> Yeah that's fine, it's a minor comment. Looks good to me.

I've started moving the utils elsewhere. Woke up wondering why I wouldn't just do that... :)

PR #474, opened 2016-03-03
Interactions and RoundRobin

meatballs:
This is the final piece of work on #284. It will:

* Add tournament_type.py
This module contains a ```round_robin``` function which is passed to the Tournament class. It defines the set of Match objects to be be played for the tournament.

The idea is that other types of tournament could be represented as functions with the same signature and passed to Tournament objects. This could be done either within the tournament_type module, if we want that type permanently within the library, or ad-hoc by individual users.

* Add ```Tournament.type```
This is a new parameter in the __init__ method for Tournament. It expects a function to be passed which defines the set of matches to play.

* Refactor the ```Tournament._play_round_robin``` method
This is replaced by ```Tournament.play_matches``` and the ```tournament_type``` module.

* Remove the now redundant RoundRobin class

* Rename the integration test ```test_round_robin``` to ```test_sample_tournaments``` and refactor its content

* Add ```Tournament.matches```
This is a list of lists - one entry per repetition each containing a list of axelrod.Match objects. Its population is controlled by the ```keep_matches``` parameter which has been added to ```Tournament.__init__``` with a default value of False.


basic_strategies.csv
```diff
@@ -0,0 +1,3 @@
+Defector,Bully,Suspicious Tit For Tat,Alternator,Tit For Tat,Anti Tit For Tat,Win-Stay Lose-Shift,Cooperator
```
drvinceknight: I don't think you wanted to commit these. They're created by the test suite (a couple of other files are similarly created)?

docs/tutorials/getting_started/interactions.rst
```diff
@@ -0,0 +1,34 @@
...
+    ...     strategies, turns=3, repetitions=1, with_interactions=True)
+    >>> results = tournament.play()
+
+We can now view the detailed interactions that occurred between the players
+for each turn in each match. (The interactions are stored within a dictionary,
+so we'll add a sort to show the entries in a sensible order).::
+
+    >>> for key in sorted(tournament.interactions):
+    ...     print('%s: %s' % (key, tournament.interactions[key]))
+    (0, 0): [('C', 'C'), ('C', 'C'), ('C', 'C')]
```
drvinceknight: Is the output of this the index of the strategies? 

- Could we change it to be the str? (Perhaps this is not a good suggestion)
- If not, could you include a sentence just saying that it's the indices of the players.

Also, how does this fit in with the sparklines? (That was one of the motivations for this refactor if I recall).

meatballs: Yes, they key in that dictionary is the pair of index numbers. It's the dictionary that's passed to payoff_matrix and cooperation_matrix to compute the results.

I suppose we could pretty it up somehow before it's stored as Tournament.interactions.

meatballs: > Could we change it to be the str? (Perhaps this is not a good suggestion)

We could, I suppose, but it feels better to me to output the actual data that's being within the library

meatballs: > Also, how does this fit in with the sparklines? (That was one of the motivations for this refactor if I recall).

The dictionary could be passed through the sparklines function if the user so wished

drvinceknight:
> We could, I suppose, but it feels better to me to output the actual data that's being within the library

Cool, I'm happy with that :) If it's not too much bother could you just add a line to the docs saying that the keys are the pairs of the indices?

docs/tutorials/getting_started/interactions.rst
drvinceknight: That's awesome. Is it worth putting that in the docs?

meatballs: Here's another idea - instead of holding the interactions within the tournament, I could just hold the match objects themselves. That way, user could do whatever they liked - including calls to the sparkline function!

drvinceknight: That sounds perfect to me.

drvinceknight: Any scope in using a tuple of players as the key?

meatballs: > Any scope in using a tuple of players as the key?

No. When the dictionary is passed to the payoff_matrix and cooperation_matrix functions, they would be unable to distinguish between two players of the same class.

meatballs: I'll have a go at using the match objects and some kind of function (possibly a property) to return a pretty version of the interactions from those objects.

drvinceknight: Sounds great. I think having the match objects returned is important
because at some point down the line or indeed for potential other
tournament types those match objects could have other information.

I'll look at it all closely next week (on Monday I've got some train time
:+1:).

axelrod/_strategy_utils.py
```diff
@@ -47,11 +48,23 @@ def limited_simulate_play(player_1, player_2, h1):
...
     update_history(player_2, h2)
 
+
 def simulate_match(player_1, player_2, strategy, rounds=10):
     """Simulates a number of matches."""
     for match in range(rounds):
         limited_simulate_play(player_1, player_2, strategy)
 
+
+def calculate_scores(p1, p2, game):
```
marcharper: Does it make sense to just pass in the histories here rather than the player objects?

axelrod/_strategy_utils.py
```diff
@@ -61,29 +74,28 @@ def look_ahead(player_1, player_2, game, rounds=10):
...
+    (not reevaluated). From:
+    https://wiki.python.org/moin/PythonDecoratorLibrary#Memoize
+    """
+
+    def __init__(self, func):
         self.func = func
         self.cache = {}
 
-   def __call__(self, *args):
+    def __call__(self, *args):
```
marcharper: Sorry about the spacing on these! That's what I get for copying from the python wiki.

axelrod/tests/integration/test_sample_tournaments.py
```diff
@@ -0,0 +1,68 @@
...
+    def setUpClass(cls):
+        cls.game = axelrod.Game()
+
+    @classmethod
+    def get_test_outcome(cls, outcome, turns=10):
+
+        # Extract the name of players from the outcome tuples,
+        # and initiate the players by getting the classes from axelrod.
+        names = [out[0] for out in outcome]
+        players = [getattr(axelrod, n)() for n in names]
```
marcharper: Just a thought for the future: we might think about having a dictionary mapping player names to classes in the axelrod module for python beginner friendliness (avoiding the need for `getattr`). Perhaps a solution in search of a problem.

axelrod/tests/unit/test_tournament.py
```diff
@@ -394,26 +412,58 @@ def test_worker(self):
...
             game=self.game,
-            turns=200,
-            repetitions=self.test_repetitions)
-        output = tournament._play_round_robin(cache_mutable=False)
-        self.assertEqual(output['payoff'], self.expected_payoff)
-        self.assertEqual(tournament.deterministic_cache, {})
+            turns=3,
+            repetitions=self.test_repetitions,
+            keep_matches=True)
+        matches = {
```
marcharper: Maybe parameterize `matches` with `turns=200` above and `turns=3` here in a function?

axelrod/tests/unit/test_tournament_type.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_round_robin(self):
+        matches = axelrod.tournament_type.round_robin(
+            players=self.players,
+            turns=test_turns,
+            deterministic_cache={}
+        )
+        match_definitions = [
+            (match) for match in matches]
+        expected_match_definitions = [
+            (0, 0),
```
marcharper: Perhaps a bit overkill with the number of lines here :)

axelrod/tournament.py
```diff
@@ -3,32 +3,64 @@
...
+            Whether a cache has been passed in from an external object
+        noise : float
+            The probability that a player's intended action should be flipped
+        with_morality : boolean
+            Whether morality metrics should be calculated
+        with_interactions : boolean
+            Whether interaction results should be included in the output
+        """
         self.name = name
+        if type is not None:
```
marcharper: Can we avoid using `type` as a variable here since it's a python builtin? (I wish it wasn't a built in.)

axelrod/tournament.py
```diff
@@ -126,21 +237,60 @@ def _process_done_queue(workers, done_queue, outcome):
...
-            cache_mutable=cache_mutable,
-            noise=self.noise)
-        return round_robin.play()
+    def _play_matches(self, matches):
+        """
+        Play the supplied matches.
+
+        Parameters
+        ----------
+        dictionary
```
marcharper: `matches: dictionary`

axelrod/tournament.py
```diff
@@ -126,21 +237,60 @@ def _process_done_queue(workers, done_queue, outcome):
...
+        A dictionary whose key is a tuple of player index numbers and the
+        corresponding value is an axelrod Match object
+
+        Returns
+        -------
+        dictionary
```
marcharper: `output: dictionary`
     mapping ... to ...

axelrod/tournament_type.py
```diff
@@ -0,0 +1,64 @@
...
+    Create a pair of Player objects.
+
+    If the two index numbers are the same, the second player object is
+    created using the clone method of the first.
+
+    Parameters
+    ----------
+    players : list
+        A list of axelrod.Player objects
+    player_1_index : integer
```
marcharper: `player1_index`, `player2_index` like the rest of the function

axelrod/tournament_type.py
```diff
@@ -0,0 +1,64 @@
...
+    deterministic_cache : dictionary
+        A cache of resulting actions for deterministic matches
+    cache_mutable : boolean
+        Whether the deterministic cache should be updated
+    noise : float
+        The probability that a player's intended action should be flipped
+
+    Returns
+    -------
+    dictionary
```
marcharper: `matches: dictionary`

docs/tutorials/getting_started/interactions.rst
```diff
@@ -0,0 +1,41 @@
+Accessing the interactions
+==========================
+
+This tutorial will show you briefly how to access the detailed interaction
+results corresponding to the tournament.
+
+As shown in :ref:`getting-started` let us create a tournament except that,
```
marcharper: Maybe something like: "To access the detailed interaction results we create a tournament as usual (see getting-started), setting the `keep_matches=True` parameter to save the interactions."

marcharper:
Looks good to me overall. The only thing I'd really like changed is the use of `type` as an unbound variable name since it conflicts with a python builtin. This could lead to weird and hard to diagnose bugs. (I know it's annoying that it's used as a builtin.) With that changed I'm :+1: for the PR. The other comments are minor.

axelrod/_strategy_utils.py
drvinceknight: I think players could be better actually. Thinking that it might be more future proof? For whatever reason, perhaps at some point we need more than just the histories to calculate the scores? (I have absolutely no idea what that would be but just a thought).

axelrod/_strategy_utils.py
meatballs: No problem - just trying to keep my linter happy so I can tell when I'm introducing proper rubbish for it to be concerned about!

.gitignore
```diff
@@ -6,4 +6,7 @@ _build/
 dist/
 MANIFEST
 Axelrod.egg-info
-.hypothesis/eval_source/
+.hypothesis/
```
drvinceknight: I actually went the other way with this here: https://github.com/Axelrod-Python/Axelrod/commit/fd1b3bdf97369ff332f54fae167ccd14ee76ccf4

That follows https://hypothesis.readthedocs.org/en/latest/database.html#sharing-your-example-database which lets the hypothesis db get shared. 

Was there a reason why it perhaps wasn't working for you? Maybe the part about adding this to the .git/config (which can't be in the commit I linked to above):

```
[merge "hypothesisdb"]
    name = Hypothesis database files
    driver = python -m hypothesis.tools.mergedbs %O %A %B
```

drvinceknight: I think there might be a better way of making the db persist on travis (David mentioned something about this on the hypothesis irc channel). Something about a cached folder... If the db is causing problems we can leave this as it is and open and I can look in to that...

axelrod/match.py
```diff
@@ -119,6 +119,6 @@ def play(self):
 
     def sparklines(self, c_symbol=u'█', d_symbol=u' '):
         return (
-            sparkline(self._player1.history, c_symbol, d_symbol) +
+            sparkline(self.player1.history, c_symbol, d_symbol) +
             u'\n' +
-            sparkline(self._player2.history, c_symbol, d_symbol))
+            sparkline(self.player2.history, c_symbol, d_symbol))
```
drvinceknight: Out of curiosity, why `_player` :arrow_right: `player`?

axelrod/tests/integration/test_sample_tournaments.py
drvinceknight: Nice idea, have opened #479 we can chat about this more there if need be :)

axelrod/tests/unit/test_tournament.py
drvinceknight: Or similarly use a hypothesis test?

axelrod/tournament.py
```diff
@@ -3,32 +3,64 @@
...
+from .tournament_type import round_robin
+from .payoff import payoff_matrix
+from .cooperation import cooperation_matrix
 
 
 class Tournament(object):
     game = Game()
 
-    def __init__(self, players, name='axelrod', game=None, turns=200,
+    def __init__(self, players, tournament_type=None, name='axelrod',
```
drvinceknight: Could the default just be `round_robin` this would be more readable? (Based on what is happening at line 48)

axelrod/tournament.py
```diff
@@ -3,32 +3,64 @@
...
+                 game=None, turns=200,
                  repetitions=10, processes=None, prebuilt_cache=False,
-                 noise=0, with_morality=True):
+                 noise=0, with_morality=True, keep_matches=False):
+        """
+        Parameters
+        ----------
+        players : list
+            A list of axelrod.Player objects
+        name : string
```
drvinceknight: Could you include a description for `tournament_type`?

Something like:

```
tournament_type : a class from the tournament_type module, default is `round_robin`
```

axelrod/tournament.py
```diff
@@ -126,21 +237,59 @@ def _process_done_queue(workers, done_queue, outcome):
...
+        matches : dictionary
+            Mapping a tuple of player index numbers to an axelrod Match object
+
+        Returns
+        -------
+        dictionary
+            Containing the payoff and cooperation matrices
+        """
+        interactions = {}
+        matches_to_keep = []
```
drvinceknight: Only create this if `self._keep_matches`?

drvinceknight:
This looks awesome @meatballs, just one or two very minor questions/requests.

Great improvement to the library.

.gitignore
meatballs: To be honest, I haven't looked into hypothesis at all and simply added this to .gitignore because I kept accidentally committing test files.

axelrod/match.py
meatballs: cock up!

axelrod/tournament.py
meatballs: Yep. I'll sort that

axelrod/tournament.py
meatballs: OK, will do

axelrod/tournament.py
meatballs: Yep, probably wise. I'll sort that

.gitignore
drvinceknight: > To be honest, I haven't looked into hypothesis at all and simply added this to .gitignore because I kept accidentally committing test files.

Let's leave it as you have it here and I'll look in to making the db persist on travis. :) :+1:

axelrod/match.py
meatballs: Because it's being called in various places from outside the Player class and so shouldn't have the underscored name.

drvinceknight: Even though we don't necessarily want it exposed to the user? (Or perhaps we do?)

Genuine questions, not at all disagreeing, just wanting to learn :)

meatballs: My understanding is that the underscore indicates whether an attribute should be considered 'private' i.e. only accessed from within the object itself. I don't think it has anything to do with whether a user should be accessing it.

drvinceknight: Cool :) Thanks.

Issue #438, opened 2015-11-24
Instantiation of LogHandlers


meatballs:
The setup_logging function in utils.py instantiates all the possible log handlers within the logHandlers dictionary. This leads to 'axelrod.log' files being created when they aren't required.

Instead, the dictionary should have the class names as its values and only the selected entry should be instantiated.

drvinceknight:
:+1: I might take a look at this (the log handlers is a part of the library I know nothing about) but if someone wants to jump in before me: be my guest.

drvinceknight:
Fixed by #480.

PR #480, opened 2016-03-07
Only instantiate the logHandler in use

meatballs:
#438

### DIFF
--- a/axelrod/utils.py
+++ b/axelrod/utils.py
@@ -10,15 +10,18 @@ def timed_message(message, start_time):
     elapsed_time = time.time() - start_time
     return message + " in %.1fs" % elapsed_time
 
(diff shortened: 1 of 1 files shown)


drvinceknight:
Looks good to me, is it possible to write a test for this?

Either way: :+1:

meatballs:
> is it possible to write a test for this?

I'm sure it is, but I was feeling too lazy to tackle it since we don't have any tests at all for utils.py!

drvinceknight:
> I'm sure it is, but I was feeling too lazy to tackle it since we don't have any tests at all for utils.py!

Cool, have opened #482 I'm happy to tackle that.

meatballs:
:+1:

Issue #413, opened 2015-11-04
Clone players at tournament level


meatballs:
We should move the cloning of players (for self-interactions) to the tournament level so that it only occurs once.

This should wait until #284 is complete.

marcharper:
Instead of once per repetition?

meatballs:
Exactly, yes!

marcharper:
Good idea -- that should be a significant speed up, yes?

meatballs:
I hope so. It would certainly reduce the number of cloning operations and they seem to take a large amount of time each.

drvinceknight:
Sounds good to me.

meatballs:
Given how we've ended up tackling #284, I think this is now impossible.

As we're keeping the Match objects, which each include a pair of Players objects with the completed history, we can't re-use Player objects between repetitions.

meatballs:
Actually, I may have to disagree with myself above. Further thinking required....

drvinceknight:
Done by #483.

PR #481, opened 2016-03-07
Add cloning of opponents to Tournament

meatballs:
#413

### DIFF
--- a/axelrod/tournament.py
+++ b/axelrod/tournament.py
@@ -14,7 +14,7 @@ class Tournament(object):
     game = Game()
 
     def __init__(self, players, tournament_type=round_robin, name='axelrod',
-                 game=None, turns=200,
+                 clone_opponents=True, game=None, turns=200,
                  repetitions=10, processes=None, prebuilt_cache=False,
                  noise=0, with_morality=True, keep_matches=False):
         """
@@ -27,6 +27,8 @@ class Tournament(object):
             an axelrod Match object
         name : string
             A name for the tournament
+        clone_opponents: boolean
+            Whether the players should be cloned to form the opposing players
         game : axelrod.Game
             The game object used to score the tournament
         turns : integer
@@ -51,6 +53,10 @@ class Tournament(object):
         if game is not None:
             self.game = game
         self.players = players
+        if clone_opponents:
+            self.opponents = self._cloned_opponents(self.players)
+        else:
+            self.opponents = self.players
         self.repetitions = repetitions
         self.prebuilt_cache = prebuilt_cache
         self.deterministic_cache = {}
@@ -77,6 +83,25 @@ class Tournament(object):
                  noise=self.noise)
             newplayers.append(player)
         self._players = newplayers
+
+    def _cloned_opponents(self, players):
+        """
+        Call the clone method for each player to produce the list of opponents
+
+        Parameters
+        ----------
+        players : list
+            A list of axelrod.Player objects
+
+        Returns
+        -------
+        list
+            Of cloned axelrod.Player objects
+        """
+        opponents = []
+        for player in players:
+            opponents.append(player.clone())
+        return opponents
 
     def play(self):
         """
@@ -131,6 +156,7 @@ class Tournament(object):
         """
         matches = self.tournament_type(
             players=self.players,
+            opponents=self.opponents,
             turns=self.turns,
             deterministic_cache=self.deterministic_cache,
             cache_mutable=True,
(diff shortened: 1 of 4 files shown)


axelrod/tournament.py
```diff
@@ -27,6 +27,8 @@ def __init__(self, players, tournament_type=round_robin, name='axelrod',
             an axelrod Match object
         name : string
             A name for the tournament
+        clone_opponents: boolean
```
drvinceknight: Can we give more context here, I'm actually not sure I understand what is going on so for a beginner using the library I think this would certainly be confusing.

drvinceknight:
> #413

I understand what you said there but I'm not sure why this is then optional? As I said in my inline comment, I think that could be confusing to a beginner (it's confusing me :)).

meatballs:
I made it possible to set clone_opponents to False simply because it's only necessary if the tournament includes matches between players of the same class. Whilst that's true for a round robin, it isn't necessarily always true.

meatballs:
Strictly speaking, it's not really true even for a round robin - it just is for our 'flavour' of round robin!

drvinceknight:
> I made it possible to set clone_opponents to False simply because it's only necessary if the tournament includes matches between players of the same class. Whilst that's true for a round robin, it isn't necessarily always true.

Ah I understand: yes that's useful... so the use case would be to create a tournament by passing a given type **and** set `clone_opponents` to False? I'm not sure I like the idea of having to do those two things (and potentially not remembering to do one of them). 

What I really like about the changes you made for #284 is that you just create a tournament type function and it takes care of everything (single source of truth)... Am I missing something?

axelrod/tournament.py
```diff
@@ -27,6 +27,10 @@ def __init__(self, players, tournament_type=round_robin, name='axelrod',
             an axelrod Match object
         name : string
             A name for the tournament
+        clone_opponents : boolean
+            Whether the players should be cloned to form the opposing players.
+            Set this to True if the tournament includes matches between players
```
drvinceknight: It's `True` by default, so perhaps this should be "Set this to False if...", also, if this is staying in can we move it to the end of the named argument list? Only that if it's so early in the list it might feel like it's something you're 'supposed' to play with but really it's only going to be used in 'specific cases'.

meatballs:
> What I really like about the changes you made for #284 is that you just create a tournament type function and it takes care of everything (single source of truth)... Am I missing something?

No, you're not missing anything - it would be 'neater' to handle this in the tournament_type function. However, the whole point of #413 is to handle the cloning at Tournament level so it's only done once. If it's done in the tournment_type, then it happens for every repetition just as it does now.

axelrod/tournament.py
meatballs: will do. hold on.....

drvinceknight:
What happens if we pass `clone_opponents` as `False` but use our `round_robin`?

meatballs:
> What happens if we pass clone_opponents as False but use our round_robin?
bad things

meatballs:
> What happens if we pass clone_opponents as False but use our round_robin?

Here's a thought - We could pass the clone_opponents param into the tournament_type function so that it can check that it's been set correctly.

drvinceknight:
> No, you're not missing anything - it would be 'neater' to handle this in the tournament_type function. However, the whole point of #413 is to handle the cloning at Tournament level so it's only done once.

Could the `tournament_type` return both: `matches, clone_players`? 

```
...
                pair, turns, deterministic_cache, cache_mutable, noise)
            matches[(player1_index, player2_index)] = match
    return matches, True
```

and then (in the tournament):

```
self.matches, self.clone_opponents = tournament_type
```

What do you think?

meatballs:
It could, but that would happen on each repetition and thus defeat the point of #413

meatballs:
Sorry - just realised that I misunderstood what you were suggesting.

drvinceknight:
> It could, but that would happen on each repetition and thus defeat the point of #413

Really? We're just passing an extra boolean when we pass the tournament_type, everything else stays the same, why would it repeat anymore than it does already here?

meatballs:
But that doesn't work because the function needs to know the value of that parameter before it can create the Match objects correctly.

drvinceknight:
> Sorry - just realised that I misunderstood what you were suggesting.

The tethering on the train means our messages are going out at the same time :) Sorry!

meatballs:
We could take this conversation over the gitter channel

PR #483, opened 2016-03-08
Re-introducing RoundRobin

meatballs:
#413 Create a TournamentType class and RoundRobin class in order to clone players once at the Tournament level rather than once per repetition

### DIFF
--- a/axelrod/tournament_type.py
+++ b/axelrod/tournament_type.py
@@ -1,36 +1,7 @@
 from .match import Match
 
 
-def pair_of_players(players, player1_index, player2_index):
-    """
-    Create a pair of Player objects.
-
-    If the two index numbers are the same, the second player object is
-    created using the clone method of the first.
-
-    Parameters
-    ----------
-    players : list
-        A list of axelrod.Player objects
-    player1_index : integer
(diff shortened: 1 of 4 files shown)


axelrod/tournament_type.py
```diff
@@ -1,63 +1,74 @@
...
+        self.players = players
+        self.turns = turns
+        self.deterministic_cache = deterministic_cache
+        self.opponents = players
+
+    @property
+    def opponents(self):
+        return self._opponents
+
+    @opponents.setter
```
drvinceknight: So is the cloning back to happening at the RoundRobin level **but** this `setter` basically caches it so that it only needs to be done once?

meatballs: Not quite. The old RoundRobin class was instantiated once per repetition. This is instantiated once per Tournament and its ```build_matches``` method is called once per repetition.

As the cloning is done on instantiation, it occurs at the Tournament level.

drvinceknight: Awesome.

drvinceknight:
Looks fantastic to me.

drvinceknight:
:+1:

PR #484, opened 2016-03-09
Very minor: hypothesis'ing a test

drvinceknight:
Follows comment by @marcharper here: https://github.com/Axelrod-Python/Axelrod/pull/474#discussion_r55115176


drvinceknight:
Not too sure what's happened with this branch, really should only be af5c5b6.

marcharper:
Looks fine. If you want to get it down to just one commit, just make a new branch and cherry-pick af5c.

drvinceknight:
That's basically equivalent to what I did. Really not sure where the extra comits are coming from. I just redid this on #485 (the commands I used are there).

drvinceknight:
Could this possibly have something to do with the history change? I see no way how but I'm confused... (as you can see on #485, I did a clean checkout from master...). Is there some history I've got on my computer that I shouldn't?

PR #489, opened 2016-03-11
Added Gambler strategy

GDKO:
Hello axelrod people,

I have added a new strategy called PSO Gambler.
Instead of using a binary outcome for the lookup table from LookerUp class, I have run a Particle Swarm Optimisation to calculate the probability for C or D based on the lookup_table. This resulted in 6 cases where the action is randomised. 
Furthermore, It will always defect the last 2 turns (similar to the Backstabber class).

I have also added some tests, based on the lookerup, backstabber and random test suites.

Let me know if there are any problems.

I had great fun using the library.

Cheers,
Georgios

### DIFF
--- a/axelrod/strategies/gambler.py
+++ b/axelrod/strategies/gambler.py
@@ -57,7 +57,7 @@ class Gambler(Player):
         # If there isn't enough history to lookup an action, cooperate.
         if len(self.history) < max(self.plays, self.opponent_start_plays):
             return C
-        # GK: Defect Last 2 turns
+        # GK: Defect Last 2 turns, Idea came from the Backstabber class
         if len(opponent.history) > (self.tournament_attributes['length'] - 3):
             return D
         # Count backward m turns to get my own recent history.
--- a/axelrod/strategies/gambler.py
+++ b/axelrod/strategies/gambler.py
@@ -1,3 +1,9 @@
+from axelrod import Actions, Player, init_args, random_choice
+from itertools import product
+
+C, D = Actions.C, Actions.D
+
+
 class Gambler(Player):
 
     name = 'Gambler'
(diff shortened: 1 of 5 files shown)


meatballs:
Looks good. Many thanks and welcome!

axelrod/strategies/gambler.py
```diff
@@ -0,0 +1,111 @@
...
+
+C, D = Actions.C, Actions.D
+
+
+class Gambler(Player):
+    """
+    A LookerUp class player which will select randomly an action in some cases.
+    It will always defect the last 2 turns.
+    Most comments and structure come from the LookerUp class.
+    New comments start with GK
```
drvinceknight: Could you move these last two lines outside of the docstring? (Perhaps an inline comment just below?) Only because the docstring will appear here: http://axelrod.readthedocs.org/en/latest/reference/all_strategies.html and without the whole code this line could be slightly confusing.

marcharper:
Very nice! Thanks for contributing :)

langner:
Very nice. I hope we see more of these types of strategies. The whole
backstabber business also got me thinking that in an evolutionary context
defecting at the end should not be viable. But is there any way to select
against backstabbers in a population dynamics setup?

drvinceknight:
What do you mean by 'select against backstabbers in an evolutionary
context'?

(It's possible to identify all strategies that make use of the game length,
we have a classifier for that. )

drvinceknight:
It's the 'select against ...'  that I'm missing :)

marcharper:
Some people study setups with e.g. reputation-based elements, or dynamics involving a punishment option are used to eliminate defectors.

This is a [great article](http://grantland.com/features/diplomacy-the-board-game-of-the-alpha-nerds/) that contains a story of a guy that built a reputation for being a cooperator over many years as a player of a game called diplomacy and defected when he had an opportunity for a rare solo victory.

langner:
> It's the 'select against ...' that I'm missing :)

By 'select against' I meant be able to defect against in the next iteration of an ecological simulation.

PR #491, opened 2016-03-11
Addressing #482: some tests for utils.py

drvinceknight:
- Tests for timed_message (simple)
- Tests for logger
        - Adding testfixtures https://pythonhosted.org/testfixtures/logging.html It looks like this is necessary but it *does* add to the requirements.txt. Is it worth it?

Furthermore I'm not completely satisfied by the tests for the logger but
it *is* at least a test.

Note that this doesn't add a test for `run_tournament`, that's very much tested in a lot of other places and everytime we run the results but I suppose that a test would be nice. I might get to that at some point.

### DIFF
--- a/axelrod/tests/unit/test_utils.py
+++ b/axelrod/tests/unit/test_utils.py
@@ -42,14 +42,3 @@ class TestSetupLogging(unittest.TestCase):
             axelrod.utils.setup_logging(logging_destination='console',
                                         verbosity=level)
             compare(logger.level, levels[level])
-
-        # Removing this as it doesn't seem to work on travis.
-        #compare([
-            #C('logging.StreamHandler',
-              #stream=sys.stderr,
-              #formatter=C('logging.Formatter',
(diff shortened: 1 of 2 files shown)


meatballs:
> Furthermore I'm not completely satisfied by the tests for the logger but
it is at least a test.

Agreed - we had nothing previously, so this is definitely worth having

drvinceknight:
@meatballs I've removed your positive review label: I had to make some tweaks. It's more or less the same situation: not necessarily as tested as would be ideal but tests nonetheless.

marcharper:
It's fine with me, I'd prefer the commented code to be in a separate commit and left in another branch, but that's a minor comment. Either way it's fine with me.

drvinceknight:
I agree. I'll remove it tomorrow (not in front of my machine).

drvinceknight:
This is basically ready to go now @meatballs.

PR #492, opened 2016-03-12
Refactoring of Gambler classes to inherit from LookerUp

marcharper:
* Remove / Refactor redundant code between LookerUp and Gambler
* Use transformer for defection on last two rounds
* Update tests for use of the transformer (it properly handles the `tournament_length == -1` case)
* Removed redundant tests for `random_choice`

We could further refactor EvolvedLookerUp and PSOGambler but I think this is a nice balance between streamlined and readable.

@meatballs The new `Match` class doesn't set the tournament length as far as I can tell. Is that correct? We might consider adding that option as an keyword argument.

Also I think there are a few places we could refactor existing code to use the Match class, particularly in the test suite (TestHeadsUp, responses_test, maybe Mindreader's simulations).

### DIFF
--- a/axelrod/strategies/gambler.py
+++ b/axelrod/strategies/gambler.py
@@ -1,18 +1,20 @@
-from axelrod import Actions, Player, init_args, random_choice
 from itertools import product
+
+from axelrod import Actions, Player, init_args, random_choice
+from axelrod.strategy_transformers import FinalTransformer
+from .lookerup import LookerUp, create_lookup_table_keys
+
 
 C, D = Actions.C, Actions.D
 
 
-class Gambler(Player):
+@FinalTransformer((D, D)) # End with two defections if tournament length is known
+class Gambler(LookerUp):
     """
     A LookerUp class player which will select randomly an action in some cases.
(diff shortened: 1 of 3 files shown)


meatballs:
> The new Match class doesn't set the tournament length as far as I can tell. Is that correct? We might consider adding that option as an keyword argument.

that's correct. It's still done by Tournament and I've been wondering whether it should move. Not really come to a conclusion in my own mind yet!

marcharper:
Ok, I may add it as an option in another PR. I think it would be a nice option for Match regardless, for testing purposes at least.

drvinceknight:
Just a thought about the match class setting the tournament length. We do want to be able to have matches without tournaments (as a nice quick tool to create one v ones).

marcharper:
For sure, I think the default would be to not set the tournament length.

meatballs:
I've had a few thoughts on this. I'll open a separate issue for it.

meatballs:
Opened #493

GDKO:
I can move the PSOGambler inside the LookerUp class. It will need only a few changes.

PR #494, opened 2016-03-19
Add scores method to Match class

marcharper:
Just adds a method to score the results of a Match.


drvinceknight:
Looks fine to me. Could you add a tiny bit to the docs?

drvinceknight:
:+1:

Issue #493, opened 2016-03-12
Setting Match/Tournament length


meatballs:
At the moment, the number of rounds per match is set by the tournament and passed to the matches within it.

This should probably be done within the Match class so that they can be created completely independently - but the Tournament needs to retain some level of control over the definition of matches within itself.

Proposal:

* Match takes an ```end_match``` function as a parameter in its constructor (rather than ```turns``` as at present).

* ```end_match``` functions takes a single parameter - the number of turns played so far - and return a boolean to indicate whether the match should stop.

* The default ```end_match``` function returns true when the number of turns played reaches 200

* The tournament can pass whatever ```end_match``` function it likes to all matches within itself

I also think this would simplify the work required for #37

marcharper:
Also we have the function `Player.set_tournament_attributes(self, length=-1, game=None, noise=0)` that `Match` can take advantage of, maybe in conjunction with a boolean that indicates whether players are aware of the match parameters.

drvinceknight:
I wonder if this makes the match class more complicated than it needs to be? 

The fact that's its easy to make a match between two players of a given number of turns (or as it is presently: a given probability) is quite nice. What problem is this solving? At present _it is_ easy to create a match independently isn't it?

drvinceknight:
(I might just be misunderstanding)

marcharper:
Some players change behavior if they are aware of the match length. Right now we handle that by setting tournament attributes; it would nice to be able to do the same with matches to be able to properly access the different behaviors.

Moreover I'd say it's more a concern of the Match class than of Tournament -- it's easy to image a tournament of three rounds of 50, 100, and 200 turns respectively. It just so happens that a traditional tournament is a single round of matches of equal length.

drvinceknight:
Right. I see a bit better. 

So for matches that won't technically have a length *as far as the players are concerned* (ones that end probabilistically which is what I've been thinking about recently) we'll have to have a separate way of dealing with that? (`match.length=False but `match.prob=.8` kind of thing?).  

I do like the user interface we currently have as described here:  http://axelrod.readthedocs.org/en/latest/tutorials/further_topics/creating_matches.html

drvinceknight:
As far as a tournament of 200, 100 and 50 length rounds would that not just be taken care of by the new functionality that @meatballs put together where a tournament type would be defined in such a way for that?

marcharper:
[...]

PR #496, opened 2016-03-20
Probabilistic ending tournaments 37

drvinceknight:
Working on 37.


meatballs:
I'm not so sure this is the way to go. I think the probabilistic ending should be part of the match definition rather than the tournament.

Different types of tournament should be manner in which matches are created and played. e.g. we might have a league system as well as the current round robin.

Whether a match has a fixed length or a random length, is surely part of the match? Yes, the tournmanent would need a parameter to define the type of matches is should create, but that should simply be passed to the match class itself.

drvinceknight:
> I'm not so sure this is the way to go. I think the probabilistic ending should be part of the match definition rather than the tournament.

This is what I thought and did before (see #495), but after thinking about it for quite a while this really does make the implementation trivial and thanks to the tournament types nicely extensible to any other type of tournament.

> Different types of tournament should be manner in which matches are created and played. e.g. we might have a league system as well as the current round robin.

Yup. I agree *but* having the Matches just defined by the number of rounds doesn't stop this.

> 
> Whether a match has a fixed length or a random length, is surely part of the match? Yes, the tournmanent would need a parameter to define the type of matches is should create, but that should simply be passed to the match class itself.

Ultimately a given match always has a fixed length and it's main defining property is it's length. I think it's much nicer and 'modular' to keep the Match as a single/simple set of rounds of a given length (http://axelrod.readthedocs.org/en/latest/reference/glossary.html#a-match). 

This goes back to the discussion on #493, if the Match class allows us to tell the players what they know everything else can be taken out of it: keeping the Matches a very simple modular part of the tournament. As you said, we want that every match can be different in a tournament, if we can control what players know about the Match we don't need to complicate the Matches. If we need to (if/and when the need arises) we can always inherit and make a different class that the tournament type could use (which I would prefer to adding more and more layers to Matches). 

The main reason I like this approach is that it doesn't add any extra layer of abstraction (which I at present don't understand the need for).

meatballs:
In that case, ProbEndRoundRobin should descend from RoundRobin. Match legnth could be determined by methods in TournamentType or even standalone functions in the tournmanent_type module.

> The main reason I like this approach is that it doesn't add any extra layer of abstraction (which I at present don't understand the need for).

For me, it's needed to avoid code duplication and I think we have that in the two descendants of TournamentType. If the abstraction is done within tournament_type, then I'm happy.

drvinceknight:
> In that case, ProbEndRoundRobin should descend from RoundRobin. Match legnth could be determined by methods in TournamentType or even standalone functions in the tournmanent_type module.

Does it make much difference from descending from `TournamentType` as I'm overwriting the `build_matches` method which is the only thing different?

The match length is determined in the `ProbEndRoundRobin` with the `sample_length` method. I don't think it needs to be in the actual parent `TournamentType` class if that's what you meant?

Completely agree about code de duplication (for example in https://github.com/Axelrod-Python/Axelrod/pull/496/files#diff-dad184e860a161b8b929a7ebbfd21c9cR328 I'm rerunning the `init` of the parent class for the `Tournament` class). Is there a specific part here where you feel it's happening (if it is I shall fix it)?

drvinceknight:
> and I think we have that in the two descendants of TournamentType.

Somehow I missed that. I see what you mean Perhaps we could create a `create_match` method in the tournament type... Which would be the only thing that would need to change.

So:

```
match = Match(pair, self.sample_length(self.prob_end),
```

would become:

```
match = self.create_match(pair),
```

Then the `ProbEnd` would only need to rewrite the `create_match` method.

Does that sound better?

meatballs:
That's pretty much what I had in mind - just hadn't bothered thinking of name!

drvinceknight:
Let me know what you think of a18630a

(I think this is far better, sorry I didn't get what you meant straight away.)

Going to work on this more later, need to start listening to the great talks at http://www.software.ac.uk/cw16/agenda :)

meatballs:
that's looking much DRYer!

meatballs:
Could you expand the docstring for sample_length? I don't know what pdf or cdf refers to.

drvinceknight:
> Could you expand the docstring for sample_length? I don't know what pdf or cdf refers to.

Yup: good point. Let me know if db0dde6 does the trick (very happy to add more if you think it's needed).

drvinceknight:
This is now ready to be looked at :)

Let me start by what is not there and that I would prefer to leave to issues and future PRs:

- Cooperation metrics (I had a hard time untangling results to get this far so I didn't even try, this could actually be simple.)
- Parallel processes (again this could actually be simple, my stamina was low). Not having this is a shame but could be a first next step (if we wanted to start running prob end tournaments in https://github.com/Axelrod-Python/tournament

What is there:

- New `ProbEndTournament` class
- Corresponding `ProbEndResults` class
- Minor tweak to the Plot class but it could in fact be nicer to create a `ProbEndPlot` class. Although perhaps this is not necessary for this tournament but for some tournaments no doubt we will want a whole different library of plots.

If you'd like to see how this looks in a notebook take a look at this branch on the notebooks repo: https://github.com/Axelrod-Python/Axelrod-notebooks/blob/37/basic-prob-end-tournament.ipynb

I really did struggle with untangling the results module. I think this could even be worth a group (virtual or irl) sprint to refactor one day perhaps? I believe that with Matches this can be made far simpler, less layered and welcoming of future tournament types (I might of course be wrong! :)). **So no doubt my interface to the results can be improved: if anyone has any suggestions let me know, but this does work so those could perhaps be issues for a future PR?**

On a brighter note: the Matches class made making the actual tournament trivial :) :+1:

axelrod/match.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, players, turns, deterministic_cache=None,
         self.result = []
         self.player1 = players[0]
```
marcharper: Do we need `players`, `player1`, and `player2`?

axelrod/result_set.py
```diff
@@ -143,3 +147,51 @@ def csv(self):
                     for rank in self.ranking]
             writer.writerow(list(map(str, data)))
         return csv_string.getvalue()
+
+
+class ProbEndResultSet(ResultSet):
+    """A class to hold the results of a tournament."""
+
+    def __init__(self, players, prob_end, repetitions,
```
marcharper: Spacing on the function definition could be better....

drvinceknight: :+1:

axelrod/match.py
drvinceknight: I agree. I've just pushed just going with `players`. That's more flexible and doesn't change any actual functionality.

axelrod/result_set.py
drvinceknight: Let me know if this is better (I'm sticking to "80 char PEP8..." but happy to change it).

marcharper:
I'm fine with updating the results interface later.

drvinceknight:
I went for a long walk with the :dog: in my attempt to stay away from a computer today. Came back with a head full of ideas so going to attempt to work those in. Don't review just yet :)

drvinceknight:
This is superseded by #507 (separate branch).

Issue #504, opened 2016-03-27
Corner Case Bug in LookerUp


marcharper:
Specifying a history length of 0 hits a subtle slicing bug.

PR #505, opened 2016-03-27
Catch self.plays == 0 corner case of LookerUp

marcharper:
Addresses #504.

### DIFF
--- a/axelrod/strategies/lookerup.py
+++ b/axelrod/strategies/lookerup.py
@@ -122,11 +122,15 @@ class LookerUp(Player):
         if len(self.history) < max(self.plays, self.opponent_start_plays):
             return C
         # Count backward m turns to get my own recent history.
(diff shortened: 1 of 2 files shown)


drvinceknight:
Any chance of a test to verify that this is caught?

marcharper:
The test will be an upcoming inclusion of multiple evolved best strategies for various combinations of the parameters. I can add one now for this case if you want. I have a branch up already on my fork with the strategies, but I wanted to try to get the best tables possible so it could all go in at once.

drvinceknight:
> I can add one now for this case if you want.

If that's not too much of a pain, I like the idea of specific tests for bugs we find. I'll leave it up to you. :+1: 

Marking this as a bug, and it's got a :+1: from me (if putting in that test is a pain just go ahead and merge).

Issue #347, opened 2015-10-06
Docstring Format


meatballs:
Our docstrings currently use a variety of formats and we should agree a standard.

My suggestion is that we use Numpydoc (https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) since it's the closest to anything we currently have.

Tasks required are:

1. Agree the standard we're adopting
1. Tidy up the existing docstrings to meet that standard
1. Add some instructions to docs/contributing.rst

drvinceknight:
I'm happy with the suggested format.

langner:
+1

marcharper:
:+1:

meatballs:
I reckon we can close this one now. Anyone object?

drvinceknight:
Fine by me.

marcharper:
Have we converted them all?

drvinceknight:
Fair point. I don't believe we have for everything. Main components: I
believe we have but not _all_ strategies for example.

drvinceknight:
I'll pick this up next.

meatballs:
What's your opinion on my last comment on #457?

I suspect using annotations might be a better option.

drvinceknight:
> What's your opinion on my last comment on #457?

So the idea would be that the docstrings would no longer need:

```
Parameters
-----------
```

or 

```
Returns
--------
```

is that the idea?

I guess when viewing the docstrings (via a notebook for example) the type annotation would be visible so that that would not be 'lost' and we would still have a docstring that describes what is going on right?

Could you explain what we would use that autodoc plugin for? We're not currently autodoc'ing anything apart from the strategies.

meatballs:
there's quite a good example here: https://pypi.python.org/pypi/sphinx-autodoc-napoleon-typehints

meatballs:
The docstrings remain but argument types and return value types move into the annotations.

drvinceknight:
> The docstrings remain but argument types and return value types move into the annotations.

Just so that I understand, the advantage of that is two fold:

1. We don't have to have `Parameters` and `Returns` which is what the numpy spec is about.
2. By getting rid of those we ensure that they are never lies (which as and when we've changed things sometimes they have become). 

> there's quite a good example here: https://pypi.python.org/pypi/sphinx-autodoc-napoleon-typehints

That's a different plugin to the one you pasted on #457 right?

[...]

PR #507, opened 2016-03-28
37 - Prob end and rewrite of the results

drvinceknight:
This adds to #496 (which could be closed).

This is mainly a rewrite of the results module. Now it is initialised by being passed a set of interactions (list of dictionaries mapping indices to matches). This has the main advantage of simplifying things. No more `results.results`, `results.output` and `results.outcome` which were various forms of holding the same data, now all data is basically in `results.matches`.

To do this:

- I added various methods to the Match class: these can be used to calculate more or less everything.
- I've rewritten all the method to calculate the results, for simplicity I have moved them out of `payoff.py` (which I think could be removed) and mostly moved everything out of `cooperation.py`. I'm not sure I see the advantage of that extra layer but perhaps I'm missing something. All result calculations are done by manipulating matches and **aim to make no assumptions about what matches are actually played**. 

This second point has the advantage of making the probabilistic end tournament **very simple** and in fact I've gotten rid of the `ProbEndResults` class as it is not at all needed.

The `Results` class is now in essence a class that computes a standard set of results on any set of matches (even the tests examples I used in the class are not a standard round robin).

One downside is that it is not taking advantage of the length of a match being constant for a standard tournament but I'd suggest that if need be it's easy to inherit and build a `Results` class that took advantage of the structure but I don't think that's at present necessary.

Furthermore if at any point we wanted to add to the `standard` set of results it's just a question of adding a method that only has one type of data source to deal with: the matches. I've kind of done an example of this: the results class now also has a match lengths attribute with just shows the lengths of matches played by each player (useless for standard tournament but neat to see expected results for prob end).

Here's a notebook (which I will at some point tweak): https://github.com/Axelrod-Python/Axelrod-notebooks/blob/37/basic-prob-end-tournament.ipynb

Looking forward to hearing thoughts on this.

PS I actually changed one doctest in the morality metrics. I believe that was not quite being calculated consistently. Might be worth having a specific look at to make sure I'm correct. It's an easy fix back if I'm wrong.


drvinceknight:
One other thing: in order to make things generic and inheritable etc... I have removed the default plot title. Our plots can be passed a title: `plot.boxplot("plot of awesomeness")`. Perhaps this can be used in the tournament manager to get the current plots? 

On #496 I had a horrible `try` `except` way of asking for forgiveness but that's not escalable to further tournaments...

No doubt I'm forgetting other little things...

marcharper:
Quite a PR! Looks good from a quick review. I'll take a close look this evening.

drvinceknight:
> Quite a PR! Looks good from a quick review. I'll take a close look this evening.

No rush, this is worth checking through carefully. Happy to add more tests, clarify etc... (not right now, right now I'm just glad to have opened the PR, I've been a bit obsessed with this since Thursday :)).

meatballs:
I should be able find time to look at this tomorrow

drvinceknight:
:+1:

axelrod/payoff.py
```diff
@@ -52,8 +52,8 @@ def player_count(interactions):
 
             n = (-1 +- sqrt(1 + 8p)) / 2
 
-        Taking only the real roots allows us to derive the number of players
-        given the number of pairs:
+        Taking only the non negative root allows us to derive the number of
```
marcharper: non-negative or nonnegative (sorry for the pedantry)

axelrod/tests/unit/test_resultset.py
```diff
@@ -1,64 +1,409 @@
...
+        self.assertEqual(len(rs.eigenmoses_rating), rs.nplayers)
+        for j, rate in enumerate(rs.eigenmoses_rating):
+            self.assertAlmostEqual(rate, self.expected_eigenmoses_rating[j])
 
     def test_csv(self):
-        rs = axelrod.ResultSet(self.players, 5, 2, self.test_outcome)
+        rs = axelrod.ResultSet(self.players, self.matches)
         self.assertEqual(rs.csv(), self.expected_csv)
+
+#class TestProbEndResultSet(TestResultSet):
```
marcharper: Can we either uncomment or delete this section of commented code?

axelrod/tournament.py
```diff
@@ -4,8 +4,8 @@
 import multiprocessing
 
 from .game import Game
-from .result_set import ResultSet
-from .tournament_type import RoundRobin
+from .result_set import ResultSet#, ProbEndResultSet
```
marcharper: Same here. If you want to keep the code just put it in another commit / branch for safe keeping?

marcharper:
Looks good to me. I really like the additions to the match class and the increased use of it throughout, as well as the probabilistic ending classes. I had just a few minor comments inline.

axelrod/tournament.py
drvinceknight: Nope, don't want to keep. Just missed it. Getting rid of it now :) :+1:

drvinceknight:
Thanks @marcharper, I've made those changes. For info in 627fe84 I've added a property based test that checks that `ProbEndTournament` runs fine for a variety of tournaments (replicating the test that found that bug with `MindReader` and `BackStabber`).

meatballs:
Sorry, got a bit overtaken by incoming stuff today. It will have to be tomorrow now

drvinceknight:
> Sorry,

No problem at all, whenever you have time :)

meatballs:
I'm starting to look at this and, overall, I think it looks good. One comment strikes me immediately:

> I have moved them out of payoff.py (which I think could be removed) and mostly moved everything out of cooperation.py

If that's the case, then could you remove the redundant methods from those two modules? If either ends up empty as a result, then remove the module.

I think the rest of my comments are likely to be picky stuff!!

drvinceknight:
> If that's the case, then could you remove the redundant methods from those two modules? If either ends up empty as a result, then remove the module.

Yeah: good shout. On it! :)

drvinceknight:
> If either ends up empty as a result, then remove the module.

bea72af has done this (removed both the modules).

axelrod/result_set.py
```diff
@@ -13,132 +14,555 @@
...
             players (list): a list of player objects.
-            turns (int): the number of turns per interaction.
-            repetitions (int): the number of time the round robin was repeated.
-            outcome (dict): returned from the Tournament class and containing
-                various sets of results for processing by this class.
+            matches (list): a list of list of completed matches
+                            (1 for each repetition)
             with_morality (bool): a flag to determine whether morality metrics
                 should be calculated.
         """
```
meatballs: This whole docstring needs to be in numpy format as we agreed in #347 

Whilst we've not bothered going through existing docstrings and changing them in one go for that ticket, we have been doing so whenever they've been 'touched' as part of other changes.

axelrod/__init__.py
```diff
@@ -10,9 +10,9 @@
 from .match import Match
 from .strategies import *
 from .tournament_type import *
-from .tournament import Tournament
+from .tournament import Tournament, ProbEndTournament
 from .tournament_manager import TournamentManager
 from .tournament_manager_factory import TournamentManagerFactory
-from .result_set import ResultSet
+from .result_set import ResultSet#, ProbEndResultSet
```
meatballs: Why is this commented at the end? Either it's needed or it's not.

axelrod/result_set.py
drvinceknight: Yup: good catch. I'm on it.

axelrod/__init__.py
drvinceknight: Sloppy by me, relic of my refactor. Fixing now.

drvinceknight:
I think 796a4ae has got the numpy formatting but let me know if not.

drvinceknight:
402ae74 gets al the numpy formatting.

I've realised that the shorter methods in the `Match` class (some new, some old) haven't stuck strictly to it. I don't personally mind (they are one line docstrings) but I can patch them up if we want.

meatballs:
Not worth the effort, I reckon

meatballs:
I'm happy with this once travis passes.

drvinceknight:
> I'm happy with this once travis passes.

:fireworks: :tada: :+1: :bow:

meatballs:
> :fireworks: :tada: :+1: :bow:

You might get to return to the real world for a while

drvinceknight:
> You might get to return to the real world for a while

I hope it's still there... 

Have just opened #509 and #510 which are logical enhancements and pretty straightforward.

PR #513, opened 2016-04-02
All plots: simple label in tnrmnt manager. Addresses #510

drvinceknight:
I use the xlabel for the heatplot as otherwise it overlapped. This was actually missing for some plots previously. As you can see I've gone for keeping things simple. About to work on #509 in which I'll make a slight tweak to have the prob end instead of the match length.

Here are some examples:

basic_strategies_boxplot
basic_strategies_payoff
basic_strategies_lengthplot svg
screenshot 2016-04-02 09 30 59

PR #514, opened 2016-04-02
Add prob end tournament type to run_axelrod

drvinceknight:
This adds to #513 and also addresses the naming of the tournaments output by the manager as discussed here: https://github.com/Axelrod-Python/tournament/issues/27

### DIFF
--- a/axelrod/tests/unit/test_utils.py
+++ b/axelrod/tests/unit/test_utils.py
@@ -51,8 +51,8 @@ class TestBuildExclusionsDict(unittest.TestCase):
         for (eb, eo, ec, eco) in itertools.product(bools, repeat=4):
             expected_dict = {
                 'basic_strategies': eb,
-                'strategies': eo,
+                'ordinary_strategies': eo,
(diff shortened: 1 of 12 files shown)


axelrod/tests/unit/test_utils.py
```diff
@@ -41,3 +41,18 @@ def test_basic_configuration_console(self):
...
                                         verbosity=level)
             compare(logger.level, levels[level])
+
+
+class TestBuildExclusionsDict(unittest.TestCase):
+    def test_build_exclusions_dict(self):
+        """This is really a regression test to make sure the names of the
+        tournaments stay consistent"""
+        bools = [True, False]
+        for (eb, eo, ec, eco) in itertools.product(*[bools for _ in range(4)]):
```
marcharper: This is a lot to unpack -- is there another way to express the argument of the product? I think there is a repeat keyword argument that may do the same thing.

drvinceknight:
defaf81 fixes an error in the exclusions.

Issue #520, opened 2016-04-05
High (too high?) memory usage when running the full tournament


drvinceknight:
When I update the tournament results I get a crash:

```
tournament(gh-pages) ✗: ./update_results.sh 0 20
-- Regenerating noisy tournament figures
/usr/local/lib/python3.4/dist-packages/matplotlib/backends/backend_gtk3agg.py:18: UserWarning: The Gtk3Agg backend is known to not work on Python 3.x with pycairo. Try installing cairocffi.
  "The Gtk3Agg backend is known to not work on Python 3.x with pycairo. "
Starting basic_strategies tournament: Turns: 200, Repetitions: 20, Strategies: 8.
Finished all basic_strategies tasks in 8.4s
Starting ordinary_strategies tournament: Turns: 200, Repetitions: 20, Strategies: 113.
Finished all ordinary_strategies tasks in 1370.4s
Starting strategies tournament: Turns: 200, Repetitions: 20, Strategies: 123.
Traceback (most recent call last):
  File "run_axelrod", line 132, in <module>
    run_tournaments(**kwargs)
  File "/home/vince/src/Axelrod/axelrod/utils.py", line 82, in run_tournaments
    manager.run_tournaments()
  File "/home/vince/src/Axelrod/axelrod/tournament_manager.py", line 65, in run_tournaments
    self._run_single_tournament(tournament)
  File "/home/vince/src/Axelrod/axelrod/tournament_manager.py", line 96, in _run_single_tournament
    self._generate_output_files(tournament, ecosystem)
  File "/home/vince/src/Axelrod/axelrod/tournament_manager.py", line 121, in _generate_output_files
    image_format=self._image_format)
  File "/home/vince/src/Axelrod/axelrod/tournament_manager.py", line 143, in _save_plots
    self._save_plot(figure, file_name)
  File "/home/vince/src/Axelrod/axelrod/tournament_manager.py", line 164, in _save_plot
    figure.savefig(file_name, bbox_inches='tight', dpi=dpi)
  File "/usr/local/lib/python3.4/dist-packages/matplotlib/figure.py", line 1565, in savefig
    self.canvas.print_figure(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/matplotlib/backend_bases.py", line 2180, in print_figure
    **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/matplotlib/backends/backend_svg.py", line 1206, in print_svg
[...]

PR #521, opened 2016-04-05
Investigating 520: reducing memory consumption

drvinceknight:
**No rush with this one: #520 needs investigating.**

This does two things in an attempt to address #520 (although it helped it did not address it on my machine).

1. No longer pass the match class to the `ResultSet`: pass interactions instead (I'm not yet sure how I feel about this, although recreating match instances to deal with the multiprocessing bug, ie the bandaid is silly):

    Whereas before we were passing a dictionary mapping index pairs to matches (which held the interactions as well as methods to analyse those interactions). Now I'm just passing the interactions (this is still in line with the refactor the ResultsSet as all analyse is done with no assumptions of the structure and with just the one base data type). This is what the interactions looks like:

    ```
    {(0,2}:[('C', 'D'), ('D', 'C')...], ...
    ```

    They used to look like:

    ```
    {(0,2), <Match>,...
    ```

    This results in the methods that the Match class use to use now also being made available to the Results class by introducing a `interactions_utils` module (which is lacking tests right now: will get to that).

2. The `tournament_type.build_matches` function is now a generator. I read that that helped with memory, seems like a good idea (doesn't hurt if we're not keeping the matches anyway).

I've also included something to the docs to show how to quickly get the matches back from the interactions (so you can get the sparklines etc...).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Axelrod.egg-info
 basic_strategies.csv
 cache.txt
 test.csv
+basic_tournament.csv
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -14,6 +14,6 @@ from .tournament import Tournament, ProbEndTournament
 from .tournament_manager import TournamentManager, ProbEndTournamentManager
 from .tournament_manager_factory import (TournamentManagerFactory,
                                          ProbEndTournamentManagerFactory)
-from .result_set import ResultSet
+from .result_set import ResultSet, ResultSetFromFile
 from .ecosystem import Ecosystem
 from .utils import run_tournaments, run_prob_end_tournaments, setup_logging
--- a/axelrod/interaction_utils.py
+++ b/axelrod/interaction_utils.py
@@ -80,3 +80,20 @@ def get_normalised_cooperation(interactions):
     normalised_cooperation = tuple([c / float(nturns) for c in cooperation])
 
     return normalised_cooperation
+
+
+def sparkline(actions, c_symbol=u'█', d_symbol=u' '):
+    return u''.join([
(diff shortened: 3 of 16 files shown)


drvinceknight:
> I've also included something to the docs to show how to quickly get the matches back from the interactions (so you can get the sparklines etc...).

I've pulled these out a bit better (fixed imports etc..., tests etc...) and the sparklines have been moved out also. Could be worth adding documentation saying how these functions can be used to analyse any list of 2-tuples of `C`s and `D`s...

drvinceknight:
To try to deal with memory by sharing out tasks, I've added the ability to write interactions directly to file without creating a results class (this is probably a good idea anyway as people might want to be able to write data and analyse using whatever they want). The fundamental block that gets written is the interactions.

I write a series of interactions for one match: `[('C'), ('D'), ('D', 'C'), (A13, A23)]` as a simple string: `CDDCA13A23` (so the length of this string will always be even).

To complement this there's a `ResultSetFromFile` results class that just has a slightly different init file (to be able to read in the data files). Here are some screenshots of this working:

https://www.dropbox.com/s/cbrnyvjlqlhmtav/Screenshot%202016-04-06%2011.36.44.png?dl=0

and gives the same results when going directly to the results class:

https://www.dropbox.com/s/7wizbrlsl1cy5kc/Screenshot%202016-04-06%2011.37.06.png?dl=0

Todo:

- [ ] Write up documentation for this
- [ ] Perhaps move the `string_to_interaction` method to the `interactions_utils` module so that users could easily analyse the data outside of the results class if wanted.
- [ ] See if this actually helps with #520...

drvinceknight:
I'm going to open another PR from this branch and clarify it.

Issue #518, opened 2016-04-03
Investigate use of alternatives to the standard lib multiprocessing tool


drvinceknight:
Following #516, a fix was produced in #517 that makes the workers for the parallel process pass back a dictionary of index_pairs to results instead of index_pairs to matches. Once these come back the matches are 'rebuilt' with the results and then passed on to the result set for analysis as usual.

This is because the standard lib multiprocessing library cannot pass back and forth anything and everything (more details at #516).

Here are some comments showing the fix in #517 and what behaviour would be preferred:

```
  216     def _process_done_queue(self, workers, done_queue, matches):
  ...
  237                 # Because of the Multiprocessing issue described in `def
  238                 # _worker` we have to rebuild the matches so that they can be
  239                 # handled in a consistent way by the results class.  Rebuild
  240                 # the matches with the data:
  241                 new_matches = self.tournament_type.build_matches(
  242                     cache_mutable=False, noise=self.noise)
  243                 for index_pair, result in results.items():
  244                     new_matches[index_pair].result = result
```

and

```
  244     def _worker(self, work_queue, done_queue):
  ...
  260             # Multiprocessing cannot pass back the `new_matches`
  261             # so we cannot do: `done_queue.put(new_matches)`
  262             # Instead we send back the results.
```

drvinceknight:
Closed by #522

PR #522, opened 2016-04-06
Replace multiprocessing with multiprocess

meatballs:
#518

Also add a module definition to decorated classes so that they can be pickled


drvinceknight:
👍

marcharper:
Great job @meatballs!

meatballs:
Thank you! Very few lines of code resulting from much reading and mucking about in the REPL!!

PR #523, opened 2016-04-06
Writing interactions to file, making separation between running tournament and analysing possible and minor memory improvements.

drvinceknight:
This does a few things which all aimed to address #520 but have otherwise improved the library:

- There is the ability to run `tournament.play(filename)` (as opposed to `tournament.play()` which returned a results set). This will write interactions to files: which makes the work "even more" reproducible.
- The main data structure handled by the results is no longer matches but a more atomic: `interactions`. This is not a new class but simple a list of 2-tuples of actions (which was basically what the match class was carrying around for the results set), ie things of the form: `[('C', 'D'), ('D', 'C'),...]`. This takes care of one aspect of the multiprocessing (workers pass back the `interactions`).
- To handle the `interactions`, I've created a new module `interaction_utils`, this is just a bunch of functions that act on a set of interactions. This is used by the `Match` class (which retains all it's previous methods: so `match.scores()`, `match.winner()` **but they are also** used by the `ResultsSet`. Finally as described in the tutorial I wrote about reading and writing data they can also be used separately (just on any list of the form `[('C', 'D'), ('D', 'C'),...]`).
- The other aspect of the multiprocessing is that we no longer pass to them a list of matches but a generator which creates the matches. I read that this helped with memory (makes sense to me).
- As the tournament can write to file, I've created a new `ResultsSetFromFile` class which will read in the interactions from file. It is good to allow for a separation of these two things (but it's not enforced, I'm breaking now previous functionality).

The **TLDT** as far as performance is concerned:

Running the tournament:

- Noisy: 15.69gb of ram
- Standard: 16.69gb of ram
- Prob end: 1.58gb of ram (using `prob_end=.1` makes this quite small)

This outputs files of the following size (the names are not coded anywhere but in scripts currently on the tournament repo):

- strategies_noisy_interactions.csv (247mb)
- strategies_std_interactions.csv (247mb) - These two files are too big for github's limit: need to open an issue to figure that out.
- strategies_probend_interactions.csv (14mb)

Analysing the tournament:

- Noisy: 9.40gb of ram
- Standard: 9.40gb of ram
- Prob end: 0.79gb of ram


axelrod/interaction_utils.py
```diff
@@ -0,0 +1,116 @@
...
+        sparkline(histories[1], c_symbol, d_symbol))
+
+
+def string_to_interactions(string):
+    """
+    Converts a compact string representation of an interaction to an
+    interaction:
+
+    'CDCDDD' -> [('C', 'D'), ('C', 'D'), ('D', 'D')]
+    """
```
marcharper: Here's another way to do it (if you want to be a fancy functional programmer):
```
>>> s = "A1B2C3"
>>> from itertools import islice
>>> list(zip(islice(s, 0, None, 2), islice(s, 1, None, 2)))
[('A', '1'), ('B', '2'), ('C', '3')]
```

marcharper:
Just curious -- is there a reason not to revert to passing matches back and forth given @meatballs latest PR? That's the root cause of the memory usage?

axelrod/interaction_utils.py
drvinceknight: I'm all for being fancy. Happy to put this in (never used islive before :)) but is there an argument for the status quo being slightly more readable?

drvinceknight:
> Just curious -- is there a reason not to revert to passing matches back and forth given @meatballs latest PR? That's the root cause of the memory usage?

I think the main reason is that this makes it easier to have reading and writing to file. This allows for a separation between data generation and analysis which come to think about it was a long time coming... This could be done by keeping the Match classes as before but I think it's nicer this way.

In essence an instance of the Match class has just been replaced with the interactions (so I don't think there's much difference memory wise?). The main purpose of the Match class for the ResultSet was to 'carry those around'. It also frees up the methods that were in the Match class to be used on any interaction (including the ones that could come out of data files).

It also is perhaps more robust? Say in the future we find a way to decorate Match classes or something like that. There is nothing complicated being passed back and forth to the workers: generators going in and lists of 2 tuples coming out.

I was super happy with the idea of passing matches back and forth but after working on this I think it makes things even nicer.

axelrod/interaction_utils.py
drvinceknight: To be clear: I'm happy to go with whatever you suggest let me know what you think. :)

marcharper: It's fine as is.

marcharper:
Ok, great. Thanks for the explanations!

axelrod/interaction_utils.py
```diff
@@ -0,0 +1,116 @@
...
+def get_final_score_per_turn(interactions, game=None):
+    """Returns the mean score per round for a set of interactions"""
+    scores = get_scores(interactions, game)
+    nturns = len(interactions)
+
+    if len(scores) == 0:
+        return None
+
+    final_score_per_turn = tuple(
+        sum([score[playeri] for score in scores]) / (float(nturns))
```
marcharper: Minor: I'd prefer e.g. `player_index` rather than `playeri` for readability

axelrod/interaction_utils.py
```diff
@@ -0,0 +1,116 @@
...
+C, D = Actions.C, Actions.D
+
+def get_scores(interactions, game=None):
+    """Returns the scores of a given set of interactions."""
+    if not game:
+        game = Game()
+    return [game.score(plays) for plays in interactions]
+
+
+def get_final_score(interactions, game=None):
```
marcharper: Minor: `compute_final_score` or similar instead of `get_final_score`? It's technically not simply fetching a score.

axelrod/interaction_utils.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def get_winner_index(interactions, game=None):
+    """Returns the index of the winner of the Match"""
+    scores = get_final_score(interactions, game)
+
+    if scores is not None:
+        if scores[0] == scores[1]:
+            return False  # No winner
+        return sorted([0, 1],
```
marcharper: Would `numpy.argmax` work here?

axelrod/result_set.py
```diff
@@ -94,12 +101,12 @@ def build_match_lengths(self):
 
         for rep in range(self.nrepetitions):
 
-            for player_pair_index, match in self.matches[rep].items():
+            for player_pair_index, interactions in self.interactions[rep].items():
                 i, j = player_pair_index
```
marcharper: `i, j = player, opponent`?

axelrod/result_set.py
```diff
@@ -14,25 +16,30 @@
...
         self.nplayers = len(players)
-        self.matches = matches
-        self.nrepetitions = len(matches)
+        self.interactions = interactions
+        self.nrepetitions = len(interactions)
 
         # Calculate all attributes:
+        self.build_all(with_morality)
+
+    def build_all(self, with_morality):
```
marcharper: Would these be better as properties? Maybe the user doesn't always want to calculate everything.

axelrod/result_set.py
```diff
@@ -633,3 +643,90 @@ def csv(self):
...
+            A tuple:
+                - First element: list of player names
+                - Second element: interactions (list of dictionaries mapping
+                  index indices to interactions)
+        """
+        players_d = {}
+        interactions_d = {}
+        with open(filename, 'r') as f:
+            for row in csv.reader(f):
+                index_pair = (int(row[0]), int(row[1]))
```
marcharper: Can you include some comments about the structure of the csv flie -- i.e. what is in `row[0]`, `row[1]`, etc.

axelrod/tests/unit/test_interaction_utils.py
```diff
@@ -0,0 +1,86 @@
+# -*- coding: utf-8 -*-
+import unittest
+import axelrod.interaction_utils as iu
+from axelrod import Actions
+
+C, D = Actions.C, Actions.D
+
+
+class TestMatch(unittest.TestCase):
+    interactions = [
```
marcharper: The spacing in this section seems like a bit much -- surely some of these fit in 80 characters. OTOH I'm not sure what PEP8 says in this cases....

axelrod/tests/unit/test_tournament.py
```diff
@@ -394,30 +397,137 @@ def test_play_matches(self, turns):
...
-        }
-        tournament._play_matches(matches)
-        for m in matches.values():
-            self.assertEqual(len(m), turns)
-            # Check that the name of the winner is in the list of names of
-            # players in the tournament (the way the matches are created these
-            # are not the same players).
-            self.assertIn(str(m.winner()), [str(p) for p in tournament.players] + ['False'])
+
+        matches = [
```
marcharper: I feel like there ought to be a more compact representation:
```
player_classes = [axelrod.Cooperator, axelrod.TitForTat, ...]
matches = []
for i, player_cls in enumerate(player_classes):
     for j, opponent_cls in enumerate(players_classes):
           matches.append( ((i, j), axelrod.match(player_cls(), opponent_cls(), turns=turns) )
```

axelrod/tournament.py
```diff
@@ -73,7 +72,7 @@ def players(self, players):
             newplayers.append(player)
         self._players = newplayers
 
-    def play(self):
+    def play(self, filename=False):
```
marcharper: `filename=None` seems more appropriate

axelrod/tournament.py
```diff
@@ -269,12 +262,54 @@ def _play_matches(self, matches):
...
-        for match in matches.values():
+        interactions = {}
+        for index_pair, match in matches:
             match.play()
+            interactions[index_pair] = match.result
+        return interactions
+
+    def _write_to_csv(self, filename):
+        """Write the interactions to csv."""
+        with open(filename, 'w') as f:
```
marcharper: maybe `csvfile` instead of `f` for readability

marcharper:
I left a bunch of minor comments, mostly just readability suggestions since the code is fairly terse in places (by python standards), and since readability is a goal of the library. None of the comments are deal-breakers -- as far as I can tell this iteration is more readable than the prior implementations.

One concern I have is that we're trading disk latency for memory usage. That's a classic solution of course but I wonder if we could also attempt to curb memory usage in other ways. I'll put some thought into it.

axelrod/interaction_utils.py
drvinceknight: Yup.

axelrod/interaction_utils.py
drvinceknight: Good idea.

axelrod/interaction_utils.py
drvinceknight: Just max will work actually. Fixed.

axelrod/result_set.py
drvinceknight: I'm not too sure, I went back and forth with this one. Not sure this is best but it's most "like we had before"...

So would you suggestion to not call `build_all` during `init` and have each of the other ones as a property? My only thought as to why that wouldn't be great is in the case of not computing the various results in the right order which would expend more memory?

marcharper: I'm not sure what the optimal approach is. If anything is particularly intense or less likely to be used then they would be good candidates for properties.

axelrod/result_set.py
drvinceknight: Yup.

axelrod/result_set.py
drvinceknight: I think I've got all these. Good be worth checking I understood correctly (c26bf8fab4e5b7ce59c146de9c5c20073c8835b3).

axelrod/tests/unit/test_interaction_utils.py
drvinceknight: Yup they all fit on one line. Much better.

axelrod/tests/unit/test_tournament.py
drvinceknight: Nice. Done.

axelrod/tournament.py
drvinceknight: Done.

axelrod/tournament.py
drvinceknight: Yup.

axelrod/result_set.py
drvinceknight: Perhaps something to investigate at a later stage perhaps?

I suggest that users in a notebook etc will be running small enough tournaments that it makes sense to just calculate everything (the order in which this sets the attribute is purposeful).

If however you're running a big tournament then you would probably want to do a one of calculation... Or perhaps that's not the right way to think about it...

>  If anything is particularly intense or less likely to be used then they would be good candidates for properties.

I would prefer it if we handled every property the same I think...

drvinceknight:
Have addressed all these @marcharper (I think! :)), big fan of all the comments about readability. Very important.

Just the one point about properties etc that I'm unsure as to what is best...

axelrod/match.py
```diff
@@ -100,72 +96,38 @@ def play(self):
 
     def scores(self, game=None):
         """Returns the scores of the previous Match plays."""
-        if not game:
-            game = Game()
-        scores = [game.score(plays) for plays in self.result]
+        scores = iu.compute_scores(self.result, game)
         return scores
```
meatballs: Why bother with the ```scores``` variable instead of simply returning the result of ```iu.compute_scores``` in a single line?

axelrod/match.py
```diff
@@ -100,72 +96,38 @@ def play(self):
...
     def final_score(self, game=None):
         """Returns the final score for a Match"""
-        scores = self.scores(game)
-
-        if len(scores) == 0:
-            return None
-
-        final_score = tuple(sum([score[playeri] for score in scores])
-                            for playeri in [0, 1])
+        final_score = iu.compute_final_score(self.result, game)
```
meatballs: Same comment as for ```scores```

axelrod/match.py
```diff
@@ -100,72 +96,38 @@ def play(self):
...
         """Returns the mean score per round for a Match"""
-        scores = self.scores(game)
-
-        if len(scores) == 0:
-            return None
-
-        final_score_per_turn = tuple(
-            sum([score[playeri] for score in scores]) / (float(self._turns))
-            for playeri in [0, 1])
+        final_score_per_turn = iu.compute_final_score_per_turn(self.result, game)
```
meatballs: Same comment as for ```scores```

drvinceknight: Yeah I agree. Will fix these now on the train.

drvinceknight:
Have fixed those @meatballs. I wonder why I did that in the first place? 😕

axelrod/tests/unit/test_tournament.py
```diff
@@ -394,30 +397,128 @@ def test_play_matches(self, turns):
...
+        for i, player_cls in enumerate(player_classes):
+            for j, opponent_cls in enumerate(player_classes):
+                players = (player_cls(), opponent_cls())
+                match = axelrod.Match(players, turns=turns)
+                matches.append(((i, j), match))
+        matches_generator = (element for element in matches)
+
+        self.assertEqual((len(list(matches_generator))), len(matches))
+
+        interactions = tournament._play_matches(matches_generator)
```
marcharper: I'm surprised that this generator works again -- shouldn't it be empty after the above assertion?

Also is there a reason to use a generator here? You've already got an iterable with matches. I could certainly see making matches in to a generator function instead of a list.

marcharper:
I just have one new question about the generator, otherwise this is :+1: from me.

axelrod/tests/unit/test_tournament.py
drvinceknight: > I'm surprised that this generator works again -- shouldn't it be empty after the above assertion?

Yeah that is weird! 

I've just seen what happens when I add:

```
411         list(matches_generator)
412         self.assertEqual((len(list(matches_generator))), len(matches))
```

I then get the expected failure:

```
    lambda data: test(*arguments, **example_kwargs)
  File "/Users/vince/src/Axelrod/axelrod/tests/unit/test_tournament.py", line 412, in test_play_matches
    self.assertEqual((len(list(matches_generator))), len(matches))
AssertionError: 0 != 16
```

Does assert check "what would happen"? I don't know here. I've searched around and can't find anything immediately about this.

I've played around with it adding:

```
 410
 411         #self.assertEqual((len(list(matches_generator))), len(matches))
 412
 413         L = (n for n in range(5))
 414         self.assertEqual(len(list(L)), 5)
 415         M = [k for k in L]
 416         self.assertEqual(list(M), [0,1,2,3,4])
```

Note that I comment out 411 (this doesn't actually fail the tests) but adding 413 and 416 does, it gives:

```
F.........
FAIL: test_play_matches (axelrod.tests.unit.test_tournament.TestTournament)
  File "/Users/vince/src/Axelrod/axelrod/tests/unit/test_tournament.py", line 392, in test_play_matches
    @example(turns=3)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/hypothesis/core.py", line 266, in wrapped_test
    lambda data: test(*arguments, **example_kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/hypothesis/executors.py", line 57, in default_new_style_executor
    return function(data)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/hypothesis/core.py", line 266, in <lambda>
    lambda data: test(*arguments, **example_kwargs)
  File "/Users/vince/src/Axelrod/axelrod/tests/unit/test_tournament.py", line 416, in test_play_matches
    self.assertEqual(list(M), [0,1,2,3,4])
AssertionError: Lists differ: [] != [0, 1, 2, 3, 4]

Second list contains 5 additional elements.
First extra element 0:
0

- []
+ [0, 1, 2, 3, 4]
```

Note that adding:

```
  417
  418         interactions = tournament._play_matches(matches_generator)
  419         self.assertEqual(len(interactions), len(matches))
```

(and removing those extra random tests) passes... I'm confused by this...

For the purpose of readability of the test, we could always just remove `self.assertEqual((len(list(matches_generator))), len(matches))`? It's not a terribly useful test, it was more for me to check the generator from list syntax had worked as I expected.

> 
> Also is there a reason to use a generator here? You've already got an iterable with matches. I could certainly see making matches in to a generator function instead of a list.

I only create a generator to make sure that the `_play_matches(matches_generator)` does indeed work on a generator and doesn't require a list.

Building the list and turning that in to the iterator seemed the simplest way to go...

drvinceknight:
I looked in to the confusing test some more and came to the conclusion of gremlins.

Let me know what you think of 6a80a6a722d365d240b33184896da3d83830d882, I've ultimately removed the test as it didn't really do anything with regards to testing actual functionality of the library. I did add another one liner that tested the `iterations` variable wasn't empty (and is in fact the expected size).

(I also tweaked the matches it creates slightly by adding in the inequality, only because I was a bit confused when the number of matches being played was 16 and not 10, which is the expected number of matches between 4 players, 10 = 4C2 + 4).

axelrod/tests/unit/test_tournament.py
marcharper: With a generator, once you've exhausted it once it usually raises `StopIterationError` and should give an empty list. However the generator function could be programmed to reset, like `range`:
```
>>> gen = range(0, 5)
>>> print(list(gen))
[0, 1, 2, 3, 4]
>>> print(list(gen))
[0, 1, 2, 3, 4]
```
Contrast with:
```
>>> gen = (i for i in range(0, 5))
>>> print(list(gen))
[0, 1, 2, 3, 4]
>>> print(list(gen))
[]
```
`_play_matches` should work with either a generator or a list, right? If the point is to test with a generator, let's simply put a comment above the generator creation block so that it's not confusing in the future.

marcharper:
Ok, this makes more sense -- now you' definitely get a fresh generator each time you call the function. I'm happy with this change.

PR #524, opened 2016-04-07
Adding latest doi link to citation file.

drvinceknight:


marcharper:
This is fine. We could go the extra mile and give the explicit citation that should be used.

drvinceknight:
My thinking behind this was that if we forgot to update the citation file
it would not be up to date with the zenodo doi. But I can just make sure I
do that every time I do a new release. I'll make that adjustment tonight
(or tomorrow :)).

drvinceknight:
Have added the bibtex with the missing components. I've also made this an rst file and included the zenodo badge. @meatballs this is quite different to what you ok'd but I think closer to what you were suggesting on gitter.

drvinceknight:
This is more or less temporary, once the paper is published, we should ask to cite that.

marcharper:
> This is more or less temporary, once the paper is published, we should ask to cite that.
porque no los dos?

marcharper:
Are you opposed to writing out a citation since it's possible that bibtex is not being used by the citer?

drvinceknight:
I don't understand. My latest push has the bibtex. 

What I'm thinking about is that the bibtex will change based on the release tag. This is now in there, saying to look at zenodo for the relevant details *but* once the paper is done we can simply point there as well (with a persistent citation at least for the paper).

drvinceknight:
> porque no los dos 

Si bueno!

marcharper:
It's good to go in for my part, I'll bounce it back to @meatballs for a recheck.

PR #527, opened 2016-04-11
Create DeterministicCache class

meatballs:
#412


langner:
Since Python 2.2 you can subclass `dict` directly.

meatballs:
> Since Python 2.2 you can subclass dict directly.

Yes, and there's a lot on SO and elsewhere that advises to do so. However, the author of Fluent Python says to use UserDict and backs it up with some in-depth stuff on the differences between CPython and PyPy which can catch you out (p76 and 352, if you have a copy). I decided he seemed to know what he was talking about and went with his recommendation.

meatballs:
I also liked the fact that UserDict gives me a separate instance of dict as an attribute. That allows me to pickle just the cached entries in the dict without having to pickle the whole object .

axelrod/deterministic_cache.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        with open(file_name, 'rb') as io:
+            data = dill.load(io)
+
+        if isinstance(data, dict):
+            self.data = data
+        else:
+            raise ValueError('Cannot read file into cache')
+
+        try:
```
marcharper: Is there something that works in py 2 and 3 like `len(self.data.values()[0])`?

meatballs: Good call!

```len(list(self.data.values())[0])``` seems to do it. The 'list' is necessary for python 3 but it also works on python 2. I'll sort it out.

File: axelrod/deterministic_cache.py [hunk 2]
```diff
@@ -0,0 +1,158 @@
...
+    from collections import UserDict
+import dill
+
+from axelrod import Player
+
+
+class DeterministicCache(UserDict):
+    """A class to cache the results of deterministic matches
+
+    For fixed length matches with no noise between pairs of deterministic
```
drvinceknight: We could make this 'for given length' and add the length of the match to the key. In fact the length should probably be there anyway.

drvinceknight:
This looks good @meatballs. My one inline comment is potentially worth another issue or it could be worth having here (up to you). 

Ultimately the match length should be in the key as we want the cache to be able to recreate any interactions (ignoring the potential structure of the tournament).

axelrod/deterministic_cache.py
```diff
@@ -0,0 +1,158 @@
...
+    of multiple repetitions.
+
+    By also storing those cached results in a file, we can re-use the cache
+    between multiple tournaments if necessary.
+
+    The cache is a dictionary mapping pairs of Player classes to a list of
+    resulting interactions. e.g. for a 3 turn Match between Cooperator and
+    Alternator, the dictionary entry would be:
+
+    (axelrod.Cooperator, axelrod.Alternator): [('C', 'C'), ('C', 'D'), ('C', 'C')]
```
drvinceknight: So my suggestion would be that the key should be:

```
(axelrod.Cooperator, axelrod. Alternator, 3):...
```

meatballs:
> My one inline comment is potentially worth another issue or it could be worth having here (up to you).

I think that's worth another issue.

drvinceknight:
> I think that's worth another issue.

Cool. In which case I'm happy :)

> len(list(self.data.values())[0]) seems to do it. The 'list' is necessary for python 3 but it also works on python 2. I'll sort it out.

Let me know when you're done, I'm happy to press 'go'.

PR #528, opened 2016-04-11
Fix DoubleCrosser and test to match description

meatballs:
DoubleCrosser says in its docstring that it defects on the last 2 rounds unconditionally but the code was defecting for the last three rounds. @uglyfruitcake confirms that the intention was for defection on the last two only.

This corrects that error and adds the use of the FinalTransformer decorator.

### DIFF
--- a/axelrod/strategies/backstabber.py
+++ b/axelrod/strategies/backstabber.py
@@ -29,7 +29,7 @@ class BackStabber(Player):
             return D
         return C
 
-
+@FinalTransformer((D, D)) # End with two defections
 class DoubleCrosser(Player):
     """
     Forgives the first 3 defections but on the fourth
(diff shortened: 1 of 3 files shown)


drvinceknight:
:+1:

drvinceknight:
In ` axelrod/tests/unit/test_tournament.py` we added the following test to just keep us on our toes with the  decorators. This isn't really necessary but could you add `DoubleCrosser` in there too (no harm)?

This would be simply adding `DoubleCrosser` to the list on line 161 of `test_tournament.py`: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_tournament.py#L161

marcharper:
Also note that the decorator only triggers if the tournament length is known.

Issue #526, opened 2016-04-10
Rename `tournament_type`


drvinceknight:
The `TournamentType` class is passed to the `Tournament` class and is 'mainly' used there to generate matches.

This is great, as it make it very easy to create a tournament using whatever rules we want to build matches.

The different types of tournaments are at the moment (currently just the one: `ProbEndTournament)` built using inheritance from the `Tournament` class. I think it would be more verbose to rename `tournament_type` to `match_generator` or something similar (**suggestions?**).

I've been thinking about this when answering a gitter query about creating a different type of tournament in which I explained that you had to build a new class of tournament and a new class of tournament_type. A new class of tournament and a new class of match generator seems more explicit.

**To be clear: I'm not suggesting any change of functionality, just a rename of a class.**

An alternative would be to move the match generation functions out of the class and just have them as methods of the tournament but I think this is not necessary and it could be nice to keep these things separated (although I'm not entirely sure why).

meatballs:
I've been thinking something similar, so yes!

(I also like 'Arena' as used in https://github.com/tscizzle/IPD_Morality  - similar to our TournamentManager)

drvinceknight:
> I've been thinking something similar, so yes!

Cool. I'll put something together. I've also thought of a nice example, I think as to why they can be kept separate (`match_generator/tournament_type` and `tournament`), I'll write that in to a tutorial on "creating new tournaments".

> 
> (I also like 'Arena' as used in https://github.com/tscizzle/IPD_Morality - similar to our TournamentManager)

I like the name too (although `TournamentManager` does say what it does...). Happy either way.

drvinceknight:
Closed by #531

PR #531, opened 2016-04-12
`tournament_type` -> `match_generator` #526

drvinceknight:
#526 

I also make a very minor change to the results and plots classes: I allow them to deal with missing values: so cases where a player doesn't not actually play a match (they get a score of 0). This is based on the simple tournament I build in the tutorial, where match ups occur with a given probability.

### DIFF
--- a/axelrod/match.py
+++ b/axelrod/match.py
@@ -19,8 +19,6 @@ class Match(object):
                 The number of turns per match
         deterministic_cache : dictionary
             A cache of resulting actions for deterministic matches
-        cache_mutable : boolean
-            Whether the deterministic cache can be updated or not
         noise : float
             The probability that a player's intended action should be flipped
         """
--- a/axelrod/match_generator.py
+++ b/axelrod/match_generator.py
@@ -19,7 +19,8 @@ class MatchGenerator(object):
             A list of axelrod.Player objects
         turns : integer
             The number of turns per match
-        deterministic_cache : dictionary
+        deterministic_cache : an instance of axelrod.DeterministicCache
+        class
             A cache of resulting actions for deterministic matches
         """
         self.players = players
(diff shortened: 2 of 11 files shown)


axelrod/match.py
```diff
@@ -19,8 +19,6 @@ def __init__(self, players, turns, deterministic_cache=None, noise=0):
                 The number of turns per match
         deterministic_cache : dictionary
             A cache of resulting actions for deterministic matches
-        cache_mutable : boolean
-            Whether the deterministic cache can be updated or not
         noise : float
```
meatballs: Whoops. Thank you!

axelrod/match_generator.py
```diff
@@ -4,22 +4,23 @@
...
+        is in charge of playing the matches and collecting the results.
 
         Parameters
         ----------
         players : list
             A list of axelrod.Player objects
         turns : integer
             The number of turns per match
-        deterministic_cache : dictionary
+        deterministic_cache : an instance of axelrod.DeterministicCache
```
meatballs: And thanks once again - seems I missed a few things!

axelrod/match_generator.py
```diff
@@ -59,8 +60,6 @@ def build_matches(self, noise=0):
 
         Parameters
         ----------
-        cache_mutable : boolean
-            Whether the deterministic cache should be updated
```
meatballs: Hmmm.... You really are tidying up the mess I left behind

axelrod/match_generator.py
```diff
@@ -97,10 +96,10 @@ def __init__(self, players, prob_end, deterministic_cache):
             A list of axelrod.Player objects
         prob_end : float
             The probability that a turn of a Match is the last
-        deterministic_cache : dictionary
+        deterministic_cache : an instance of axelrod.DeterministicCache
```
meatballs: Well, this is embarrassing

meatballs:
👍

PR #534, opened 2016-04-13
Moran process

### LINKED ISSUE #224: Add Moran process
At this point it would be fairly easy (I think) to add in support for population dynamics, with the Moran process and/or the common imitative-switching process. If there is interest I can get something started.

These simulations take a much longer time of course but allow the empirical computation of e.g. fixation probabilities.

marcharper:
Closes  #224 

The Moran process is somewhat of a hybrid of a match and a tournament. It's sort of a `n >= 2` match, but also has several rounds. The process is widely studied in the literature.

### DIFF
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -7,6 +7,7 @@ import numpy as np
 from .deterministic_cache import DeterministicCache
 from .match import Match
 from .player import Player
+from .random_ import randrange
 
 
 def fitness_proportionate_selection(scores):
@@ -58,7 +59,7 @@ class MoranProcess(object):
         # Fitness proportionate selection
         j = fitness_proportionate_selection(scores)
         # Randomly remove a strategy
-        i = random.randrange(0, len(self.players))
+        i = randrange(0, len(self.players))
         # Replace player i with clone of player j
         self.players[i] = self.players[j].clone()
         self.populations.append(self.population_distribution())
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -11,7 +11,17 @@ from .random_ import randrange
 
 
 def fitness_proportionate_selection(scores):
-    """Randomly selects an individual proportionally to score."""
+    """Randomly selects an individual proportionally to score.
+
+    Parameters
+    ----------
+    scores: Any sequence of real numbers
+
+    Returns
+    -------
+    An index of the above list selected at random proportionally to the list
+    element divided by the total.
+    """
     csums = np.cumsum(scores)
     total = csums[-1]
     r = random.random() * total
@@ -29,6 +39,7 @@ class MoranProcess(object):
         self.populations = []
         self.populations.append(self.population_distribution())
         self.score_history = []
+        self.num_players = len(self.players)
 
     @property
     def _stochastic(self):
@@ -67,7 +78,7 @@ class MoranProcess(object):
     def _play_next_round(self):
         """Plays the next round of the process. Every player is paired up
         against every other player and the total scores are recorded."""
-        N = len(self.players)
+        N = self.num_players
         scores = [0] * N
         for i in range(N):
             for j in range(i + 1, N):
@@ -102,7 +113,7 @@ class MoranProcess(object):
 
     def play(self):
         """Play the process out to completion."""
-        while True: # O_o
+        while True:
             try:
                 self.__next__()
             except StopIteration:
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8 -*-
+from collections import Counter
+import random
+
+import numpy as np
+
+from .deterministic_cache import DeterministicCache
+from .match import Match
+from .player import Player
+
+
+def fitness_proportionate_selection(scores):
+    """Randomly selects an individual proportionally to score."""
+    csums = np.cumsum(scores)
+    total = csums[-1]
+    r = random.random() * total
+
+    for i, x in enumerate(csums):
+        if x >= r:
+            return i
+
+class MoranProcess(object):
+    def __init__(self, players, turns=100, noise=0):
+        self.turns = turns
+        self.noise = noise
+        self.players = list(players) # initial population
+        self.winner = None
+        self.populations = []
+        self.populations.append(self.population_distribution())
+        self.score_history = []
+
+    @property
+    def _stochastic(self):
+        """
+        A boolean to show whether a match between two players would be
+        stochastic
+        """
+        if self.noise:
+            return True
+        else:
+            return any(p.classifier['stochastic'] for p in self.players)
+
+    def __next__(self):
+        """Iterate the population:
+        - play the round's matches
+        - chooses a player proportionally to fitness (total score) to reproduce
+        - choose a player at random to be replaced
+        - update the population
+        """
+        # Check the exit condition, that all players are of the same type.
+        population = self.populations[-1]
+        classes = set(p.__class__ for p in self.players)
+        if len(classes) == 1:
+            self.winner = str(self.players[0])
+            raise StopIteration
+        scores = self._play_next_round()
+        # Update the population
+        # Fitness proportionate selection
+        j = fitness_proportionate_selection(scores)
+        # Randomly remove a strategy
+        i = random.randrange(0, len(self.players))
+        # Replace player i with clone of player j
+        self.players[i] = self.players[j].clone()
+        self.populations.append(self.population_distribution())
+
+    def _play_next_round(self):
+        """Plays the next round of the process. Every player is paired up
+        against every other player and the total scores are recorded."""
+        N = len(self.players)
+        scores = [0] * N
+        for i in range(N):
+            for j in range(i + 1, N):
+                player1 = self.players[i]
+                player2 = self.players[j]
+                player1.reset()
+                player2.reset()
+                match = Match((player1, player2), self.turns, noise=self.noise)
+                match.play()
+                match_scores = np.sum(match.scores(), axis=0) / float(self.turns)
+                scores[i] += match_scores[0]
+                scores[j] += match_scores[1]
+        self.score_history.append(scores)
+        return scores
+
+    def population_distribution(self):
+        """Returns the population distribution of the last iteration."""
+        player_names = [str(player) for player in self.players]
+        counter = Counter(player_names)
+        return counter
+
+    next = __next__  # Python 2
+
+    def __iter__(self):
+        return self
+
+    def reset(self):
+        """Reset the process to replay."""
+        self.winner = None
+        self.populations = [self.populations[0]]
+        self.score_history = []
+
+    def play(self):
+        """Play the process out to completion."""
+        while True: # O_o
+            try:
+                self.__next__()
+            except StopIteration:
+                break
+
+    def __len__(self):
+        return len(self.populations)
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -5,7 +5,7 @@ import random
 import numpy as np
 
 from .deterministic_cache import DeterministicCache
-from .match import Match
+from .match import Match, is_stochastic
 from .player import Player
 from .random_ import randrange
 
@@ -35,7 +35,7 @@ class MoranProcess(object):
         self.turns = turns
         self.noise = noise
         self.players = list(players) # initial population
-        self.winner = None
+        self.winning_strategy_name = None
         self.populations = []
         self.populations.append(self.population_distribution())
         self.score_history = []
@@ -47,10 +47,7 @@ class MoranProcess(object):
         A boolean to show whether a match between two players would be
         stochastic
         """
-        if self.noise:
-            return True
-        else:
-            return any(p.classifier['stochastic'] for p in self.players)
+        is_stochastic(self.players, self.noise)
 
     def __next__(self):
         """Iterate the population:
@@ -63,7 +60,7 @@ class MoranProcess(object):
         population = self.populations[-1]
         classes = set(p.__class__ for p in self.players)
         if len(classes) == 1:
-            self.winner = str(self.players[0])
+            self.winning_strategy_name = str(self.players[0])
             raise StopIteration
         scores = self._play_next_round()
         # Update the population
@@ -107,7 +104,7 @@ class MoranProcess(object):
 
     def reset(self):
         """Reset the process to replay."""
-        self.winner = None
+        self.winning_strategy_name = None
         self.populations = [self.populations[0]]
         self.score_history = []
 
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -47,7 +47,7 @@ class MoranProcess(object):
         A boolean to show whether a match between two players would be
         stochastic
         """
-        is_stochastic(self.players, self.noise)
+        return is_stochastic(self.players, self.noise)
 
     def __next__(self):
         """Iterate the population:
--- a/axelrod/random_.py
+++ b/axelrod/random_.py
@@ -14,7 +14,8 @@ def random_choice(p=0.5):
     return Actions.D
 
 def randrange(a, b):
-    """Python 2 / 3 compatible randrange."""
+    """Python 2 / 3 compatible randrange. Returns a random integer uniformly
+    between a and b (inclusive)"""
     c = b - a
     r = c * random.random()
     return a + int(r)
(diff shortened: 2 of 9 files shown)


axelrod/moran.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def reset(self):
+        """Reset the process to replay."""
+        self.winner = None
+        self.populations = [self.populations[0]]
+        self.score_history = []
+
+    def play(self):
+        """Play the process out to completion."""
+        while True: # O_o
```
drvinceknight: Why this inline comment?

axelrod/random_.py
```diff
@@ -12,3 +12,9 @@ def random_choice(p=0.5):
     if r < p:
         return Actions.C
     return Actions.D
+
+def randrange(a, b):
+    """Python 2 / 3 compatible randrange."""
```
drvinceknight: Could you write a bit more saying what this takes and returns (numpy syntax? If we're going to stick to that.).

docs/tutorials/getting_started/index.rst
```diff
@@ -16,3 +16,4 @@ Contents:
    noisy_tournaments.rst
    ecological_variant.rst
    command_line.rst
+   moran.rst
```
drvinceknight: This should come before `command_line.rst`.

docs/tutorials/getting_started/moran.rst
```diff
@@ -0,0 +1,49 @@
...
+    >>> players = [axl.Cooperator(), axl.Defector(),
+    ...               axl.TitForTat(), axl.Grudger()]
+    >>> mp = axl.MoranProcess(players)
+    >>> mp.play()
+    >>> mp.winner   # doctest: +SKIP
+
+    Defector
+
+You can access some attributes of the process, such as the number of rounds::
+    >>> len(mp)  # doctest: +SKIP
```
drvinceknight: Does this build ok? I think you need a space before the code indent?

docs/tutorials/getting_started/moran.rst
```diff
@@ -0,0 +1,49 @@
...
+    Counter({'Defector': 4})]
+
+The scores in each round::
+    >>> import pprint
+    >>> pprint.pprint(mp.score_history)  # doctest: +SKIP
+    [[6.0, 7.0800000000000001, 6.9900000000000002, 6.9900000000000002],
+    [6.0, 7.0800000000000001, 6.9900000000000002, 6.9900000000000002],
+    [3.0, 7.04, 7.04, 4.9800000000000004],
+    [3.04, 3.04, 3.04, 2.9699999999999998],
+    [3.04, 3.04, 3.04, 2.9699999999999998]]
```
drvinceknight: Could you add a sentence or two explaining the connection/difference to the ecological variant (what insights are gained from one over the other?)? Is the ecological variant 'kind off' a moran process in expectation for a given cost function?

docs/tutorials/getting_started/moran.rst
```diff
@@ -0,0 +1,49 @@
...
+    [Counter({'Defector': 1, 'Cooperator': 1, 'Grudger': 1, 'Tit For Tat': 1}),
+    Counter({'Defector': 1, 'Cooperator': 1, 'Grudger': 1, 'Tit For Tat': 1}),
+    Counter({'Defector': 2, 'Cooperator': 1, 'Grudger': 1}),
+    Counter({'Defector': 3, 'Grudger': 1}),
+    Counter({'Defector': 3, 'Grudger': 1}),
+    Counter({'Defector': 4})]
+
+The scores in each round::
+    >>> import pprint
+    >>> pprint.pprint(mp.score_history)  # doctest: +SKIP
```
drvinceknight: On `further_topics/probabilistict_end_tournaments.rst` we have:

```
>>> results = tournament.play()
>>> m = results.payoff_matrix
>>> for row in m:  
...     print([round(ele, 1) for ele in row]) # Rounding output  
```

Perhaps do that for consistency?

axelrod/random_.py
marcharper: I could, but it's the same as `randrange` in the `random` module. But we can't use that and have our tests work on both python 2 and 3.

axelrod/moran.py
marcharper: It's not necessary

axelrod/moran.py
```diff
@@ -0,0 +1,112 @@
...
+import numpy as np
+
+from .deterministic_cache import DeterministicCache
+from .match import Match
+from .player import Player
+from .random_ import randrange
+
+
+def fitness_proportionate_selection(scores):
+    """Randomly selects an individual proportionally to score."""
```
drvinceknight: Could you add a bit more to the docstring here?

axelrod/random_.py
drvinceknight: Yup, I know why it's here. Just asking for a 1 line saying: `returns a random float between a and b (inclusive)`.

docs/tutorials/getting_started/moran.rst
marcharper: They are pretty different. As far as I know the ecological variant used by Axelrod gets little attention while the Moran process is in thousands of papers. One is not a special case of the other AFAIK but I'll admit to not having studied the Axelrod version much.

axelrod/random_.py
drvinceknight: Sorry: `random integer`

axelrod/moran.py
```diff
@@ -0,0 +1,112 @@
...
+    csums = np.cumsum(scores)
+    total = csums[-1]
+    r = random.random() * total
+
+    for i, x in enumerate(csums):
+        if x >= r:
+            return i
```
drvinceknight: I think this can be done simpler with `numpy.random.choice` as in https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategy_transformers.py#L296

http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.random.choice.html

axelrod/moran.py
```diff
@@ -0,0 +1,112 @@
...
+        # Randomly remove a strategy
+        i = randrange(0, len(self.players))
+        # Replace player i with clone of player j
+        self.players[i] = self.players[j].clone()
+        self.populations.append(self.population_distribution())
+
+    def _play_next_round(self):
+        """Plays the next round of the process. Every player is paired up
+        against every other player and the total scores are recorded."""
+        N = len(self.players)
```
drvinceknight: Set a `self.num_players` on init so this doesn't get recounted every play? (minor)

axelrod/moran.py
marcharper: Sure, but I'm not sure what to add really beyond the name of the function. [Wikipedia](https://en.wikipedia.org/wiki/Fitness_proportionate_selection) has an extensive explanation.

It's sort of a weighted `randrange` -- if you put in a constant vector then it would be `randrange(0, len(scores))`.

axelrod/random_.py
marcharper: Done.

axelrod/moran.py
```diff
@@ -0,0 +1,112 @@
...
+            self.winner = str(self.players[0])
+            raise StopIteration
+        scores = self._play_next_round()
+        # Update the population
+        # Fitness proportionate selection
+        j = fitness_proportionate_selection(scores)
+        # Randomly remove a strategy
+        i = randrange(0, len(self.players))
+        # Replace player i with clone of player j
+        self.players[i] = self.players[j].clone()
```
drvinceknight: Is this going to be ok? Can we have a test with strategies we've had issues with before? In the same way we have a test for tournaments with strategies `Backstabber, ThueMorse` etc...

axelrod/moran.py
marcharper: In this version the population size is constant but there are many variations in which that's not the case. I can go ahead and change it if you'd like.... I'm guessing we'd have another class for variants.

axelrod/moran.py
drvinceknight: I wasn't overly clear:

```
Input:

    scores: list an order list of scores for each player

Returns:

    An index of the above list based on the proportional probability of each sore.
```

I realise this is trivial, just trying to future proof things as much as possible.

axelrod/moran.py
marcharper: Hmm.. I'm not sure which is more simple. `scores` will still have to be normalized, something like:
```
np.random.choice(len(scores), 1, p=np.array(scores) / sum(scores)
```
Technically scores could be all zeros in my implementation -- that shouldn't happen for IPD but it could for another game matrix. So we'd need a check or try/except block for that, and it might be better as is.

docs/tutorials/getting_started/moran.rst
drvinceknight: Perhaps you could put something like what you just wrote there?

```
Given the evolutionary basis of the Moran process it can be compared to the ecological variant <link to that doc>. Whilst that variant was used by Axelrod in his original works, the Moran process is now much wider studied in the literation.

Note that the Moran process limits to the replicator equation used by the ecological variant as the population size heads toward infinity.
```

Or something like that?

drvinceknight:
This looks really great @marcharper, I'm heading to class now. Going to take another look at it later.

axelrod/moran.py
marcharper: I thought about that -- I was hoping you could work some hypothesis magic :). I think we should be ok as far as strategy pairings go since I'm using the match class. So if they pass their clone tests and match pairings they should be fine here.

axelrod/moran.py
drvinceknight: No no, if it can vary than I'm with you! Keep it like this :) 👍

axelrod/moran.py
drvinceknight: > I thought about that -- I was hoping you could work some hypothesis magic :)

Sure thing: I'll send you a PR (probably in time for your tomorrow morning).

axelrod/moran.py
drvinceknight: > it might be better as is.

I think it is. Leave it.

axelrod/moran.py
```diff
@@ -0,0 +1,123 @@
...
+    @property
+    def _stochastic(self):
+        """
+        A boolean to show whether a match between two players would be
+        stochastic
+        """
+        if self.noise:
+            return True
+        else:
+            return any(p.classifier['stochastic'] for p in self.players)
```
meatballs: This method is almost identical to the one in the Match class. I think they both ought to be abstracted out somewhere.

docs/tutorials/getting_started/moran.rst
drvinceknight: That check box confused the hell out of me: like "I don't remember writing that..." but yeah I've seen the addition. Thanks.

axelrod/moran.py
drvinceknight: https://github.com/marcharper/Axelrod/pull/3

drvinceknight:
@marcharper the `winner` method returns the str of the `player`. This is in contrast to the `Match` class which returns the actual winning player instance. I understand why this is (this is an evolutionary process so an actual winner is not a particular player but a strategy). Could we change the name of this method to something that clarifies that:

-`winning_strategy_name`
- `winner_name`

? Not too sure what but I don't like that the type of `Match.winner()` and `Moran.winner()` is different.

drvinceknight:
One of my comments on the docs seemed to have been described as 'on an outdated diff':

On further_topics/probabilistict_end_tournaments.rst we have:

```
>>> results = tournament.play()
>>> m = results.payoff_matrix
>>> for row in m:  
...     print([round(ele, 1) for ele in row]) # Rounding output  
```

Perhaps do that in the docs here for consistency?

docs/tutorials/getting_started/moran.rst
drvinceknight: You don't. Have built these locally (see my PR on your branch).

axelrod/moran.py
marcharper: The standard is not to vary, so I changed it.

docs/tutorials/getting_started/moran.rst
```diff
@@ -0,0 +1,57 @@
...
+
+You can access some attributes of the process, such as the number of rounds::
+
+    >>> len(mp)  # doctest: +SKIP
+    6
+
+The sequence of populations::
+
+    >>> import pprint
+    >>> pprint.pprint(mp.populations)  # doctest: +SKIP
```
drvinceknight: Given the base print below, do we want the same here for consistency? (minor)

drvinceknight:
A part from my pprint/print comment inline my one remaining comment about the docs would be whether it should sit in `getting_started` or if it was better suited `further_topics` where for example the probabilistic ending tournament sits?

marcharper:
I think pprint does a better job in less code in this case.

drvinceknight:
> I think pprint does a better job in less code in this case.

Fine.

drvinceknight:
> whatever you think is best is fine.

Can you move it to `further_topics` then? It feels better there to me. I would put it between Morality Metrics and Probabilistic Ending Tournaments.

marcharper:
Ok, but I'd say the same logic applies to the ecological variant, no?

drvinceknight:
> Ok, but I'd say the same logic applies to the ecological variant, no?

I think that given that it comes out of the standard tournament/results class (it's a simple manipulation of the payoff matrix) I'd suggest that it should stay there. The moran process is as you said a different type of interaction all together...

marcharper:
I've moved it but I don't agree with your position. IMO the main value of the library comes from
* the library of strategies
* the test suite
* the classes that run matches and tournaments

The ecological variant is somewhat of a side curiosity that isn't going to excite anyone about the library; the Moran process and other population dynamics that are actively studied would actually be of interest to other researchers. Just because the ecological variant falls out of the results for the tournament code doesn't really make it a first class citizen. I could code the Moran process to use the tournament objects to run its matches and achieve the same effect.

drvinceknight:
> I've moved it but I don't agree with your position. IMO the main value of the library comes from

Thanks :) Have marked this with a positive review. 

> 
> the library of strategies
> the test suite
> the classes that run matches and tournaments

I completely agree. I would suggest a 4th strength is the accessibility of the library: a few people have been introduced to dev, game theory and open source through it (something I'm completely delighted about).

> The ecological variant is somewhat of a side curiosity that isn't going to excite anyone about the library; the Moran process and other population dynamics that are actively studied would actually be of interest to other researchers. Just because the ecological variant falls out of the results for the tournament code doesn't really make it a first class citizen. I could code the Moran process to use the tournament objects to run its matches and achieve the same effect.

I see what you're saying and understand/agree that from a high level game theoretic research point of view (which is certainly what the library is) that the moran process is far more of interest. However, I think the getting started tutorial should perhaps be aimed also at people who have very little knowledge of game theory. Explaining that the eco variant is based on the payoff matrix is simpler and more immediately accessible than a different interaction type: as you said the moran process is a hybrid of an n player match and a tournament.

I do not at all mean that the moran process is a second citizen (far from it, and apologies if that is how I came across with this awesome contribution of yours), I just don't feel it's what's needed in the introduction (which is not only aimed at experts in game theory but also complete novices). I am 98% sure that with some of my students for whom I've said "work through the intro", I would have probably said "except the moran process", only because it requires them to create a different class to the Tournament, Results, Plot (some of my students have come to the library with not the strongest code/game theory background). 

No doubt experts/researchers would perhaps even come looking for the moran process and I don't think it's too hard to find (just as for the probabilistic ending tournaments). 

Anyway, sorry, I realise you were trying to put this to bed (perhaps we can open issue if we should chat about this further: onboarding is important to get right). I just wanted to make sure you didn't think I thought this was a second class citizen. I really do think it's awesome and am excited to learn more about it. The fact that it's heavily used in the literature is awesome. :) 👍

drvinceknight:
> Tournament, Results, Plot (some of my students have come to the library with not the strongest code/game theory background).

Having written that and realised that we need the `Eco` class...

Perhaps moving the ecological variant to the Further topics is the way to go. I'll leave it up to you :), either of these is fine by me:

- As is.
- Both ecological and Moran in further (perhaps this further simplifies the intro which is even better).

marcharper:
Let's discuss in a future PR / on gitter. I think this one is good to go in pending a second review.

langner:
Good stuff!

My opinion about the docs: create something like `beyond_basics.rst` which mentions both moran and eco and links to pages that expound on those topics. I would also put noisy tournaments in that bucket. In my mind, `getting started` should contain only the absolute basics; it should be geared towards people that have just learned about IPD yesterday, maybe this is also the first they've seen the words "game" and "theory" next to each other.

drvinceknight:
@meatballs this is good to go I believe (assuming you're happy). The conversation about the docs has been moved to #536 :)

PR #537, opened 2016-04-14
Move Setting of attributes from tournament to match

meatballs:
#493 

This PR moves the responsibility for passing match attributes (e.g. length) from the tournament class to the match itself. (They were previously known as tournament_attributes and have also been renamed).

It doesn't add any of the potential new functionality discussed in #493.

### DIFF
--- a/axelrod/match.py
+++ b/axelrod/match.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from axelrod import Actions
+from axelrod import Actions, Game
 from .deterministic_cache import DeterministicCache
 
 import axelrod.interaction_utils as iu
@@ -12,9 +12,11 @@ def is_stochastic(players, noise):
     of the players involved is stochastic."""
     return (noise or any(p.classifier['stochastic'] for p in players))
 
+
 class Match(object):
 
-    def __init__(self, players, turns, deterministic_cache=None, noise=0):
+    def __init__(
+        self, players, turns, game=None, deterministic_cache=None, noise=0):
         """
         Parameters
         ----------
@@ -22,6 +24,8 @@ class Match(object):
             A pair of axelrod.Player objects
         turns : integer
                 The number of turns per match
+        game : axelrod.Game
+            The game object used to score the match
         deterministic_cache : dictionary
             A cache of resulting actions for deterministic matches
         noise : float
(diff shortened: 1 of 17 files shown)


drvinceknight:
This looks great @meatballs, I think on #493 one of the things that we talked about was:

> a boolean that indicates whether players are aware of the match parameters.

I think this could be done by passing a dictionary to a match mapping attribute names to booleans. By default:

```
information={"length"=True, "game"=True, "noise"=True}
```

Then here we could have something like:

```
@players.setter
def players(self, players):
      """Ensure that players are passed the match attributes"""
      newplayers = []

     attributes = self.get_information(self.information)
     for player in players:
          player.set_match_attributes(
               length=attributes["turns"],
               game=attributes["game"],
               noise=attributes["noise"])
           newplayers.append(player)
```

The alternative to `True` could be what we actually want the players to think the match is. So for example we could tell the players they're playing 400 turns but are in fact player 200 etc

```
information={"length"="inf", "game"=True, "noise"=.5}
```

So that would be where the players think they're playing for an infinite amount of time, there is a noise of .5 etc...

So the `attributes = self.get_information(self.information)` part, that method would just go through and create a dictionary of what information the players are given. (I'm sure there are perhaps better names for these). That could perhaps be done on init...

axelrod/match.py
```diff
@@ -12,46 +12,70 @@ def is_stochastic(players, noise):
...
+            The game object used to score the match
         deterministic_cache : dictionary
             A cache of resulting actions for deterministic matches
         noise : float
             The probability that a player's intended action should be flipped
         """
         self.result = []
-        self.players = list(players)
+        self.turns = turns
         self._classes = (players[0].__class__, players[1].__class__)
```
drvinceknight: What is this for? (I know you haven't added it with this: just curious.)

meatballs:
> I think on #493 one of the things that we talked about was: .......

@drvinceknight Yes - we mentioned several other ideas on  #493 but I prefer to break things down - especially when it comes to refactoring existing functionality vs adding new functionality.

I'd rather keep this PR purely for the former, get it right and get it in before moving on to adding anything new.

axelrod/match.py
meatballs: I wasn't sure myself, so I left it well alone!!

I'm not certain when the players tuple was converted to a list or why.

drvinceknight: Cool... :)

drvinceknight:
> I'd rather keep this PR purely for the former, get it right and get it in before moving on to adding anything new.

Fair enough, I thought that was a pretty minor part of this but either way this is an improvement so it's a 👍 from me.

drvinceknight:
(Gosh, just re read the conversation on #493: that was a doozy! Good call on doing this step separately first.)

meatballs:
We could break those new ideas out of #493 and into separate issues

drvinceknight:
Good call. We should close #493 once this is in and then open other issues as necessary.

Issue #536, opened 2016-04-14
Docs Discussion


marcharper:
@langner @drvinceknight | follow up from #534

We've been chatting in gitter about updating the docs... I'm inclined to agree now that the first page docs ought to be as easy as possible to digest for newcomers. In that regard I suggest that we focus on matches before anything else. IMO they are the easiest chunk to understand, and we could even point at a "gentle intro" jupyter notebook (perhaps a simplified version of [this one](https://github.com/Axelrod-Python/Axelrod-notebooks/blob/master/Matches.ipynb)).

We should also consider  a "for the experts" or "advanced topics in detail" page.

drvinceknight:
For reference I went and grabbed this issue: #299 which is what started the first revamp of the docs. The very first set of documentation was terrible, it was a long "story". Following #299 we moved towards a more modular setup which I think has worked well so far and unless I'm misunderstanding is not what we're discussing? I think it's a question of classification that we need to agree on?

> I'm inclined to agree now that the first page docs ought to be as easy as possible to digest for newcomers. In that regard I suggest that we focus on matches before anything else. IMO they are the easiest chunk to understand, and we could even point at a "gentle intro" jupyter notebook (perhaps a simplified version of this one).

This sounds like a great idea. We could have an extra  "creating a match" here: http://axelrod.readthedocs.org/en/latest/tutorials/getting_started/index.html (so that section would have 3 sub sections: Installing, creating a match, creating a tournament.).

and point towards:

http://axelrod.readthedocs.org/en/latest/reference/description.html

I don't think it would need a lot for this introductory section. We want to keep things light so people can get started. 

> 
> We should also consider a "for the experts" or "advanced topics in detail" page.

I agree. Here is where the classification/naming/labelling/description is important. I agree we should have a better landing page that describes what the "sections" enable you to do. I think describing who the sections are for is not ideal as self classification can go wrong (expert game theorist, never used python etc...). So I would prefer labelling what the section enable you to do.


Our current landing page http://axelrod.readthedocs.org/en/latest/index.html, says nothing.

My suggested labelling of the sections would be something more verbose:

1. "Getting started" -> "New to Python and/or Game Theory"
2. "Further topics" -> "Contemporary research topics"
3. "Advanced topics" -> "Further tools in the library"

(Based on that, I would actually suggest that the Moran process would go to 2. and the eco would stay in 1. I think we could also move `Classification of strategies` and `Reading and writing interactions to file` to 3.)

[...]

PR #539, opened 2016-04-15
Docs reorganisation

drvinceknight:
#536 

Main:

- Relabelling of tutorial sections and some corresponding reorganisation.
- Landing page with 3 code snippets showing Matches, Tournaments and Moran processes.
- Initial tutorials are now: installation, building a match and then tournament. This replaces the current getting started section: http://axelrod.readthedocs.org/en/latest/tutorials/getting_started/getting_started.html which had installation and a tournament. I don't think this is necessary given the landing page.

Minor:

- Removing the ecological example from prob end.
- Adding an example to the new tournament docs.
- Minor rst syntax fixing and new images (boxplots replaced with violon plots as matplotlib now can handle violon for some of the ones that had boxplots kicking in).
- Changing strategies -> players in various parts (this is clearer, a strategy instance is a player)


marcharper:
Very nice :) Thanks for taking care of this!

PR #540, opened 2016-04-16
Adds match length to deterministic cache key

### LINKED ISSUE #533: Cache key to include Match length
Following #527 we have a nice cache class. This currently is designed for tournaments of constant match length and maps: pairs of players (when both players are not stochastic) to outcomes.

```
(axelrod.Cooperator, axelrod.Alternator): [('C', 'C'), ('C', 'D'), ('C', 'C')]
(axelrod.TitForTat, axelrod.Alternator): [('C', 'C'), ('C', 'D'), ('D', 'C')]
```

A small change would allow us to map pairs of players and match lengths to outcomes:

```
(axelrod.Cooperator, axelrod.Alternator, 3): [('C', 'C'), ('C', 'D'), ('C', 'C')]
(axelrod.TitForTat, axelrod.Alternator, 5): [('C', 'C'), ('C', 'D'), ('D', 'C'), ('C', 'D'), ('D', 'C')]
```

The main goal of this would be to continue the move away from assumptions about the tournament structure which will continue to make new tournaments easily. For example it would be easy to create a tournament where half matches are a single turn and the other half are a full 200 (or whatever).

Basically, whenever a match between deterministic players of any length happens, the cache could be used (or updated). The structure of the tournament would be irrelevant.

meatballs:
Closes #533

### DIFF
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -54,7 +54,7 @@ class DeterministicCache(UserDict):
 
         if not self._is_valid_key(key):
             raise ValueError(
-                'Key must be a pair of deterministic axelrod Player classes')
+                'Key must be a tuple of 2 deterministic axelrod Player classes and an integer')
 
         if not self._is_valid_value(value):
             raise ValueError(
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -43,7 +43,6 @@ class DeterministicCache(UserDict):
         """
         UserDict.__init__(self)
         self.mutable = True
-        self.turns = None
         if file_name is not None:
             self.load(file_name)
 
@@ -63,9 +62,6 @@ class DeterministicCache(UserDict):
 
         UserDict.__setitem__(self, key, value)
 
-        if self.turns is None:
-            self.turns = len(value)
-
     def _is_valid_key(self, key):
         """Validate a proposed dictionary key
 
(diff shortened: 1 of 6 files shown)


axelrod/deterministic_cache.py
```diff
@@ -81,13 +77,18 @@ def _is_valid_key(self, key):
...
             return False
 
-        # The tuple should be a pair
-        if len(key) != 2:
+        # The tuple should be a triplet
+        if len(key) != 3:
             return False
 
-        # Each item in the pair should be a subclass of axelrod.Player
+        # The tripley should be a pair of axelrod.Player sublclasses and an
```
drvinceknight: typo: "triplet"

axelrod/match.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, players, turns, game=None, deterministic_cache=None,
         """
         self.result = []
         self.turns = turns
-        self._classes = (players[0].__class__, players[1].__class__)
+        self._cache_key = (players[0].__class__, players[1].__class__, turns)
```
drvinceknight: Really like this rename.

drvinceknight:
This looks awesome @meatballs. Will be able to use the cache for the prob end tournament now :)

Suggestion: instead of importing from future to deal with the py2 division could you just force float division in relevant places in the tests:

```
(3/float(turns), 2/float(turns))
```

Mainly because this is how it's done in a lot of other spots and also it's a bit more verbose?

meatballs:
> Suggestion: instead of importing from future to deal with the py2 division could you just force float division in relevant places in the tests:

Yep, that's a better plan. I'll do that now

drvinceknight:
Assuming travis is content, this is a big 👍 from me. Thanks @meatballs!

meatballs:
> Assuming travis is content...

Always a big assumption when it comes to getting any my code working on python 2 and 3!

PR #544, opened 2016-04-16
Remove line to set cache mutabilty to false

### LINKED ISSUE #543: Add cache to prob end tournament
Following #540 the cache can be used for more or less every match that has the potential to be repeated (regardless of structure of the tournament).

Thus it can be used in the prob end tournament and perhaps could also be used 'more' elsewhere (I'm not completely familiar with the defaults, it might make sense to tweak those?).

meatballs:
closes #543


drvinceknight:
Heh, Heh, that was easy... 👍

PR #546, opened 2016-04-16
Add deterministic cache to Moran process an a bug fix

marcharper:
Also fix a bug in the reset function.

### DIFF
--- a/axelrod/tests/unit/test_moran.py
+++ b/axelrod/tests/unit/test_moran.py
@@ -100,3 +100,9 @@ class TestMoranProcess(unittest.TestCase):
         # Check that players reset
         for player in mp.players:
             self.assertEqual(len(player.history), 0)
+
+    def test_cache(self):
+        p1, p2 = axelrod.Cooperator(), axelrod.Defector()
+        mp = MoranProcess((p1, p2))
+        mp.play()
+        self.assertEqual(len(mp.deterministic_cache), 1)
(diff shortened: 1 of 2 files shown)


drvinceknight:
Can you add a test (at least to that the cache initialises properly)?

(Really going offline for the night now :))

axelrod/tests/unit/test_moran.py
```diff
@@ -98,5 +98,11 @@ def test_reset(self):
         self.assertEqual(mp.winning_strategy_name, None)
         self.assertEqual(mp.score_history, [])
         # Check that players reset
-        for player in mp.players:
-            self.assertEqual(len(player.history), 0)
+        for player, intial_player in zip(mp.players, mp.initial_players):
+            self.assertEqual(str(player), str(intial_player))
+
+    def test_cache(self):
```
drvinceknight: Could you test both (passing a cache and it creating a new one)?

```
       p1, p2 = axelrod.Cooperator(), axelrod.Defector()
       mp = MoranProcess((p1, p2))
       mp.play()
       self.assertEqual(len(mp.deterministic_cache), 1)

       # Check that can pass a pre built cache
       cache = axelrod.DeterministicCache()
       mp = MoranProcess((p1, p2), deterministic_cache=cache)
       self.assertEqual(cache, mp.deterministic_cache)
```

drvinceknight:
A part from my suggestion for that extra little test this is a 👍  from me.

drvinceknight:
Have opened #548 **but** that sits on top of this.

PR #548, opened 2016-04-17
Adding tutorial about cache and also ability to pass a cache directly to the tournament

### LINKED ISSUE #547: Write tutorial for cache
To go in the "Further capabilities in the library" section.

drvinceknight:
Closes #547 

**This is on top of #546, so that should be merged/reviewed first. (Merging this will merge #546)**

Note that to make things in line with the match and the Moran process I've added the ability to pass a cache directly to the `Tournament` class. So the following works (as shown in the tutorial):

```
>>> cache = axl.DeterministicCache(filename)  # Read in some prebuilt cache
>>> players = [axl.GoByMajority(), axl.Alternator(), axl.Cooperator()]
>>> tournament = axl.Tournament(players, turns=200, repetitions=5, deterministic_cache=cache)
>>> results = tournament.play()
```

I haven't gotten rid of the `prebuilt_cache` argument as I didn't want to get in to the `TournamentManager` (I think that's where it's used). That could potentially be made simpler with this though?

### DIFF
--- a/docs/tutorials/advanced/using_the_cache.rst
+++ b/docs/tutorials/advanced/using_the_cache.rst
@@ -8,24 +8,34 @@ different result, for deterministic strategies, when there is no noise there is
 no need to re run the match. The library has a :code:`DeterministicCache` class
 that allows us to quickly replay matches.
 
-Here is how to create a new empty cache::
-
-    >>> import axelrod as axl
-    >>> cache = axl.DeterministicCache()
-    >>> len(cache)
-    0
 
 Caching a Match
 ---------------
 
-We can use this cache in a Match, let us also time the execution::
+To illustrate this, let us time the play of a match **without** a cache::
 
+    >>> import axelrod as axl
     >>> import timeit
     >>> def run_match():
     ...     p1, p2 = axl.GoByMajority(), axl.Alternator()
-    ...     match = axl.Match((p1, p2), deterministic_cache=cache, turns=200)
+    ...     match = axl.Match((p1, p2), turns=200)
     ...     return match.play()
-    >>> time_with_empty_cache = timeit.timeit(run_match, number=500)
+    >>> time_with_no_cache = timeit.timeit(run_match, number=500)
+    >>> time_with_no_cache  # doctest: +SKIP
+    2.2295279502868652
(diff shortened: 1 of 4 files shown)


docs/tutorials/advanced/using_the_cache.rst
```diff
@@ -0,0 +1,99 @@
+.. _using-the-cache:
+
+Using the cache
+===============
+
+Whilst for stochastic strategies every repetition of a Match will give a
+different result. For deterministic strategies, when there is no noise there is
```
meatballs: Whilst for stochastic strategies, every repetition of a Match will give a different result, for deterministic strategies, .....

docs/tutorials/advanced/using_the_cache.rst
```diff
@@ -0,0 +1,99 @@
...
+    >>> time_with_empty_cache = timeit.timeit(run_match, number=500)
+
+We can take a look at the cache::
+
+    >>> cache  # doctest: +ELLIPSIS
+    {(<class 'axelrod.strategies.gobymajority.GoByMajority'>, <class 'axelrod.strategies.alternator.Alternator'>, 200): [('C', 'C'), ..., ('C', 'D')]}
+    >>> len(cache)
+    1
+
+Note, that is maps triplets of players and match lengths to interactions.
```
meatballs: typo here, I think

meatballs: Might be better as:

Note: This maps a triplet of 2 player classes and the match length to the resulting interactions.

meatballs:
Just minor wording comments on the docs from me. Everything else looks good.

drvinceknight:
> Just minor wording comments on the docs from me. Everything else looks good.

Think I've picked them all up. Thanks :) 👍

drvinceknight:
Just walking the puppy and figured out something that's not quite right with this. Will be a two second fix.

drvinceknight:
I wasn't timing the run without the cache properly (`timeit` was in fact just running that once without a cache and then every other time with a cache). Have fixed this now (and it makes for an even more impressive time difference :)).

PR #556, opened 2016-04-19
Prob End Tournament can now take a cache

### LINKED ISSUE #555: Probabilistic ending tournament cannot take a cache.
In #548 I added the ability for the `Tournament` class to take a cache. I forgot to add it to the `ProbEndTournament`. PR incoming (simple fix).

drvinceknight:
Closes #555

Also removed test and method that can now just inherit from base class.

### DIFF
--- a/axelrod/tournament.py
+++ b/axelrod/tournament.py
@@ -310,8 +310,8 @@ class ProbEndTournament(Tournament):
 
     def __init__(self, players, match_generator=ProbEndRoundRobinMatches,
                  name='axelrod', game=None, prob_end=.5, repetitions=10,
-                 processes=None, prebuilt_cache=False, noise=0,
-                 with_morality=True):
+                 processes=None, deterministic_cache=None, prebuilt_cache=False,
+                 noise=0, with_morality=True):
         """
         Parameters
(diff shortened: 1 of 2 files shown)


axelrod/tournament.py
```diff
@@ -310,8 +310,8 @@ class ProbEndTournament(Tournament):
 
     def __init__(self, players, match_generator=ProbEndRoundRobinMatches,
                  name='axelrod', game=None, prob_end=.5, repetitions=10,
-                 processes=None, prebuilt_cache=False, noise=0,
-                 with_morality=True):
+                 processes=None, deterministic_cache=None, prebuilt_cache=False,
```
marcharper: Is it necessary to have both `deterministic_cache` and also `prebuilt_cache`?

drvinceknight: At the moment: yes, the tournament manager and factory classes I believe
make use of `prebuilt_cache`. I'm being a bit lazy with it and just not
touching those two classes for now... (I see the plan being for
`prebuilt_cache` to go eventually, it's superceded by the ability to just
give a cache or not).

Issue #552, opened 2016-04-18
Refactor Factory Classes?


marcharper:
Does anyone have thoughts about replacing the `TournamentManagerFactory` classes (and maybe the `TournamentManagers`) with scripts that directly call the `Tournament` objects? There are at least a few benefits:
* The library itself serves as examples of how to run tournaments in a way more typical of an external user
* Simplifies the library by removing some classes. The factory classes are only used internally and are not documented for external usage
* It's easier to follow the logic of what's going on
* It's arguably more pythonic

Of course there is significant potential downside:
* [if it ain't broke don't fix it](https://en.wiktionary.org/wiki/if_it_ain't_broke,_don't_fix_it)

drvinceknight:
> Does anyone have thoughts about replacing the TournamentManagerFactory classes (and maybe the TournamentManagers) with scripts that directly call the Tournament objects? 

Yes please.

> It's easier to follow the logic of what's going on

I have to admit I found it quite confusing going through the `TournamentManager` to the `TournamentManagerFactory` when I was adding prob end to the `run_axelrod` script. Given that the tournament api is quite simple now I think this level of abstraction is not needed but perhaps I'm missing something.

For info, when it came to running the tournament on our university cluster, as I had to do it in two intermediate steps (write data, then analyse), to make sure I knew what was happening this is what I had, (repeated for each of the 3 tournaments):

- This runs the tournament: https://github.com/Axelrod-Python/tournament/blob/gh-pages/run_noisy.py
- This analyses it: https://github.com/Axelrod-Python/tournament/blob/gh-pages/analyse_noisy.py

I realise there's repetition for the plotting but I think we could make `run_axelrod` much simpler by simply having a `plot_all` function that takes a `result` set and then 1 function for each tournament (`run_basic_strategies`, `run_strategies`,...) type we wanted (that would basically be 1 line of code), then the arguments being read in could just call those functions. I realise there will be some minor duplication of code but I think that the gain in readability is worth it.

These functions wouldn't need to be part of the library themselves, they just make use of the library. We could put them at the top of `run_axelrod` which I think would make it more readable.

Perhaps a 1st step would be to refactor `run_axelrod` (given all the dependencies, we could also now use docopt perhaps?), if we have all the functionality without the need for the `Manager` and `Factory`, classes we could remove them?

meatballs:
[...]

PR #557, opened 2016-04-19
Re write of run_axelrod.

drvinceknight:
Here is my attempt at addressing the discussion of #552. **Note this is branched from the bug fix PR of #556.**

This does not use the Manager or ManagerFactory class, no doubt some
duplication can be reduced but perhaps this is an OK balance? 

The only functionality lost from the current run_axelrod (I believe) is
the logger. I'm not entirely sure I understand the need for it (so
please correct me, perhaps the logger object itself can be lifted in to
here if it's necessary?).

This also adds docopt as a dependency.

### DIFF
--- a/run_axelrod
+++ b/run_axelrod
@@ -1,135 +1,147 @@
 #!/usr/bin/env python
+"""Run axelrod
 
-"""A script to run the Axelrod tournament.
+Usage:
+    run_axelrod [-h] [-t TURNS] [-e PROBEND] [-r REPETITIONS] [-n NOISE] [-o
+    OUTPUTDIRECTORY] [-i IMAGE-FORMAT] [-c CACHE] [-p PROCESSES] [--rc] [--xb] [--xo] [--xc] [--xa]
 
-The code for strategies is present in `axelrod/strategies`.
+Options:
+    -h --help                    show a help message
+    -t TURNS                     turns per match [default: 200]
+    -e PROBEND                   probability of a match ending (if provided, overrides turns) [default: ""]
+    -r REPETITIONS               repetitions of each tournament [default: 100]
+    -n NOISE                     noise level [default: 0]
+    -o OUTPUT-DIRECTORY          output directory [default: None]
+    -i IMAGE-FORMAT              image format for output, e.g. svg, jpg, png [default: svg]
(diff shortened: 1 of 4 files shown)


run_axelrod
```diff
@@ -1,135 +1,147 @@
...
+
+    # Parse arguments
+    turns = int(arguments['-t'])
+    repetitions = int(arguments['-r'])
+    prob_end = float(arguments['-e'])
+    noise = float(arguments['-n'])
+    processes = eval(arguments['-p'])
+
+    output_directory = arguments['-o']
+    if output_directory == "None":
```
marcharper: I'm guessing this is a `doctopt` thing, to return `"None"` instead of `None`?

run_axelrod
```diff
@@ -1,135 +1,147 @@
...
+        output_directory = None
+    cache_file = arguments['-c']
+    if cache_file == "None":
+        cache_file = None
+
+    image_format = arguments['-i']
+    rebuild_cache = arguments['--rc']
+    exclude_basic = arguments['--xb']
```
marcharper: Perhaps this is not the time, but I'd prefer the API to be what tournaments *to* run rather than what tournaments *not to* run.

drvinceknight: Fine by me. So default would be to run nothing?

run_axelrod
drvinceknight: Yeah, this felt a bit icky, could get around it by changing the defaults to something like ` "" ` but that felt ickier...

run_axelrod
marcharper: I guess it depends on the purpose. Maybe running the basic tournament is fine, and it's quick anyway. But the other tournaments are much longer.

It's just easier to understand if you have to tell it what you want it to do. Otherwise you have to know everything it wants to do by default and then scale it back.

drvinceknight: Yup, I'm happy with do nothing by default myself (otherwise it's creating
files you might not want) and then just add in the options for what you
want.

I think the exclude made sense back in the day when this was a quick one
stop 'update everything' tool. Now that we have so many strategies and
counting perhaps it's not as apt.

drvinceknight:
Closing this based on latest on #552

PR #559, opened 2016-04-20
Removing run_axelrod and factor/manager classes

drvinceknight:
Closes 552

- Remove run_axelrod.
- Removing manager, managerfactory and utils.
- Removing corresponding tests.
- Removing prebuilt cache argument from tournament.
- Removing tutorial on command line.
- Minor modifications to README.rst.


drvinceknight:
Have opened: https://github.com/Axelrod-Python/tournament/issues/33 on the tournament repo to keep track of this.

marcharper:
I'm happy with it. The drop in coverage seems mostly due to the plots not being made, so maybe we can add back some tests. I know it's hard to test plots other than just making them and checking that it doesn't crash.

It may make sense now to move the plotting functionality into member functions of ResultsSet, but that can be another discussion.

drvinceknight:
Don't merge, I'll add in those tests for the plots :)

PR #560, opened 2016-04-21
Writing tournaments as each set of interactions occurs (currently: only at the end of all repetitions)

drvinceknight:
This aims to further address #520.

I transpose the current data file format (rows now correspond to repetitions and columns to matches).


drvinceknight:
On #520 @marcharper has had a better idea.

PR #567, opened 2016-04-29
Add self.init_args to Lookerup

marcharper:


marcharper:
Travis should fail, second commit will add `@init_args` to the child class to fix.

drvinceknight:
Is there a test of some sort we can include that checks that it's being done right? Perhaps that the player clones correctly?

drvinceknight:
It would be good to confirm what you said on gitter: that it's not currently being cloned correctly.

drvinceknight:
With and without this extra line I get (py3):

```
>>> import axelrod as axl
>>> lookup_table = {('', 'C', 'D') : "D", ('', 'D', 'D') : "D", ('', 'C', 'C') : "C", ('', 'D', 'C') : "C",}
>>> p = axl.LookerUp(lookup_table)
>>> p.init_args == p.clone().init_args
True
>>> p = axl.LookerUp()
>>> p.init_args == p.clone().init_args
True
```

Is that not what this should be addressing (ie True should only be happening for the case with your fix?).

Similarly, with or without:

```
>>> p = axl.EvolvedLookerUp()
>>> p.init_args == p.clone().init_args
True
```

marcharper:
We need to check that a non-default table is being cloned correctly.

marcharper:
I'm not sure what's going on then, will take a close look tonight.

drvinceknight:
FYI Just in case, I just retried the above (both on master and with this branch) with the following table:

```
lookup_table = {('', 'C', 'D') : "D", ('', 'D', 'D') : "D", ('', 'C', 'C') : "D", ('', 'D', 'C') : "D",}
```

And got the same results (I tried that as I saw the message you deleted via email about trying with a non default table, the other one is the default one for the lookerup).

PR #572, opened 2016-05-07
Parallel Matches 

marcharper:
Re:  #520. We'll still need to add a `BigResultsSet` and to consider breaking out the parallel aspects of the tournament class, but this PR already makes a lot of progress on the memory-consumption front.

Some notes:
* Memory consumption is vastly reduced while running, however:
* Large tournaments will still use a lot of memory loading in the resultset~~, which happens automatically at the end~~
* All results are written to disk (tempfile or supplied filename), tournament no longer has an `interactions` attribute
* Caching is done on a match-by-match basis and no longer loaded from a file

Could use some help with the last few tests @drvinceknight .


drvinceknight:
Sweet, great job getting this done I'll take a look through it tomorrow! 👍  👏 

> Could use some help with the last few tests @drvinceknight .

Will sort those out: no problemo :)

marcharper:
Looks like I fixed most of them, but we could use some more to replace a few that I removed. Also `test_namespace` in `test_strategy_transformers.py` fails locally but passes on travis.

marcharper:
Just did a run with:
```
python3 run_axelrod --rc -p 4 -o assets -r 40 -t 200 --xc --xb --xa
```
Memory usage per thread was < 70 MB for 16 minutes until the loading of the results set, which got up to about 4Gb and took about 6 minutes.

```
Running tournament with 113 strategies

real	22m9.149s
user	72m53.744s
sys	0m6.780s
```

drvinceknight:
> Not sure why but test_namespace in test_strategy_transformers.py fails locally yet passes on travis.

Could you check that this happens for you when running `./test` but not when running them separately `python3 -m unittest discover axelrod.tests`? That's the case for me, perhaps that gives a clue as to why it fails?

axelrod/tournament.py
```diff
@@ -48,19 +48,23 @@ def __init__(self, players, match_generator=RoundRobinMatches,
...
-    def play(self, filename=None):
+        if not filename:
+            self.outputfile = NamedTemporaryFile(mode='w')
+            filename = self.outputfile.name
+        else:
+            self.outputfile = open(filename, 'w')
+        self.writer = csv.writer(self.outputfile)
+        self.filename = filename
+
+    def play(self):
```
drvinceknight: Perhaps:


```
def play(self, build_results=True):
    ...
```


So that you can pass `build_results=False` then you don't build the results set (an error/warning would be raised if no filename is given and `build_results` is set to `True`)? Just for the case that you'd want the data but not actually want to analyse it (for example my final year student this year was analysing the results with R so there would not have been the need for the higher memory consumption for the Results set).

drvinceknight:
This tutorial needs to be rewritten: http://axelrod.readthedocs.io/en/latest/tutorials/advanced/making_tournaments.html 

The doctests for it passed as the new methods created there aren't actually the ones we used. I'm putting together a PR to your branch for it.

axelrod/match_generator.py
```diff
@@ -56,39 +60,48 @@ class RoundRobinMatches(MatchGenerator):
...
+                match_params = self.build_single_match_params(noise)
+                index_pair = (player1_index, player2_index)
+                yield (index_pair, match_params, self.repetitions)
 
-    def build_single_match(self, pair, noise=0):
+    def build_single_match_params(self, noise=0):
         """Create a single match for a given pair"""
-        return Match(
-            pair, self.turns, self.game, self.deterministic_cache, noise)
+        return (self.turns, self.game, None, noise)
```
drvinceknight: Can we get rid of the `None` here? Quite confusing without the diff.

axelrod/match_generator.py
```diff
@@ -151,3 +162,8 @@ def sample_length(self, prob_end):
             return float("inf")
         except ValueError:
             return 1
+
+    def estimated_size(self):
```
drvinceknight: Same comment as above.

axelrod/match_generator.py
```diff
@@ -56,39 +60,48 @@ class RoundRobinMatches(MatchGenerator):
...
+    def build_single_match_params(self, noise=0):
         """Create a single match for a given pair"""
-        return Match(
-            pair, self.turns, self.game, self.deterministic_cache, noise)
+        return (self.turns, self.game, None, noise)
+
+    def __len__(self):
+        """Calculates the number of matches."""
+        n = len(self.players)
+        num_matches = int(n * (n + 1) // 2 + n)
```
drvinceknight: Is this correct? In 6ee95f4 I add a test that compares `len(match_generator)` to `len(list(match_generator.build_match_chunks()))`. I would have expected these to be equal? What are you counting here?

drvinceknight:
I've opened #574 which has the tutorial re write but also some minor things that just seemed easier to code then to explain (feel free to close PR and cherry pick what you want @marcharper).

axelrod/tests/unit/test_match_generator.py
```diff
@@ -41,71 +45,75 @@ class TestRoundRobin(unittest.TestCase):
...
+        match = axelrod.Match(*match_params)
+        self.assertIsInstance(match, axelrod.Match)
+        self.assertEqual(len(match), test_turns)
+
+    @given(repetitions=integers(min_value=1, max_value=test_repetitions))
+    @example(repetitions=test_repetitions)
+    def test_build_match_chunks(self, repetitions):
+        rr = axelrod.RoundRobinMatches(self.players, test_turns, test_game, repetitions)
+        chunks = list(rr.build_match_chunks())
+        #match_definitions = [tuple(list(index_pair) + [repetitions]) for (index_pair, match_params, repetitions) in chunks]
```
drvinceknight: Not necessary.

axelrod/match_generator.py
```diff
@@ -106,14 +119,12 @@ def __init__(self, players, prob_end, game, deterministic_cache):
...
         """
         super(ProbEndRoundRobinMatches, self).__init__(
-            players, turns=float("inf"), game=game,
-            deterministic_cache=deterministic_cache)
+            players, turns=float("inf"), game=game, repetitions=repetitions)
         self.prob_end = prob_end
 
-    def build_single_match(self, pair, noise=0):
+    def build_single_match_params(self, noise=0):
         """Create a single match for a given pair"""
```
drvinceknight: """Create a single set of match parameters"""

axelrod/result_set.py
```diff
@@ -651,16 +651,8 @@ def csv(self):
...
-
-    Initialised by a csv file of the format:
-
-
-    [p1index, p2index, p1name, p2name, p1rep1ac1p2rep1ac1p1rep1ac2p2rep1ac2,
-    ...]
-    [0, 1, Defector, Cooperator, DCDCDC, DCDCDC, DCDCDC,...]
-    [0, 2, Defector, Alternator, DCDDDC, DCDDDC, DCDDDC,...]
-    [1, 2, Cooperator, Alternator, CCCDCC, CCCDCC, CCCDCC,...]
+    """A class to hold the results of a tournament. Reads in a CSV file produced
```
drvinceknight: Could you add the new format of the file back in to the docstring?

drvinceknight: No, scrap that, just saw that it's in the read csv method. That's probably enough.

axelrod/tournament.py
```diff
@@ -88,57 +88,41 @@ def _build_result_set(self):
...
+        chunks = self.match_generator.build_match_chunks()
+
+        for chunk in chunks:
+            interactions = self._play_matches(chunk)
+            self._write_interactions(interactions)
         return True
 
-    def _run_parallel_repetitions(self, interactions):
+    def _write_interactions(self, interactions):
+        """Either write to memory or to file"""
```
drvinceknight: Change the docstring (not accurate) **although** is it perhaps worth having the ability to write memory? It's not that much extra code? This is more of a question than a request: what do you think? (Perhaps there's not too much to gain from doing that as the ResultsSet has all the interactions in memory anyway).

axelrod/tournament.py
```diff
@@ -154,14 +138,14 @@ def _run_parallel_repetitions(self, interactions):
         done_queue = Queue()
         workers = self._n_workers()
 
-        for repetition in range(self._parallel_repetitions):
-            work_queue.put(repetition)
+        chunks = self.match_generator.build_match_chunks()
+        for chunk in chunks:
+            # This is crap: I'm going through the generator, ideally want the
```
drvinceknight: Probably worth removing this comment which was just me when I threw this together quickly as a proof of concept. Given the speedups and memory checks you've done this is perhaps not as bad as I thought at the time.

axelrod/tournament.py
```diff
@@ -304,7 +264,8 @@ class ProbEndTournament(Tournament):
 
     def __init__(self, players, match_generator=ProbEndRoundRobinMatches,
                  name='axelrod', game=None, prob_end=.5, repetitions=10,
-                 processes=None, deterministic_cache=None, noise=0,
+                 processes=None,
+                 noise=0,
                  with_morality=True):
```
drvinceknight: `filename=None)` is missing here. Could we add a test that checks that as well?

axelrod/tournament.py
```diff
@@ -325,17 +286,14 @@ def __init__(self, players, match_generator=ProbEndRoundRobinMatches,
...
-        deterministic_cache : instance
-            An instance of the axelrod.DeterministicCache class
         with_morality : boolean
             Whether morality metrics should be calculated
         """
         super(ProbEndTournament, self).__init__(
             players, name=name, game=game, turns=float("inf"),
             repetitions=repetitions, processes=processes,
-            deterministic_cache=deterministic_cache,
             noise=noise, with_morality=with_morality)
```
drvinceknight: Will need filename here.

docs/tutorials/advanced/reading_and_writing_interactions.rst
```diff
@@ -7,70 +7,108 @@ a :code:`filename` argument to the :code:`play` method of a tournament::
...
+This can be transformed in to the usual interactions by zipping:
 
-    >>> string = 'CCDCCDDC'
-    >>> axl.interaction_utils.string_to_interactions(string)
-    [('C', 'C'), ('D', 'C'), ('C', 'D'), ('D', 'C')]
+    >>> list(zip("CCDC", "CDCD"))
+    [('C', 'C'), ('C', 'D'), ('D', 'C'), ('C', 'D')]
 
 This should allow for easy manipulation of data outside of the capabilities
 within the library, but it is also possible to generate a standard result set
```
drvinceknight: If we go with my suggestion ([on the play method of the tournament](https://github.com/Axelrod-Python/Axelrod/pull/572/files#diff-dad184e860a161b8b929a7ebbfd21c9cR67)) of the possibility of not creating the results set (`play(build_results=False)`) then we could add that here (just briefly mentioning that it's possible and might be of interest due to high memory consumption for large tournaments).

docs/tutorials/advanced/using_the_cache.rst
```diff
@@ -65,25 +65,8 @@ We can write the cache to file::
...
-    >>> tournament = axl.Tournament(players, turns=200, repetitions=5, deterministic_cache=cache)
-    >>> results = tournament.play()
-
-The cache has not only been used for this but has now also been updated with the
-new matches::
-
-    >>> len(cache)
-    6
+Tournaments will automatically create caches as needed on a match by match
+basis.
```
drvinceknight: Is there any point to keeping the ability to pass a cache to a tournament? This cache could perhaps be copied and sent out to each worker (and potentially brought back together at the end?). Perhaps not for this PR.

axelrod/tests/unit/test_tournament.py
```diff
@@ -391,132 +277,98 @@ def test_play_matches(self, turns):
...
-                        match = axelrod.Match(players, turns=turns)
-                        yield ((i, j), match)
-
-        matches_generator = make_generator()
-        interactions = tournament._play_matches(matches_generator)
-
-        self.assertEqual(len(interactions), 10)
+        def make_chunk_generator():
+            """
+            A generator that returns player index pairs and match objects for a
```
drvinceknight: We can probably remove the docstring here? (for the test?)

axelrod/tournament.py
marcharper: For sure, I almost did this but I thought to maybe wait for BigResultSet.

If `filename=None` and `build_results=False` then no data will be recorded. Should we assume the user knows what they are doing, or have some kind of warning?

axelrod/match_generator.py
marcharper: I went with `len(rr)` as the number of match pairings, which is n*(n+1) / 2 for all the i-j pairings, plus n more for the i-i pairings. `estimated_size` gives the total number of matches with repetitions and turns accounted for, so it's a measure of the memory footprint.

We can change this if you feel a different quantity is more natural for `len`.

axelrod/match_generator.py
marcharper: It's for the deterministic cache, which I'm not passing in any more (for this PR at least). Perhaps a comment is ok for now, or a keyword argument? It's possible that someone would want to do so... otherwise we could reorder the parameters of `Match` but I didn't want to do that.

axelrod/tournament.py
marcharper: I took it out since it's just as fast to write to disk and it seemed more simple to just use a temp file instead. I think if we have "build_results" as suggested above then maybe we don't need this.

docs/tutorials/advanced/using_the_cache.rst
marcharper: It could still be useful. Hard to say for sure without further testing. Not doing so for the standard tournament doesn't seem to cause any loss in performance.

marcharper:
The namespace test still fails for me locally when running `.\test`:

```
ERROR: test_namespace (unit.test_strategy_transformers.TestTransformers)
  File "/home/user/repos/axelrod/Axelrod/axelrod/tests/unit/test_strategy_transformers.py", line 365, in test_namespace
    axelrod.tests.unit.test_strategy_transformers.TestClass)
AttributeError: module 'axelrod.tests.unit' has no attribute 'test_strategy_transformers'
```

axelrod/tournament.py
marcharper: With `estimated_size` and `BigResultSet` we could make the decision for the user based on the likely memory footprint.

axelrod/tests/unit/test_tournament.py
```diff
@@ -177,100 +159,18 @@ def test_parallel_play(self):
...
+            #name=self.test_name,
+            #players=self.players,
+            #game=self.game,
+            #turns=200,
+            #repetitions=self.test_repetitions,
+            #processes=2)
+        #tournament._run_parallel()
+        #self.assertEqual(len(tournament.interactions), 15)
+        #for r in tournament.interactions.values():
+            #self.assertEqual(len(r), self.test_repetitions)
```
drvinceknight: 4b07252ed5672dc39d9066956d7ec07ddc902ee1 adds some tests for this. Basically I just check that the `True` is returned by the method itself but also using mocking that the `write_interactions` is called the amount of times we expect.

axelrod/tournament.py
drvinceknight: I would have a warning or perhaps if `filename=None` then we automatically set `build_results` to True with perhaps a warning as well?

axelrod/match_generator.py
drvinceknight: Keyword argument gets my vote :)

axelrod/match_generator.py
drvinceknight: Shouldn't the number of i-j pairings plus i-i pairings be the number of chunks generated (which is what my test is checking)? 

Either way, I think len should be the number of chunks generated: if only from the point of view of the progress bar. I think that for any generator if I would `len(generator) == len(list(generator))` to be True?

axelrod/tournament.py
drvinceknight: Cool, just change the `docstring` then :)

> With estimated_size and BigResultSet we could make the decision to keep interactions in memory for the user based on the likely memory footprint.

Ah I see what that's for now. Could you add a bit more to the docstring? And perhaps make it `_estimated_size` (not sure about that: up to you).

docs/tutorials/advanced/using_the_cache.rst
drvinceknight: Cool, something to think about: not for now :)

drvinceknight:
> The namespace test still fails for me locally when running .\test:

That's the same behaviour for me. What about `python3 -m unittest discover axelrod.tests`: I don't get that failure when calling the unittests directly.

axelrod/tournament.py
```diff
@@ -88,57 +88,41 @@ def _build_result_set(self):
...
-            self._run_single_repetition(interactions)
+        chunks = self.match_generator.build_match_chunks()
+
+        for chunk in chunks:
+            interactions = self._play_matches(chunk)
+            self._write_interactions(interactions)
         return True
 
-    def _run_parallel_repetitions(self, interactions):
+    def _write_interactions(self, interactions):
```
drvinceknight: 0eb187f646c834eb00811e4d15daaf6f8820504b adds a specific test for `write_interactions` (which is extensively tested throughout). Perhaps overkill but I think it's still nice to have.

marcharper:
That command works fine.

axelrod/tournament.py
marcharper: I moved it back into `play`.

axelrod/match_generator.py
marcharper: I think it's just an off by one error, should have been n(n-1) / 2 + n

docs/tutorials/advanced/using_the_cache.rst
marcharper: If someone comes up with a really expensive deterministic strategy then it could be useful for the interactions to be precomputed. But for now it seems better to just let `Match` handle caching, and to have a leaner cache (to reduce hash lookups). We could subset a loaded cache and pass that through.

drvinceknight: 👍

axelrod/match_generator.py
drvinceknight: 👍 

I really should have been able to see that (my PhD was in combinatorics, playing around with those stupid coefficients was all I did for 3 years)...

drvinceknight:
I'm pretty happy :+1: 

I started writing a couple of comments for some remaining things but realised it would be quicker to just open #575. Feel free to consider that a set of comments if you'd rather: it's just some tests and some docstrings that are no longer accurate (I think).

drvinceknight:
This is a go from me :+1: 

PS I think a doctest on the tutorial for making new tournaments might need to be skipped because there is a random situation that will cause a bug (had to rerun a test on my own branch). Just mentioning it in case travis complains (I'm heading to bed).

marcharper:
Thanks @drvinceknight!

marcharper:
I think we'll need to use a random number seed to make sure that test always passes.

drvinceknight:
> I think we'll need to use a random number seed to make sure that test always passes.

Good call: #577

meatballs:
If we are no longer passing a cache into the tournament, shouldn't we get rid of the load and save methods on the DeterministicCache class?

drvinceknight:
> If we are no longer passing a cache into the tournament, shouldn't we get rid of the load and save methods on the DeterministicCache class?

The cache can still be used by a `Match` and by the `MoranProcess` so I think it's worth keeping. Always useful for users who might be doing things outside of a tournament? 

**But** the documentation change for `using_the_cache.rst` here doesn't make that clear (I missed this first time around @marcharper). 

I suggest modifying the `Caching a Moran Process` to something like the following ("First let us..."):

```
A prebuilt cache can also be used in a Moran process (by default a new cache is used). First let us read in the previously built cache:

>>> cache = axl.DeterministicCache("cache.txt")  #  <- This is a new line, was in the Tournament part
>>> players = [axl.GoByMajority(), axl.Alternator(),
...            axl.Cooperator(), axl.Grudger()]
>>> mp = axl.MoranProcess(players, deterministic_cache=cache)
>>> populations = mp.play()
>>> mp.winning_strategy_name   
Defector
```

and where it currently says: "Again we see that the cache has been augmented...": that should be "We see that the cache has been augmented...".

meatballs:
OK, seems reasonable!

Vince Knight wrote:

marcharper:
The cache on disk may still be useful, here or in other tournament types. I suggest we investigate further after this PR is in. And it could certainly be useful for the Moran process with a lot of types (and that may not be as parallelizable).

PR #574, opened 2016-05-08
Not for master: Adding tutorial, and a test that fails (I think this is the expected behaviour)

drvinceknight:


marcharper:
I rebased onto parallel; merging.

PR #578, opened 2016-05-09
Adding a progress bar

### LINKED ISSUE #558: Progress bar.
I'm playing around with a progress bar using https://pypi.python.org/pypi/tqdm, I'll throw in some gifs below of it working as a proof of concept. If we like it, I'll open a PR (my thinking is to make it optional).

drvinceknight:
Closes #558

This is built on top of #572.

Works for both parallel and serial. Here's a gif showing it (code from https://gist.github.com/drvinceknight/2f42efc2388f41d3da7931ffc10aee6f):

941gayzwqh

### DIFF
--- a/axelrod/tournament.py
+++ b/axelrod/tournament.py
@@ -7,6 +7,8 @@ from multiprocessing import Process, Queue, cpu_count
 from tempfile import NamedTemporaryFile
 import warnings
 
+import tqdm
+
 from .game import Game
 from .match import Match
 from .match_generator import RoundRobinMatches, ProbEndRoundRobinMatches
@@ -66,7 +68,7 @@ class Tournament(object):
         # Save filename for loading ResultSet later
         self.filename = filename
 
-    def play(self, build_results=True, filename=None):
+    def play(self, build_results=True, filename=None, progress_bar=None):
         """
         Plays the tournament and passes the results to the ResultSet class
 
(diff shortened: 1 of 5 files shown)


marcharper:
Very nice :)

axelrod/tournament.py
```diff
@@ -66,7 +68,7 @@ def setup_output_file(self, filename=None):
         # Save filename for loading ResultSet later
         self.filename = filename
 
-    def play(self, build_results=True, filename=None):
+    def play(self, build_results=True, filename=None, progress_bar=None):
```
meatballs: If progress_bar is a boolean, why not use False as the default value?

axelrod/tournament.py
```diff
@@ -76,20 +78,28 @@ def play(self, build_results=True, filename=None):
...
         filename : string
             name of output file
+        progress_bar : bool
+            Whether or not to create a progress bar which will be updated
 
         Returns
         -------
         axelrod.ResultSet
         """
+        if progress_bar is not None:
```
meatballs: and this could become ```if not progress_bar```

axelrod/tournament.py
```diff
@@ -107,13 +117,25 @@ def _build_result_set(self):
...
+        progress_bar : bool
+            Whether or not to update the tournament progress bar
+        """
         chunks = self.match_generator.build_match_chunks()
 
         for chunk in chunks:
             results = self._play_matches(chunk)
             self._write_interactions(results)
+
+            if progress_bar is not None:
```
meatballs: same comment as previous

axelrod/tournament.py
```diff
@@ -202,6 +234,9 @@ def _process_done_queue(self, workers, done_queue):
                 stops += 1
             else:
                 self._write_interactions(results)
+
+                if progress_bar is not None:
```
meatballs: same again

drvinceknight:
Good call @meatballs, have made that change (and also added some more tests that check if the progress bar was or wasn't created).

drvinceknight:
Have added a line to the docs about the progress bar @meatballs: 08a0bb5

I don't feel strongly about it. Just a thought. Happy to throw it out...

meatballs:
:+1:

Vince Knight wrote:

marcharper:
It looks good for me, my only comment is that we might want progress bars on by default. Regardless this is :+1: from me.

drvinceknight:
I'm easy.

Note that with False by default when running the test suite you get 2 or 3
progress bars appear in the output. Changing it to True by default would
have a BUNCH more of them. Not the end of the world.

Any preference @meatballs?

meatballs:
How about True by default but the test suite passes in False?

drvinceknight:
That would work. I'll do that now on the train. :)

drvinceknight:
> How about True by default but the test suite passes in False?

That's done. A few progress bar still appear (to test that it actually works and doesn't break anything) but most are turned off.

Issue #580, opened 2016-05-10
Move `processes` argument from `tournament.init` to `tournament.play`


drvinceknight:
This seems more sensible to me from the point of view of what a tournament is and being able to choose how to play it.

meatballs:
👍

PR #582, opened 2016-05-11
Move processes argument from Tournament.__init__ to tournament.play

marcharper:
Re #580

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -135,6 +135,9 @@ paper please list it here:
   other environments
   <http://vknight.org/unpeudemath/gametheory/2015/11/28/Experimenting-with-a-high-performing-evolved-strategy-in-other-environments.html>`_
   by `drvinceknight <https://twitter.com/drvinceknight>`_
+- `An open reproducible framework for the study of the iterated prisoner's
+  dilemma <https://arxiv.org/abs/1604.00896>_`: a pre print of a paper describing this
+  library (20 authors).
 
 Contributors
 ============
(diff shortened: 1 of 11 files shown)


marcharper:
The build fails on Python 3.4 while building matplotlib, I suspect it's just a travis problem.

drvinceknight:
Have kicked off the build again. I like that you've moved repetitions also.

marcharper:
Actually it just looks that way from the diff, but it could be a good idea to move repetitions as well.

drvinceknight:
Oh yeah! I see that now :) As it's needed for the generator actually perhaps leave it? (haven't had coffee yet) in a way that is part of the tournament and I can imagine a tournament generator that samples the repetitions for each match?

marcharper:
Let's think about it for another PR.

drvinceknight:
Yup. Weird the build failed to install matplotlib again... Have set it off again, I'll take a proper look this morning. I'm guessing it's some temporary problem with Travis....

marcharper:
I can't see any reason why it would fail on just 3.4, so my next thought was some kind of Travis issue as well.

drvinceknight:
It's not getting to the code right, just failing to install matplotlib. For sure a Travis issue...

drvinceknight:
#585 should fix the py3.4 bug (issue with an upstream bug in pyparsing).

drvinceknight:
👍 from me once py3.4 is happy.

Issue #573, opened 2016-05-08
BigResultSet


marcharper:
For large tournaments, loading all the interactions into memory isn't feasible. @drvinceknight had the idea on #520 to have a special class for this case.

drvinceknight:
👍 and perhaps with `BigResultSet` there would be a way to speed things up by parallelising them? (No idea: just a thought.)

drvinceknight:
Hey @marcharper have you played around with HDF5 http://docs.h5py.org/en/latest/index.html 

Looks like this would involve storing the file in a different format (perhaps we could do that with a `big` switch or something...)  and might have other implications. 

Just a thought.

marcharper:
My next idea was to load the data using generators, computing wins, score differences, and such to reduce the memory footprint, possibly going over the file more than once.

I'm open to other options of course. I previously tried a somewhat similar approach with `shelve` (trying to use the standard library if possible). It's really just about the database build time, if an hdf5 file can be constructed relatively quickly then it makes sense, otherwise simply making multiple passes over a text file is likely faster.

drvinceknight:
I think the idea of generators sounds better to me. Saw this on a thread on reddit so thought it was worth mentioning :) Perhaps something to fall back on if generators doesn't do the trick :)

marcharper:
Definitely something to consider when we build out an API that saves large numbers of results over time.

drvinceknight:
FYI, I've had a little go at this, still very much a work in progress and mainly just playing around.

The main input to the `ResultsSet` is a dictionary mapping tuples of player index pairs to lists of interactions. So I've got below a class that simply mimics that dictionary using generators and never really (I think) having anything in memory.

Here it is:

```
import csv

class InteractionsFromFile:
    """A class that behaves like a dictionary but reads from file"""

    def __init__(self, filename):
        self.file = filename
    
    def __getitem__(self, index_pair):
        """Gives the dictionary like behaviour"""
        interactions = []
        with open(self.file, 'r') as file:
            reader = csv.reader(file)
            for row in reader:
                if index_pair == (int(row[0]), int(row[1])):
                    interactions.append(list(zip(row[4], row[5])))
        return interactions
            
    def keys(self):
        """Mimics the keys.values()"""
        generator = self.key_generator()
        return generator   
          
    def key_generator(self):
        """A generator for the keys.
        I use a cache to keep track as you go over the file if it's a new key or not."""
        self.keys_cache = set()
[...]

PR #583, opened 2016-05-11
BigResultsSetFromFile

marcharper:
First attempt at a BigResultsSet re: #573.

### DIFF
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -13,5 +13,5 @@ from .strategies import *
 from .deterministic_cache import DeterministicCache
 from .match_generator import *
 from .tournament import Tournament, ProbEndTournament
-from .result_set import ResultSet, ResultSetFromFile
+from .result_set import ResultSet, ResultSetFromFile, BigResultSetFromFile
 from .ecosystem import Ecosystem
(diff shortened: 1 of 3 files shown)


drvinceknight:
#585 should fix the py3.4 bug (issue with an upstream bug in pyparsing).

marcharper:
I haven't actually tested this on a large tournament, obviously we don't want to add that to Travis. I'll run some longer tournaments locally and report back.

drvinceknight:
:+1: Looking neat. Hadn't seen this kind of stuff before.

marcharper:
Unfortunately this turned out to be really slow and we're better off reloading from disk as needed.

drvinceknight:
How slow is slow because if the memory consumption is not crazy (currently the full tournament uses up 10GB of Ram) it could still be worthwhile? 

This isn't about speed but about memory.

marcharper:
It took more than 30 minutes to load the 8gb file into the shelve (I stopped it before it finished), and took around 1.5 gb of memory. I'm sure we can do better by simply wrapping the interactions object into a generator and reading it off disk multiple times (at least once for `scores` and once for `wins`). I was just trying to change as little code as possible with this PR and it didn't work out as well as I'd hoped.

drvinceknight:
Cool. :)

PR #591, opened 2016-05-14
New Strategies

marcharper:
This PR contains the following strategies:
* Adaptive
* Handshake
* CyclerDC and CyclerDDC (used in the literature)
* 8 Finite State Machine strategies: Fortress3, Fortress4, Predator, Raider, Ripoff, SolutionB1, SolutionB5, Thumper

I'm open to ideas about better testing for the FSM strategies.


marcharper:
The build error appears to be another matplotlib issue with travis.

drvinceknight:
> The build error appears to be another matplotlib issue with travis.

I think this might be a hypothesis tournament being chosen that isn't returning a result in a given amount of time.

This is on the failing build (py2):

```
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
  0%|                                                     | 0/3 [00:00<?, ?it/s].
```

This is on a passing one (py3):

```
Matplotlib is building the font cache using fc-list. This may take a moment.
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
  0%|                                                     | 0/6 [00:00<?, ?
```

The default hypothesis settings have been overwritten to not time out so if it's taking time this could clash with travis which is waiting to get a response. Going to fork this branch and try a couple of things. There was a reason for overriding the travis timeout but I'm not sure what it was.

Also, I need to set up a persisting hypothesis database on travis. I'm not too sure how to do this but will figure it out.

axelrod/strategies/adaptive.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self):
+        Player.__init__(self)
+        self.initial_plays = [C] * 6 + [D] * 5
+        self.reset()
+
+    def score_last_round(self, opponent):
+        try:
+            game = self.match_attributes["game"]
+        except AttributeError:
```
drvinceknight: Could you add a docstring for this explaining that this uses the game attribute if it is known by the player (neat to have that).

axelrod/strategies/adaptive.py
```diff
@@ -0,0 +1,49 @@
...
+            game = Game()
+        if len(self.history):
+            last_round = (self.history[-1], opponent.history[-1])
+            scores = game.score(last_round)
+            self.scores[last_round[0]] += scores[0]
+
+    def strategy(self, opponent):
+        # Update scores from the last play
+        self.score_last_round(opponent)
+        # Begin by playing the sequence C,C,C,C,C,C,D,D,D,D,D
```
drvinceknight: Any merit in having this initial sequence as a parameter? With 6Cs and 5 Ds as default?

axelrod/strategies/handshake.py
```diff
@@ -0,0 +1,33 @@
...
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self):
+        Player.__init__(self)
+        self.initial_plays = [C, D]
```
drvinceknight: Parametrize?

axelrod/tests/unit/test_adaptive.py
```diff
@@ -0,0 +1,50 @@
...
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_strategy(self):
+        # Test initial play sequence
+        self.responses_test([], [], [C] * 6 + [D] * 5)
```
drvinceknight: Could we have some specific tests for some of the methods: for example check that if the strategy does not know the game that the default one is used?

drvinceknight:
I've made some initial comments on the simpler ones (which all look great). Going to need a bit more time to look through the FSMs.

Re travis: tests seem to be hanging locally for me too. Hypothesis timeout doesn't seem to be the culprit...

drvinceknight:
The new strategies will need to be added to: http://axelrod.readthedocs.io/en/latest/reference/all_strategies.html in `docs/reference/all_strategies.rst`

drvinceknight:
I think I've fixed the build error. It was an issue with the integration test for the full tournament (nothing to do with hypothesis): building the results set on py2 was taking a huge amount of time (even on my machine: actually never got there). See #593.

marcharper:
A hypothesis test timed out.

axelrod/strategies/adaptive.py
marcharper: I'm not aware of another variant in the literature.

axelrod/strategies/handshake.py
marcharper: Same as above -- if we find another such strategy, then sure.

drvinceknight: I don't understand the drawback of parametrising this. This would be:

```
+    def __init__(self, initial_plays=[C, D]):
+        Player.__init__(self)
+        self.initial_plays = initial_plays
```

Seems simple enough and harmless, also immediately means that it can be used for to see effect of initial plays. Same comment for above.

drvinceknight:
> A hypothesis test timed out.

I'll add a timeout=0 for those tests.

drvinceknight:
https://github.com/Axelrod-Python/Axelrod/pull/594

marcharper:
Thanks @drvinceknight !

I didn't see any easy way to test the more complex FSM strategies. However I did put in sample tables for Cooperator, Defector, TFT, and WSLS, and tested various outcomes. So the biggest risk is just transcription error for the tables on my part.

In a few of the papers there were a few tables of expected outcomes (e.g. Fortress3 vs Fortress3). In those cases I put in explicit tests for the expected history.

drvinceknight:
Sounds great. I'm looking forward to looking through the FSM stuff. Looks really interesting, probably won't be today but hopefully by end of tomorrow (although there's a non zero probability that I'm lying to myself as that's far more interesting than marking so might end up doing it tonight anyway :)).

marcharper:
No rush. I've emailed the authors of those papers to see if they have any other recommendations. There were a few strategies that I couldn't track down exactly (S7 and Fortress5). They generated thousands of strategies for their papers so I'm hoping there are a few interesting ones laying about.

axelrod/strategies/handshake.py
marcharper: There's no drawback, I just added the strategy for completeness with the literature. We can generalize now if you'd like, or leave it as an easy PR for someone else.

drvinceknight: There's probably merit in opening an issue to take stock of all the
strategies and see what can be parametrized.

I'd prefer to see this done here already (seems like it might as well) but
up to you/second reviewer.

Hoping to get to this properly this evening :)

axelrod/strategies/adaptive.py
meatballs: or, perhaps, using the initial_sequence decorator?

marcharper: Sure we could do either. I'm not opposed, but as far as I know there's no real need to do so. This is a "for completeness" strategy to reproduce other works.

axelrod/tests/unit/test_adaptive.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_scoring(self):
+        player = axelrod.Adaptive()
+        opponent = axelrod.Cooperator()
+        player.play(opponent)
+        player.play(opponent)
+        self.assertEqual(3, player.scores[C])
+        game = axelrod.Game(-3, 10, 10, 10)
+        player.set_match_attributes(game=game)
```
drvinceknight: So this doesn't test that the exception is caught when the match attributes are not set.

Something like:

```
+        player = axelrod.Adaptive()
+        player.play(opponent)
+        player.play(opponent)
+        self.assertEqual(3, player.scores[C])
+        game = axelrod.Game(-3, 10, 10, 10)
+        #player.set_match_attributes(game=game)  Need to remove this to make sure it works.
```

axelrod/tests/unit/test_handshake.py
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    def test_strategy(self):
+        # Test initial play sequence
+        self.responses_test([], [], [C, D])
+
+        self.responses_test([C, D], [C, D], [C])
+        self.responses_test([C, D], [C, C], [D])
+        self.responses_test([C, D], [D, C], [D])
+        self.responses_test([C, D], [D, D], [D])
```
drvinceknight: Throw in one extra line with some arbitrary set of histories of size 5 say? Just to show that the play depends on the handshake.

axelrod/strategies/finite_state_machines.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def move(self, opponent_action):
+        """Computes the response move and changes state."""
+        next_state, next_action = self.state_transitions[(self.state, opponent_action)]
+        self.state = next_state
+        return next_action
+
+
+class FSMPlayer(Player):
+    """Abstract base class for finite state machine players."""
```
drvinceknight: Can you throw a good reference here for the general idea? (If there is one)

axelrod/tests/unit/test_finite_state_machines.py
```diff
@@ -0,0 +1,236 @@
+"""Tests for Finite State Machine Strategies."""
+import unittest
+
+import axelrod
+from .test_player import TestHeadsUp, TestPlayer
+
+C, D = axelrod.Actions.C, axelrod.Actions.D
+
```
drvinceknight: Could you add some tests for the abstract class for the FSM as well please?

axelrod/strategies/finite_state_machines.py
```diff
@@ -0,0 +1,277 @@
...
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self, transitions, initial_state, initial_action):
+        Player.__init__(self)
+        self.initial_action = initial_action
+        self.fsm = SimpleFSM(transitions, initial_state)
+
+    def strategy(self, opponent):
```
drvinceknight: It's possible that an `FSMPlayer` be ill defined. If the `transitions` are not complete:

For example:

```
transitions = ((1, D, 2, D),
                      (1, C, 1, D),
                      (2, C, 1, D),
                      (2, D, 3, C),
                      (3, C, 3, C))
```

Could we either:

- Include a check for this in the strategy
- Include a generic test in the strategy `TestFSMPlayer` class that will test that all states are in the transitions. Something like checking that `(x, y, *, *)` exists for all `y in [C, D]` and `x` for which `(*, *, x, *)` is in `transitions`.

The second one of those would be my preferred approach.

drvinceknight:
I'm happy with the particular response tests for the FSM (just the one comment about the completeness of the transitions).

docs/reference/all_strategies.rst
```diff
@@ -42,6 +45,9 @@ Here are the docstrings of all the strategies in the library.
 .. automodule:: axelrod.strategies.defector
    :members:
    :undoc-members:
+.. automodule:: axelrod.strategies.finite_state_machines
+   :members:
+   :undoc-members:
 .. automodule:: axelrod.strategies.forgiver
    :members:
    :undoc-members:
```
drvinceknight: missing handshake?

axelrod/strategies/adaptive.py
```diff
@@ -0,0 +1,50 @@
...
+
+class Adaptive(Player):
+    """Start with a specific sequence of C and D, then play the strategy that
+    has worked best, recalculated each turn."""
+
+    name = 'Adaptive'
+    classifier = {
+        'memory_depth': float('inf'),  # Long memory
+        'stochastic': False,
+        'makes_use_of': set(),
```
drvinceknight: This should include the game.

You'll then need to update the [classification of strategies documentation](https://github.com/Axelrod-Python/Axelrod/blob/master/docs/tutorials/advanced/classification_of_strategies.rst).

axelrod/strategies/adaptive.py
drvinceknight: Using the initial sequence decorator would not make changing the initial sequence available to the user (and note that for the handshake strategy it wouldn't do the trick as we need to know the length of the handshake so to speak).

That's my request (for here and handshake): make these nice generic strategies, the default value of which is one that adds to the completeness of the library. 

I still don't understand why we wouldn't do this, even if it's not used it's nice to have and seems like good practice (a part from adding one small extra test it's not that much extra work). **However, this should not delay this PR from going in** it's a simple thing that could be added at a later time. If there's opposition perhaps easiest to just leave it as it is :)

axelrod/tests/unit/test_adaptive.py
marcharper: Doesn't it test the two different paths? Otherwise the score be different.

axelrod/tests/unit/test_handshake.py
marcharper: These already test that -- if the handshake doesn't match the next move is D.

axelrod/strategies/finite_state_machines.py
marcharper: Sure, I just looked at wikipedia. Maybe I can find a better reference...

axelrod/strategies/finite_state_machines.py
marcharper: OK

docs/reference/all_strategies.rst
marcharper: Good catch

axelrod/strategies/adaptive.py
marcharper: Yes it should.

axelrod/strategies/finite_state_machines.py
drvinceknight: Even if you just throw the wiki link in there I think that's ok.

axelrod/tests/unit/test_adaptive.py
drvinceknight: Hmmm... Yeah you're right that should do...  My comment was mainly motivated by the fact that the coverage report said it didn't hit those lines... Leave it as it is.

axelrod/tests/unit/test_handshake.py
drvinceknight: Yup but this doesn't quite check that it will C or D forever. 

So just add some more to the history so we know this isn't just checking the particular case?

```
+        self.responses_test([C, D] * 2, [D, C] * 2, [D])
+        self.responses_test([C, D] * 2, [C, D] * 2, [C])
```

axelrod/strategies/adaptive.py
marcharper: The reasons why not:
* Simpler code
* Fewer cases to test
* No need for `@init_args`
* It may never get used anyway

But I'm not really against it (just not really for it), so there's a commit generalizing the arguments for both strategies.

axelrod/tests/unit/test_adaptive.py
marcharper: If coverage says it's not hit I'll try to add another test.

axelrod/tests/unit/test_finite_state_machines.py
marcharper: In addition to the four explicit tests? Right now you can't actually instantiate the FSMPlayer without supplying arguments (but we can pick a default of course).

drvinceknight: That was my thinking (in addition), I know it's a bit overkill. For the
tests just choose some super simple argument perhaps?

marcharper: Ok I've made it TFT for testing purposes.

axelrod/strategies/finite_state_machines.py
marcharper: I think I've covered what you wanted, take a look.

marcharper:
I'm not sure about the memory depth. Only the last round of play is explicitly used, but the state itself is the result of many rounds of play. TFT is a finite state machine and has depth 1, but so is TF2T, which has depth 2.

For an 8 state machine you could imagine that the next play could depend on the exact sequence of the last 8 moves. My intuition is that it's something like the longest path between any two states. I would think we could compute the depth exactly for small machines.

drvinceknight:
> I'm not sure about the memory depth. Only the last round of play is explicitly used, but the state itself is the result of many rounds of play. TFT is a finite state machine and has depth 1, but so is TF2T, which has depth 2.
> 
> For an 8 state machine you could imagine that the next play could depend on the exact sequence of the last 8 moves. My intuition is that it's something like the longest path between any two states. I would think we could compute the depth exactly for small machines.

Yeah super good point. I was caught up in thinking that it's just remembering what happened last time but in fact the state is remembering "more" in effect... I think the longest path makes sense. Because we know there's two paths from each state (C or D) I wonder if the calculation can be done in a simpler fashion...

axelrod/strategies/finite_state_machines.py
drvinceknight: Yeah you nailed it. 👍

drvinceknight:
> Because we know there's two paths from each state (C or D) I wonder if the calculation can be done in a simpler fashion...

There's not (drew some of these graphs and realised pretty quick).

Here's one approach: http://algorithmdojo.blogspot.co.uk/2013/02/longest-path-in-graph.html

I suggest hard coding the length in the player class and then automating the test class. Using a function like `check_state_transitions` that would build the graph and examine the paths and check that we have hard coded the correct length. Probably no need to use the approach in that link but just use a brute force method (we're not going to have huge graphs and it would just be for the tests).

marcharper:
Unfortunately I think it's even more complicated than I thought. If you look at the representation of Pavlov/WSLS in Ashlock2008.pdf you'll see that there are two states but it's a memory-one strategy.

I took a "best guess" at each for now.

drvinceknight:
> I took a "best guess" at each for now.

Cool. Let's go with this for now but perhaps open an issue about investigating the memory length of finite state machines? If anything it sounds like a neat problem to look in to...

drvinceknight:
This PR gets my 👍 . Looking forward to seeing what the full tournament looks like :)

PR #592, opened 2016-05-15
Setting a 5 minute timeout on hypothesis strategies.

drvinceknight:


drvinceknight:
Don't merge this: doesn't seem to be fixing anything.

PR #593, opened 2016-05-15
Make results set not build for full tournament in the integration test.

drvinceknight:
This was causing travis to timeout on py2 (it was also hanging on py2 for my own machine).

Have reduced the size of things as well: less turns. This test just needs to check that everything runs, everything else is well tested elsewhere.

### DIFF
--- a/axelrod/tests/integration/test_tournament.py
+++ b/axelrod/tests/integration/test_tournament.py
@@ -1,6 +1,6 @@
 import unittest
-import os
 import axelrod
+import warnings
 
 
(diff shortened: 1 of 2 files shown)


drvinceknight:
This should also make the full test suite run quicker...

drvinceknight:
(The progress bar really helped to debug this 👍  :))

drvinceknight:
So b12ef4c seems to do the trick. I couldn't quite get 55623ef to work because the error messages where still being caught by travis or something... I think 55623ef could be more elegant but it doesn't really matter.

PR #598, opened 2016-05-17
Reducing max examples and using a timeout.

drvinceknight:


drvinceknight:
Don't merge.

drvinceknight:
I've set two builds going, I can't seem to find anything online that doesn't indicate that travis will just fix itself but I'm really not sure...

@marcharper this PR reduces the number of examples needed by hypothesis. If this still bugs out you can try changing the timeout option to something other than 0 (no timeout). I'm heading to bed now but will be around tomorrow to take another look if this hasn't fixed itself...

PR #603, opened 2016-05-30
602

### LINKED ISSUE #602: Results set processing shouldn't be in the progress bar

drvinceknight:
Closes #602

This implements a progress bar for the results set. This progress bar is not very granular but it is precise.

### DIFF
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -15,7 +15,7 @@ except ImportError:
     from io import StringIO
 
 
-def update_pbar(method):
+def update_progress_bar(method):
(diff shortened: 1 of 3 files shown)


axelrod/result_set.py
```diff
@@ -14,10 +15,25 @@
     from io import StringIO
 
 
+def update_pbar(method):
```
marcharper: can we write `update_progress_bar` instead?

drvinceknight: Yup: good call! Done :) :+1:

PR #607, opened 2016-05-31
Fix PEP8 errors and redundant imports

meatballs:


drvinceknight:
Assuming travis is happy: 👍  from me.

PR #608, opened 2016-05-31
Refactors match_generator

meatballs:
to move noise parameter into __init__


drvinceknight:
I think the doctests might fail, either way, you'll need to move the noise out of the class created here: https://github.com/Axelrod-Python/Axelrod/blob/master/docs/tutorials/advanced/making_tournaments.rst

drvinceknight:
Looks good to me. 👍

PR #609, opened 2016-05-31
Add match_attributes to match class

### LINKED ISSUE #538: Setting information passed to players by matches
The conversations on #493 and #537 sparked the idea that a tournament could define the set of information which its matches could pass on to the players. Currently, the length, game and noise are always passed to all players.

e.g.
```
{
    'length': True
    'game': True
    'noise': True
}
```

Alternatively, we could pass the actual values which we want the players to use (which may or may not be the correct values).

meatballs:
closes #538


drvinceknight:
Could have some tests showing checking behaviour when passing non defaults.

drvinceknight:
The prob end tournament now tells the players the length of the match (which it shouldn't).

I'm not sure where this got introduced (not in this PR I don't think), but this PR would allow us to pass a `turns: float('inf')` so that the players don't know the match length. 

I think it would need the tournament to have a: `self.build_match` method which the prob end tournament could overwrite... It would be good to add a test to the prob end tournament that checks that players don't know the match lengths...

meatballs:
is that not a separate bug which needs another issue created?

drvinceknight:
> is that not a separate bug which needs another issue created?

I think it is indeed a separate thing, it's definitely not introduced by this right? (Still got my head deep in marking so if you could just confirm I'm not talking nonsense)...

meatballs:
It's not something that this PR introduced.

axelrod/tests/unit/test_match.py
```diff
@@ -32,6 +32,19 @@ def test_init(self, turns, game):
         self.assertEqual(match.noise, 0)
         self.assertEqual(match.game.RPST(), game.RPST())
 
+    @given(turns=integers(min_value=1, max_value=200), game=games())
+    @example(turns=5, game=axelrod.DefaultGame)
+    def test_init(self, turns, game):
```
drvinceknight: Can you change the name of this method to `test_init_with_non_default_match_attributes` or perhaps something that's less of a mouthful (but something that isn't `test_init`)?

meatballs: Agghhhh!! Cut and paste strikes again!

drvinceknight:
Good to go from me. This clears the way for a nice fix of #610. 👍

meatballs:
That last commit fixes #610

drvinceknight:
Cool, I've opened #611, as a PR to this branch with a single test that just checks that players do indeed not know the length of the match. Done using the final transformer on the cooperators and checking their cooperation rating.

PR #611, opened 2016-05-31
Test: players do not know length in prob end

drvinceknight:
Just to check that the expected behaviour has kicked in. A cooperation
rating of 1 implies that they never defected as expected.


meatballs:
Sorry, but I don't like that as a unit test as it involves too much other code from within the library.

I already included a unit test in the main PR, so I don't really see the need for this at all. If it has to be there, then it's an integration test.

drvinceknight:
Fair enough, I'll move it in to the integration test now.

I saw your test but that doesn't test that the tournament/players do indeed behave as expected so I'd like to see it in (it should have been in before really as we have not actually been testing one of the defining properties of the prob end tournament).

Moving it now :)

drvinceknight:
34805e5 moves it (I think you're right and this is much more an integration test) and also throws in some `progress_bar=False` that tidy things up a bit :)

PR #612, opened 2016-05-31
Adding an integration test

drvinceknight:
An integration test just to confirm that the players behave as expected in probabilistic end tournaments. This ensures that the correct information is being passed (and also in a way as an integration test it checks that that one decorator is behaving as expected in this case).

Thanks Owen for pointing out that this was not at all a unit test :) 👍

### DIFF
--- a/axelrod/tests/integration/test_tournament.py
+++ b/axelrod/tests/integration/test_tournament.py
@@ -2,6 +2,7 @@ import unittest
 import axelrod
 import tempfile
 
+from axelrod.strategy_transformers import FinalTransformer
 
 class TestTournament(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


axelrod/tests/integration/test_tournament.py
```diff
@@ -66,12 +67,28 @@ def test_noisy_tournament(self):
...
         tournament = axelrod.Tournament(players, turns=20, repetitions=10,
                                         with_morality=False, noise=0.75)
-        results = tournament.play()
+        results = tournament.play(progress_bar=False)
         self.assertEqual(results.ranked_names[0], "Cooperator")
+
+
+class TestProbEndTournament(unittest.TestCase):
+    def test_players_do_not_know_match_length(self):
+        """Create two players who should cooperate on last two turns if they
```
marcharper: defect? or don't know when the last turns are?

drvinceknight: yup: good catch. Fixing.

PR #618, opened 2016-06-03
Adding a progress bar to reading data.

### LINKED ISSUE #617: Add a progress bar to the ResultsSetFromFile
We have a progress bar for the actual play of the tournaments and for the analysis of the results (once the interactions have been read in to memory). This is good as it gives feedback (especially for very large tournaments when you might not know for sure that everything is actually running).

When redoing some of the plots for the paper, I read in the whole tournament data and was left with a long time with no user feedback: we don't have a progress bar for the reading of the data from file.

Have almost finished implementing a solution for this (one that includes the possibility of quickly reading the size of the data before actually reading it in). Just got to write a couple of tests and make sure it's all doing what it's supposed to do but here's a screenshot:

screenshot 2016-06-03 14 53 23

drvinceknight:
Closes #617 

- Done by keeping track of the interactions in the tournament and
  passing this to the ResultSetFromFile.
- If this is created by directly reading from file (so that we do not
  know how many rows of data there are) then a quick count of
  the size of the file is done (this is slightly inefficient but is
  worth it as it gives some user feedback).

Here's a gif of it reading in the file when it's not given the size of the file (there's a small delay as it counts the size of the file):

Here's a gif of it reading it in when it's given the size of the file:

We can also call it with `progress_bar=False`:

### DIFF
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -707,7 +707,8 @@ class ResultSetFromFile(ResultSet):
     by the tournament class.
     """
 
-    def __init__(self, filename, progress_bar=True):
+    def __init__(self, filename, progress_bar=True,
+                 num_interactions=False):
         """
         Parameters
         ----------
@@ -715,8 +716,14 @@ class ResultSetFromFile(ResultSet):
                 name of a file of the correct file.
             progress_bar : bool
                 Whether or not to create a progress bar which will be updated
+            num_interactions : int
+                If the number of interactions is known this helps with
+                displaying the progress bar, if it is not known and progress_bar
+                is set to True then an initial count will take place
         """
-        self.players, self.interactions = self._read_csv(filename)
+        self.players, self.interactions = self._read_csv(filename,
+                                                         progress_bar,
+                                                         num_interactions)
         self.nplayers = len(self.players)
         self.nrepetitions = len(list(self.interactions.values())[0])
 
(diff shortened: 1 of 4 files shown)


axelrod/result_set.py
```diff
@@ -707,16 +707,23 @@ class ResultSetFromFile(ResultSet):
     by the tournament class.
     """
 
-    def __init__(self, filename, progress_bar=True):
+    def __init__(self, filename, progress_bar=True,
+                 num_interactions=False):
```
marcharper: `None` instead of `False` makes more sense?

axelrod/tournament.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, players, match_generator=RoundRobinMatches,
         self.name = name
         self.turns = turns
         self.noise = noise
+        self.num_interactions = 0
```
marcharper: Is this computable from the match generator?

marcharper:
Looks good, I just had a couple of minor comments.

axelrod/result_set.py
drvinceknight: Yup.

axelrod/tournament.py
drvinceknight: I thought about that but then it would only work for the case of generators for which we know the size. However doing it this way (counting the rows we write) ensures that the size of the data file is always known.

drvinceknight:
I seem to have found an error, it's not quite behaving as it's supposed to. Don't merge just yet :)

drvinceknight:
> I seem to have found an error, it's not quite behaving as it's supposed to. Don't merge just yet :)

c420b54 fixes it. This is good to go. Reapplying your 1-positive-review label @marcharper (took it off just to make sure this wouldn't be merged whilst I did the one line fix).

Issue #387, opened 2015-10-26
Tournaments to Reproduce


marcharper:
* http://www.graham-kendall.com/papers/lhk2011.pdf

marcharper:
So for this issue we really just need to add the remaining strategies. Most of them are fairly straight-forward.

marcharper:
Strategies still needed:
* adaptive tft
* contrite tft
* gradual
* naive prober
* remorseful prober
* reverse pavlov
* soft grudger

drvinceknight:
I'm happy to pick these up, been a while since I wrote some strategies :) I think some might already be in there...

drvinceknight:
Hey @nikoleta-v3 I've been working on the remaining strategies listed by @marcharper and have `reverse pavlov` and `soft grudger` done but perhaps as a 'getting started with the library' exercise you could implement one of the others? We'll do it together over the next couple of days (perhaps tomorrow - Wednesday) but take a look at the paper linked to at the top and see what you think of the 5 others and pick one out.

Nikoleta-v3:
@drvinceknight Yes it would be a great help for starting and motivate me as I am going through the code. Will make sure to go through the paper today and chose one.

marcharper:
See also #379 for more strategies that we need. Just note that these range from fairly easy to quite hard to implement (typically because they are not sufficiently described).

drvinceknight:
@Nikoleta-v3 and I have had a look and believe that adaptive tit for tat is equivalent to: [GoByMajority](http://axelrod.readthedocs.io/en/latest/_modules/axelrod/strategies/gobymajority.html#GoByMajority).

drvinceknight:
Editing the list above: #626 implements Gradual and I have the other two (reverse pavlov
soft grudger) written on a branch.

PR #624, opened 2016-06-08
Implement Gradual strategy

Nikoleta-v3:
This is one of the strategies from the paper mentioned on #387 

Note that the strategy is not completely defined in the paper but @drvinceknight and I tested 3 varieties and this is the one that matched the results.

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -273,3 +273,20 @@ class OmegaTFT(Player):
         Player.reset(self)
         self.randomness_counter = 0
         self.deadlock_counter = 0
+
+class Gradual(Player):
+    """A player that punishes defection with a growing number of defections."""
+
+    name = "Gradual"
+    classifier = {
+        'memory_depth': float('inf'),  # Long memory, memory-2
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    @staticmethod
+    def strategy(opponent):
+        return D if D in opponent.history[-2:] else C
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -274,12 +274,18 @@ class OmegaTFT(Player):
         self.randomness_counter = 0
         self.deadlock_counter = 0
 
+
 class Gradual(Player):
-    """A player that punishes defection with a growing number of defections."""
+    """
+    A player that punishes defections with a growing number of defections
+    but after punishing enters a calming state and cooperates no matter what
+    the opponent does for two rounds.
+
+    http://perso.uclouvain.be/vincent.blondel/workshops/2003/beaufils.pdf """
 
     name = "Gradual"
     classifier = {
-        'memory_depth': float('inf'),  # Long memory, memory-2
+        'memory_depth': float('inf'),
         'stochastic': False,
         'makes_use_of': set(),
         'inspects_source': False,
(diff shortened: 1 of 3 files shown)


meatballs:
Looking at it now.....

axelrod/strategies/titfortat.py
```diff
@@ -273,3 +273,59 @@ def reset(self):
...
+    calming = False
+    punishing = False
+    punishment_count = 0
+    punishment_limit = 0
+
+    def strategy(self, opponent):
+
+        if self.calming:
+            self.calming = False
+            return 'C'
```
marcharper: Let's use `C` and `D` from `Axelrod.Actions` instead of explicit `'C'` and `'D'` (should already be imported at the top of the file). Same for the lines just below.

axelrod/strategies/titfortat.py
```diff
@@ -273,3 +273,59 @@ def reset(self):
...
+    classifier = {
+        'memory_depth': float('inf'),
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    calming = False
```
marcharper: Just so I understand -- there are actually three states, calming, punishing, or neither?

marcharper: Also, these four initialization lines should be in `def __init__(self)` since they can differ for multiple instances of the class. When variables are defined at the class level they are shared by all instances of the class.

In other words, we need an `__init__` function that does just what you've done in `reset` already:

```
def __init__(self):
    Player.__init__(self)
    self.calming = False
    self.punishing = False
    self.punishment_count = 0
    self.punishment_limit = 0
```

axelrod/strategies/titfortat.py
meatballs: 👍 (@marcharper seems to be able to type faster than I can)

axelrod/tests/unit/test_titfortat.py
```diff
@@ -272,3 +272,50 @@ def test_rounds(self):
...
+    }
+
+    def test_strategy(self):
+        """Starts by cooperating."""
+        self.first_play_test(C)
+
+    def test_effect_of_strategy(self):
+        """Punishes defection with a growing number of defections and calms the opponent
+            with two Cooperations in a row"""
+        self.responses_test([C], [C], [C])
```
marcharper: I believe `responses_test` can take an `attr` dictionary that will check for attribute values. That is, you can pass `attr = {"calming": True}` as an argument and at the end of the sequence of plays the test function will check that these attributes have been set on the player. Just an FYI for test writing in the future. Look at the tests for `Grudger` for some examples.

marcharper:
Looks good -- nice first contribution! Let's make sure to add an `__init__` function for this new player. 

The failing test is a doctest:

```

FAIL: ./docs/tutorials/advanced/classification_of_strategies.rst

Doctest: classification_of_strategies.rst

  File "/opt/python/3.4.2/lib/python3.4/doctest.py", line 2193, in runTest

    raise self.failureException(self.format_failure(new.getvalue()))

AssertionError: Failed doctest test for classification_of_strategies.rst

  File "./docs/tutorials/advanced/classification_of_strategies.rst", line 0

File "./docs/tutorials/advanced/classification_of_strategies.rst", line 32, in classification_of_strategies.rst

Failed example:

    len([s for s in axl.strategies if s().classifier['memory_depth']==1])

Expected:

    19

Got:

    20

```

I'm not sure why that happened since the new player has infinite `memory_depth`.

Nikoleta-v3:
Thank you for the comments. I will go back to add the stuff you have proposed and I will try to look why it's failing that one test.

Nikoleta-v3:
I  am closing this PR cause I opened it to the wrong branch.  ^_^

PR #626, opened 2016-06-08
Adding Gradual

Nikoleta-v3:
Third attempt at #624


Nikoleta-v3:
@marcharper  @meatballs Thank you for your comments! I believe I have addressed all of them. 
Also answering to your previous comment : yes it seems that there are three states. The first state player cooperates until the opponent defects. Then the player goes to punishing state, followed by the calming state (cooperates twice).

Issue #379, opened 2015-10-22
Desired New strategies


marcharper:
While researching other strategies, I have come across a number of relatively simple strategies that would be nice to have. I'm leaving them here as introductory tasks for new contributors. Cross-ref #302.

Implemented in [PRISON](http://www.lifl.fr/IPD/ipd.frame.html) (look in `classics.str`):
* per_ddc
* gradual
* gradual_killer
* soft_tf2t
* c_then_per_dc
* per_cccdcd
* prober4
* doubler
* soft_spiteful
* slow_tft
* easy_go
* better_and_better
* worse_and_worse, worse_and_worse2, worse_and_worse3

From CoopSim:
* ContriteTFT
* TwoTitsForTwoTats
* Others that you find interesting

From [this page](http://www.prisoners-dilemma.com/strategies.html):
* Soft Grudger
* Adaptive Tit For Tat
* Adaptive
* APavlov
* Omega Tit For Tat (see also [here](http://arxiv.org/pdf/cs/0609017.pdf))

From the mythical tournament preliminary to Axelrod #1:
* Analogy

From Axelrod tournament #2:
* Tranquilizer

drvinceknight:
This is a brilliant issue: :+1:

drvinceknight:
This strategy was mentioned on reddit (Random TitForTat and/or Grudger: defects with random probability):

https://www.reddit.com/r/GAMETHEORY/comments/480xb3/how_to_beat_this_strategy_of_prisoners_dilemma/

Adam-Flammino:
Would you mind if I started working on a few of these? I'm sure I'll be slow since I'm new to programming and I'm going to school while working full time, but it looks like they've been posted for awhile. If there are any strategies on the list that are particularly easy to implement I'd be happy to start with one of those.

drvinceknight:
Absolutely @Adam-Flammino: please do!

I suggest that when you decide on a strategy you run it past us just to make sure it hasn't been implemented yet (it's possible that they're on the list but have already been implemented).

It could also be worth checking here: http://axelrod.readthedocs.io/en/latest/reference/all_strategies.html

That's the list of the strategies that are definitely in the library :)

As far as a suggestion for an initial one to go for, perhaps pick "Nasty Tit For Tat" which is described as "tit-for-tat but attempts to exploit non-retaliatory strategies by playing a DD with some probability". If you think there's another one in the list that you like the look of please do go for that though :)

[...]

PR #629, opened 2016-06-11
Implemented a naive prober strategy (like tit for tat, but randomly defects).

314pe:
Implemented as randomly defecting strategy for #379 (with a relatively small probability) which was decorated with Tit For Tat style defection using RetaliateUntilApologyTransformer. At first, I wanted to start with a Tit For Tat and decorate that with a random flip transformer, but that would flip some Ds to Cs. Is this good or maybe if I could add some better tests, it would be visible how this strategy is inaccurate?


marcharper:
This one looks good to me. I'll give @drvinceknight a chance to take a look since some of the commits are mine. Thanks @314pe !

drvinceknight:
This looks great! Thanks @314pe :)

PR #630, opened 2016-06-13
Adding naive prober to strategies.

drvinceknight:
#629 didn't add the strategy to the list of strategies. Pinging @314pe just so that you see where this should have been :)


meatballs:
Sorry, should have noticed theses in the review.

drvinceknight:
As should I. No problem at all :)

PR #638, opened 2016-06-19
Fixing reset method for FSMs

### LINKED ISSUE #637: Finite state machine players don't reset properly
```
>>> import axelrod as axl
>>> tft = axl.TitForTat()
>>> predator = axl.Predator()
>>> predator.fsm.state
1
>>> m = axl.Match((tft, predator), 2)
>>> m.play()
[('C', 'C'), ('C', 'D')]
>>> predator.fsm.state
 2
>>> predator.reset()
>>> predator.fsm.state
2
```

Stumbled on this working on #636 (writing a hypothesis strategy that contrite TfT reduces to TfT in 0 noise) so the above is reduced from seeing that when playing the same match again we get a different output:

```
>>> m = axl.Match((tft, predator), 2)
>>> m.play()
[('C', 'C'), ('C', 'C')]
```

Am going to work on a fix now and include a hypothesis test that checks that random deterministic matches give the same outcomes.

drvinceknight:
Closes #637 

This also adds an integration test that checks that repeating matches with deterministic well behaved players gives the same outcome (this is how I found #637).

This 98d1837 adds the test (which should fail assuming hypothesis stumbles on the same error I was getting locally). f263ac5949e8fc57c5b5afb0a76dd5bce825d8ec fixes the finite state machines.

### DIFF
--- a/axelrod/strategies/finite_state_machines.py
+++ b/axelrod/strategies/finite_state_machines.py
@@ -54,6 +54,7 @@ class FSMPlayer(Player):
             initial_state = 1
             initial_action = C
         Player.__init__(self)
+        self.initial_state = initial_state
         self.initial_action = initial_action
         self.fsm = SimpleFSM(transitions, initial_state)
 
@@ -66,6 +67,10 @@ class FSMPlayer(Player):
             # for the strategy to function
             self.state = self.fsm.state
             return action
+
+    def reset(self):
+        Player.reset(self)
+        self.fsm.state = self.initial_state
 
 
 class Fortress3(FSMPlayer):
--- a/axelrod/tests/integration/test_matches.py
+++ b/axelrod/tests/integration/test_matches.py
@@ -1,0 +1,25 @@
+"""Tests for some expected match behaviours"""
+import unittest
+import axelrod
+
+from hypothesis import given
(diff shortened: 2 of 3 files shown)


drvinceknight:
(Feel free to squash the commits with the merge: there should only be 2 commits here anyway, got 4 because I'm yet to get my git-foo black belt.)

axelrod/strategies/finite_state_machines.py
```diff
@@ -54,8 +54,9 @@ def __init__(self, transitions=None, initial_state=None, initial_action=None):
             initial_state = 1
             initial_action = C
         Player.__init__(self)
+        self.initial_state = initial_state
         self.initial_action = initial_action
-        self.fsm = SimpleFSM(transitions, initial_state)
+        self.fsm = SimpleFSM(transitions, self.initial_state)
```
marcharper: Does this change make a difference or is it just style?

marcharper:
Good catch.

axelrod/strategies/finite_state_machines.py
drvinceknight: Nope, no difference. Not sure why I did that, I would actually argue the style before was better: changing now and cleaning up the commit history too :)

PR #639, opened 2016-06-20
Adding contrite TfT

### LINKED ISSUE #636: Adding Contrite TfT
This strategy plays like TfT in an environment with no noise. **If there is noise** it has the following added capability:

> If it is defected against because of noise (ie as a mistake) then it responds by cooperating twice.

This requires a bit more information as we don't currently have a way for a player to know if an opponent's play was returned because of noise or not. @marcharper pointed out the [`TrackHistoryTransformer`](http://axelrod.readthedocs.io/en/latest/tutorials/advanced/strategy_transformers.html) which allows a player to track it's own intended play. Perhaps something like that could be used.

drvinceknight:
Closes #636 

A strategy that plays like TfT if there is no noise. If there is noise and it defects because of noise (intending to cooperate), if the opponent cooperates but defects as a result of the noisy defection then it cooperates to try and bring back mutual cooperation.

Wasn't sure if there was a better way to handling the renaming from the `TrackHistoryTransformer` (without my change it adds a prefix to the strategy)?

Note that this is on top of #638 (as it needed that fix to work for the 'reduce to tft' test to work).

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -365,7 +365,7 @@ class ContriteTitForTat(Player):
         if not opponent.history:
             return C
 
-        # If contrite and but managed to cooperate: apologise.
+        # If contrite but managed to cooperate: apologise.
         if self.contrite and self.history[-1] == C:
             self.contrite = False
             return C
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -1,5 +1,5 @@
 from axelrod import Actions, Player, init_args, flip_action
-from axelrod.strategy_transformers import TrackHistoryTransformer
+from axelrod.strategy_transformers import StrategyTransformerFactory, history_track_wrapper
 
 C, D = Actions.C, Actions.D
 
@@ -334,7 +334,11 @@ class Gradual(Player):
         self.punishment_limit = 0
 
 
-@TrackHistoryTransformer
+Transformer = StrategyTransformerFactory(
+    history_track_wrapper, name_prefix=None)()
+
+
+@Transformer
 class ContriteTitForTat(Player):
     """
     A player that corresponds to Tit For Tat if there is no noise. In the case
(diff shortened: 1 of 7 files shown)


axelrod/strategies/titfortat.py
```diff
@@ -331,3 +332,52 @@ def reset(self):
...
+        'manipulates_state': False
+    }
+    contrite = False
+
+    def strategy(self, opponent):
+
+        if not opponent.history:
+            return C
+
+        # If contrite and but managed to cooperate: apologise.
```
meatballs: Just a typo in the comment - 'and but'

drvinceknight: Fixing now :)

marcharper:
> Wasn't sure if there was a better way to handling the renaming from the TrackHistoryTransformer (without my change it adds a prefix to the strategy)?

You could just make a new instance:
```
transformer = StrategyTransformerFactory(
    history_track_wrapper, name_prefix="")()
```

The prefix is there to prevent overwriting the original class; in this case it doesn't matter but it may in other contexts. I'd prefer to leave the default with some prefix. Actually, the transformer is supposed to take a name_prefix parameter:

```
        def __call__(self, PlayerClass):
            """
            Parameters
            ----------
            PlayerClass: A subclass of axelrod.Player, e.g. Cooperator
                The Player Class to modify
            name_prefix: str
                A string to prepend to the Player and Class name
```

so that the following would work:

```
@TrackHistoryTransformer(name_prefix='')
class MyPlayer(object):
    name = "MyName"
print(MyPlayer().name)
```

but as you can see it's not implemented properly.

drvinceknight:
Thanks: have fixed that using your suggestion.

> The prefix is there to prevent overwriting the original class; in this case it doesn't matter but it may in other contexts. I'd prefer to leave the default with some prefix.

Me too, I like having the prefix. Would be great to be able to pass the prefix (or `None`) to the decorator as you said there. I'll open an issue for it (also, note that `FinalTransformer` does not have a prefix (used by `Backstabber`). That should be addressed too.

drvinceknight:
Have opened #642

PR #648, opened 2016-06-28
[#647] Add git hook script

### LINKED ISSUE #647: Consider using a git hook to prepend the issue number to a commit message
So that:

* Commits relating to an issue appear in on that issue's page
* Anyone using a gui git client has clickable links to the issue in their git log

This can be achieved manually but a hook removes the effort.

meatballs:
For consideration!!

fixes #647


utils/prepare-commit-msg.py
```diff
@@ -0,0 +1,38 @@
...
+#     ln -s ./utils/prepare-commit-msg.py prepare-commit-msg
+#
+# TODO - the above doesn't work on Windows
+
+import sys
+import re
+from subprocess import check_output
+
+# By default, the hook will check to see if the branch name starts with
+# 'issue-' and will then prepend whatever follows in the commit message.
```
drvinceknight: Just noting that this would be different to current practice? (Where the branch name is just the issue number) Not that this has ever been a strict requirement.

Not a problem, arguably having `issue-` is nicer/better.

utils/prepare-commit-msg.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+
+# This script is intended to be used a a git hook and will prepend the ticket
+# number to a commit message in the correct format for Github/Gitlab to parse.
```
drvinceknight: Can we add something like `(use of this is optional and is meant to be helpful)` at the top or something (just thinking of the beginner contributor who might stumble upon this and think it's something they HAVE to do.

meatballs: Yep - I'll put some words in. Need to find out the Windows command too.

drvinceknight:
If this goes in, I'm not sure if having a new directory is __potentially__ less or more confusing (to the very beginner contributor) than a file at the root level (where we already have `.mailmap`, `.editorconfig` etc...). So perhaps this could be `.prepare-commit-msg.py`?

(Not terribly offended by the new directory either, just a thought.)

meatballs:
> So perhaps this could be .prepare-commit-msg.py?

Yep, good idea

drvinceknight:
Whilst not overly excited about this, I do think your commit message on this PR look quite nice. I'll probably end up using the damn thing! :) 👍

meatballs:
> Whilst not overly excited about this, I do think your commit message on this PR look quite nice. I'll probably end up using the damn thing! 

Just something I've already had to write for a work project. Seemed a shame to waste it once the contract finished!

PR #650, opened 2016-07-06
Soft tf2t: rebase

marcharper:
Hi @ranjinidas : I rebased your branch onto the latest master so that it is easier for us to review. We'll also need to add some tests, but we can take care of that if needed.

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -387,13 +387,13 @@ class ContriteTitForTat(Player):
         self.contrite = False
         self._recorded_history = []
 
-class SlowTitForTat(Player):
+
+class SoftTitForTwoTats(Player):
     """
-    A player co-operates twice, then if the opponent makes the same move twice, 
-    the player mimics its move
+    A player co-operates except for when the opponent defects twice
     """
 
-    name = 'Slow Tit For Tat'
+    name = 'Soft Tit For Two Tats'
     classifier = {
         'memory_depth': 2,  # Four-Vector = (1.,0.,1.,0.)
         'stochastic': False,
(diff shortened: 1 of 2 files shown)


drvinceknight:
Unless I'm misunderstanding the strategy you have written is already implemented as `TitFor2Tats`: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/titfortat.py#L40

Is that right @ranjinidas ?

axelrod/strategies/titfortat.py
```diff
@@ -386,3 +386,31 @@ def reset(self):
...
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent):
+        if len(self.history) == 0:
+            return C
+
+        if opponent.history[-2] == [D, D]:
```
drvinceknight: I think you wanted this to be `opponent.history[-2:] == [D, D]` so that it sliced the list. Otherwise `opponent.history[-2]` is just the second from last element which would either be `C` or `D`.

drvinceknight:
If you'd like a hand at all feel free to jump in to the chat room here: https://gitter.im/Axelrod-Python/Axelrod 

There's usually a couple of us around and we're happy to help :)

ranjinidas:
Hi Vince and Marc,

Thanks for your replies! I didn't realize Tit for Two Tats and Slow Tit for
Two Tats were two different things. It was open in the Desired New
Strategies Issue, but thanks for letting me now!

I'm a beginner in Python and OSS, and I will try again with another
strategy, but thank you for being so supportive.

Best,
Ranjini

drvinceknight:
As far as I can tell soft tit for two tats isn't there but slow tit for 2 tats is and is defined as:

> Plays [c,c], then if opponent plays two consecutive time the same move plays its move.

Looking through your commits, it looks like you started writing this strategy here: d4dbe8af0e442fd1db4e0984966dc0614076b23e (you can click on that) but then you changed it here: 	3ee3876 to become tit for 2 tats.

> I'm a beginner in Python and OSS, and I will try again with another
strategy, but thank you for being so supportive.

I think implementing slow tit for 2 tats is probably a good choice of strategy :) We're happy to help out  :)

ranjinidas:
@drvinceknight Thanks for the feedback. :) I'll go back to working on that strategy.

drvinceknight:
Closing this as it's superseded by #659 (we can always open it again if need be).

PR #653, opened 2016-07-08
651 - Tests for random seeding and a random seeding function

### LINKED ISSUE #651: Add tests for reproducibility of stochastic results
I'd like to add a property based test similar to https://github.com/Nikoleta-v3/Axelrod/blob/635/axelrod/tests/integration/test_matches.py but that checks that when setting a seed we get the same results for stochastic strategies (those tests there only check for 'well behaved deterministic' strategies. I have a hunch that this isn't actually true as we're not setting numpy's seed...

If that's the case and both seeds need to be set, we should document how to get reproducible results and either:

1. Indicate that you should set both seeds (numpy and stdlib);
2. Write a little helper function that does this (and just document it: `axelrod.seed()` or something like that...

drvinceknight:
Closes #651 

Firstly this adds a test to `integration/test_matches`. Simply checks that if two stochastic matches are played with the same seed they give the same result. 

Note that both the numpy and stdlib seed need to be set.

Have also added a function that sets both of those: `axelrod.set_seed` just to make it 'less easy' to only set one of the seeds...

### DIFF
--- a/axelrod/tests/unit/test_random_.py
+++ b/axelrod/tests/unit/test_random_.py
@@ -1,9 +1,10 @@
 """Test for the random strategy."""
 
+import numpy
 import random
 import unittest
 
-from axelrod import random_choice, Actions
+from axelrod import random_choice, set_seed, Actions
 
 C, D = Actions.C, Actions.D
 
@@ -16,3 +17,18 @@ class TestRandom_(unittest.TestCase):
         self.assertEqual(random_choice(), C)
         random.seed(2)
         self.assertEqual(random_choice(), D)
+
+    def test_set_seed(self):
+        """Test that numpy and stdlib random seed is set by set seed helper
+        function"""
+
+        numpy_random_numbers = []
+        stdlib_random_numbers = []
+        for _ in range(2):
+            set_seed(0)
+            numpy_random_numbers.append(numpy.random.random())
+            stdlib_random_numbers.append(random.random())
+
+        self.assertEqual(numpy_random_numbers[0], numpy_random_numbers[1])
+        self.assertEqual(stdlib_random_numbers[0], stdlib_random_numbers[1])
+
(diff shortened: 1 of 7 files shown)


marcharper:
Are we concerned that setting the same exact seed may have unexpected consequences?

drvinceknight:
I'm not sure I understand what you mean? As in the same seed for numpy and stdlib?

marcharper:
Yeah that's what I meant. I presume they use different generators / algorithms but I don't really know.

drvinceknight:
> I presume they use different generators / algorithms but I don't really know.

Yeah I believe they're two completely different generators (which is why you need to set both seeds). 

I don't believe they are any unintended circumstances but just in case I've just added 7ee1d83 which adds 2 further integration tests that checks not that two matches repeat but that actual tournaments repeat as well.

axelrod/tests/unit/test_random_.py
```diff
@@ -16,3 +17,18 @@ def test_return_values(self):
         self.assertEqual(random_choice(), C)
         random.seed(2)
         self.assertEqual(random_choice(), D)
+
+    def test_set_seed(self):
+        """Test that numpy and stdlib random seed is set by set seed helper
```
marcharper: `set_seed` would make this sentence less confusing.

drvinceknight: Yeah good call.

Before I change that: what do you think about the name of the function: is `set_seed` the right way to go? Should it actually just be `seed`?

So there's `random.seed`, `numpy.random.seed` and then equivalently `axelrod.seed`?

marcharper: Yeah that's a good idea :)

drvinceknight: Done!

PR #654, opened 2016-07-13
635 Implementing spatial tournaments

### LINKED ISSUE #635: Implementing Spatial Tournaments
Spatial Tournaments are usually implemented on a lattice, where usually players only play their neighbours(on a lattice). I will be working on implementing a spatial tournament that is more generally defined by a graph(players play who ever they are connected to). 

@drvinceknight  and I are talking about using networkx for the graph part. The graph would only be an input to a new match generator class. As a dependency networkx is not too heavy but it would be good to hear everyone’s thoughts(we could use a dictionary, but that would be a bit messy).

After this to match the literature I will explore various evolutionary processes one of which could be a spatial Moran process. But that's for later. 

If anyone is interested here is some of the literature : 

-[Nowak and May, 1992 ](http://www.nature.com/nature/journal/v359/n6398/abs/359826a0.html)
-[Lindgren and Nordahl, 1994](http://www.sciencedirect.com/science/article/pii/0167278994902895)
-[Maciver, 2014](http://project-archive.inf.ed.ac.uk/ug4/20140533/ug4_proj.pdf)

Nikoleta-v3:
Closes #635. 

-  Implemented a new tournament class and a new match generator class for spatial tournaments 
-  Appropriate tests for the new classes have been written
-  Created a doc file in the further issues for spatial tournaments. 

After considering all the thoughts on how to implement a spatial structure tournament, I have decided to
simple consider a list that includes the edges. The new class  `SpatialMatches()`  creates
chunks where players - nodes  only play players that are connected to with an edge. A tournament 
class `SpatialTournament` was also implemented to pass the `edges` argument and `SpatialMatches`
as the `match_generator`.   An error message was also included. Just in case a  player is not interacting with anyone . If you have such a player is better to not include him in the tournament at all. 

Before creating these classes I had written some tests. For `SpatialMatches`, `SpatialTournament` , as well as some property tests.  Also a test for comparing the results of a complete multi-graph and a round robin tournament. 

In the result set, seems that in the loop statements the self interaction have been counted twice each time. So an `if` statement had to be altered to an `elif`.
Also there was one test in the result_tests that seem to have been passing by change, so I altered the
implementation a little bit. 

Furthermore, I have added in the docs of the Axelrod library a small explanation and example of how to create spatial tournaments. All tests seem to be working fine !


axelrod/match_generator.py
```diff
@@ -167,3 +167,61 @@ def estimated_size(self):
...
+        ----------
+        players : list
+            A list of axelrod.Player objects
+        turns : integer
+            The number of turns per match
+        game : axelrod.Game
+            The game object used to score the match
+        repetitions : int
+            The number of repetitions of a given match
+        edges : dictionary
```
marcharper: Can you explain what this dictionary looks like? In the tests below it looks like a list.

marcharper:
Can we add a test that has a set of deterministic players on a complete graph compared to the non-spatial tournament? We should get the same results in both cases.

Nikoleta-v3:
First of all, it is a list ! I just happened to be calling it a dictionary for a long time... I will correct that, there seems to be more than once. Also for your second point. It has already been implemented, is the test for the complete multigraph.

drvinceknight:
> Also for your second point. It has already been implemented, is the test for the complete multigraph.

Here: https://github.com/Axelrod-Python/Axelrod/pull/654/files#diff-74e02134acd0ef3b29c84d6643837201R593 (line 593):

```
+        def test_complete_tournament(self, strategies, turns):
+            """
+            A test to check that a spatial tournament on the complete multigraph
+            gives the same results as the round robin.
+            """
```

marcharper:
Great, sorry I'm not familiar with the term multigraph.

Is there an easy demo case that compares to the other tournaments in the documentation? There is an example that shows how many cooperators you need to overcome a defector, is there a topology that changes the number of cooperators needed (a cycle perhaps)? That would be a nice addition to the docs.

drvinceknight:
> Great, sorry I'm not familiar with the term multigraph.

It could be better to just rename it to just 'complete graph'?

> 
> There is an example that shows how many cooperators you need to overcome a defector

If I remember correctly, I think that example is no longer there. It was in the early form of the docs which were far too complicated, before we simplified the tutorials to being more modular. Not saying it's not worth including an example like that here, just pointing it out.

Nikoleta-v3:
I would be happy to work on something like that ! But maybe @drvinceknight is right, is not to technical for the specific section. It could go to http://axelrod-tournament.readthedocs.io/en/latest/, as an "Research Example - something"  ?.

drvinceknight:
> But maybe @drvinceknight is right, is not to technical for the specific section. It could go to http://axelrod-tournament.readthedocs.io/en/latest/, as an "Research Example - something" ?.

I think I would prefer the tournament repo to stay as it is. I _also_ think the docs should not try and do too much (as they did previously) **but** perhaps there is an interesting topolgy that could be shown.

Perhaps run two tournaments: one complete graph (remarking that this corresponds to the round robin), and then another that perhaps isolates `Defector` a bit (perhaps it only plays `TitForTat` which serves as a bridge/separator to the other strategies that continue to all play each other)? And then that would show how the cooperators do better?

axelrod/match_generator.py
```diff
@@ -167,3 +167,61 @@ def estimated_size(self):
...
+    def __init__(self, players, turns, game, repetitions, edges):
+
+        player_indices = list(range(len(players)))
+        node_indices = sorted(set([node for edge in edges for node in edge]))
+        if player_indices != node_indices:
+            raise ValueError("The graph edges do not include all players.")
+
+        self.edges = edges
+        super(SpatialMatches, self).__init__(players, turns, game, repetitions)
+        """
```
marcharper: Let's move the doc string up to just under the `class SpatialMatches` line. Can we clean up the language a bit? I'll put suggestions inline below.

axelrod/match_generator.py
```diff
@@ -167,3 +167,61 @@ def estimated_size(self):
...
+        A class that generates spatial structure matches. In spatial structure
```
marcharper: A class that generates *spatially-structured matches.* In these matches, players interact only with their neighbors rather than the entire population. This reduces to a well-mixed population when the spatial graph is a complete graph.

axelrod/match_generator.py
```diff
@@ -167,3 +167,61 @@ def estimated_size(self):
...
+        matches the players do not play against every other player, instead they
+        only play their neighbors.
+
+        In literature the authors interpreted spatial tournaments topology
```
marcharper: A common case in the literature is a square lattice or grid. This tournament type works for an arbitrary graph.

drvinceknight:
> "Research Example - something" ?.

It could certainly be a great notebook to add to https://github.com/Axelrod-Python/Axelrod-notebooks

axelrod/match_generator.py
```diff
@@ -167,3 +167,61 @@ def estimated_size(self):
...
+        as a 2D square lattice. We want to create a class that will not take as
+        an argument only a 2D lattice but any given graph.
+
+        After considering various python packages for graphs, we ended up using
```
marcharper: Remove this paragraph?

axelrod/match_generator.py
```diff
@@ -167,3 +167,61 @@ def estimated_size(self):
...
+        a simple dictionary that includes the edges.So the players - nodes will
+        only play players that are connected to with an edge.
+
+        We have chosen to implement spatial structure tournaments because
```
marcharper: See the following references:
* Nowak and May (1992) <http://www.nature.com/nature/journal/v359/n6398/abs/359826a0.html
* Lindgren and Nordahl (1994) <http://www.sciencedirect.com/science/article/pii/0167278994902895>
* Brauchli, Killingback and Doebeli (1999) <http://www.zoology.ubc.ca/~doebeli/reprints/Doe38.pdf>

If you want to include more information about the literature and history I'd suggest putting that in the documentation instead.

axelrod/tests/unit/test_match_generator.py
```diff
@@ -207,3 +207,39 @@ def test_len(self, prob_end):
...
+        self.assertEqual(sorted(match_definitions), sorted(expected_match_definitions))
+
+    def test_len(self):
+        edges = [(0, 1), (1, 2), (3, 4)]
+        sp = axelrod.SpatialMatches(
+            self.players, test_turns, test_game, test_repetitions, edges)
+        self.assertEqual(len(sp), len(list(sp.build_match_chunks())))
+        self.assertEqual(len(sp), len(edges))
+
+"""
```
marcharper: Let's remove this if it's not going to be used.

axelrod/tests/unit/test_property.py
```diff
@@ -3,7 +3,8 @@
 import axelrod
 from axelrod.tests.property import (strategy_lists,
                                     matches, tournaments,
-                                    prob_end_tournaments, games)
```
marcharper: I see you are in the @meatballs school of spacing :)

axelrod/tests/unit/test_resultset.py
```diff
@@ -394,3 +392,488 @@ class TestDecorator(unittest.TestCase):
     def test_update_progress_bar(self):
         method = lambda x: None
         self.assertEqual(axelrod.result_set.update_progress_bar(method)(1), None)
+
+
+class TestResultSet_SpatialStructure(TestResultSet):
```
marcharper: Minor comment: `TestResultSet_SpatialStructure` PEP8 says no underscores for class names (just CamelCase).

axelrod/tests/unit/test_resultset.py
```diff
@@ -394,3 +392,488 @@ class TestDecorator(unittest.TestCase):
...
+                0.7683498199278325,
+                0.5488212999484519
+            ]
+
+        cls.expected_csv = (
+            'Defector,Tit For Tat,Alternator\n3.4,2.6,1.5\n3.4,2.6,1.5\n3.4,2.6,1.5\n')
+
+    def test_match_lengths(self):
+        """
+        Overwriting match lengths because of edges
```
marcharper: Can you expand this doc string a bit?

axelrod/tests/unit/test_resultset.py
```diff
@@ -394,3 +392,488 @@ class TestDecorator(unittest.TestCase):
...
+
+                for j, length in enumerate(opp):
+                    edge = (i, j)
+                    # Specific test for example match setup
+                    if edge in self.edges or edge[::-1] in self.edges :
+                        self.assertEqual(length, self.turns)
+                    else:
+                        self.assertEqual(length, 0)
+
+class TestResultSet_SpatialStructure_Two(TestResultSet_SpatialStructure):
```
marcharper: PEP8: Underscores

axelrod/tournament.py
```diff
@@ -335,3 +335,43 @@ def __init__(self, players, match_generator=ProbEndRoundRobinMatches,
...
+        ----------
+        players : list
+            A list of axelrod.Player objects
+        match_generator : class
+            A class that must be descended from axelrod.MatchGenerator
+        name : string
+            A name for the tournament
+        game : axelrod.Game
+            The game object used to score the tournament
+        edges : dictionary
```
marcharper: list of tuples

marcharper:
This is a nice contribution overall!

I put in lots of little comments above. It would be nice to have a more complex example somewhere in the docs or notebooks. It might also be interesting to augment the tournament repository with some new tournaments, say on a big cycle, a grid like in the literature, or some other interesting case.

Also two ideas for the future:
* probabilistic ending + spatial games, and
* Moran Process on directed graphs (I'll tackle this one)

Nikoleta-v3:
Thank you @marcharper ! I will go  through the comments and make the corrections. Also as suggested I will create a few notebooks.  

We were considering evolution in general but suggested that could be left for another time. I do believe from October I will spend a lot of time with Spatial tournaments so I appreciate the ideas !

drvinceknight:
> It might also be interesting to augment the tournament repository with some new tournaments, say on a big cycle, a grid like in the literature, or some other interesting case.

Sure: my main question would be how to pick the particular graph for the tournament repo (if a cycle: how do we order it?). Perhaps a discussion for another issue: have opened https://github.com/Axelrod-Python/tournament/issues/36

> probabilistic ending + spatial games, and

Yes. Maybe you could look in to that at some point @Nikoleta-v3 (not necessarily this Summer: there's a lot to do based on our chat this am although it would be pretty straightforward to implement).

I'll throw an issue together for it :)

> Moran Process on directed graphs (I'll tackle this one)

Yes :) 👍

Nikoleta-v3:
@marcharper I believe I have addressed all of your comments ! But I am still working on some notebooks for the documentation and for my dissertation as well.

PR #658, opened 2016-07-16
Caching the hypothesis db.

### LINKED ISSUE #595: Make the hypothesis failure data base persist on travis
Just a question of figuring out where the db is and how to tell travis to remember it (I think this says how: https://docs.travis-ci.com/user/caching/). (No changes to the library, just better use of hypothesis and travis).

drvinceknight:
(I believe) this closes #595.

It caches the hypothesis failure database. It just means that travis will keep track of parts of the test space for which tests previously failed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,11 @@ python:
   - 2.7
   - 3.4
   - 3.5
+
+cache:
+  directories:
+    - $TRAVIS_BUILD_DIR/.hypothesis
+
 before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start


marcharper:
How do we know if it works?

drvinceknight:
> How do we know if it works?

Excellent question. I've spoken to the maintainer of hypothesis and this is how he said to do it. Based on the travis documentation about making things persist it seems to make sense.

Ultimately I don't think there is a way of knowing (even if the same failed example was triggered twice: that wouldn't necessarily confirm it)... At worst it's not working which would not actually change anything from the status quo.

marcharper:
I guess we'll find out soon!

PR #659, opened 2016-07-16
Added Slow TF2T

ranjinidas:
Corrected strategy from last pull request.

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -387,13 +387,13 @@ class ContriteTitForTat(Player):
         self.contrite = False
         self._recorded_history = []
 
-class SlowTitForTat(Player):
+
+class SoftTitForTwoTats(Player):
     """
-    A player co-operates twice, then if the opponent makes the same move twice, 
-    the player mimics its move
+    A player co-operates except for when the opponent defects twice
     """
 
-    name = 'Slow Tit For Tat'
+    name = 'Soft Tit For Two Tats'
     classifier = {
         'memory_depth': 2,  # Four-Vector = (1.,0.,1.,0.)
         'stochastic': False,
@@ -404,14 +404,11 @@ class SlowTitForTat(Player):
     }
 
     def strategy(self, opponent):
-
-        #Co-operate twice 
-        if self.history[-1] == C:
+        if len(self.history) == 0:
             return C
 
-        #If opponent makes the same move twice, mimc
-        if opponent.history[-2] == opponent.history [-1]:
-            return opponent.history[-1]
+        if opponent.history[-2] == [D, D]:
+            return D
         
         return C
 
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -405,8 +405,9 @@ class SlowTitForTwoTats(Player):
     }
 
     def strategy(self, opponent):
+        
         #Play [c, c]
-        if len(self.history)<2:
+        if len(self.history) < 2:
             return C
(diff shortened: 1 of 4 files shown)


axelrod/strategies/titfortat.py
```diff
@@ -386,3 +386,33 @@ def reset(self):
...
+
+class SlowTitForTwoTats(Player):
+    """
+    A player that plays [c,c], then if the opponent plays the same move twice, 
+    plays that move 
+    """
+
+    name = 'Slow Tit For Two Tats'
+    classifier = {
+        'memory_depth': 2,  # Four-Vector = (1.,0.,1.,0.)
```
drvinceknight: The `# Four-Vector = (1., 0., 1., 0.)` isn't needed anymore.

drvinceknight:
This is looking good @ranjinidas! Thanks!

We're still missing a couple of tests, we'd be happy to help:

1. Would you like to do this yourself?
2. Would you like one of us to walk you through it?
3. Would you like one of us to do it?

Whatever you would prefer :)

Also, as a side note. I think (although I'm not sure) the errors you were having might have been because the doctests failed half way through and there were some files written that were not deleted... That's just a hunch though: they all passed fine on travis (a system hooked in to github that runs the test suite for every PR).

axelrod/strategies/titfortat.py
```diff
@@ -386,3 +386,33 @@ def reset(self):
         Player.reset(self)
         self.contrite = False
         self._recorded_history = []
+
+
+class SlowTitForTwoTats(Player):
+    """
+    A player that plays [c,c], then if the opponent plays the same move twice,
```
drvinceknight: Could we write "A player that plays C twice, ...". This is not terribly important: just looks nicer to me :)

ranjinidas:
Hi @drvinceknight, thanks for reviewing this! I have a test written but can check with you or someone else on Gitter before adding it in.

drvinceknight:
Awesome: I'm online now if that's helpful :)

axelrod/strategies/titfortat.py
```diff
@@ -386,3 +386,33 @@ def reset(self):
...
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent):
+        #Play [c, c]
+        if len(self.history)<2:
```
drvinceknight: Tiny bit of [PEP8](https://www.python.org/dev/peps/pep-0008/):

```
if len(self.history) < 2:
```

drvinceknight:
This looks great to me! 👍 

I've left one very minor PEP8 comment. We have a two core dev review policy so once one of the others has taken a look (and possibly added their own comments etc) this will go in :) Thanks a lot for working on this.

axelrod/strategies/titfortat.py
ranjinidas: Added this in! Thanks!

axelrod/strategies/titfortat.py
```diff
@@ -386,3 +386,34 @@ def reset(self):
...
+        'stochastic': False,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent):
+        
+        #Play [c, c]
```
marcharper: # Start with two cooperations

marcharper:
Looks good to me, we can sweep up any PEP8 issues sometime later.

PR #662, opened 2016-07-16
Non default games fixed

### LINKED ISSUE #661: Tournament is only played with default game
We pass a game to the tournament and this is recorded correctly but at no point is it passed to the result set.

```
>>> game = axl.Game(p=-1, r=-1, s=-1, t=-1)
>>> game
Axelrod game: (R,P,S,T) = (-1, -1, -1, -1)
>>> tournament = axl.Tournament(players, game=matching_pennies, turns=2, repetitions=2)
>>> tournament.game
Axelrod game: (R,P,S,T) = (-1, -1, -1, -1)
```

but:

```
>>> results = tournament.play(processes=0)
>>> results.normalised_scores
[[1.875, 1.5], [3.5, 3.0], [2.0, 2.25], [1.75, 2.375], [2.875, 2.875]]  # All positive
```

We just need to have the `ResultSet` take a game parameter and have this passed to it from the `Tournament` during play.

I'm working on that now :)

drvinceknight:
Closes #661 

The result set takes a game and the tournament is able to pass it a game.

### DIFF
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -5,6 +5,7 @@ import tqdm
 from numpy import mean, nanmedian, std
 
 from . import eigen
+from .game import Game
 import axelrod.interaction_utils as iu
 
 try:
@@ -32,8 +33,9 @@ def update_progress_bar(method):
 
 class ResultSet(object):
     """A class to hold the results of a tournament."""
+    game = Game()
 
-    def __init__(self, players, interactions, progress_bar=True):
+    def __init__(self, players, interactions, progress_bar=True, game=None):
         """
         Parameters
         ----------
@@ -45,6 +47,8 @@ class ResultSet(object):
             progress_bar : bool
                 Whether or not to create a progress bar which will be updated
         """
+        if game is not None:
+            self.game = game
         self.players = players
         self.nplayers = len(players)
         self.interactions = interactions
@@ -162,7 +166,8 @@ class ResultSet(object):
         for index_pair, repetitions in self.interactions.items():
             if index_pair[0] != index_pair[1]:  # Ignoring self interactions
                 for repetition, interaction in enumerate(repetitions):
-                    final_scores = iu.compute_final_score(interaction)
+                    final_scores = iu.compute_final_score(interaction,
+                                                          self.game)
                     for player in range(2):
                         player_index = index_pair[player]
                         player_score = final_scores[player]
@@ -211,7 +216,8 @@ class ResultSet(object):
                     player_index = index_pair[player]
 
                     for rep, interaction in enumerate(repetitions):
-                        winner_index = iu.compute_winner_index(interaction)
+                        winner_index = iu.compute_winner_index(interaction,
+                                                               self.game)
                         if winner_index is not False and player == winner_index:
                             wins[player_index][rep] += 1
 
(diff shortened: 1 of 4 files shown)


axelrod/result_set.py
```diff
@@ -32,8 +33,9 @@ def wrapper(*args):
 
 class ResultSet(object):
     """A class to hold the results of a tournament."""
+    game = Game()
```
meatballs: Why bother with this?

axelrod/result_set.py
```diff
@@ -45,6 +47,8 @@ def __init__(self, players, interactions, progress_bar=True):
             progress_bar : bool
                 Whether or not to create a progress bar which will be updated
         """
+        if game is not None:
+            self.game = game
```
meatballs: this could just be self.game = Game()

axelrod/result_set.py
```diff
@@ -687,9 +703,10 @@ class ResultSetFromFile(ResultSet):
     """A class to hold the results of a tournament. Reads in a CSV file produced
     by the tournament class.
     """
+    game = Game()
```
meatballs: Same comment as previously

axelrod/result_set.py
```diff
@@ -702,6 +719,8 @@ def __init__(self, filename, progress_bar=True,
                 displaying the progress bar, if it is not known and progress_bar
                 is set to True then an initial count will take place
         """
+        if game is not None:
+            self.game = game
```
meatballs: ```self.game = game()```?

axelrod/result_set.py
drvinceknight: I don't think that would do the trick actually.

However, what could be nicer than current is to have:

```
__init__(self, players, interactions, progress_bar=True, game=Game()):
```

and then:

```
self.game = game
```

The current setup mimics the tournament class setup which sets the class attribute to be the default game and then changes it if a different one is passed in init.

meatballs: Can't remember why off the top of my head, but that's an anti-pattern and not recommended.

meatballs: Why would my suggestion not work?

drvinceknight:
Have fixed the merge conflict.

axelrod/result_set.py
meatballs: http://docs.python-guide.org/en/latest/writing/gotchas/

drvinceknight: > Why would my suggestion not work?

I might be misunderstanding. Is your suggestion to replace the two lines:

```
        if game is not None:
              self.game = game
```

with the single line:

```
self.game = Game()
```

If that's the case then it would not allow us to pass a non default game as done in this test https://github.com/Axelrod-Python/Axelrod/pull/662/files#diff-fea0766a5edc2285adc3af49e295100cR176

(That's obviously a nonsense game but theoretically different games could be used to see effect of R, P, S, T on tournament etc...)

drvinceknight: FYI, this is how things are currently done in the tournament class: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tournament.py#L18.

drvinceknight: Cool re that antipattern. The alternative I see (if we don't like setting the class attribute) is in the init:

```
if game is None:
    self.game = Game
else:
   self.game = game
```

I can change that here and also in the tournament class.

meatballs: Yes! That!

drvinceknight: Cool. On it's way :)

meatballs: Which is what I should have suggested in the first place, but was too lazy to type out in full. sorry.

drvinceknight:
581b753 and 428cc40 change the class attribute setting to all happen in the init.

axelrod/result_set.py
drvinceknight: No problemo :)

PR #669, opened 2016-07-22
Add warning for matplotlib in virtual env on Mac

### LINKED ISSUE #668: Matplotlib within virtual environments on OSX
Matplotlib does not work (easily) within virtual environments on OSX. Full explanation at: http://matplotlib.org/faq/virtualenv_faq.html

If axelrod is installed within a virtualenv on Mac, matplotlib is installed but then a runtime error occurs on import.

meatballs:
Fix #668

### DIFF
--- a/axelrod/plot.py
+++ b/axelrod/plot.py
@@ -11,7 +11,11 @@ except ImportError:
     matplotlib_installed = False
 except RuntimeError:
     matplotlib_installed = False
-    warnings.warn('matplotlib does work with virtual environments on MacOS.')
+    warnings.warn(
+        'Matplotlib failed to import and so no plots will be produced. This ' +
+        'could be caused by using a virtual environment on OSX. See ' +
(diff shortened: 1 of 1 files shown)


axelrod/plot.py
```diff
@@ -8,7 +9,9 @@
     from mpl_toolkits.axes_grid1 import make_axes_locatable
 except ImportError:
     matplotlib_installed = False
-
+except RuntimeError:
+    matplotlib_installed = False
+    warnings.warn('matplotlib does work with virtual environments on MacOS.')
```
drvinceknight: Perhaps something like:

"If you are using a virtual environment this might be causing the error: matplotlib does **not**..." 

(Only bold facing the not because I think that's what you mean?)

Just thinking in case a run time error kicks in for a different reason?

meatballs: yep, dumb stupid typo! I'll sort it...

drvinceknight:
Looks good to me. :+1:

PR #674, opened 2016-07-31
Probabilistic ending spatial tournament.

drvinceknight:
This also adds two hypothesis test decorators (for the two spatial tournaments).

### DIFF
--- a/axelrod/match_generator.py
+++ b/axelrod/match_generator.py
@@ -168,6 +168,7 @@ class ProbEndRoundRobinMatches(RoundRobinMatches):
         size = self.__len__() * (1. / self.prob_end) * self.repetitions
         return size
 
+
 class SpatialMatches(RoundRobinMatches):
     """
     A class that generates spatially-structured matches.
@@ -207,3 +208,42 @@ class SpatialMatches(RoundRobinMatches):
 
     def __len__(self):
         return len(self.edges)
+
+
+class ProbEndSpatialMatches(SpatialMatches, ProbEndRoundRobinMatches):
+    """
+    A class that generates spatially-structured prob ending matches.
+    In these matches, players interact only with their neighbors rather than the
+    entire population. This reduces to a well-mixed population when the spatial
+    graph is a complete graph.
+
+    Parameters
+    ----------
+    players : list
+        A list of axelrod.Player objects
+    prob_end : float
+        The probability that a turn of a Match is the last
+    game : axelrod.Game
+        The game object used to score the match
+    repetitions : int
+        The number of repetitions of a given match
+    edges : list
+        A list of tuples containing the existing edges
+    """
+
+    def __init__(self, players, prob_end, game, repetitions, edges):
+
+        player_indices = list(range(len(players)))
+        node_indices = sorted(set([node for edge in edges for node in edge]))
+        if player_indices != node_indices:
+            raise ValueError("The graph edges do not include all players.")
+
+        self.edges = edges
+        ProbEndRoundRobinMatches.__init__(self, players, prob_end,
+                                          game, repetitions, edges)
+
+    def build_single_match_params(self):
+        """
+        Creates a single set of match parameters.
+        """
+        return ProbEndRoundRobinMatches.build_single_match_params(self)
--- a/axelrod/match_generator.py
+++ b/axelrod/match_generator.py
@@ -169,6 +169,31 @@ class ProbEndRoundRobinMatches(RoundRobinMatches):
         return size
 
 
+def test_graph_is_connected(edges, players):
+    """
+    Test if a set of edges defines a complete graph on a set of players.
+
+    This is used by the spatial tournaments and will raise a ValueError if the
+    graph would have some players not playing anyone.
+
+    Parameters:
+    -----------
+
+        edges : a list of 2 tuples
+        players : a list of player names
+    """
+    # Check if all players are connected.
+    player_indices = set(range(len(players)))
+    node_indices = set()
+    for edge in edges:
+        for node in edge:
+            node_indices.add(node)
+
+    if player_indices != node_indices:
+        raise ValueError("The graph edges do not include all players.")
+
+
+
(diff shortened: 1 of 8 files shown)


axelrod/match_generator.py
```diff
@@ -207,3 +208,42 @@ def build_match_chunks(self):
...
+    repetitions : int
+        The number of repetitions of a given match
+    edges : list
+        A list of tuples containing the existing edges
+    """
+
+    def __init__(self, players, prob_end, game, repetitions, noise, edges):
+
+        player_indices = list(range(len(players)))
+        node_indices = sorted(set([node for edge in edges for node in edge]))
```
marcharper: This line might be less confusing as nested for loops.

axelrod/match_generator.py
```diff
@@ -207,3 +208,42 @@ def build_match_chunks(self):
...
+        if player_indices != node_indices:
```
marcharper: If I understand correctly then these lines are just a check and the results are not needed. It might be more efficient to use sets and check for membership rather than build two lists, sort one, and check if they are equal. I think the current solution is `O(n^2)` and my proposal is better.

axelrod/match_generator.py
```diff
@@ -207,3 +208,42 @@ def build_match_chunks(self):
...
+    game : axelrod.Game
+        The game object used to score the match
+    repetitions : int
+        The number of repetitions of a given match
+    edges : list
+        A list of tuples containing the existing edges
+    """
+
+    def __init__(self, players, prob_end, game, repetitions, noise, edges):
+
```
marcharper: A comment here might improve readability.

axelrod/tests/property.py
```diff
@@ -145,6 +142,123 @@ def prob_end_tournaments(draw, strategies=axelrod.strategies,
...
+        The minimum number of strategies to include
+    max_size : integer
+        The maximum number of strategies to include
+    min_turns : integer
+        The minimum number of turns
+    max_turns : integer
+        The maximum number of turns
+    min_noise : float
+        The minimum noise value
+    min_noise : float
```
marcharper: `max_noise` (you were so close!)

axelrod/tests/property.py
```diff
@@ -145,6 +142,123 @@ def prob_end_tournaments(draw, strategies=axelrod.strategies,
...
+        The minimum number of strategies to include
+    max_size : integer
+        The maximum number of strategies to include
+    min_prob_end : float
+        The minimum probability of a match ending
+    max_prob_end : float
+        The maximum probability of a match ending
+    min_noise : float
+        The minimum noise value
+    min_noise : float
```
marcharper: `max_noise`

marcharper:
Looks good overall. One general comment -- there are a lot `axelrod.*` statements, you might consider importing some more things.

axelrod/match_generator.py
drvinceknight: Sure.

axelrod/match_generator.py
drvinceknight: Yes: good point. Will fix.

drvinceknight:
Thanks @marcharper, will address today or tomorrow :) 👍

drvinceknight:
> One general comment -- there are a lot axelrod.* statements, you might consider importing some more things.

I think these were all in `property.py`, I think I've changed them all but let me know if not :)

Is the idea that it's better to avoid full library imports in the library itself?

marcharper:
> Is the idea that it's better to avoid full library imports in the library itself? 

I just prefer shorter lines in a case like this. It can speed up loops to import a deeper reference but I don't think that's a concern here.

drvinceknight:
Makes sense. Thanks :)

drvinceknight:
defcc38 adds/improves tests for noise.

axelrod/match_generator.py
meatballs: This looks like a repeat of the code for ```SpacialMatches.__init__``` If that's the case, then it ought to be a separate function that gets called by both init methods.

drvinceknight: Good catch. Will fix. 👍

drvinceknight: ff721b8

axelrod/match_generator.py
```diff
@@ -168,6 +168,32 @@ def estimated_size(self):
         size = self.__len__() * (1. / self.prob_end) * self.repetitions
         return size
 
+
+def test_graph_is_connected(edges, players):
+    """
```
meatballs: I think this might be better as a function returning a boolean - call it ```graph_is_connected``` and return True or False. The init methods then just need a line to raise an error if False.

axelrod/match_generator.py
```diff
@@ -191,11 +217,7 @@ class SpatialMatches(RoundRobinMatches):
 
     def __init__(self, players, turns, game, repetitions, edges):
 
-        player_indices = list(range(len(players)))
-        node_indices = sorted(set([node for edge in edges for node in edge]))
-        if player_indices != node_indices:
-            raise ValueError("The graph edges do not include all players.")
-
+        test_graph_is_connected(edges, players)
```
meatballs: so this would become:
```
if not graph_is_connected():
    raise ValueError("The graph edges do not include all players.")
````

meatballs: I know that means a slight repeat of code, but it makes the function more readily understood, I reckon

drvinceknight: I agree. Just out with the little dude 🐶, will fix in a bit.

drvinceknight: ef9d93f

axelrod/tests/property.py
meatballs: Just this one left to fix and I reckon this is good to go

drvinceknight: Indeed! I thought I'd gotten it but obviously not. Fixed: f890e36

PR #677, opened 2016-08-01
Noise in Spatial Tournament

### LINKED ISSUE #673: Noise not working for spatial match tournaments
The Tournament class takes a noise parameter but this is not passed to the corresponding match generator. 

Here we see that the last parameter is 0 (it should be .5, the noise):

```
>>> players = [s() for s in axl.demo_strategies]
>>> edges = [(i, i + 1) for i in range(len(players) - 1)]
>>> tournament = axl.SpatialTournament(players=players, turns=5, edges=edges, noise=.5)
>>> params = tournament.match_generator.build_single_match_params()
>>> params
(5, Axelrod game: (R,P,S,T) = (3, 1, 0, 5), None, 0)
```

Here is the corresponding code for a Round Robin:

```
>>> players = [s() for s in axl.demo_strategies]
>>> tournament = axl.Tournament(players=players, turns=5, noise=.5)
>>> params = tournament.match_generator.build_single_match_params()
>>> params
(5, Axelrod game: (R,P,S,T) = (3, 1, 0, 5), None, 0.5)
```

It should be a straight forward fix (just adding a test and making sure the argument is passed properly to the generator).

@Nikoleta-v3 would you like to fix this (I'm happy to if you're busy)?

Nikoleta-v3:
Closes #673 

The bug seems to have been my fault as I was not passing the argument noise to either
the `SpatialMatches` generator or the `SpatialTournament` class. 

It has been corrected and also I added a test for the noise in the `test_match_generator` arcooding to @drvinceknight work in the `TestProbEndSpatialTournament`. 

Locally all 1595 tests seems to be running fine. Let's see what Travis has to say.

### DIFF
--- a/axelrod/tests/unit/test_match_generator.py
+++ b/axelrod/tests/unit/test_match_generator.py
@@ -225,7 +225,8 @@ class TestSpatialMatches(unittest.TestCase):
             self.players, turns, test_game, repetitions, noise, edges)
         chunks = list(sp.build_match_chunks())
         match_definitions = [tuple(list(index_pair) + [repetitions])
-                             for (index_pair, match_params, repetitions) in chunks]
+                             for (index_pair, match_params, repetitions, noise)
+                                                                      in chunks]
         expected_match_definitions = [(edge[0], edge[1], repetitions)
                                       for edge in edges]
 
--- a/axelrod/tests/unit/test_match_generator.py
+++ b/axelrod/tests/unit/test_match_generator.py
@@ -44,7 +44,7 @@ class TestTournamentType(unittest.TestCase):
         with self.assertRaises(NotImplementedError):
             len(tt)
 
-    def _params(self):
+    def test_build__params(self):
         tt = axelrod.MatchGenerator(
             self.players, test_turns, test_game, test_repetitions)
         with self.assertRaises(NotImplementedError):
--- a/axelrod/tests/unit/test_match_generator.py
+++ b/axelrod/tests/unit/test_match_generator.py
@@ -44,7 +44,7 @@ class TestTournamentType(unittest.TestCase):
         with self.assertRaises(NotImplementedError):
             len(tt)
 
-    def test_build_match_params(self):
+    def _params(self):
         tt = axelrod.MatchGenerator(
             self.players, test_turns, test_game, test_repetitions)
         with self.assertRaises(NotImplementedError):
@@ -99,7 +99,7 @@ class TestRoundRobin(unittest.TestCase):
 
     @given(repetitions=integers(min_value=1, max_value=test_repetitions))
     @example(repetitions=test_repetitions)
-    def test_build_match_chunks(self, repetitions):
+    def _chunks(self, repetitions):
         rr = axelrod.RoundRobinMatches(self.players, test_turns, test_game, repetitions)
         chunks = list(rr.build_match_chunks())
         match_definitions = [tuple(list(index_pair) + [repetitions]) for (index_pair, match_params, repetitions) in chunks]
@@ -125,7 +125,7 @@ class TestProbEndRoundRobin(unittest.TestCase):
     @given(repetitions=integers(min_value=1, max_value=test_repetitions),
            prob_end=floats(min_value=0, max_value=1))
     @example(repetitions=test_repetitions, prob_end=.5)
-    def test_build_match_chunks(self, repetitions, prob_end):
+    def _chunks(self, repetitions, prob_end):
         rr = axelrod.ProbEndRoundRobinMatches(
             self.players, prob_end, test_game, repetitions)
         chunks = list(rr.build_match_chunks())
@@ -135,7 +135,7 @@ class TestProbEndRoundRobin(unittest.TestCase):
         self.assertEqual(sorted(match_definitions), sorted(expected_match_definitions))
 
     @given(prob_end=floats(min_value=0.1, max_value=0.5))
-    def test_build_matches_different_length(self, prob_end):
+    def es_different_length(self, prob_end):
         """
         If prob end is not 0 or 1 then the matches should all have different
         length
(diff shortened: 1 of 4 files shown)


axelrod/tests/unit/test_match_generator.py
```diff
@@ -44,7 +44,7 @@ def test_len(self):
         with self.assertRaises(NotImplementedError):
             len(tt)
 
-    def test_build_match_params(self):
+    def _params(self):
```
drvinceknight: you seem to have lost `test_build_match_`. I expect this test was not running (always make sure your test fails before it passes :)).

axelrod/tests/unit/test_match_generator.py
```diff
@@ -218,10 +218,11 @@ def setUpClass(cls):
     @given(repetitions=integers(min_value=1, max_value=test_repetitions),
            turns=integers(min_value=1, max_value=test_turns))
     @example(repetitions=test_repetitions, turns=test_turns)
-    def test_build_match_chunks(self, repetitions, turns):
+    def _chunks(self, repetitions, turns):
```
drvinceknight: Same comment up until here.

axelrod/tournament.py
```diff
@@ -376,4 +376,4 @@ def __init__(self, players, edges, match_generator=SpatialMatches,
 
         self.edges = edges
         self.match_generator = SpatialMatches(
-            players, turns, self.game, repetitions, edges)
+            players, turns, self.game, repetitions, self.noise, edges)
```
drvinceknight: Could you add an assertion in the init test that checks that when a spatial tournament is initialised with noise, it's match generator has the same noise value?

```
tournament = axelrod.SpatialTournament(...)
self.assertEqual(tournament.match_generator.noise, noise) 
```

Something like that.

axelrod/tests/unit/test_match_generator.py
```diff
@@ -44,7 +44,7 @@ def test_len(self):
         with self.assertRaises(NotImplementedError):
             len(tt)
 
-    def test_build_match_params(self):
+    def test_build__params(self):
```
drvinceknight: I see you fixed this but I would just fix it (and the others) back to what it was (just read the diff here):

```
test_build_match_params
```

axelrod/tests/unit/test_match_generator.py
Nikoleta-v3: no idea how I did this -_-

axelrod/tests/unit/test_match_generator.py
```diff
@@ -99,7 +99,7 @@ def test_build_single_match_params(self):
 
     @given(repetitions=integers(min_value=1, max_value=test_repetitions))
     @example(repetitions=test_repetitions)
-    def test_build_match_chunks(self, repetitions):
+    def _chunks(self, repetitions):
```
drvinceknight:
```
test_build_match_chunks
```

axelrod/tests/unit/test_match_generator.py
```diff
@@ -125,7 +125,7 @@ def setUpClass(cls):
     @given(repetitions=integers(min_value=1, max_value=test_repetitions),
            prob_end=floats(min_value=0, max_value=1))
     @example(repetitions=test_repetitions, prob_end=.5)
-    def test_build_match_chunks(self, repetitions, prob_end):
+    def _chunks(self, repetitions, prob_end):
```
drvinceknight: If this has got the same name as the other one it shouldn't as they actually overwrite so rename this to something different...

axelrod/tests/unit/test_match_generator.py
```diff
@@ -135,7 +135,7 @@ def test_build_match_chunks(self, repetitions, prob_end):
         self.assertEqual(sorted(match_definitions), sorted(expected_match_definitions))
 
     @given(prob_end=floats(min_value=0.1, max_value=0.5))
-    def test_build_matches_different_length(self, prob_end):
+    def es_different_length(self, prob_end):
```
drvinceknight:
```
test_build_matches_different_length
```

axelrod/tests/unit/test_match_generator.py
drvinceknight: Same comment.

axelrod/tournament.py
Nikoleta-v3: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_tournament.py#L595

`self.assertEqual(tournament.noise, 0.2)`

Should I add an extra test anyway or does this cover it ? I had this test written initial without realising I was not passing noise.

drvinceknight: Yeah add another one, that just tests that noise has been passed to the tournament, not that the tournament has passed the noise to it's generator. Which is what the bug is here.

So have:

```
self.assertEqual(tournament.noise, tournament.match_generator.noise)
```

marcharper:
@Nikoleta-v3 Please see the PR I opened on your branch.

drvinceknight:
> We need to add a test that actually runs a spatial tournament to catch the bug I just fixed...

I wondered how the doctest caught something our unit tests didn't... Any idea why this test does not catch whatever the bug was: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_tournament.py#L621

(Haven't really spoken to @Nikoleta-v3 today, I know she's busy with some writing.)

marcharper:
I'm not sure really but it was a tuple unpacking issue for match chunks so I would think that one of the other tests would have invoked it.

drvinceknight:
Strange. Definitely worth including the example from the doc test as a unit test to make sure it doesn't happen again (and that we're not counting on the doctests to find bugs) :)

drvinceknight:
I have opened https://github.com/Nikoleta-v3/Axelrod/pull/2 with the tests that I feel are necessary. Note that I opened that to your master branch as I didn't go through your fixes: we can chat about this when we catch up today.

drvinceknight:
Have spoken to @Nikoleta-v3 and she's asked that I just submit my fix to this as she's had a bunch of data back that needs analysing. I'm closing this and about to open another PR.

Nikoleta-v3:
@marcharper  Thank you for the comments ! I read everything through :smile:

Issue #225, opened 2015-07-20
Add sources for the definitions of strategies


marcharper:
Right now we have a lot of strategies (expecting more, I presume), and there are a few issues that make sense to stay on top of: (1) identifying the original source of a strategy, and (2) tracking multiple names for a given strategy.

Ideally we would have a list of sources that we can cite in various places -- the strategy overview docs as well as the doc strings for the various strategies. I would also help to have an index of strategies of sources indicating what has been implemented, including strategies that have multiple names. That way if someone is either looking for an existing strategy (and doesn't know that WSLS and Pavlov are the same thing) or wants to add a strategy, they can tell if it's already present without digging through all the code (as it's not necessarily obvious which file a strategy would even be in).

I'm imaging an documentation *.rst file that has an alphabetical list of strategy names and alternates along with the library name and citations to the original source of the strategy. Of course I'm open to other suggestions.

Right now the overview of strategy doc is fine but it is organized around existing tournaments, and there are many strategies in the library already that are not in one of these tournaments (and many still from the tournaments that need to implemented).

drvinceknight:
> Ideally we would have a list of sources that we can cite in various places -- the strategy overview docs as well as the doc strings for the various strategies. I would also help to have an index of strategies of sources indicating what has been implemented, including strategies that have multiple names. That way if someone is either looking for an existing strategy (and doesn't know that WSLS and Pavlov are the same thing) or wants to add a strategy, they can tell if it's already present without digging through all the code (as it's not necessarily obvious which file a strategy would even be in).

I'm not too sure I understand how this differs from the current list of strategies that we're working on. #199 was brought in (unless I'm misunderstanding) with the same goal that you're suggesting. I think it was there decided that indexing by previous tournaments followed by the newer ones would be the way to go.

> 
> I'm imaging an documentation *.rst file that has an alphabetical list of strategy names and alternates along with the library name and citations to the original source of the strategy. Of course I'm open to other suggestions.

I think that including the sources in the docstrings and using the semi-automatically generated index would take care of this would it not?

Perhaps I'm missing something but I also don't want to fragment the documentation of strategies too much.

For example for Djaxelrod we have https://github.com/Axelrod-Python/DjAxelrod/issues/35 which is going to enable a hover box (or similar) which would read the docstrings... 

I just think that keeping the index short and snappy (ie semi automatically generated) as well as the curated context list is enough no? Delighted to discuss though, am perhaps overlooking something. :) :+1: 

[...]

PR #683, opened 2016-08-08
Added strategies Desperate, Hopeless, Willing, and Grim. 

ranjinidas:
All four strategies are contained in the file mutual, since they follow a similar theme. The commits became a little messy after I had to undo a push.


marcharper:
Hi @ranjinidas! Thanks for the contribution.

I think you may have added the build directory post installation to the repository. Can you try adding a new commit that removes the files in the build directory. If you need help with that just let me know.

Also I believe that GRIM is the same as Grudger -- I think that the other three strategies are not yet in the library :) Let's check that before merging.

drvinceknight:
These look great @ranjinidas, your Grim strategy is not the same as the Grudger strategy in the library but as @marcharper said that is another common name for Grudger (a strategy that will defect forever if ever defected against). Did you have a reason for that particular name? Perhaps we could find another one.

I don't think the other strategies are in there yet but once you remove the build files it'll be easier to look through and be sure :)

The tests failing on travis are two simple doctests:

1. `len([s for s in axl.strategies if s().classifier['memory_depth']==1])` is returning 27 now and was returning 21 (not sure why there's a jump of 6 strategies...)
2. The other is in the `docs/tutorials/advanced/setting_a_seed.rst`, everything is still working but the way I've written that isn't great. I'm going to open a PR to your branch with a way that makes that right.

drvinceknight:
The [PR is open](https://github.com/ranjinidas/Axelrod/pull/1) but it looks like it's grabbed quite a few commits (not too sure why, I branched from your branch...). The change it makes is pretty simple so you might just want to implement it yourself. 

You can see it here: https://github.com/ranjinidas/Axelrod/pull/1/commits/52452f75569ad96a08f524b42fcfe53990fae33b

Let us know if we can help with anything else :) 👍

drvinceknight:
> The PR is open but it looks like it's grabbed quite a few commits (not too sure why, I branched from your branch...). The change it makes is pretty simple so you might just want to implement it yourself.
> 
> You can see it here: ranjinidas@52452f7

Ignore all that (sorry!).

I see now that your master branch is a few commits behind the master so the change I'm suggesting wouldn't make much sense as that file isn't there. I've opened a simple PR that once it  goes in will make that test pass.

ranjinidas:
Hi @marcharper and @drvinceknight! Thanks for the feedback. I got the strategies here: http://www.rug.nl/research/gelifes/tres/publications/_pdf/vdberg_weissing_prsb_2015.pdf, where Grim was defined as "only cooperates after mutual cooperation". Do you have any suggestions for what I could change the name to? 

Also, sorry about getting the build directory in there. I'll remove it and commit again.

marcharper:
@drvinceknight Are you sure they aren't the same? Based on the reference, it looks like a memory-one strategy with four-vector (1, 0, 0, 0), which seems like a strategy that cooperates initially and then defects forever if ever defected against. (Notice that all of the strategies in the reference are memory-one and that we have many; would be nice to have them all).

I think we're supposed to assume that GRIM cooperates on the first move (despite the description in the reference and @ranjinidas's implementation) -- otherwise it will never cooperate. Based on that assumption I think it is the same as grudger.

Finally, we (me, I suppose) should prioritize #225 and also a dictionary or convenient reference in the docs for strategies that have multiple common names.

drvinceknight:
Sorry you're completely correct. It is the same as Grudger! So it can be
removed.

drvinceknight:
After @meatballs merged #685 the only test that is now failing is this doctest:

```
>>> len([s for s in axl.strategies if s().classifier['memory_depth']==1])
21
```

That's in: `docs/tutorials/advanced/classification_of_strategies.rst`.

Your branch is returning 27. Not entirely sure why that's gone up by 6 but perhaps once you've removed the build files it'll be easier to figure it out. :) 👍

drvinceknight:
This looks great now. As far as I can see it's just that one doctest that needs fixing (now indeed expecting 24) 👍

ranjinidas:
Thanks!

marcharper:
Two more things, if you don't mind:
* Can we double check what the strategies do on the first move? Just want to make sure we've got it right, since for these types of strategies the often either always cooperate first or behave randomly
* Can you squash commits e4e4fd4 through 4280e15 to a single commit? This will keep the files in the build directory from being added to the git tracking objects. To do so, run the following commands, assuming your repositories
```
git branch backup
git rebase -i origin master
```
This will bring up a list of commits. You'll need to change "pick" to "squash"  for commits 82a27cb through 4280e15. Git will ask you to modify the commit message (the default is fine).

Then run `git push -f` (to your master branch on github). If you need help with this just let me know. I've got a local copy of your branch, but I want to make sure you get credit for your work (so it's best if you squash the commits).

drvinceknight:
The mutual module needs to be added to the documentation so that the three strategies are automatically added to the index in the docs. This is explained in the last paragraph here http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/adding_the_new_strategy.html  (on my phone and about to head to bed, can be more helpful tomorrow if needed :)).

ranjinidas:
@marcharper The strategy does not specify a first move. I thought since each strategy definition says that the player /always/ plays a move unless prompted otherwise, it would be safe to have the start move set to the default move. I can remove the first move test, though, if you think I should. 

@drvinceknight Added it in! :)

marcharper:
In the reference it says: " In addition, a separate locus determines an individual’s
behaviour in the first round; this locus can take on any value
of the unit interval..."

and also

"At the beginning of each simulation, the loci of all indi-
viduals were initialized at random: initial values for the
locus that determines the behaviour in the first round were
drawn from a normal distribution with mean 0.5 and stan-
dard  deviation  0.1;  the  four  binary  loci  in  the  1 : 1
architecture were assigned a 0 or a 1 with equal probability;
and the nine loci encoding the connection weights and
thresholds in the neural network architecture were assigned
values that were drawn from a normal distribution with
mean 0 and standard deviation [sigma_N]"

So it looks like the first move is random somehow.  Maybe that's what we should do: `random_choice()` from our `axelrod.random_.py` file.

ranjinidas:
Okay, I will look into that! Thank you! I'm also having some issues with the rebase. I mentioned them on Gitter, but will bring it over here if you prefer.

marcharper:
Closing re: #686, please move discussion there and re-open if there are any issues with my rebase.

PR #692, opened 2016-08-11
Add long_run_time classifier

### LINKED ISSUE #690: Add long run time to player classifier
See chat on #689, idea is to add `long_run_time` to the classifier dict.  We could then change:

```
long_run_time_strategies = [MetaMajority, MetaMinority, MetaWinner,
                   MetaMajorityMemoryOne, MetaWinnerMemoryOne,
                   MetaMajorityFiniteMemory, MetaWinnerFiniteMemory,
                   MetaMajorityLongMemory, MetaWinnerLongMemory, MetaMixer]
```

to be

```
long_run_time_strategies = [s for s in all_strategies if not s.classifier['long_run_time']]
```

This could be `False` by default so we'd only need to actually add it to the `long_run_time_strategies`. All the tests are in place that should pick it up for the lists and we could add a test to the global test player class too...

drvinceknight:
Closes #690.

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -73,6 +73,7 @@ class Player(object):
         'stochastic': False,
         'memory_depth': float('inf'),
         'makes_use_of': None,
+        'long_run_time': False,
         'inspects_source': None,
         'manipulates_source': None,
         'manipulates_state': None
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -12,21 +12,20 @@ from .meta import (
     MetaMixer
     )
 
-all_strategies.append(MetaHunter)
+all_strategies.extend([MetaHunter, MetaMajority, MetaMinority, MetaWinner,
+                       MetaMajorityMemoryOne, MetaWinnerMemoryOne,
+                       MetaMajorityFiniteMemory, MetaWinnerFiniteMemory,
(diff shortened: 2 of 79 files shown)


marcharper:
This is good. Maybe we should have a standard run time metric, like run time versus TFT over 100 runs or similar.

drvinceknight:
> Maybe we should have a standard run time metric, like run time versus TFT over 100 runs or similar.

I quite liked the timing ranking you put together here https://github.com/Axelrod-Python/Axelrod-notebooks/blob/master/Axelrod-Strategy-Times.ipynb. Could we not just re run that if and when a strategy comes in that we think might have a long run time?

marcharper:
Sure, I'm just suggesting an official definition for newcomers. In the spreadsheet associated to that notebook I standardized the values by dividing by the run time of one of the deterministic strategies. So I'm suggesting that we say some like run time is considered long if it's > 10x TFT (or whatever), and link to the notebook for a timing function of some kind that compares e.g. 100 matches of TFT vs TFT and 100 of TFT vs new_strategy.

We can discuss outside the PR, I think it's fine as is.

drvinceknight:
Cool I see what you mean. Sounds good, perhaps a number of standard
deviations from the mean time of all strategies? Just a suggestion, I don't
have a strong opinion on what definition/guideline we go for.

PR #686, opened 2016-08-10
Added strategies Desperate, Willing, Hopeless, Grim

marcharper:
Adding strategies/tests Desperate, Hopeless, Willing and Grim in mutual.py

This is a rebase of PR #683 by @ranjinidas

PR #693, opened 2016-08-13
Pushing Mutual strategies from mutual2

ranjinidas:


ranjinidas:
Should I just remove first_play_test altogether? There are two failing right now because of random_choice().

marcharper:
What we usually do is to set the random seed to something specific so that we know what the move will be each time. We'll add extra tests as needed, so feel free to remove the failing tests for now.

ranjinidas:
Okay, just removed them and I'll look out for the extra tests and random seed so I know what to use next time. Thanks!

marcharper:
Great I'll take it from here.

marcharper:
I moved the new commits to #686 .

PR #694, opened 2016-08-15
Unique strategy id and index

### LINKED ISSUE #688: Unique ID for Player
The axelrod-api project currently generates a unique id for each Player class and a dictionary with the id as its key. These could both be done within the library itself and used more widely than just the api.

The unique id is currently produced using a function which strips the whitespace from the player name and converts it to lower case. This could be done with a property in the Player class.

The dictionary mapping id to class is done using a comprehension based on the existing strategies lists and could be added to ```strategies.__init__.py```

The unique id function is currently ```strategy_id``` within https://github.com/Axelrod-Python/axelrod-api/blob/master/api/core/serializers.py

The dictionary is a class variable within https://github.com/Axelrod-Python/axelrod-api/blob/master/api/core/views.py

meatballs:
closes #688

### DIFF
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -10,13 +10,18 @@ from .meta import (
     MetaMajorityMemoryOne, MetaWinnerMemoryOne, MetaMajorityFiniteMemory,
     MetaWinnerFiniteMemory, MetaMajorityLongMemory, MetaWinnerLongMemory,
     MetaMixer
-    )
+)
 
-all_strategies.extend([MetaHunter, MetaMajority, MetaMinority, MetaWinner,
-                       MetaMajorityMemoryOne, MetaWinnerMemoryOne,
-                       MetaMajorityFiniteMemory, MetaWinnerFiniteMemory,
-                       MetaMajorityLongMemory, MetaWinnerLongMemory, MetaMixer])
+all_strategies.extend([
+    MetaHunter, MetaMajority, MetaMinority, MetaWinner, MetaMajorityMemoryOne,
+    MetaWinnerMemoryOne, MetaMajorityFiniteMemory, MetaWinnerFiniteMemory,
+    MetaMajorityLongMemory, MetaWinnerLongMemory, MetaMixer])
 
+
+def strategy_id(strategy):
+    return strategy.name.lower().replace(' ', '_')
+
+strategy_index = {strategy_id(s): s for s in all_strategies}
 
 # Distinguished strategy collections in addition to
 # `all_strategies` from _strategies.py
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -105,12 +105,20 @@ class Player(object):
         self.id = self._unique_id()
 
     def _unique_id(self):
+        """
+        Returns a unique id based on the player name and any initialisation
+        arguments.
+        """
         id = [self.name.replace(' ', '_').lower()]
         if self.init_args:
             id.extend(['_', self._hashed_arguments()])
         return ''.join(id)
 
     def _hashed_arguments(self):
+        """
+        Returns a url safe, md5 hash of the player's initialisation arguments
+        serialised as a json string.
+        """
         serialised_args = json.dumps(self.init_args).encode('utf-8')
         md5_args = hashlib.md5(serialised_args).digest()
         base64_args = base64.urlsafe_b64encode(md5_args).decode('utf-8')
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -102,6 +102,7 @@ class Player(object):
         self.defections = 0
         self.init_args = ()
         self.set_match_attributes()
+        self.id = self._unique_id()
 
     def _unique_id(self):
         id = [self.name.replace(' ', '_').lower()]
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -23,12 +23,19 @@ def is_basic(s):
     inspects_source = s.classifier['inspects_source']
     manipulates_source = s.classifier['manipulates_source']
     manipulates_state = s.classifier['manipulates_state']
-    return (not stochastic) and (not inspects_source) and (not manipulates_source) and (not manipulates_state) and (depth in (0, 1))
+    return (
+        not stochastic
+        and not inspects_source
+        and not manipulates_source
+        and not manipulates_state
+        and depth in (0, 1)
+    )
 
 
 def obey_axelrod(s):
     """
-    A function to check if a strategy obeys Axelrod's original tournament rules.
+    A function to check if a strategy obeys Axelrod's original tournament
+    rules.
     """
     classifier = s.classifier
     return not (
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -1,6 +1,9 @@
 from functools import wraps
 import random
 import copy
+import json
+import hashlib
+import base64
 
 from axelrod import Actions, flip_action
 from .game import DefaultGame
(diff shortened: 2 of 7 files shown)


marcharper:
This is my fault, I removed the encoding header line in my last cleanup PR. Just add `# -*- coding: utf-8 -*-` to `test_player.py` to fix it.

axelrod/strategies/__init__.py
```diff
@@ -10,14 +10,19 @@
...
+all_strategies.extend([
+    MetaHunter, MetaMajority, MetaMinority, MetaWinner, MetaMajorityMemoryOne,
+    MetaWinnerMemoryOne, MetaMajorityFiniteMemory, MetaWinnerFiniteMemory,
+    MetaMajorityLongMemory, MetaWinnerLongMemory, MetaMixer])
 
 
+def strategy_id(strategy):
+    return strategy.name.lower().replace(' ', '_')
+
+strategy_index = {strategy_id(s): s for s in all_strategies}
```
drvinceknight: Is `index` the right word? Makes me think of a number... Not terribly offended by index though.

drvinceknight:
This looks good to me.

- Do we want to add this id as an attribute to the strategies? Thinking of the use case of writing them to data files.
- Could we have a test for `strategy_index`? Perhaps in here: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_classification.py#L131, a check that all strategies have ids that are keys and also just hard checking a few of them?

axelrod/strategies/__init__.py
meatballs: It makes me think of an index at the back of a book, or a database index - something where you look things up!!

Not too bothered myself, if there's a better suggestion.

meatballs:
> Do we want to add this id as an attribute to the strategies? Thinking of the use case of writing them to data files.

Not sure I quite understand the use case - I don't have any great objection, just want to understand!!

It would need to be a class attribute, if that makes any difference

meatballs:
> Could we have a test for strategy_index? Perhaps in here: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_classification.py#L131, a check that all strategies have ids that are keys and also just hard checking a few of them?

OK. I'll put something together

meatballs:
I've added two tests for the strategy index:

1. ensure that all strategies have an id that is included in the index. This test passes OK.

1. ensure that the lengths of the all_strategies list and the strategy_index dictionary are equal. (i.e. there are no duplicate strategy IDs). This test fails.

meatballs:
Code to flush out the duplicates:
```
import axelrod as axl
import collections

ids = [axl.strategy_id(s) for s in axl.all_strategies]
print([(item, count) for item, count in collections.Counter(ids).items() if count > 1])
```

which gives:
```
[('player', 16), ('cycler', 5)]
```

meatballs:
```
print([s for s in axl.all_strategies if s.name in ('Player', 'Cycler')])
```
gives
```
[<class 'axelrod.strategies.cycler.CyclerCCCCCD'>, <class 'axelrod.strategies.cycler.CyclerCCCD'>, <class 'axelrod.strategies.cycler.CyclerCCD'>, <class 'axelrod.strategies.cycler.CyclerDC'>, <class 'axelrod.strategies.cycler.CyclerDDC'>, <class 'axelrod.strategies.gobymajority.GoByMajority'>, <class 'axelrod.strategies.gobymajority.GoByMajority10'>, <class 'axelrod.strategies.gobymajority.GoByMajority20'>, <class 'axelrod.strategies.gobymajority.GoByMajority40'>, <class 'axelrod.strategies.gobymajority.GoByMajority5'>, <class 'axelrod.strategies.gobymajority.HardGoByMajority'>, <class 'axelrod.strategies.gobymajority.HardGoByMajority10'>, <class 'axelrod.strategies.gobymajority.HardGoByMajority20'>, <class 'axelrod.strategies.gobymajority.HardGoByMajority40'>, <class 'axelrod.strategies.gobymajority.HardGoByMajority5'>, <class 'axelrod.strategies.retaliate.LimitedRetaliate'>, <class 'axelrod.strategies.retaliate.LimitedRetaliate2'>, <class 'axelrod.strategies.retaliate.LimitedRetaliate3'>, <class 'axelrod.strategies.retaliate.Retaliate'>, <class 'axelrod.strategies.retaliate.Retaliate2'>, <class 'axelrod.strategies.retaliate.Retaliate3'>]
```

meatballs:
the remaining 16 with an id of 'player' are those that use init_args to set the name. Since they only do so on instantiation, the class attribute still has 'Player' as its name'

axelrod/strategies/__init__.py
drvinceknight: No I think now that you've said that I'm cool with index. Ignore me :)

drvinceknight:
> Not sure I quite understand the use case - I don't have any great objection, just want to understand!!
> 
> It would need to be a class attribute, if that makes any difference

Class attribute would be fine. 

So at the moment @Nikoleta-v3 is dealing with a dataset of the following form (that's she's generating):

```
Player 1 name, Player 2 name,...
Cooperator, Tit For Tat,...
```

where this dictionary becomes useful is if instead the data set was:


```
Player 1 id, Player 2 id,...
id1, id2,...
```

So that if we were to read in the data set we could obtain the classes by going:

```
strategy_index[id1]
```

**So...** to obtain the id in the first place we would need to write: `player.id` (or whatever) to the data set). 

This was what I was thinking you meant when wrote (#668):

> The unique id is currently produced using a function which strips the whitespace from the player name and converts it to lower case. This could be done with a property in the Player class.

Without the players themselves having the id I'm not sure I understand the use case for the index dictionary? (Been a long day and I've just had a walk in the sun so I might be missing something).

> the remaining 16 with an id of 'player' are those that use init_args to set the name. Since they only do so on instantiation, the class attribute still has 'Player' as its name'

That's a pity. Perhaps if we indeed put a the id as a class property, then we could override it for these tricky ones and change the generation of the index to just grab the id from the class itself?

```
strategy_index = {s.id: s for s in all_strategies}
```

One other point (when we get there): could you add a brief line or two about this index to the docs: http://axelrod.readthedocs.io/en/latest/tutorials/advanced/classification_of_strategies.html

marcharper:
Isn't the strategy name already a unique id at the class level? A new class attribute won't help with:
* Multiple instances, or
* Different parameters

drvinceknight:
Good point.

drvinceknight:
> Good point.

The use case I described above could actually easily be taken care of using this dictionary:

```
d = {s().name:s for s in axl.strategies}
```

Perhaps, if what this PR is doing is creating a dictionary that maps lower case names with stripped white space (needed for the api right? for urls?) to classes, then this is better located in the api and doesn't add anything to the library?

meatballs:
> Perhaps, if what this PR is doing is creating a dictionary that maps lower case names with stripped white space (needed for the api right? for urls?) to classes, then this is better located in the api and doesn't add anything to the library?

Possibly, but it's also adding the index - yes, we could do that with the name as the key, but then I'd just end up repeating virtually the same index in the api but using the unique_id instead.

meatballs:
so, if the index is no use, then this should just be in the api. However, if the index is useful then both should be in the library to keep things DRY between it and the api.

drvinceknight:
I agree, both here or both in the api.

My current point of view is that as is the index is not useful. To use it,
a library user needs to know the name (stripped or not or equivalent) which
gives  access to the class but that's equivalent to knowing the class and
calling it directly.

meatballs:
I think I'm coming round to that view myself.

drvinceknight:
I suggest we close this PR?

Looking through the commit messages it looks like you might have fished out a couple of things that were wrong with some `.name`s? Are those worth cherry picking?

meatballs:
Yep. I'll close this PR and submit a new one. I need the names fixing in order for the indexing to work properly in the api.

PR #698, opened 2016-08-24
Add build directory to gitignore.

drvinceknight:
This has caught people out on a couple of recent PRs (#697 and #693). I don't actually
know what this directory is for and/or where it comes from (never seen
it on my machine).


marcharper:
I think it's from `python setup.py install`?

PR #699, opened 2016-08-26
Filtering Strategies

### LINKED ISSUE #696: Strategy Filtering by Dictionary
Within the api, there are a set of functions to filter strategies from their classifier entries using a dictionary. (This is because parameters within a url are passed as dictionary).

For example, the url https://axelrod-api.herokuapp.com/strategies/?stochastic=false&inspects_source=true passes the following dictionary:
```
{
    'stochastic': 'false',
    'inspects_source': 'true'
}
```
and it's then used to generate the list of strategies which satisfy those filtering criteria.

We could implement that functionality within the library itself, so that we could do something like:
```
filter = {
    'stochastic': 'false',
    'inspects_source': 'true'
}
print(axelrod.filtered_strategies(filter))
```

This wouldn't add anything that couldn't already be done with a list comprehension. The same filtering can be done already with:
```
print([[s in axl.strategies if not s.classifier[’stochastic’] and s.classifier[‘inspects_source']]))
```

However, it would add an extra capability to the library rather than it existing only within the api.

meatballs:
This code fixes #696 

I've not written the documentation yet, which is why this PR is not ready to merge.

I'd appreciate thoughts and comments on the code and the tests. Once we're happy with those, I'll finish off the PR by completing the documentation.

### DIFF
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -38,6 +38,17 @@ def filtered_strategies(filterset, strategies=all_strategies):
     strategy classes which pass all of those filters from the given list of
     strategies.
 
+    e.g.
+
+    For the filterset dict:
+        {
+            'stochastic': True,
+            'min_memory_depth': 2
+        }
+
+    the function will return a list of all deterministic strategies with a
+    memory_depth of 2 or more.
+
     Parameters
     ----------
         filterset : dict
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -1,5 +1,6 @@
 from ..player import is_basic, obey_axelrod
 from ._strategies import *
+from ._filters import passes_filterset
 
 # `from ._strategies import *` import the collection `strategies`
 # Now import the Meta strategies. This cannot be done in _strategies
@@ -29,3 +30,9 @@ long_run_time_strategies = [s for s in all_strategies if
 cheating_strategies = [s for s in all_strategies if not obey_axelrod(s())]
 
 ordinary_strategies = strategies  # This is a legacy and will be removed
+
+
+def filtered_strategies(filterset, strategies=all_strategies):
+    return [
+        s for s in strategies
+        if passes_filterset(s, filterset)]
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -33,6 +33,30 @@ ordinary_strategies = strategies  # This is a legacy and will be removed
 
 
 def filtered_strategies(filterset, strategies=all_strategies):
+    """
+    Applies the filters defined in the given filterset dict and returns those
+    strategy classes which pass all of those filters from the given list of
+    strategies.
+
+    Parameters
+    ----------
+        filterset : dict
+            mapping filter name to criterion.
+            e.g.
+                {
+                    'stochastic': True,
+                    'min_memory_depth': 2
+                }
+        strategies: list
+            of subclasses of axelrod.Player
+
+    Returns
+    -------
+        list
+
+        of subclasses of axelrod.Player
+
+    """
     return [
         s for s in strategies
         if passes_filterset(s, filterset)]
--- a/axelrod/strategies/_filters.py
+++ b/axelrod/strategies/_filters.py
@@ -1,0 +1,81 @@
+from distutils.util import strtobool
+from collections import namedtuple
+import operator
+
+
+def passes_boolean_filter(strategy, value, classifier):
+    if isinstance(value, str):
+        filter_value = strtobool(value)
+    else:
+        filter_value = value
+
+    return strategy.classifier[classifier] == filter_value
+
+
+def passes_operator_filter(strategy, value, classifier, operator):
+    if isinstance(value, str):
+        filter_value = int(value)
+    else:
+        filter_value = value
+
+    classifier_value = strategy.classifier[classifier]
+    if (isinstance(classifier_value, str) and
+            classifier_value.lower() == 'infinity'):
+        classifier_value = float('inf')
+
+    return operator(classifier_value, filter_value)
+
+
+def passes_in_list_filter(strategy, value, classifier):
+    return value in strategy.classifier[classifier]
+
+
+def passes_filterset(strategy, filterset):
+    """
+    Parameters
+    ----------
+        strategy : a descendant class of axelrod.Player
+        filterset : dictionary
+
+    Returns
+    -------
+        boolean
+    """
+    FilterFunction = namedtuple('FilterFunction', 'function kwargs')
+    filter_functions = {
+        'stochastic': FilterFunction(
+            function=passes_boolean_filter,
+            kwargs={'classifier': 'stochastic'}),
+        'long_run_time': FilterFunction(
+            function=passes_boolean_filter,
+            kwargs={'classifier': 'long_run_time'}),
+        'manipulates_state': FilterFunction(
+            function=passes_boolean_filter,
+            kwargs={'classifier': 'manipulates_state'}),
+        'manipulates_source': FilterFunction(
+            function=passes_boolean_filter,
+            kwargs={'classifier': 'manipulates_source'}),
+        'inspects_source': FilterFunction(
+            function=passes_boolean_filter,
+            kwargs={'classifier': 'inspects_source'}),
+        'min_memory_depth': FilterFunction(
+            function=passes_operator_filter,
+            kwargs={'classifier': 'memory_depth', 'operator': operator.ge}),
+        'max_memory_depth': FilterFunction(
+            function=passes_operator_filter,
+            kwargs={'classifier': 'memory_depth', 'operator': operator.le}),
+        'makes_use_of': FilterFunction(
+            function=passes_in_list_filter,
+            kwargs={'classifier': 'makes_use_of'})
+    }
+    passes_filters = []
+
+    for filter, filter_function in filter_functions.items():
+
+        if filterset.get(filter, None) is not None:
+            kwargs = filter_function.kwargs
+            kwargs['strategy'] = strategy
+            kwargs['value'] = filterset[filter]
+            passes_filters.append(filter_function.function(**kwargs))
+
+    return all(passes_filters)
--- a/axelrod/strategies/_filters.py
+++ b/axelrod/strategies/_filters.py
@@ -37,8 +37,8 @@ def passes_operator_filter(strategy, classifier_key, value, operator):
 
 def passes_in_list_filter(strategy, classifier_key, value):
     """
-    Tests whether a given value exists in the list returned from the given
-    strategy's classifier dict for the given classifier_key.
+    Tests whether a given list of values exist in the list returned from the
+    given strategy's classifier dict for the given classifier_key.
 
     Parameters
     ----------
@@ -46,15 +46,19 @@ def passes_in_list_filter(strategy, classifier_key, value):
         classifier_key: string
             Defining which entry from the strategy's classifier dict is to be
             tested (e.g. 'makes_use_of').
-        value: string
-            The value against which the strategy's classifier dict entry is to
+        value: list
+            The values against which the strategy's classifier dict entry is to
             be tested.
 
     Returns
     -------
         boolean
     """
-    return value in strategy.classifier[classifier_key]
+    result = True
+    for entry in value:
+        if entry not in strategy.classifier[classifier_key]:
+            result = False
+    return result
 
 
 def passes_filterset(strategy, filterset):
--- a/axelrod/strategies/_filters.py
+++ b/axelrod/strategies/_filters.py
@@ -207,12 +207,12 @@ def passes_filterset(strategy, filterset):
     # Loop through each of the entries in the filter_functions dict and, if
     # that filter is defined in the supplied filterset, call the relevant
     # function and record its result in the passes_filters list.
-    for filter, filter_function in filter_functions.items():
+    for _filter, filter_function in filter_functions.items():
 
-        if filterset.get(filter, None) is not None:
+        if filterset.get(_filter, None) is not None:
             kwargs = filter_function.kwargs
             kwargs['strategy'] = strategy
-            kwargs['value'] = filterset[filter]
+            kwargs['value'] = filterset[_filter]
             passes_filters.append(filter_function.function(**kwargs))
 
     # Return True if the strategy passed all the supplied filters
--- a/axelrod/strategies/_filters.py
+++ b/axelrod/strategies/_filters.py
@@ -5,7 +5,7 @@ import operator
 def passes_operator_filter(strategy, classifier_key, value, operator):
     """
     Tests whether a given strategy passes a filter for a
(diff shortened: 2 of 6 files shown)


axelrod/strategies/_filters.py
```diff
@@ -0,0 +1,179 @@
...
+        'manipulates_state': FilterFunction(
+            function=passes_boolean_filter,
+            kwargs={'classifier_key': 'manipulates_state'}),
+        'manipulates_source': FilterFunction(
+            function=passes_boolean_filter,
+            kwargs={'classifier_key': 'manipulates_source'}),
+        'inspects_source': FilterFunction(
+            function=passes_boolean_filter,
+            kwargs={'classifier_key': 'inspects_source'}),
+        'min_memory_depth': FilterFunction(
```
drvinceknight: I think I've seen why you've done this so my question might not make terrible sense but could we perhaps also include `memory_depth` in here? So that it picks out exactly the strategies with a given memory depth? I'm not sure my suggestion is a good one, I just like the idea of the player classifier keys having the same keys (as much as possible) as this filter key... (I'm thinking out loud here)

meatballs: It starts to get tricky if there are also filters on min_memory_depth and/or max_memory_depth and those conflict with whatever is set as an equality check, so I left it out!

axelrod/tests/unit/test_filters.py
```diff
@@ -0,0 +1,155 @@
...
+        deterministic_filterset = {
+            'stochastic': False
+        }
+
+        uses_length_filterset = {
+            'stochastic': True,
+            'makes_use_of': 'length'
+        }
+
+        self.assertEqual(
```
drvinceknight: Could you also include some tests that show using the filter gives the same results as using some list comprehensions? (This might be more helpful to me right now just so that I see how it all works).

meatballs: Will do

axelrod/strategies/_filters.py
drvinceknight: Yeah I can see that...

axelrod/tests/unit/test_filters.py
meatballs: Actually, I don't like that as a unit test - it depends too much on other stuff within the library.

I don't mind adding something into the integration tests, if you want.

meatballs:
> This might be more helpful to me right now just so that I see how it all works

That suggests my code/docstrings/comments aren't sufficiently clear. Anywhere in particular I can add/amend something that might help?

axelrod/tests/unit/test_filters.py
drvinceknight: Yeah as an integration test: sure.

drvinceknight:
> That suggests my code/docstrings/comments aren't sufficiently clear. Anywhere in particular I can add/amend something that might help?

I expect what you've got is probably just fine, there's just a lot there. This was just me looking through things briefly and was looking for/expecting those tests. Once I get a moment to put my head to this properly I'm sure it'll clear it up :) :+1: 

(I do think we want those tests though: ultimately if there are two ways of getting some sort of filtered strategy set we need to be checking that they both give the same one. But sure: integration test :+1: .)

meatballs:
> if there are two ways of getting some sort of filtered strategy set we need to be checking that they both give the same one

Agreed! 👍

axelrod/strategies/__init__.py
```diff
@@ -29,3 +30,9 @@
 cheating_strategies = [s for s in all_strategies if not obey_axelrod(s())]
 
 ordinary_strategies = strategies  # This is a legacy and will be removed
+
+
+def filtered_strategies(filterset, strategies=all_strategies):
```
drvinceknight: A short docstring?

axelrod/strategies/_filters.py
```diff
@@ -0,0 +1,179 @@
...
+    Tests whether a given strategy passes a boolean equality filter for a
+    given key in its classifier dict.
+
+    Parameters
+    ----------
+        strategy : a descendant class of axelrod.Player
+        classifier_key: string
+            Defining which entry from the strategy's classifier dict is to be
+            tested (e.g. 'stochastic' or 'inspects_source').
+        value: string or boolean
```
drvinceknight: Why not just accept a boolean here?

axelrod/strategies/__init__.py
meatballs: Ah yes - I meant to come back to that but forgot!

drvinceknight:
Have had a proper look now. Looks ok to me :+1: :smile:  

My only comment is that it's a shame that the classifier dictionary and the filter dictionary don't share the same set of keys but I understand why you've done it that way. I expect you've thought about it longer than I have and there's no other way around it :)

axelrod/strategies/_filters.py
drvinceknight: Could that be dealt with in the api? So the api, would convert the string to a boolean? Basically just this `filter_value = strtobool(value)` would need to be in the api right?

Just thinking that it feel a bit strange in the library, we could also pass integers that would be converted to booleans right? (Not that we'd want to). From the library's point of view we're only expecting booleans here.

axelrod/strategies/_filters.py
```diff
@@ -0,0 +1,179 @@
...
+    Tests whether a given strategy passes a filter for a
+    given key in its classifier dict using a given inequality operator.
+
+    Parameters
+    ----------
+        strategy : a descendant class of axelrod.Player
+        classifier_key: string
+            Defining which entry from the strategy's classifier dict is to be
+            tested (e.g. 'memory_depth').
+        value: string or int
```
drvinceknight: Similar comment here? Could the string be converted to an int at the api level?

axelrod/strategies/_filters.py
meatballs: It could, but it's a right nuisance. I'd have to loop through the incoming dict converting strings sometimes, but not others. I felt it was much cleaner to do it in the actual filter function.

meatballs: I've thought on this one a little further and changed my mind. I agree that it would be cleaner to do the string conversion in the API. I'll sort that out.

axelrod/tests/unit/test_filters.py
```diff
@@ -0,0 +1,147 @@
...
+            'stochastic': True
+        }
+
+        deterministic_filterset = {
+            'stochastic': False
+        }
+
+        uses_length_filterset = {
+            'stochastic': True,
+            'makes_use_of': 'length'
```
marcharper: This needs to be `['length']` I think.

marcharper:
I think the "makes use of" filter example needs to be `["length"]` instead of `"length"`.

axelrod/tests/unit/test_filters.py
meatballs: No. The filter is checking that 'length' exists within the 'makes_use_of' list.

marcharper: So what if someone wants to check for two makes_use_of properties?

meatballs: At the moment, it's not supported by the filtering dict. You'd have to use a comprehension.

It should be possible to add a filter to do this - Django's QueryDict allows multiple values for the same key, so the url would be something like: http://<host>/strategies?makes_use_of=game&makes_use_of=length

meatballs: I'll amend the list filter so that it takes another list as input and checks that all items from the given list are present in the strategy's classifier. I'll work out how to handle passing that list in the API.

meatballs:
> My only comment is that it's a shame that the classifier dictionary and the filter dictionary don't share the same set of keys but I understand why you've done it that way. I expect you've thought about it longer than I have and there's no other way around it :)

How about this....?

My original thought was not to include a ```memory_depth``` filter because I wanted to have ```min_memory_depth``` and ```max_memory_depth``` so that ranges were easily done. The problem I envisaged with having all three was having to write the validation that checked they were compatible.

So, how about we don't bother with the validaion? If you put junk like, for example:
```
filterset = {
   memory_depth=2,
   min_memory_depth=4
}
```
then you'll get nothing back and you'll have to work out why all by yourself!!

drvinceknight:
I like that: sound good to me.

axelrod/tests/unit/test_filters.py
meatballs: Done!!

meatballs:
I think I've picked up all the comments here. Anything I've missed?

If not, it's time to move on to the documentation and, for that, I think there is one main question:

Which is the default/preferred method of filtering and which is the 'we also have' option?

For me, the answer to that question is that the filterset dict should be the main method that is explained in the docs but that we also end with something that says you can also use list comprehensions and gives some examples.

drvinceknight:
> Which is the default/preferred method of filtering and which is the 'we also have' option?
> 
> For me, the answer to that question is that the filterset dict should be the main method that is explained in the docs but that we also end with something that says you can also use list comprehensions and gives some examples.

I think the default method of obtaining a list of strategies should be the filterset **but** the docs should also explain that each strategy has this classifier. So I suggest that http://axelrod.readthedocs.io/en/latest/tutorials/advanced/classification_of_strategies.html **starts with (just writing the corresponding code snippets to illustrate the examples and explanations that should be included)**:

```
>>> import axelrod as axl
>>> expected_dictionary = {'manipulates_state': False, 'makes_use_of': set([]), 'long_run_time': False, 'stochastic': False, 'manipulates_source': False, 'inspects_source': False, 'memory_depth': 0}  # Order of this dictionary might be different on your machine
>>> axl.Cooperator.classifier == expected_dictionary
```

```
>>> s = axl.Cooperator()
>>> s.classifier == expected_dictionary
True
```

and perhaps to show specific use cases a handful of things like:

```
>>> s = axl.TitForTat()
>>> s.classifier['memory_depth']
1
>>> s = axl.Random()
>>> s.classifier['stochastic']
True
```


Then I suggest there is no mention of the list comprehension approach (people would be able to do that themselves if they wanted to based on the above) and the current list comprehension examples in the docs are replaced with a filterset approach. 

How does that sound?

axelrod/strategies/__init__.py
```diff
@@ -29,3 +30,44 @@
...
+    Applies the filters defined in the given filterset dict and returns those
+    strategy classes which pass all of those filters from the given list of
+    strategies.
+
+    e.g.
+
+    For the filterset dict:
+        {
+            'stochastic': True,
+            'min_memory_depth': 2
```
marcharper: Can we add an example of makes use of to the docstring, since it's a bit different than the others?

axelrod/strategies/_filters.py
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    # A list of boolean values to record whether the strategy passed or failed
+    # each of the filters in the supplied filterset.
+    passes_filters = []
+
+    # Loop through each of the entries in the filter_functions dict and, if
+    # that filter is defined in the supplied filterset, call the relevant
+    # function and record its result in the passes_filters list.
+    for filter, filter_function in filter_functions.items():
```
marcharper: can we change `filter` to `filter_` since `filter` is a python keyword?

docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -20,37 +28,82 @@ Note that instances of the class also have this classifier::
...
 
-    >>> len([s for s in axl.strategies if s().classifier['memory_depth']==1])
+    >>> filterset = {
+    ...     'memory_depth': 1
+    ... }
+    >>> strategies = axl.filtered_strategies(filterset)
+    >>> len(strategies)
     24
 
+Mutliple filter can be specified within the filterset dictionary. To specify a
```
marcharper: multiple

docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -20,37 +28,82 @@ Note that instances of the class also have this classifier::
...
 strategies::
 
-    >>> len([s for s in axl.strategies if s().classifier['stochastic']])
-    43
+    >>> filterset = {
+    ...     'stochastic': True
+    ... }
+    >>> strategies = axl.filtered_strategies(filterset)
+    >>> len(strategies)
+    35
```
drvinceknight: Why is the number different here?

docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -20,37 +28,82 @@ Note that instances of the class also have this classifier::
...
 length of each match of the tournament::
 
-    >>> len([s() for s in axl.strategies if 'length' in s().classifier['makes_use_of']])
-    10
+    >>> filterset = {
+    ...     'makes_use_of': ['length']
+    ... }
+    >>> strategies = axl.filtered_strategies(filterset)
+    >>> len(strategies)
+    4
```
drvinceknight: Why is it different?

(Am sure it's not a problem just want to understand.)

drvinceknight:
This looks good to me. I like the docs :) 

Just the two questions as to why the filterset approach gave different numbers in the docs (not suggesting this needs to be changed, just wondering why and if it's indicating something that we've missed).

meatballs:
The difference in the numbers is not because of the filtering. It's caused by the dynamic classifications which means that you get different results at class level than you do at instance level.

The previous docs used list comprehensions at instance level. The filtering approach works at class level.

Personally, I think that's confusing and we should get rid of the dynamic stuff - but that's a matter for a separate debate!!

docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -20,37 +28,82 @@ Note that instances of the class also have this classifier::
...
+    >>> filterset = {
+    ...     'stochastic': True
+    ... }
+    >>> strategies = axl.filtered_strategies(filterset)
+    >>> len(strategies)
+    35
+
 
-Or indeed find out how many strategy only use 1 turn worth of memory to
+Or, to find out how many strategy only use 1 turn worth of memory to
```
drvinceknight: strategies

axelrod/strategies/__init__.py
meatballs: There is a separate filter function for that - `passes_in_list_filter`

drvinceknight:
> The difference in the numbers is not because of the filtering. It's caused by the dynamic classifications which means that you get different results at class level than you do at instance level.

Cool, makes sense :) This PR has a 👍  from me.

> 
> Personally, I think that's confusing and we should get rid of the dynamic stuff - but that's a matter for a separate debate!!

Indeed: a separate discussion but I think it actually would be confusing to get rid of it. The `Random` strategy is stochastic but a player playing the `Random` strategy with `p=0` is not acting stochastically. But yes, not relevant here, the docs are good like this: 👍

meatballs:
> Indeed: a separate discussion but I think it actually would be confusing to get rid of it. The Random strategy is stochastic but a player playing the Random strategy with p=0 is not acting stochastically. 

Ah yes. Perhaps we need an 'even more advanced' tutorial!!

marcharper:
Nice addition @meatballs !

PR #700, opened 2016-08-26
Strategy names and PEP8

meatballs:
- Ensures that each Strategy class has a unique name (required for the API)
- Fixes some PEP8 errors

### DIFF
--- a/axelrod/strategies/cycler.py
+++ b/axelrod/strategies/cycler.py
@@ -68,7 +68,7 @@ class Cycler(Player):
         """
         Player.__init__(self)
         self.cycle = cycle
-        self.name = "Cycler {0}".format(cycle)
+        self.name = "Cycler {}".format(cycle)
         self.classifier['memory_depth'] = len(cycle) - 1
 
     def strategy(self, opponent):
--- a/axelrod/strategies/cycler.py
+++ b/axelrod/strategies/cycler.py
@@ -2,6 +2,7 @@ from axelrod import Actions, Player, init_args
 
 import copy
 
+
 class AntiCycler(Player):
     """
     A player that follows a sequence of plays that contains no cycles:
@@ -77,6 +78,8 @@ class Cycler(Player):
 
 
 class CyclerDC(Cycler):
+
+    name = 'Cycler DC'
     classifier = copy.copy(Cycler.classifier)
     classifier['memory_depth'] = 1
 
@@ -86,6 +89,8 @@ class CyclerDC(Cycler):
 
 
 class CyclerCCD(Cycler):
+
+    name = 'Cycler CCD'
     classifier = copy.copy(Cycler.classifier)
     classifier['memory_depth'] = 2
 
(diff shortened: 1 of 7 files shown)


axelrod/strategies/cycler.py
```diff
@@ -67,7 +68,7 @@ def __init__(self, cycle="CCD"):
         """
         Player.__init__(self)
         self.cycle = cycle
-        self.name += " " + cycle
+        self.name = "Cycler {0}".format(cycle)
```
marcharper: FYI you can leave off the 0 in `{0}`.

axelrod/strategies/cycler.py
```diff
@@ -77,6 +78,8 @@ def strategy(self, opponent):
 
 
 class CyclerDC(Cycler):
+
+    name = 'Cycler DC'
```
marcharper: Do we need to set the name explicitly in addition to the dynamic assignment above?

axelrod/strategies/gobymajority.py
```diff
@@ -80,6 +82,7 @@ class GoByMajority40(GoByMajority):
     """
     GoByMajority player with a memory of 40.
     """
+    name = 'Go By Majority 40'
```
marcharper: Same question -- is the explicit name needed?

axelrod/tests/unit/test_player.py
```diff
@@ -1,10 +1,10 @@
-import copy
+# -*- coding: utf-8 -*-import copy
```
marcharper: I think this isn't quite right...

axelrod/tests/unit/test_player.py
```diff
@@ -259,7 +259,7 @@ def classifier_test(self, expected_class_classifier=None):
         self.assertTrue('memory_depth' in player.classifier,
                         msg="memory_depth not in classifier")
         self.assertTrue('stochastic' in player.classifier,
-                        msg="stochastic not in classifier")
+                        msg="stochastic n§ot in classifier")
```
marcharper: Did you want that extra character?

axelrod/strategies/cycler.py
meatballs: Yep - will do

axelrod/strategies/cycler.py
meatballs: Yes. the dynamic assignment only works at instance level and the API needs a unique id for each class.

axelrod/strategies/gobymajority.py
meatballs: same response!

axelrod/tests/unit/test_player.py
meatballs: you're right. It isn't. I'll sort it.

axelrod/tests/unit/test_player.py
meatballs: nope!

drvinceknight:
> Ensures that each Strategy class has a unique name (required for the API)

Should/could we have a test for this? (Sorry)

meatballs:
> Should/could we have a test for this? (Sorry)

yes, probably - I had these tests on #694 which is where theses issues were highlighted in the first place.

meatballs:
I think I've covered everything on this one. Anything I've missed?

drvinceknight:
👍

PR #705, opened 2016-09-09
Create a code of conduct for the project

meatballs:


meatballs:
I'd like to have this one in before we do the PyCon UK spring next week. Anyone have any thoughts on it?

marcharper:
No objections from me, though it's the sort of thing that I wish we didn't need to have.

It may benefit from a note in the enforcement section that says something like "While we pledge to take all reports seriously, please give others the benefit of the doubt initially: text is a medium lacking nuance and unintentional miscommunication is a common occurrence."

Also I wonder if there's a better place for this at the project level. However I don't see a way to have a readme for the entire Axelrod organization, so I guess this is the best place for now.

Two questions:
* Are we not sufficiently following e.g. github's code of conduct? Or is this simply a case of explicit is better than implicit?
* Would it be better to adopt the same code of conduct as e.g. github or python?

drvinceknight:
> Also I wonder if there's a better place for this at the project level. However I don't see a way to have a readme for the entire Axelrod organization, so I guess this is the best place for now.

I think I like it here **but** we do have: http://axelrod-python.github.io/ (source: https://github.com/Axelrod-Python/Axelrod-Python.github.io) which is an option?

> 
> Two questions:
> 
> Are we not sufficiently following e.g. github's code of conduct? Or is this simply a case of explicit is better than implicit?

I think explicit is better than implicit.

> Would it be better to adopt the same code of conduct as e.g. github or python?

I believe @meatballs has sourced this from somewhere.

My one comment would be: can we have a pointer from the contributing guidelines to the CoC (nothing much, just a simple sentence): http://axelrod.readthedocs.io/en/latest/tutorials/contributing/guidelines.html

marcharper:
I'm happy enough with this (and I know @meatballs wants to get it in soon) so I gave it a positive review. Feel free to make the changes suggested in the comments above, it won't change my opinion either way.

drvinceknight:
OK, I'll merge. My request for a pointer from the contribution guidelines can also be added at another stage.

Issue #701, opened 2016-08-28
Include method to write a summary data set to file


drvinceknight:
The results set used to have a method to write some output to csv. The output was a mid way point between the interactions and a summary (it held the total score of each strategy for each repetition).

As we now have all the interactions written by default I suggest adding a method to the results set that gives a very quick summary data set. Here is the function that I have in mind (would move it to a method of the result set):

```
import axelrod as axl
from numpy import median
from collections import namedtuple
import csv

def summary_data(results, filepath=None):
    """
    Obtain summary of performance of each strategy: 
    ordered by rank, including median normalised score and cooperation rating.
    
    Parameters
    ----------
    
        results : axelrod.results 
        filepath : a filepath to which to write the data
        
    Output
    ------
    
        A list of the form:
        
        [[player name, median score, cooperation_rating],...]
        
        If `filepath` is passed then a summary data file with the following headers will be written:
        
        "Rank", "Name", "Median-score-per-turn", "Cooperation-rating"
    """
    
    median_scores = map(median, results.normalised_scores)
    player = namedtuple("Player", ["Rank", "Name", "Median_score", "Cooperation_rating"])    

    summary_data = [perf for perf in zip(results.players, median_scores, results.cooperating_rating)]
    summary_data = [player(rank, *summary_data[i]) for rank, i in enumerate(results.ranking)]
    
    if filepath is not None:
        with open(filepath, 'w') as csvfile:
            writer = csv.writer(csvfile)
            writer.writerow(player._fields)
            for player in summary_data:
                writer.writerow(player)
    
    return summary_data
```

This would give:

```
>>> axl.seed(0)
>>> players = [s() for s in axl.demo_strategies]
>>> tournament = axl.Tournament(players)
>>> results = tournament.play()
>>> sd = summary_data(results, "test.csv")
>>> sd
[Player(Rank=0, Name='Defector', Median_score=2.4924999999999997, Cooperation_rating=0.0),
 Player(Rank=1, Name='Grudger', Median_score=2.4900000000000002, Cooperation_rating=0.503875),
 Player(Rank=2, Name='Tit For Tat', Median_score=2.308125, Cooperation_rating=0.62675),
 Player(Rank=3, Name='Cooperator', Median_score=1.869375, Cooperation_rating=1.0),
 Player(Rank=4, Name='Random: 0.5', Median_score=1.8218750000000004, Cooperation_rating=0.495125)]
```

and `tmp.csv` would have:

```
[...]

PR #707, opened 2016-09-15
Add summary method to result set

drvinceknight:
Addresses one part of #701

Have not added the part to write to file, thought that could be left outside of the library (users could choose to write it to csv, json etc...: once we've got the list of named tuples it's easy enough to do)? Not terribly sure though so happy enough to add it in.

At the moment I've just included the following tournament results:

- Rank
- Name
- Score
- Cooperation rating
- Wins

Easy enough to add more.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,5 @@ Axelrod.egg-info
 basic_strategies.csv
 cache.txt
 test.csv
+summary.csv
 basic_tournament.csv
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -1,6 +1,7 @@
 import csv
 import tqdm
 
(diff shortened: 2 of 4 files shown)


marcharper:
I think it would be nice to have the write to csv method, as you say it's easy with namedtuples, and probably future-proof-able.

drvinceknight:
Cool I'll add that in today if I get a quiet moment (pyconuk has started :)).

drvinceknight:
I've added the ability to write csv. At a later stage it would be simple enough to tweak that further to write to other data formats. 

FWIW: I've just seen a demo of the new code review thing that github have implemented. It looks pretty nice!  https://help.github.com/articles/about-pull-request-reviews/

Issue #138, opened 2015-04-01
Duplicate Strategies


meatballs:
As more strategies are created, it becomes more difficult to spot when a new strategy is a duplication of an existing one. Simple source code comparison is no good as there are many different ways to code a given strategy.

Instead, we could:
- Create a set of players whose strategies are entirely deterministic and do not depend on the opponent in any way (e.g. Cooperator, Defector, 'Cooperate for 5 turn, defect for 3 and then cooperate until the end'
- Run a tournament with those players plus a single strategy under test only
- Record the score of the single strategy somewhere permanently
- Do the above for all existing strategies
- For any new strategy, run the tournament above. If the score matches any existing, have a look at the code for those and judge whether the strategy is indeed a duplicate (or whether a new 'test' strategy could be written to differentiate between them)

drvinceknight:
Yup, I think that sounds good, as I said on gitter this only flags and we'd still need to take a look. Really like the idea. 

This could be added to the test suite right? So if the tests pass we know the strategy is different, if the test fails:

1. We look through the code and conclude that it is the same.
2. We look through the code and conclude that a new strategy should be added to the test suite to differentiate.

Yeah this would be awesome...

meatballs:
This just narrows down the list to check a new strategy against. It's not intended to give a definitive answer.

I suppose I came up with it because several new strategies have come in whilst I've been doing 'infrastructure' stuff and I have no idea how they work. I'd have no clue whether any new strategy was similar to those already!

drvinceknight:
> This just narrows down the list to check a new strategy against. It's not intended to give a definitive answer.

Yup. I love the idea of adding them to the test suite and so failing `_is_a_new_strategy` is something that can be fixed by modifying the tests by adding a new strategy to the test tournament :)

Or perhaps giving the potential for the strategy to be flagged as `new=True` which means it does not need to go through the test tournament. I am just thinking that perhaps there will be some strategies that act in a particular way which makes them hard to differentiate but that we KNOW are different (so that label would overide).

drvinceknight:
This will need to be documented with an example I think...

meatballs:
I suspect there might be clever mathematical / statistical jiggery-pokery that could be done in order to calculate degrees of similarity from the results. Now if only I could think of how to lay my hands on the sorts of people who might know how to do that.......

drvinceknight:
If only... That would be a very cool problem to look at. Let's talk about
it.

[...]

PR #717, opened 2016-09-19
Issue 138 - Adds functions to calculate the state distribution of a Match

theref:
Related to #138 #701

### DIFF
--- a/axelrod/interaction_utils.py
+++ b/axelrod/interaction_utils.py
@@ -9,6 +9,7 @@ This is used by both the Match class and the ResultSet class which analyse
 interactions.
 """
 import csv
+from collections import Counter
 
 from .game import Game
 from axelrod import Actions
@@ -85,6 +86,29 @@ def compute_normalised_cooperation(interactions):
     normalised_cooperation = tuple([c / float(num_turns) for c in cooperation])
 
     return normalised_cooperation
+
+
+def compute_state_distribution(interactions):
+    """
+    Returns the count of each state for a set of interactions.
+    """
+    if len(interactions) == 0:
+        return None
+    return Counter(interactions)
+
+
+def compute_normalised_state_distribution(interactions):
+    """
+    Returns the normalized count of each state for a set of interactions.
+    """
+    normalized_count = Counter(interactions)
+    total = sum(normalized_count.values(), 0.0)
+    # By starting the sum with 0.0 we make sure total is a floating point value,
+    # avoiding the Python 2 floor division behaviour of / with integer operands (Stack Overflow)
+
+    for key in normalized_count:
+        normalized_count[key] /= total
+    return normalized_count
 
 
 def sparkline(actions, c_symbol=u'█', d_symbol=u' '):
--- a/axelrod/interaction_utils.py
+++ b/axelrod/interaction_utils.py
@@ -101,6 +101,9 @@ def compute_normalised_state_distribution(interactions):
     """
     Returns the normalized count of each state for a set of interactions.
     """
+    if len(interactions) == 0:
+        return None
+
     normalized_count = Counter(interactions)
     total = sum(normalized_count.values(), 0.0)
     # By starting the sum with 0.0 we make sure total is a floating point value,
--- a/axelrod/interaction_utils.py
+++ b/axelrod/interaction_utils.py
@@ -91,8 +91,20 @@ def compute_normalised_cooperation(interactions):
 def compute_state_distribution(interactions):
     """
     Returns the count of each state for a set of interactions.
+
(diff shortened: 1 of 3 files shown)


axelrod/match.py
```diff
@@ -159,6 +159,12 @@ def normalised_cooperation(self):
         """Returns the count of cooperations by each player per turn"""
         return iu.compute_normalised_cooperation(self.result)
 
+    def state_distribution(self):
+        return iu.compute_state_distribution(self.result)
```
drvinceknight: Could we have a brief docstring for these two methods (similar to the ones in the interactions_utils).

axelrod/interaction_utils.py
```diff
@@ -87,6 +88,32 @@ def compute_normalised_cooperation(interactions):
     return normalised_cooperation
 
 
+def compute_state_distribution(interactions):
+    """
+    Returns the count of each state for a set of interactions.
+    """
```
meatballs: The docstring needs the full definition of the arguments and return value

axelrod/interaction_utils.py
```diff
@@ -87,6 +88,32 @@ def compute_normalised_cooperation(interactions):
...
+    if len(interactions) == 0:
```
meatballs: This could just be ```if not interactions:```

axelrod/interaction_utils.py
```diff
@@ -87,6 +88,32 @@ def compute_normalised_cooperation(interactions):
...
+        return None
+    return Counter(interactions)
+
+
+def compute_normalised_state_distribution(interactions):
+    """
+    Returns the normalized count of each state for a set of interactions.
+    """
```
meatballs: Docstring needs to define arguments and return value

axelrod/interaction_utils.py
```diff
@@ -87,6 +88,32 @@ def compute_normalised_cooperation(interactions):
...
+    if len(interactions) == 0:
```
meatballs: ```if not interactions:```

axelrod/match.py
```diff
@@ -159,6 +159,12 @@ def normalised_cooperation(self):
...
+
+    def normalised_state_distribution(self):
```
meatballs: Docstring needed here

axelrod/interaction_utils.py
theref: I just copied what was done in some of the other functions...should I change them all?

drvinceknight: You don't need to change them all (unless you wanted to!) but if you could keep the new ones according the convention that would be much appreciated. We have #347 as an ongoing issue to back-fix the formatting.

axelrod/interaction_utils.py
```diff
@@ -87,6 +88,32 @@ def compute_normalised_cooperation(interactions):
...
+        return None
+
+    normalized_count = Counter(interactions)
+    total = sum(normalized_count.values(), 0.0)
+    # By starting the sum with 0.0 we make sure total is a floating point value,
+    # avoiding the Python 2 floor division behaviour of / with integer operands (Stack Overflow)
+
+    for key in normalized_count:
+        normalized_count[key] /= total
+    return normalized_count
```
meatballs: How about:
```
interactions_count = Counter(interactions)
total = sum(interactions_count.values(), 0.0)
# By starting the sum with 0.0 we make sure total is a floating point value,
# avoiding the Python 2 floor division behaviour of / with integer operands (Stack Overflow)
return Counter([count/total for count in interactions_count])
```

meatballs: this might be a bit more 'pythonic' with a comprehension. I've not tested it, but something like:
```
interactions_count = Counter(interactions)
total = sum(interactions_count.values(), 0.0)
# By starting the sum with 0.0 we make sure total is a floating point value,
# avoiding the Python 2 floor division behaviour of / with integer operands (Stack Overflow)

return Counter({key: value / total for key, value in interactions_count.items()})
```

drvinceknight:
The failure is not your fault @theref, it's hypothesis stumbling upon a particular case where it's a tie and due to floating point errors the names do not get ordered the same. This commit on my own branch should address that: 970b5443867f55528bbf2d8e0e93447cb64e6c1c

drvinceknight:
> This commit on my own branch should address that: 970b544

I'm just about to open another PR that I think offers a better fix still.

drvinceknight:
@theref I've opened #725, you might want to grab the commit from there for now.

drvinceknight:
@theref that's now been merged in. If you merge or rebase on to master you should be good to go :) 👍

Issue #719, opened 2016-09-20
Add ability to keep everything in memory


drvinceknight:
#718 adds a fix for using temporary files on Windows (see this python issue: [tempfile.NamedTemporaryFile not particularly useful on	Windows](https://bugs.python.org/issue14243)). 

Another thing to consider is that for small tournaments it's perfectly fine to run them completely in memory. This would not require much work at all, just if a given switch is not passed we write to an interactions dictionary as opposed to a temporary file. 

Could even be that we do this by default unless a filename is passed (which works fine on Windows), so doing away with temporary files altogether.

PR #713, opened 2016-09-19
Fix for temporary files on windows.

Ducksual:
Closes #680

Issue #718, opened 2016-09-20
Fix parallel processing on windows


drvinceknight:
After @Ducksual's great work on #713 I believe this is the only remaining bug on Windows.

We're now running CI on appveyor so we do have a test runner: https://ci.appveyor.com/project/drvinceknight/axelrod/build/1.0.17/job/nb5j3y5wrxkho04a

Ducksual:
This might be related to the fact that windows uses 'spawn' instead of 'fork'. I'd be interested to know if some of these tests fail on Linux if you use set_start_method('spawn') somewhere.

https://docs.python.org/3.5/library/multiprocessing.html#contexts-and-start-methods

Going through the current master and removing all of the block multiprocessing on Windows flags gives me 3 test failures. (Possibly more if I've missed some flags).

```
======================================================================
ERROR: test_parallel_play (axelrod.tests.unit.test_tournament.TestTournament)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\ducks\PycharmProjects\Axelrod\axelrod\tests\unit\test_tournament.py", line 259, in test_parallel_play
    scores = tournament.play(processes=2, progress_bar=False).scores
  File "C:\Users\ducks\PycharmProjects\Axelrod\axelrod\tournament.py", line 116, in play
    self._run_parallel(processes=processes, progress_bar=progress_bar)
  File "C:\Users\ducks\PycharmProjects\Axelrod\axelrod\tournament.py", line 230, in _run_parallel
    self._start_workers(workers, work_queue, done_queue)
  File "C:\Users\ducks\PycharmProjects\Axelrod\axelrod\tournament.py", line 266, in _start_workers
    process.start()
  File "C:\Miniconda3\envs\axelrod\lib\multiprocessing\process.py", line 105, in start
    self._popen = self._Popen(self)
  File "C:\Miniconda3\envs\axelrod\lib\multiprocessing\context.py", line 212, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "C:\Miniconda3\envs\axelrod\lib\multiprocessing\context.py", line 313, in _Popen
    return Popen(process_obj)
  File "C:\Miniconda3\envs\axelrod\lib\multiprocessing\popen_spawn_win32.py", line 66, in __init__
    reduction.dump(process_obj, to_child)
  File "C:\Miniconda3\envs\axelrod\lib\multiprocessing\reduction.py", line 59, in dump
[...]

PR #721, opened 2016-09-20
Tournaments can run on Windows

drvinceknight:
Started with #719 but has a bit more:

- Have added the ability to keep everything in memory: currently this is the default and temporary files are not used at all but I want to think about that some more (might be more efficient to keep the temporary files as default unless we're on windows which we could just detect and in that case keep things in memory). I'm not overly exited with the fix in #713 as it risks keeping files around if anything is interrupted...
- There are some tests for the above that I need to write ideally.
- Have just skipped the parallel processing tests if on windows (have also skipped testing 2.7 and 3.4 on windows because life is short and they were throwing weird bugs). I feel happier to say that parallel processing is not supported on Windows than making hacky changes...

I believe this will pass all CI tests... Heading to bed: will work on some more tomorrow :)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,11 @@
 environment:
   matrix:
-    - PYTHON: "C:\\Python27"
-    - PYTHON: "C:\\Python34"
+    #- PYTHON: "C:\\Python27"
+    #- PYTHON: "C:\\Python34"
     - PYTHON: "C:\\Python35"
 install:
   - "%PYTHON%\\python.exe -m pip install -r requirements.txt"
 build: off
 test_script:
   - "%PYTHON%\\python.exe -m unittest discover"
+  - "%PYTHON%\\python.exe doctests.py"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,5 @@
 environment:
   matrix:
-    #- PYTHON: "C:\\Python27"
-    #- PYTHON: "C:\\Python34"
     - PYTHON: "C:\\Python35"
 install:
   - "%PYTHON%\\python.exe -m pip install -r requirements.txt"
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+import sys
 
 # The order of imports matters!
 from .actions import Actions, flip_action
@@ -15,3 +16,5 @@ from .match_generator import *
 from .tournament import Tournament, ProbEndTournament, SpatialTournament, ProbEndSpatialTournament
 from .result_set import ResultSet, ResultSetFromFile
 from .ecosystem import Ecosystem
+
+on_windows = sys.platform.startswith("win")
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
-import sys
+import os
 
-on_windows = sys.platform.startswith("win")
+on_windows = os.name == 'nt'
 
 # The order of imports matters!
 from .actions import Actions, flip_action
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -1,6 +1,8 @@
 from __future__ import absolute_import
 import sys
 
+on_windows = sys.platform.startswith("win")
+
 # The order of imports matters!
 from .actions import Actions, flip_action
 from .random_ import random_choice, seed
@@ -16,5 +18,3 @@ from .match_generator import *
 from .tournament import Tournament, ProbEndTournament, SpatialTournament, ProbEndSpatialTournament
 from .result_set import ResultSet, ResultSetFromFile
 from .ecosystem import Ecosystem
-
-on_windows = sys.platform.startswith("win")
--- a/axelrod/tests/integration/test_tournament.py
+++ b/axelrod/tests/integration/test_tournament.py
@@ -1,7 +1,7 @@
 import unittest
 import axelrod
-import tempfile
 import filecmp
+import sys
 
 from axelrod.strategy_transformers import FinalTransformer
 
@@ -34,9 +34,9 @@ class TestTournament(unittest.TestCase):
         tournament = axelrod.Tournament(name='test', players=strategies,
                                         game=self.game, turns=2,
                                         repetitions=2)
-        tmp_file = tempfile.NamedTemporaryFile()
+        filename = "test_outputs/test_tournament.csv"
         self.assertIsNone(tournament.play(progress_bar=False,
-                                          filename=tmp_file.name,
+                                          filename=filename,
                                           build_results=False))
 
     def test_serial_play(self):
@@ -50,6 +50,8 @@ class TestTournament(unittest.TestCase):
         actual_outcome = sorted(zip(self.player_names, scores))
         self.assertEqual(actual_outcome, self.expected_outcome)
 
+    @unittest.skipIf(sys.platform.startswith("win"),
+                     "Parallel processing not supported on Windows")
     def test_parallel_play(self):
         tournament = axelrod.Tournament(
             name=self.test_name,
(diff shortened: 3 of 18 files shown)


meatballs:
>  I feel happier to say that parallel processing is not supported on Windows than making hacky changes...

Agreed - we could even wrap the parallel processing in a condition that checks the os

drvinceknight:
> we could even wrap the parallel processing in a condition that checks the os

Good call, will put something like that together now :) :+1:

drvinceknight:
I think this is in good shape for a review now:

- Have added an `in_memory` variable to `tournament.play`. Default is `False`, if `True` it basically just uses an internal interactions dictionary.
- Have added a `axelrod.on_windows` that on loading of the library detects if it's running on a Windows machine.
- If running on a windows machine **and no filename is provided** then `in_memory` will automatically be set to `True`. This is because `NamedTemporaryFiles` are not very useful on Windows (see this known bug: https://bugs.python.org/issue14243).
- I have added test skips, basically dropping support for parallel processing on Windows. There were other bugs that @Ducksual found with parallel processing: I have opened #718 and suggest that that's fixed as a separate issue.

Have added some tests and also a short documentation tutorial about parallel processing.

I suggest that once this is OK'd and in:

- We make a note on #718 to say that once it's fixed we fix the docs which in two places would say that parallel processing is not supported on Windows.
- We open an issue to refactor `ResultSet` so that it's as efficient as `ResultSetFromFile`. Currently it reads the dictionary more often than it needs to. This shouldn't be hard: everything is basically already in place in `ResultSetFromFile`, that just needs to be tweaked so that it worked with a dictionary.

meatballs:
What about the wrapping of the parallel processing in a condition so that it only happens if not windows?

drvinceknight:
> What about the wrapping of the parallel processing in a condition so that it only happens if not windows?

Yup: I missed that (thought I had done it), appveyor just pointed it out. Just pushing that now :)

axelrod/__init__.py
```diff
@@ -1,4 +1,7 @@
 from __future__ import absolute_import
+import sys
+
+on_windows = sys.platform.startswith("win")
```
meatballs: Any reason you've gone with sys.platform rather than os.name?

drvinceknight: Not an educated reason: would `os.name` be better? I have to check for `nt` or something like that right?

meatballs: I've always used ```os.name == 'nt' ``` based on answers at http://stackoverflow.com/questions/4553129/when-to-use-os-name-sys-platform-or-platform-system#11674977

drvinceknight: Cool: I'll change that 👍

meatballs: but I wondered if you knew something that I don't !!

meatballs:
Big 👍 from me for this one. That was a painful day....

Ducksual:
I don't know what you're talking about. Windows is easy! ;-)

In all seriousness I tried to be fairly careful to clean up the files but it would have been really nice if NamedTemporaryFile actually did something useful in Windows (or if TemporaryDirectory had existed before Python 3.2).

Multiprocessing on windows is a separate nightmare.

appveyor.yml
```diff
@@ -1,10 +1,11 @@
 environment:
   matrix:
-    - PYTHON: "C:\\Python27"
-    - PYTHON: "C:\\Python34"
+    #- PYTHON: "C:\\Python27"
```
marcharper: Can we delete these if they won't be used?

axelrod/tests/integration/test_tournament.py
```diff
@@ -34,9 +33,9 @@ def test_full_tournament(self):
         tournament = axelrod.Tournament(name='test', players=strategies,
                                         game=self.game, turns=2,
                                         repetitions=2)
-        tmp_file = tempfile.NamedTemporaryFile()
+        filename = "test_outputs/test_tournament.csv"
```
marcharper: Are we going to run into issues if the directory doesn't exist?

drvinceknight:
> In all seriousness I tried to be fairly careful to clean up the files but it would have been really nice if NamedTemporaryFile actually did something useful in Windows (or if TemporaryDirectory had existed before Python 3.2).

You had done a great job, and it's highly appreciated. You figured out all the bugs 👍 

> 
> Multiprocessing on windows is a separate nightmare.

We'll get there eventually (hopefully)!

Either way hopefully this will all be in place so that you could contribute actual Game Theory next :)

axelrod/tournament.py
```diff
@@ -93,11 +102,15 @@ def play(self, build_results=True, filename=None,
             self.progress_bar = tqdm.tqdm(total=len(self.match_generator),
                                           desc="Playing matches")
 
-        self.setup_output_file(filename)
+        if on_windows and (filename is None):
```
marcharper: So if the tournament takes up a lot of memory on windows, it won't work?

Ducksual:
I actually checked and NamedDirectory wouldn't have helped (at least in cleaning up the files on windows). As far as I can tell the implementation of NamedTemporaryFile on windows with the locking behaviour is the only way to guarantee the file will be deleted. Without locking something else could open the file and then Windows wouldn't allow it to be removed.

axelrod/__init__.py
drvinceknight: Nope: I just got scared away by the `nt` :) Have changed this with 72238a3

axelrod/tests/integration/test_tournament.py
drvinceknight: The directory should exist, I've put a `README.md` file in there so cloning the library should put it there right?

appveyor.yml
drvinceknight: Yup.

axelrod/tournament.py
drvinceknight: Yup: it it takes up a lot of memory then it will crash. It will work fine if a filename is provided though.

drvinceknight:
> I doubt many windows users would use the library without a GUI and it makes the code more complex to maintain

I know of 3 that have (with anaconda, Jupyter etc I actually would hope we would not close the door on windows users). I expect I'll have students doing so too. Important to get it working natively on windows even if it does make it harder for us. 

> 
> Has anyone tested the library with bash on windows? That might be a better way to have some support on windows, and who knows, maybe multiprocessing works?

That's not really running on Windows though. I expect that if it runs on bash it might not pick up the `on_windows` flag? Apparently that's a full ubuntu thing? (I have no idea).

appveyor.yml
drvinceknight: Have pushed that.

meatballs:
> I doubt many windows users would use the library without a GUI 

Several of the students who came to the sprint were on windows - and I suspect that's often the case for young undergrads coming to university with their shiny new laptops!

Ducksual:
Bash on windows is Ubuntu related - but still beta and a little more work to install than Anaconda. I'm not sure if all the Linux features would work on it or not but I'm installing it now so I can find out.

drvinceknight:
> Bash on windows is Ubuntu related - but still beta and a little more work to install than Anaconda. I'm not sure if all the Linux features would work on it or not but I'm installing it now so I can find out.

Thanks @Ducksual: that's helpful. 

Either way I think that in the interest of diversity we should work to offer native support. This PR is a first step there, we can always open more PRs to improve it and reduce the complexity :)

drvinceknight:
> I'm installing it now so I can find out.

@Ducksual it would be interesting to know what the following does on bash for windows...:

```
>>> import os
>>> os.name
```

Ducksual:
>>> import os
>>> os.name
'posix'
>>>

drvinceknight:
Neat :)

Ducksual:
It's kind of neat but a little intimidating if you haven't used linux/bash before. After changing your machine to developer mode and going through the install you still don't have any gui support for linux.

drvinceknight:
> It's kind of neat but a little intimidating if you haven't used linux/bash before.

Oh absolutely. Bash for windows isn't really a solution. I was just curious :)

drvinceknight:
> Bash for windows isn't really a solution.

Although it might be for some people as it looks like you would be able to run parallel tournaments with it.

marcharper:
Thanks @ Ducksual @drvinceknight @meatballs ! Great work everyone.

PR #720, opened 2016-09-20
New Strategy From the PRISON (http://www.lifl.fr/IPD/ipd.frame.html)

Huaraz2:
implemented 'Worse and worse' strategy as suggested in #379. Made this strategy during the PyconUK sprint yesterday.

PR #724, opened 2016-09-25
New Strategy From the PRISON (http://www.lifl.fr/IPD/ipd.frame.html)

Huaraz2:
Implemented `Worse and Worse` strategy as suggested in #379 and `Knowledgeable Worse and Worse` which I started during the PyconUK Axelrod sprint. This pr is the same as #720 but now hopefully the appveyer test should pass


Huaraz2:
I have absolutely no idea as to why the tests are failing, could it be because I am using `sys.version_info` in my test function?

drvinceknight:
The error messages indicate that you haven't quite coded your strategy right for the case when match lengths are not known by the players. 

There should not be a need to use `sys.version` either. 

I will look through this today and fix it.

drvinceknight:
I've opened https://github.com/Huaraz2/Axelrod/pull/1 to your branch @Huaraz2. 

However, I'm not actually sure this is the correct strategy. I couldn't find this strategy there and looking through what I could find in the literature I can't seem to find any details (please point me towards them though).

Huaraz2:
Thanks for the pr @drvinceknight, I found this strategy in the [classics.str](https://github.com/Huaraz2/prison/blob/master/classics.str) file in the prison package that you can get from the downloads page in the link I have attached to this pr. As far as I understood it, each mactch in the tournament would last for 1000 turns, so I read the strategy as: it defects with probability of current_turn / expected_length. I now do not think this is the case and therefore will rename this one to `WorseAndWorseRandom` and make `WorseAndWorse` for real.

axelrod/strategies/worse_and_worse.py
```diff
@@ -0,0 +1,57 @@
+import random
+
+from axelrod import Actions, Player, random_choice
+
+C, D = Actions.C, Actions.D
+
+class WorseAndWorse (Player):
+    """
+    Defects with probability of 'current turn / 1000'. Therefore
+    it is more and more likely to defect as the round goes on.
```
drvinceknight: Could you add:

```
Source code available at the download tab of [PRISON1998].
```

axelrod/strategies/worse_and_worse.py
```diff
@@ -0,0 +1,57 @@
...
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent):
+        current_round = len(self.history) + 1
+        if random.randint(0, 1000) < current_round:
```
drvinceknight: Can you rewrite this so that it's like the other strategy? IE so that is uses `random_choice`:

```
        probability = 1 - float(current_round) / 1000
        return random_choice(probability)
```

When you do that you can remove the `import random` at the top of this file.

axelrod/strategies/worse_and_worse.py
```diff
@@ -0,0 +1,57 @@
...
+            return D
+        return C
+
+
+class WorseAndWorseRandom (Player):
```
drvinceknight: Not a fan of this name: it makes it sound like `WorseAndWorse` is not random. As this is a version of `WorseAndWorse` that has some knowledge I suggest you call it `KnowledgeableWorseAndWorse`.

axelrod/strategies/worse_and_worse.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    This strategy is based of 'Worse And Worse' but will defect with probability
+    of 'current turn / total no. of turns'.
+
```
drvinceknight: This still needs a names section:

```
    Names:
        - Knowledgeable Worse and Worse: Original name by Adam Phol
```

axelrod/tests/unit/test_worse_and_worse.py
```diff
@@ -0,0 +1,97 @@
...
+    def test_strategy(self):
+        """
+        Test that the strategy gives expected behaviour
+        """
+
+        axelrod.seed(1)
+        opponent = axelrod.Cooperator()
+        player = axelrod.WorseAndWorse()
+        match = axelrod.Match((opponent, player), turns=5)
+        self.assertEqual(match.play(), [('C', 'C'),
```
drvinceknight: Once you've made the expected change, this might change but either way it would be good if you picked a seed that didn't just have `WorseAndWorse` cooperate (ie choose one that gives a mix of C and D). Change the strategy first though.

docs/reference/all_strategies.rst
```diff
@@ -120,6 +123,6 @@ Here are the docstrings of all the strategies in the library.
 .. automodule:: axelrod.strategies.titfortat
    :members:
    :undoc-members:
-.. automodule:: axelrod.strategies.gradualkiller
```
drvinceknight: Do not delete `gradualkiller`.

Huaraz2: I didn't, if you look further up you will see that I moved it so that the list was in alphabetical order again.

drvinceknight: Ah great: my mistake :) 👍

axelrod/strategies/worse_and_worse.py
Huaraz2: Would this line go at the end of line 10 or do you want it as part of the names section?

drvinceknight: In between line 10 and the names section. So:

```
Defects with probability ...
is more likely ...

Source code available ...

Names...
```

axelrod/strategies/worse_and_worse.py
```diff
@@ -10,11 +8,12 @@ class WorseAndWorse (Player):
     it is more and more likely to defect as the round goes on.
 
     Names:
-        - worse_and_worse: [PRISON1998]
+        - worse_and_worse: Source code available at the download tab of
+                            [PRISON1998].
```
drvinceknight: Does this indent render ok? (you can run `make html` to check from the `docs` directory, you might have to pip install a sphinx theme...)

I think the rst syntax should be:

```
        - worse_and_worse: Source code available at the download tab of
          [PRISON1998].
```

axelrod/strategies/worse_and_worse.py
Huaraz2: Thanks for the clarification, I have now made this change.

axelrod/strategies/worse_and_worse.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def strategy(self, opponent):
+        current_round = len(self.history) + 1
+        probability = 1 - float(current_round) / 1000
+        return random_choice(probability)
+
+
+class KnowledgeableWorseAndWorse (Player):
+    """
+    This strategy is based of 'Worse And Worse' but will defect with probability
```
drvinceknight: Typo: should be 'is based on'

axelrod/strategies/worse_and_worse.py
```diff
@@ -0,0 +1,59 @@
...
+class WorseAndWorse (Player):
+    """
+    Defects with probability of 'current turn / 1000'. Therefore
+    it is more and more likely to defect as the round goes on.
+
+    Source code available at the download tab of [PRISON1998]
+
+
+    Names:
+        - worse_and_worse: [PRISON1998].
```
drvinceknight: You need the `_` at the end of the `[PRISON1998]`. So this should be `[PRISON1998]_` same for the one above (line 10).

Also can you change it to be:

```
- Worse and Worse: [PRISON1998]_
```

axelrod/strategies/worse_and_worse.py
Huaraz2: I have now made these changes

axelrod/strategies/_strategies.py
```diff
@@ -195,12 +195,14 @@
     Willing,
     WinShiftLoseStay,
     WinStayLoseShift,
+    WorseAndWorse,
+    KnowledgeableWorseAndWorse,
```
drvinceknight: Can you put this in alphabetical order please.

Huaraz2: Done

Huaraz2:
Thanks Vince, it's been fun :+1:

PR #725, opened 2016-09-26
Fix floating point error in a hyp test

drvinceknight:
In some cases floating point error will make this particular test throw
a different rank for the strategies.


drvinceknight:
@meatballs I would suggest we perhaps merge this in? It's arguably a bug, the change is very minor and @marcharper is busy at the moment (just thinking as this perhaps holds back #717...).

PR #732, opened 2016-09-29
Human Strategy

### LINKED ISSUE #728: Human Strategy
A strategy that presents the history of a match in the terminal and prompts for keyboard input rather than deriving its own action.

meatballs:
A strategy that prompts for keyboard input rather than deriving its own action.

closes #728

### DIFF
--- a/axelrod/strategies/human.py
+++ b/axelrod/strategies/human.py
@@ -1,8 +1,19 @@
-from __future__ import print_function
 from axelrod import Actions, Player, init_args
-import sys
+from prompt_toolkit import prompt
+from prompt_toolkit.validation import Validator, ValidationError
 
 C, D = Actions.C, Actions.D
+
+
+class ActionValidator(Validator):
+
+    def validate(self, document):
+        text = document.text
+
+        if text and text not in ['C', 'D']:
+            raise ValidationError(
+                message='Action must be C or D',
+                cursor_position=0)
 
 
 def human_input():
@@ -16,19 +27,7 @@ def human_input():
         Either 'C' or 'D'
     """
 
-    # Input functions changed between python 2 and python 3. This condition
-    # checks the python version being used and sets the correct function
-    # accordingly.
-    if sys.version_info[0] >= 3:
-        get_input = input
-    else:
-        get_input = raw_input
-
-    prompt = 'Action [C or D]: '
-
-    action = get_input(prompt)
-    while action.upper() not in ['C', 'D']:
-        action = input(prompt)
+    action = prompt('Action [C or D]: ', validator=ActionValidator())
 
     return action.upper()
 
--- a/axelrod/strategies/human.py
+++ b/axelrod/strategies/human.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 from axelrod import Actions, Player, init_args
 from builtins import input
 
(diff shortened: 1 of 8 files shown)


axelrod/strategies/human.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @init_args
+    def __init__(self, name='Human', c_symbol='C', d_symbol='D'):
+        """
+        Parameters
+        ----------
+        name: string
+            The name of the human player
+        c_symbol: string
+            The symbol to denote cooperation
```
drvinceknight: Perhaps:

```
The symbol to denote cooperation at the interactive prompt
```

?

axelrod/strategies/human.py
```diff
@@ -0,0 +1,153 @@
...
+        d_symbol: string
+            The symbol to denote defection
```
drvinceknight: Same as above?

docs/reference/all_strategies.rst
```diff
@@ -72,6 +72,9 @@ Here are the docstrings of all the strategies in the library.
 .. automodule:: axelrod.strategies.handshake
    :members:
    :undoc-members:
+.. automodule:: axelrod.strategies.human
```
drvinceknight: I don't know the answer to this question **but** should this be documented/listed in `all_strategies`? I wonder if it would be best left outside of it as that index gives a nice list of 'research' strategies (the cheaters being an immediate counter example to that point of view). No strong feelings from me just think it's worth thinking about.

docs/tutorials/getting_started/human_interaction.rst
```diff
@@ -0,0 +1,15 @@
...
+It is possible to play interactively using the Human strategy::
+
+    >>> import axelrod as axl
+    >>> me = axl.Human(name='me')
+    >>> players = [axl.TitForTat(), me]
+    >>> match = axl.Match(players, turns=3)
+    >>> match.play() #doctest: +SKIP
+
+You will be prompted for the action to play at each turn and will be able to
+access the results of the match as described in :ref:`creating_matches`
```
drvinceknight: Is it worth showing some example of the prompt? It needn't be too extensive and it would not include the `>>>` so it wouldn't be picked up by the doctests. So something like (after what you have already written):

---

This will then prompt the user for inputs of `C` or `D`:

```
Starting new match
Turn 1 action [C or D] for me: C

Turn 1: me played C, opponent played C
Turn 2 action [C or D] for me: D

Turn 2: me played D, opponent played C
Turn 3 action [C or D] for me: C
[('C', 'C'), ('C', 'D'), ('D', 'C')]
```

after this the `match` object can be manipulated as described in http://axelrod.readthedocs.io/en/latest/tutorials/getting_started/match.html

---

docs/reference/all_strategies.rst
meatballs: I'll take it out whilst we think further....

axelrod/strategies/human.py
```diff
@@ -0,0 +1,160 @@
...
+                '{}Turn {}: {} played {}, opponent played {}'.format(
+                    linesep, len(self.history), self.name,
+                    self.symbols[self.history[-1]],
+                    self.symbols[self.opponent_history[-1]])
+            )
+        else:
+            toolbar = None
+            print_statement = '{}Starting new match'.format(linesep)
+
+        return {
```
marcharper: Is there a reason to return a dictionary rather than a tuple? Just curious.

marcharper:
This is :+1:  from me, just the one simple question.

axelrod/strategies/human.py
meatballs: I like having the readable keys and I was too lazy to create a NamedTuple!!

PR #733, opened 2016-09-30
Add version number to library

### LINKED ISSUE #731: Add version number to library
So that we can run `axelrod.__version__` or something like that.

This should be trivial, nice to not duplicate what is in `setup.py` obviously.

drvinceknight:
Closes #731.

We cannot import __version__ from __init__.py in setup.py. This would
load all of axelrod which at that point does not have any dependencies
installed. The approach here is suggested in the later answers here:
http://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package

Have written a test: this means that when bumping the number we will
need to bump it in two places (`version.py` and `test_version.py`). If
we think this is OK I will update the wiki for the release process to
include: "run the tests" which probably should be there anyway.


drvinceknight:
The tests pass on appveyor but I'm not sure that appveyor is running the setup.py script. Would the way I'm reading the file in setup.py there work on windows because of the "/"?

drvinceknight:
> The tests pass on appveyor but I'm not sure that appveyor is running the setup.py script. Would the way I'm reading the file in setup.py there work on windows because of the "/"?

I ran the appveyor tests with 25543c1, it worked fine. (I can revert that commit if we want but perhaps no harm in checking that installing the library works on windows?)

axelrod/tests/unit/test_version.py
```diff
@@ -0,0 +1,11 @@
...
+Tests the version number
+"""
+from axelrod import __version__
+import unittest
+
+expected_version = "1.11.0"
+
+class TestVersion(unittest.TestCase):
+    def test_version(self):
+        self.assertEqual(__version__, expected_version)
```
meatballs: I don't really see the point of this test. It's just testing that Python correctly returns the value of an attribute, not any logic that we have coded into the library.

drvinceknight: Yeah, I'm not terribly sure this was needed, mainly put it in as I was just checking everything was being imported:

1. I replace the test with a check that the attribute is there (so that `axelrod.__version__` exists basically)?
2. Remove it?

meatballs: Might as well have option 1. No harm in that and at least there's then only one place to update the version.

drvinceknight: 21ec838 does that now.

setup.py
```diff
@@ -4,9 +4,12 @@
 with open('requirements.txt') as f:
     requirements = f.read().splitlines()
 
+# Read in the version number
+exec(open('axelrod/version.py', 'r').read())
+
```
meatballs: Why can't this just be `from axelrod.version import __version__`?

drvinceknight: I tried that but it still trips up on the dependencies:

```
(tmp) Axelrod(731) ✗: python setup.py install
  File "setup.py", line 9, in <module>
    from axelrod.version import __version__
  File "/Users/vince/src/Axelrod/axelrod/__init__.py", line 9, in <module>
    from .random_ import random_choice, seed
  File "/Users/vince/src/Axelrod/axelrod/random_.py", line 2, in <module>
    import numpy
ImportError: No module named 'numpy'
(tmp) Axelrod(731) ✗: cat setup.py
from setuptools import setup

# Read in the requirements.txt file
with open('requirements.txt') as f:
    requirements = f.read().splitlines()

# Read in the version number
#exec(open('axelrod/version.py', 'r').read())
from axelrod.version import __version__

setup(
    name='Axelrod',
    version=__version__,
    install_requires=requirements,
    author='Vince Knight, Owen Campbell, Karol Langner, Marc Harper',
    author_email=('axelrod-python@googlegroups.com'),
    packages=['axelrod', 'axelrod.strategies', 'axelrod.tests'],
    url='http://axelrod.readthedocs.org/',
    license='The MIT License (MIT)',
    description='Reproduce the Axelrod iterated prisoners dilemma tournament',
)
```

The `exec` approach is suggested here because of this issue: http://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package

meatballs: Of course! It's trying to load `__init__.py` first

PR #742, opened 2016-10-14
Implement state distribution in result set.

### LINKED ISSUE #738: Add state distributions to result set
Following #737 we can now add to the result sets (base and 'from file') consistently (before that they were being built in two separate ways: now code is not duplicated).

The match class and interaction utils also now have functions to calculate the state distribution for a match: counts of turns spent in each of `CC, CD, DC, DD`. 

It should be easy enough to add that to the result set. I suggest we have, similar to `results.cooperation` and `results.normalised_cooperation`:

```
results.state_distribution  # A list of lists (1 for each player);of lists (1 for each opponent); of 4 tuples of total (across all repetitions) of counts of `CC, CD, DC, DD`
results.normalised_state_distribution  # # A list of lists (1 for each player);of lists (1 for each opponent); of 4 tuples of total (across all repetitions) of normalised counts of `CC, CD, DC, DD`
```

I also suggest we add 4 more columns to the output of  `results.summarise`, which are the mean (across all opponents) normalised counts for each state.

drvinceknight:
Closes #738

Add state distribution and normalised state distribution to the result set.

Also adds the normalised state distribution to the summary.

### DIFF
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -600,7 +600,8 @@ class ResultSet(object):
         summary_data = list(zip(self.players, median_scores,
                                 self.cooperating_rating, median_wins))
 
-        summary_data = [self.player(rank, *summary_data[i], *state_prob[i]) for
+        summary_data = [self.player(rank, *(list(summary_data[i])
+                                            + state_prob[i])) for
                         rank, i in enumerate(self.ranking)]
 
         return summary_data
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -354,8 +354,8 @@ class ResultSet(object):
             counters = []
             for counter in player:
                 total = sum(counter.values())
-                counters.append(Counter({key: value / total for key, value in
-                                         counter.items()}))
+                counters.append(Counter({key: float(value) / total for
+                                         key, value in counter.items()}))
             norm.append(counters)
         return norm
 
@@ -750,7 +750,7 @@ class ResultSet(object):
                 p = sum([opp[state] for j, opp in enumerate(player) if i != j])
                 counts.append(p)
             try:
-                counts = [c / sum(counts) for c in counts]
+                counts = [float(c) / sum(counts) for c in counts]
             except ZeroDivisionError:
                 counts = [0 for c in counts]
             state_prob.append(counts)
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -7,6 +7,9 @@ from numpy import mean, nanmedian, std
 from . import eigen
 from .game import Game
 import axelrod.interaction_utils as iu
+from axelrod import Actions
+
+C, D = Actions.C, Actions.D
 
 
 def update_progress_bar(method):
@@ -142,7 +145,7 @@ class ResultSet(object):
         return players
 
     @update_progress_bar
-    def build_eigenmoses_rating(self):
+    def _build_eigenmoses_rating(self):
         """
         Returns:
         --------
@@ -156,7 +159,7 @@ class ResultSet(object):
         return eigenvector.tolist()
 
     @update_progress_bar
-    def build_eigenjesus_rating(self):
+    def _build_eigenjesus_rating(self):
         """
         Returns:
         --------
@@ -170,7 +173,7 @@ class ResultSet(object):
         return eigenvector.tolist()
 
     @update_progress_bar
-    def build_cooperating_rating(self):
+    def _build_cooperating_rating(self):
         """
         Returns:
         --------
@@ -201,7 +204,7 @@ class ResultSet(object):
                 in zip(self.cooperation, lengths)]
 
     @update_progress_bar
-    def build_vengeful_cooperation(self):
+    def _build_vengeful_cooperation(self):
         """
         Returns:
         --------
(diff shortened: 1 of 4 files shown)


axelrod/result_set.py
```diff
@@ -326,6 +326,26 @@ def build_ranking(self):
         return sorted(range(self.nplayers),
                       key=lambda i: -nanmedian(self.normalised_scores[i]))
 
+    def build_normalised_state_distribution(self):
```
marcharper: Needs an underscore? Seems like we're being inconsistent with the build* functions

axelrod/result_set.py
```diff
@@ -326,6 +326,26 @@ def build_ranking(self):
...
+            Normalised state distribution. A list of lists of counter objects:
+
+            Dictionary where the keys are the states and the values are a
+            normalized counts of the number of times that state occurs.
+        """
+        norm = []
+        for player in self.state_distribution:
+            counters = []
+            for counter in player:
+                total = sum(counter.values(), 0.0)
```
marcharper: Is the 0.0 necessary here? Is it possible for sum to be zero?

axelrod/result_set.py
```diff
@@ -401,6 +423,13 @@ def _update_cooperation(self, p1, p2, cooperations):
         self.cooperation[p1][p2] += cooperations[0]
         self.cooperation[p2][p1] += cooperations[1]
 
+    def _update_state_distribution(self, p1, p2, counter):
+        self.state_distribution[p1][p2] += counter
```
marcharper: Docstring?

axelrod/result_set.py
```diff
@@ -401,6 +423,13 @@ def _update_cooperation(self, p1, p2, cooperations):
...
+
+        counter[('C', 'D')], counter[('D', 'C')] = (counter[('D', 'C')],
```
marcharper: I like the swap but not the line break.... I guess there's no getting around that, but maybe it would be more clear if broken after the = sign

marcharper: Could shorten with `C, D = actions.C, actions.D` at the top of file...

axelrod/result_set.py
```diff
@@ -548,13 +580,28 @@ def summarise(self):
         median_wins = map(nanmedian, self.wins)
 
         self.player = namedtuple("Player", ["Rank", "Name", "Median_score",
-                                            "Cooperation_rating", "Wins"])
+                                            "Cooperation_rating", "Wins",
+                                            "CC_rate", "CD_rate", "DC_rate",
+                                            "DD_rate"])
+
+        states = [('C', 'C'), ('C', 'D'), ('D', 'C'), ('D', 'D')]
```
marcharper: Could use `C, D = actions.C, actions.D`

axelrod/result_set.py
```diff
@@ -548,13 +580,28 @@ def summarise(self):
...
+        state_prob = []
+        for i, player in enumerate(self.normalised_state_distribution):
+            counts = []
+            for state in states:
+                counts.append(sum([opp[state] for j, opp in enumerate(player)
```
marcharper: Two lines here? One for the sum, one for the append so there's no line break

axelrod/result_set.py
```diff
@@ -548,13 +580,28 @@ def summarise(self):
...
+                                   if i != j]))
+            try:
+                counts = [c / sum(counts) for c in counts]
```
marcharper: Any change for an integer division error here?

axelrod/result_set.py
```diff
@@ -548,13 +580,28 @@ def summarise(self):
...
+
+        summary_data = list(zip(self.players, median_scores,
+                                self.cooperating_rating, median_wins))
 
-        summary_data = [perf for perf in zip(self.players,
-                                             median_scores,
-                                             self.cooperating_rating,
-                                             median_wins)]
-        summary_data = [self.player(rank, *summary_data[i]) for
+        summary_data = [self.player(rank, *(list(summary_data[i])
```
marcharper: This is a bit hard to follow... could use a readability upgrade?

docs/tutorials/getting_started/tournament_results.rst
```diff
@@ -20,6 +20,10 @@ This tutorial will show you how to access the various results of a tournament:
 - Payoff difference means: the mean score differences.
 - Cooperation counts: the number of times each player cooperated.
 - Normalised cooperation: cooperation count per turn.
+- Normalised cooperation: cooperation count per turn.
+- State distribution: the count of each type of state of a match
```
marcharper: Might be good to explain what we mean by state distribution -- that it's two rounds of play rather than some other amount.

drvinceknight: I've added that in the wording below. Repeated it for both which might be a bit overkill, happy to tweak further :)

axelrod/result_set.py
drvinceknight: Added this and a bunch of others that were missing (whilst doing this I got annoyed at my prior lazy self).

axelrod/result_set.py
drvinceknight: 👍

axelrod/result_set.py
drvinceknight: Looks like it wasn't needed. Have removed it.

axelrod/result_set.py
drvinceknight: What did you mean by an integer division error? (I might have misunderstood.)

drvinceknight: I know what you mean. Am pushing a better hypothesis test and a fix.

PR #747, opened 2016-10-22
Implement doubler strategy from PRISON project

mturzanska:
Related to https://github.com/Axelrod-Python/Axelrod/issues/379

### DIFF
--- a/axelrod/strategies/doubler.py
+++ b/axelrod/strategies/doubler.py
@@ -5,9 +5,8 @@ C, D = Actions.C, Actions.D
 
 class Doubler(Player):
     """
-    Cooperates except when the opponent defected and
-    opponent's cooperations are less than twice as often
-    as opponent's defections.
+    Cooperates except when the opponent has defected and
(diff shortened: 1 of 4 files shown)


axelrod/strategies/doubler.py
```diff
@@ -0,0 +1,34 @@
+from axelrod import Actions, Player
+
+C, D = Actions.C, Actions.D
+
+
+class Doubler(Player):
+    """
+    Cooperates except when the opponent defected and
+    opponent's cooperations are less than twice as often
```
drvinceknight: Could this be changed to:

```
and opponent's cooperation count is less than twice their defections.
```

@meatballs what do you think? Your judgement is way better than mine with these things.

meatballs: How about:
```
Cooperates except when the opponent has defected and
the opponent's cooperation count is less twice their defection count.
```

drvinceknight:
Thanks @mturzanska 👍

PR #750, opened 2016-10-28
Fix PEP8 errors in Tit For Tat

meatballs:


.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
```
drvinceknight: Should we not just gitignore this directory?

meatballs: Didn't notice that creep in there. I'll sort it.

marcharper:
Looks fine to me, can you rebase onto the new master?

Issue #744, opened 2016-10-16
Track CC, CD, DC, DD in player class 


marcharper:
Currently we track cooperations and defections for efficiency for several strategies. Some strategies also use the context counts, so we might consider doing the same for these as well. See [Retaliator](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/retaliate.py for an example.

mturzanska:
Could I work on this one?

drvinceknight:
> Could I work on this one?

Please do!

drvinceknight:
Cosed by #752

PR #752, opened 2016-10-30
Track CC, CD, DC, DD in Player class

mturzanska:
Related to https://github.com/Axelrod-Python/Axelrod/issues/744

Add play_counts property to Player class (a dictionary counting occurrences of CC, CD, DC, DD).
Update the dictionary after each play.
Add tests for the property.

TODO:
Use play_counts property in existing strategies (e.g. Prober4, Retaliate)

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -53,10 +53,10 @@ def update_history(player, move):
         player.defections += 1
 
 
-def update_play_counts(player, move, reply):
-    """Updates play_counts following play. """
-    last_round = (move, reply)
-    player.play_counts[last_round] += 1
+def update_state_distribution(player, action, reply):
+    """Updates state_distribution following play. """
+    last_turn = (action, reply)
+    player.state_distribution[last_turn] += 1
 
 
 def init_args(func):
@@ -104,7 +104,7 @@ class Player(object):
                 self.classifier[dimension] = self.default_classifier[dimension]
         self.cooperations = 0
         self.defections = 0
-        self.play_counts = defaultdict(int)
+        self.state_distribution = defaultdict(int)
         self.init_args = ()
         self.set_match_attributes()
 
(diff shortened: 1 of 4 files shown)


axelrod/tests/unit/test_player.py
```diff
@@ -60,6 +64,9 @@ def test_play(self):
         self.assertEqual(p1.defections, 0)
         self.assertEqual(p2.cooperations, 0)
         self.assertEqual(p2.defections, 2)
+        # Test play counts
+        self.assertEqual(p1.play_counts, {(C, D): 2})
+        self.assertEqual(p2.play_counts, {(D, C): 2})
```
drvinceknight: Could we have a couple more tests leading to a 'fuller' dictionary please.

axelrod/player.py
```diff
@@ -52,6 +53,12 @@ def update_history(player, move):
         player.defections += 1
 
 
+def update_play_counts(player, move, reply):
```
drvinceknight: In the glossary <http://axelrod.readthedocs.io/en/latest/reference/glossary.html#a-play> we have defined a play as "a single player choosing an action."

In the results set and the interaction utils we have called these counters "state distributions". Could we call them the same thing here (for consistency).

- `update_state_distribution`
- `player_state_distribution`

drvinceknight: Yup: that's better than my suggestion :) 👍

marcharper:
Since we're counting C and D as well for each player (in attributes cooperations and defections), should we change those to a dictionary as well? (Not necessarily in this PR -- there will be a lot of places that need to change.)

drvinceknight:
> Since we're counting C and D as well for each player (in attributes cooperations and defections), should we change those to a dictionary as well? (Not necessarily in this PR -- there will be a lot of places that need to change.)

Good idea but yeah: this could/should be another issue.

drvinceknight:
Thanks for the changes @mturzanska, this looks good to me now.

PR #754, opened 2016-11-02
Moran Process with Mutation

marcharper:
I also made some improvements to the Moran process implementation -- it better handles unique player types now by using the string representations of the players. In the previous implementation the process would not have distinguished between `Random(0.5)` and `Random(0.25)`, for example.

Re #754

### DIFF
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -82,7 +82,7 @@ class MoranProcess(object):
             d[str(p)] = p
         mt = dict()
         for key in sorted(keys):
-            mt[key] = [v for (k, v) in d.items() if k != key]
+            mt[key] = [v for (k, v) in sorted(d.items()) if k != key]
         self.mutation_targets = mt
 
     def set_players(self):
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -81,7 +81,7 @@ class MoranProcess(object):
         for p in players:
             d[str(p)] = p
         mt = dict()
-        for key in keys:
+        for key in sorted(keys):
             mt[key] = [v for (k, v) in d.items() if k != key]
         self.mutation_targets = mt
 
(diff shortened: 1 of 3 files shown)


axelrod/moran.py
```diff
@@ -31,7 +31,33 @@ def fitness_proportionate_selection(scores):
...
+        Parameters
+        ----------
+        players, iterable of axelrod.Player subclasses
+        turns: int, 100
+            The number of turns in each pairwise interaction
+        noise: float, 0
+            The background noise, if any. Randomly flips plays with probability `noise`.
+        deterministic_cache: axelrod.DeterministicCache, None
+            A optional prebuilt deterministic cache
+        mutation_rate: float, 0
```
drvinceknight: A lot of the literature I've seen write this in a generic way with a matrix $M$ for mutation where $M_ij$ is the probability if mutating from species $i$ to $j$. Is that worth doing (perhaps at a later date)?

More of a question than a request. We could simply have that the mutation rate could either be a float or a 2 d array... (Or whatever: not suggest this needs to be done here).

axelrod/moran.py
```diff
@@ -40,10 +66,24 @@ def __init__(self, players, turns=100, noise=0, deterministic_cache=None):
...
         else:
             self.deterministic_cache = DeterministicCache()
+        # Build the set of mutation targets
+        # Determine the number of unique types (players)
+        keys = set([str(p) for p in players])
+        # Create a dictionary mapping each type to a set of representatives of the other types
+        d = dict()
+        for p in players:
+            d[str(p)] = p
+        mt = dict()
```
drvinceknight: could we make `mt`: `mutation_matrix` or similar?

marcharper: Sure

Issue #746, opened 2016-10-18
Dual of finite state machine players


drvinceknight:
In [Ashlock and Kim 2008](http://ieeexplore.ieee.org/document/4492964/) the dual of a strategy is defined as:

> Strategy $A'$ is said to be the dual of strategy $A$ if $A$ and $A'$ can be written as finite-state machines that are identical except that their responses are reversed. 

Note: this does not correspond to flipping the actions (using the `FlippedTransformer` for example), but needs to actually go in to the finite state machine and change the response (not the states).

This would be an easy addition to do for the generic finite state machine strategies (could be done nicely by writing a function that maps a finite state machine to the corresponding dual finite state machine) and could also be implemented for a few others (for example `Pavlov`/`WSLS` is it's own dual when you change the initial move from `C` to `D`).

This is relevant to the work being done on #138 as the dual strategies are used for the fingerprinting approach described in Ashlock's work. FYI the more we (@theref is working on this with @Nikoleta-v3 and I) work on #138 the more we feel that an alternative approach is necessary (and will no doubt be obtained).

meatballs:
> ....the more we feel that an alternative approach is necessary

An alternative to what? 

Ashlock's work itself? (i.e Is there some issue with how the fingerprinting is done)?
Using Ashlock's work for #138? (i.e The fingerprinting is fine, but doesn't solve our issue)?

drvinceknight:
> An alternative to what?
> 
> Ashlock's work itself? (i.e Is there some issue with how the fingerprinting is done)?
> Using Ashlock's work for #138? (i.e The fingerprinting is fine, but doesn't solve our issue)?

No issue with how it's done, just that it's not the best approach for #138 in general. He does what he describes but what he describes seems to be particularly aimed at finite state machine strategies and not immediately applicable to "any" strategy. My hunch is also that it's not terribly robust but time will tell as we evaluate it more. We've got a couple of ideas for other approaches for #138 that we'll explore in time.

marcharper:
Dual Strategy Transformer?

drvinceknight:
Thought about that but it would have to only apply to finite state machines (dual is only defined on fsm strategies). But there are also some strategies like `WSLS` that could have a dual although it's not currently written as a fsm. 

My thinking was to write a transformer for the fsm and build the fsm strategies using it. We could still expose the fsm transformer.  What do you think?

theref:
[...]

PR #758, opened 2016-11-05
Adds Dual Transformer and Joss-Ann Transformer #746

theref:
Adds Dual Transformer and Joss-Ann Transformer. These are both required when creating the fingerprint of a strategy as described in http://doi.org/10.1109/ITW.2010.5593352

These are no tests for these yet, but I thought I'd open a pull request to make sure people are happy with the implementation.

Edit - there are now tests.


axelrod/player.py
```diff
@@ -134,6 +134,12 @@ def strategy(self, opponent):
         """This is a placeholder strategy."""
         raise NotImplementedError()
 
+    def dual_strategy(self, opponent):
+        from copy import deepcopy
+        fresh_opponent = deepcopy(opponent)
+        self.original.play(fresh_opponent)
```
meatballs: should this be ```self.original_player```?

axelrod/strategy_transformers.py
```diff
@@ -150,24 +153,55 @@ def generic_strategy_wrapper(player, opponent, proposed_action, *args, **kwargs)
...
+
+
 def noisy_wrapper(player, opponent, action, noise=0.05):
     """Applies flip_action at the class level."""
     r = random.random()
     if r < noise:
         return flip_action(action)
     return action
 
+
```
meatballs: 👍  for fixing PEP8 errors!

axelrod/strategy_transformers.py
```diff
@@ -309,9 +350,46 @@ def mixed_wrapper(player, opponent, action, probability, m_player):
 
     return action
 
+
 MixedTransformer = StrategyTransformerFactory(
     mixed_wrapper, name_prefix="Mutated")
 
+
+def joss_ann_wrapper(player, opponent, proposed_action, probability):
+    """Wraps the players strategy function to produce the Joss-Ann.
```
meatballs: Why is it called the Joss-Ann? Is that a reference to some paper somewhere? If so, can we have a link in here?

axelrod/strategy_transformers.py
```diff
@@ -309,9 +350,46 @@ def mixed_wrapper(player, opponent, action, probability, m_player):
...
+        probability[:] = [i / sum(probability) for i in probability]
+
+    remaining_probability = max(0, 1 - probability[0] - probability[1])
+    probability += (remaining_probability,)
+    options = [C, D, proposed_action]
+    action = choice(options, p=probability)
+    return action
+
+
+JossAnnTransformer = StrategyTransformerFactory(joss_ann_wrapper,
```
meatballs: Can we have both parameters on one new line to match the other transformer definitions

axelrod/strategy_transformers.py
```diff
@@ -150,24 +153,55 @@ def generic_strategy_wrapper(player, opponent, proposed_action, *args, **kwargs)
...
 FlipTransformer = StrategyTransformerFactory(
     flip_wrapper, name_prefix="Flipped")
 
 
+def dual_wrapper(player, opponent, proposed_action):
+    """Wraps the players strategy function to produce the Dual.
+
+    The Dual of a strategy will return the exact opposite set of moves to the
+    original strategy when both are faced with the same history.
+
```
meatballs: This needs a reference to Ashlock's paper so we can see where this comes from.

axelrod/strategy_transformers.py
drvinceknight: Can we include the bibliographic reference too as well (so with `[Ashlock...]_` and the relevant info in the bibliography.rst file.

axelrod/strategy_transformers.py
```diff
@@ -309,9 +350,46 @@ def mixed_wrapper(player, opponent, action, probability, m_player):
...
+    action: an axelrod.Action, C or D
+    """
+    if sum(probability) > 1:
+        probability[:] = [i / sum(probability) for i in probability]
+
+    remaining_probability = max(0, 1 - probability[0] - probability[1])
+    probability += (remaining_probability,)
+    options = [C, D, proposed_action]
+    action = choice(options, p=probability)
+    return action
```
drvinceknight: Could we not implement this as a particular case of the Mixed Strategy transformer (by mixing between the probe and Cooperator and Defector)? This would reduce duplication.

axelrod/player.py
drvinceknight: I believe this method is not actually necessary (the transformer does not use it). Can it just be removed?

axelrod/strategy_transformers.py
drvinceknight: 👍

theref:
> Can we include the bibliographic reference too as well (so with [Ashlock...]_ and the relevant info in the bibliography.rst file.

So should I put a link to the relevant paper (in both the Dual and JossAnn Transformers) and then also add something to `/Axelrod/docs/reference/bibliography.rst`?

drvinceknight:
Yes please.

drvinceknight:
(And in the docs you'll be able to point to the reference in the bibliography.  Sphinx will render it nicely.)

PR #759, opened 2016-11-05
Create fingerprint.py and a class for Ashlocks Fingerprint

theref:
This relies on #746.

Again, no tests written but I wanted to check the code written was acceptable first. I know @drvinceknight has a few things that could be changed but wanted the discussions to take place on here.


axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+        Returns
+        ----------
+        probe_coords : ordered dictionary
+            An Ordered Dictionary where the keys are tuples representing each
+            coordinate, eg. (x, y). The value is automatically set to `None`.
+        """
+        coordinates = list(product(np.arange(0, 1, granularity),
+                                   np.arange(0, 1, granularity)))
+
+        probe_coords = OrderedDict.fromkeys(coordinates)
```
drvinceknight: I don't understand why this needs to be an OrderedDict. Would a simple dictionary not do?

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+        return probe_coords
+
+    def create_probes(self, probe, granularity):
+        """Creates a set of probe strategies over the unit square.
+
+        Construncts probe strategies that correspond to (x, y) coordinates. The
```
drvinceknight: Constructs

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+                edge = (1, index + 2)
+            else:
+                edge = (0, index + 2)
+            coord_scores[coord] = edge_scores[edge]
+
+        ser = pd.Series(list(coord_scores.values()),
+                        index=pd.MultiIndex.from_tuples(coord_scores.keys()))
+        data_frame = ser.unstack().fillna(0)
+        data_frame.shape
+        return data_frame
```
drvinceknight: A pandas data frame is an amalgamation of dictionaries and numpy, I suggest that in this particular case, just having the data as a dictionary mapping tuples to values would be simpler (and removes the issue of needing to add pandas as a dependancy for the library).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+            correspond to coordinates. The cell values are the score of the
+            original/dual strategy playing the probe with parameters that match
+            the coordinate.
+        """
+        edge_scores = {key: cfspt(value[0])[0] for key, value in
+                       results.interactions.items()}
+
+        coord_scores = OrderedDict.fromkeys(probe_coords)
+        for index, coord in enumerate(coord_scores.keys()):
+            if sum(coord) > 1:
```
drvinceknight: Perhaps a short inline comment describing what's happening with the two cases here (playing the dual or not).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+        """Plot the results of the spatial tournament.
+
+        Parameters
+        ----------
+        col_map : str, optional
+            A matplotlib colour map, full list can be found at
+            http://matplotlib.org/examples/color/colormaps_reference.html
+        """
+        self.data = self._generate_data(self.results, self.probe_players.keys())
+        sns.heatmap(self.data, cmap=col_map)
```
drvinceknight: The choice of Default need to be carefully thought through.

For @marcharper and @meatballs: @theref has created a known fingerprint using all of the colormaps available. They don't all show the characteristics described in the paper so we need to pick a good default.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+        ----------
+        strategy : class
+            A class that must be descended from axelrod.strategies.
+        probe : class
+            A class that must be descended from axelrod.strategies
+        """
+        self.strategy = strategy
+        self.probe = probe
+
+    def create_probe_coords(self, granularity):
```
drvinceknight: `granularity` should be `steps`.

An increase of granularity here would imply a less precise image/fingerprint but that's not usually what is meant by granularity.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+import pandas as pd
+import seaborn as sns
+import matplotlib.pyplot as plt
+from itertools import product
+from axelrod.strategy_transformers import JossAnnTransformer, DualTransformer
+from axelrod.interaction_utils import compute_final_score_per_turn as cfspt
+from collections import OrderedDict
+
+
+class Fingerprint():
```
drvinceknight: A docstring will be necessary.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(self):
+        pass
+
+    def plot(self):
+        pass
+
+
+class AshlockFingerprint(Fingerprint):
```
drvinceknight: A docstring will be necessary.

drvinceknight:
@marcharper, @meatballs and anyone else: to give some context (@theref and I have been working on this): This implements a general fingerprint class and a particular fingerprint class as defined by Ashlock. Ashlock's fingerprint is in fact an analytical function so this is actually a simulation of Ashlock's fingerprint. Because of that, it's actually an extension of Ashlock's fingerprint as it can be used on any strategy (and not just finite state machines strategies). 

@theref  , @Nikoleta-v3 and I are working on some of the theoretic basis for that which at some point will need to be referenced **but the basic idea is**: if you want to fingerprint S, you need to play S against a strategy that depends on coordinates in the unit square (either a Joss-Ann transformation or the Dual of the Joss-Ann, depending on where you are in the unit square).

drvinceknight:
@theref if you rebase this on to you branch for #758 you'll have the transformers here which will help the review (and merging this will automatically merge #758).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+            A list containing tuples of length 2. All tuples will have either 0
+            or 1 as the first element. The second element is the index of the
+            corresponding probe (+2 to allow for including the Strategy and it's
+            Dual).
+        """
+        edges = []
+        for index, coord in enumerate(probe_dict.keys()):
+            #  Add 2 to the index because we will have to allow for the Strategy
+            #  and it's Dual
+            if sum(coord) > 1:
```
drvinceknight: An inline comment here to explain the two cases briefly perhaps.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+            else:
+                probe_dict[coord] = JossAnnTransformer((x, y))(probe)()
+        return probe_dict
+
+    def create_edges(self, probe_dict):
+        """Creates a set of edges for a spatial tournament.
+
+        Construncts edges that correspond to the probes in `probe_dict`. Probes
+        whose coordinates sum to less/more than 1 will have edges that link them
+        to 0/1 correspondingly.
```
drvinceknight: Clarify that 0 correspond to the Joss Ann and 1 to the Dual of the Joss Ann.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+        self.probe_players = self.create_probes(self.probe, granularity)
+        self.edges = self.create_edges(self.probe_players)
+        original = self.strategy()
+        dual = DualTransformer()(self.strategy)()
+        probes = self.probe_players.values()
+        tourn_players = [original, dual] + list(probes)
+        spatial_tourn = axl.SpatialTournament(tourn_players, turns=turns,
+                                              repetitions=repetitions,
+                                              edges=self.edges)
+        print("Begin Spatial Tournament")
```
drvinceknight: No need for these print statements, pass a `progress_bar` argument to the `play` method (and to the `fingerprint` method).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+            A class that must be descended from axelrod.strategies
+        """
+        self.strategy = strategy
+        self.probe = probe
+
+    def create_probe_coords(self, granularity):
+        """Creates a set of coordinates over the unit square.
+
+        Construncts (x, y) coordinates that are separated by a step equal to
+        `granularity`. The coordinates are over the unit squeare which implies
```
Nikoleta-v3: Square?

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+
+        Parameters
+        ----------
+        col_map : str, optional
+            A matplotlib colour map, full list can be found at
+            http://matplotlib.org/examples/color/colormaps_reference.html
+        """
+        self.data = self._generate_data(self.results, self.probe_players.keys())
+        sns.heatmap(self.data, cmap=col_map)
+        plt.show()
```
Nikoleta-v3: Well I believe @drvinceknight covered some of my issues. Such as the print and the order dictionary. Looking forward to the tests. My only further comment is that the documentation could be better, I am losing you at some parts.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+        pass
+
+
+class AshlockFingerprint(Fingerprint):
+    def __init__(self, strategy, probe):
+        """
+        Parameters
+        ----------
+        strategy : class
+            A class that must be descended from axelrod.strategies.
```
meatballs: `axelrod.strategies` is a list. I think you mean `axelrod.Player`

axelrod/fingerprint.py
drvinceknight: @Nikoleta-v3, if you could point out parts that you found unclear this would be helpful. :) 👍

axelrod/fingerprint.py
Nikoleta-v3: Constructs

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+            An Ordered Dictionary where the keys are tuples representing each
+            coordinate, eg. (x, y). The value is a `JossAnnTransformer` with
+            parameters that correspond to (x, y).
+
+        Returns
+        ----------
+        edges : list of tuples
+            A list containing tuples of length 2. All tuples will have either 0
+            or 1 as the first element. The second element is the index of the
+            corresponding probe (+2 to allow for including the Strategy and it's
```
Nikoleta-v3: +2 is not clear to me when I read this.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+            A results set for a spatial tournament.
+        probe_coords : list of tuples
+            A list of tuples of length 2, where each tuple represents a
+            coordinate, eg. (x, y).
+
+        Returns
+        ----------
+        data_frame : pandas.core.frame.DataFrame
+            A pandas DataFrame object where the row and column headers
+            correspond to coordinates. The cell values are the score of the
```
Nikoleta-v3: Once you remove pandas, **maybe** some of the information of the details you are keeping and creating this table can go higher up.

axelrod/fingerprint.py
meatballs: I wonder if some of the methods in here might be better as module level functions. They could then be tested in complete isolation without the need to create a fingerprint instance.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,206 @@
...
+            coordinate, eg. (x, y). The value is a `JossAnnTransformer` with
+            parameters that correspond to (x, y).
+        """
+        probe_dict = self.create_probe_coords(granularity)
+        for coord in probe_dict.keys():
+            x, y = coord
+            if x + y > 1:
+                probe_dict[coord] = JossAnnTransformer((1 - y, 1 - x))(probe)()
+            else:
+                probe_dict[coord] = JossAnnTransformer((x, y))(probe)()
```
meatballs: For me, the clue in this one is that it doesn't use `self` anywhere within the method.

drvinceknight: > In my opinion, yes. (But this sort of thing is always a matter of opinion)!

I'm happy for this to be popped outside so that it's at the module level too: particularly useful (as @meatballs says) for testing.

From a user pov, as it's not really user facing, it doesn't really need to be in the class either (it's going to be called by other methods that the user will use right?).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,226 @@
...
+        ----------
+        strategy : class
+            A class that must be descended from axelrod.Player
+        probe : class
+            A class that must be descended from axelrod.Player
+        """
+        self.strategy = strategy
+        self.probe = probe
+
+    def create_probe_coords(self, step):
```
meatballs: This method looks like another candidate for being a module level function.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,226 @@
...
+        probe_dict : ordered dictionary
+            An Ordered Dictionary where the keys are tuples representing each
+            coordinate, eg. (x, y). The value is a `JossAnnTransformer` with
+            parameters that correspond to (x, y).
+        """
+        probe_dict = OrderedDict((coord, create_jossann(coord, probe))
+                                 for coord in probe_coords)
+        return probe_dict
+
+    def create_edges(self, probe_dict):
```
meatballs: This method looks like another candidate for being a module level function.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,226 @@
...
+            #  Add 2 to the index because we will have to allow for the Strategy
+            #  and it's Dual
+            if sum(coord) > 1:
+                edge = (1, index + 2)
+            else:
+                edge = (0, index + 2)
+            edges.append(edge)
+        return edges
+
+    def fingerprint(self, turns=50, repetitions=10, step=0.01, processes=None):
```
meatballs: This method has a lot of side-effects but returns nothing. That makes it tricky to test and even more difficult to debug when things don't work as expected.

I think it needs refactoring into one or more functions/methods with defined return values, tests, and no side-effectd.

meatballs:
Several of these functions (which look a lot nicer now 😄 ) return lists of similar looking tuples. Might a named tuple be worthwhile here?

axelrod/fingerprint.py
```diff
@@ -0,0 +1,247 @@
...
+        for index, coord in enumerate(coordinates):
+            if sum(coord) > 1:
+                edge = (1, index + 2)
+            else:
+                edge = (0, index + 2)
+            coord_scores[coord] = edge_scores[edge]
+
+        return coord_scores
+
+    def plot(self, col_map=None):
```
meatballs: Same comment as for fingerprint method

axelrod/fingerprint.py
```diff
@@ -0,0 +1,247 @@
...
+        Tuples of length 2 representing each coordinate, eg. (x, y)
+
+    Returns
+    ----------
+    probe_dict : ordered dictionary
+        An Ordered Dictionary where the keys are tuples representing each
+        coordinate, eg. (x, y). The value is a `JossAnnTransformer` with
+        parameters that correspond to (x, y).
+    """
+    probe_dict = OrderedDict((coord, create_jossann(coord, probe))
```
drvinceknight: As we've got the coords, could we not do without using an `OrderedDict`? (Basically just repeating my previous comment and what we discussed on Monday).

drvinceknight: (If there's a reason I'm missing: please mention it here.)

theref: i'm slowly removing OrderedDicts throughout the code, but it needed the tests and a bit of refactoring first

drvinceknight:
> Several of these functions (which look a lot nicer now 😄 ) return lists of similar looking tuples. Might a named tuple be worthwhile here?

My suggestion would be that as they're always just the same 2-tuples of coordinates `(x, y)` that it's not really necessary here.

axelrod/fingerprint.py
drvinceknight: Could plot and fingerprint not be interpreted as a verb?

(I'm not overly bothered either way.)

(We also have all the plot methods in the `Plot` class which are just called by the name of the type of plot, so keeping it as `plot` or `fingerprint` would be in line with that.)

meatballs:
> My suggestion would be that as they're always just the same 2-tuples of coordinates (x, y) that it's not really necessary here.

That's precisely my reason for suggesting it. I think that e.g. `coordinates.x` is far more readable when using the tuple.

drvinceknight:
> That's precisely my reason for suggesting it. I think that e.g. coordinates.x is far more readable when using the tuple.

Sure, not overly opinionated on this one either, 2-tuples as point coordinates seems fairly standard to me but having a named tuple doesn't do any harm.

axelrod/fingerprint.py
drvinceknight: 👍  Cool, apologies, I'll stop commenting for now and just wait until you say it's ready to be looked at :)

drvinceknight:
> Sure, not overly opinionated on this one either, 2-tuples as point coordinates seems fairly standard to me but having a named tuple doesn't do any harm.

I do love a named tuple :) 👍 

I think they're my favourite thing after `enumerate`. Every time I use `enumerate` I give myself a little high five. :)

meatballs:
Over to you @theref !! Whichever you prefer!

axelrod/fingerprint.py
theref: > Could plot and fingerprint not be interpreted as a verb?
> 
> 
> (We also have all the plot methods in the Plot class which are just called by the name of the type of plot, so keeping it as plot or fingerprint would be in line with that.)

This is what I thought, I'd be in favour of keeping it as is.

meatballs: OK!

theref:
I haven't played with named tuples before so I'm going to see how it fits in 🇺🇸  :haircut_man:

axelrod/fingerprint.py
theref: Please feel free to comment as I'm going! Stops me from heading down the wrong path :)
I was just commenting to log thought process etc

theref:
@drvinceknight @meatballs any feedback on the tests i'm writing would be really appreciated (if you have time)

axelrod/fingerprint.py
```diff
@@ -0,0 +1,247 @@
...
+        #  and it's Dual
+        if sum(coord) > 1:
+            edge = (1, index + 2)
+        else:
+            edge = (0, index + 2)
+        edges.append(edge)
+    return edges
+
+
+class Fingerprint():
```
meatballs: Is there an intention to do something with the class later on? It doesn't appear to be doing anything right now.

meatballs:
The tests look good so far to me

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def test_fingerprint(self):
+        pass
+
+    def test_generate_data(self):
+        af = AshlockFingerprint(self.strategy, self.probe)
+        edges, players = af.construct_tournament_elements(0.25)
+        spatial_tournament = axl.SpatialTournament(players, turns=10,
+                                                   repetitions=2,
+                                                   edges=edges)
```
meatballs: Minor nitpicking. Let's tidy this up to:
```
spatial_tournament = axl.SpatialTournament(
    players, turns=10, repetitions=2, edges=edges)
```

or

```
spatial_tournament = axl.SpatialTournament(
    players, turns=10,
    repetitions=2,
    edges=edges
)
```

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,107 @@
...
+        results = spatial_tournament.play(progress_bar=False,
+                                          keep_interactions=True)
```
meatballs: Similar nitpicking comment

axelrod/fingerprint.py
theref: The intention would be to add our own fingerprint class (next semesters focus) which would inherit from `Fingerprint` ie `class CampbellKnightFingerprint(Fingerprint)`

At that point we may decide it's better to have some functions in the original class. It's more of a just in case at the moment

theref: 👍

meatballs: Also, if you do end up putting it back later, its signature should be `class Fingerprint(object):`

I also think you should consider having any placeholder methods raise `NotImplementedError` rather than simply passing.

axelrod/tests/unit/test_fingerprint.py
theref: http://legacy.python.org/dev/peps/pep-0008/#indentation
this seems to suggest my method is perfectly ok?

meatballs: OK, I suppose so. God it's ugly, though!

theref: I find it much easier to read! I'd be happy to go with this instead:
```
spatial_tournament = axl.SpatialTournament(players, 
                                           turns=10,
                                           repetitions=2,
                                           edges=edges)
```

meatballs: Go with whatever you prefer. I don't really have a leg to stand on here.

Personally, I find that format struggles once the names of the function you're calling gets too long. Then you end using one of the formats I suggested anyway and, therefore, I prefer to stick to somethign that will always work so I don't have get used to reading lots of different layouts.

It's all personal preference, so feel free to ignore me.

axelrod/fingerprint.py
```diff
@@ -210,28 +203,28 @@ def generate_data(self, interactions, coordinates):
...
 
-        coord_scores = {coord: None for coord in coordinates}
-        for index, coord in enumerate(coordinates):
-            if sum(coord) > 1:
+        coordinate_scores = {coord: None for coord in coordinates}
+        for index, coordinate in enumerate(coordinates):
+            if sum(coordinate) > 1:
                 edge = (1, index + 2)
             else:
                 edge = (0, index + 2)
```
meatballs: This block (lines 220-224) is a repeat of lines 109-115. That suggests to me there's a function here that needs to be pulled out to module level.

drvinceknight:
> I haven't played with named tuples before so I'm going to see how it fits in

Yeah my suggestion would be to create a `point` named tuple: the example in the official docs is probably exactly what you want https://docs.python.org/3/library/collections.html#collections.namedtuple

drvinceknight:
> Yeah my suggestion would be to create a point named tuple: the example in the official docs is probably exactly what you want https://docs.python.org/3/library/collections.html#collections.namedtuple

Haha, ignore that I see you had already implemented it.

drvinceknight:
Let me know when this is ready to be looked at: tests look like they're on the right track :)

This will also need a doc tutorial.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,237 @@
...
+    step : float
+        The separation between each coordinate. Smaller steps will
+        produce more coordinates that will be closer together.
+
+    Returns
+    ----------
+    coordinates : list of tuples
+        Tuples of length 2 representing each coordinate, eg. (x, y)
+    """
+    Point = namedtuple('Point', 'x y')
```
meatballs: This could be defined at the module level and then used throughout instead of having a mixture of named and anonymous tuples.

drvinceknight:
@theref at some point in this PR, could you add some images of the fingerprints please (not in the code) just drop them here as comments.

Perhaps throw in wsls (and potentially others from that PhD thesis) analytically but then also the output of this? This will help with reviewing.

Similarly, I remember you saying that you're not plotting with the same orientation as Axelrod's work, perhaps nicer to flip that for consistency? (Not sure. Just a thought.)

theref:
@drvinceknight @meatballs the functions `finerprint` `plot` and `read_interactions` are all missing tests because I'm not sure of the best way to go about testing them. Any help with this would be appreciated, otherwise I'm happy with the code.

drvinceknight:
> finerprint plot and read_interactions are all missing tests because I'm not sure of the best way to go about testing them

Create instances and run the methods in question, then test that the corresponding attributes are what they should be.

There will be similar tests in the `ResultSet` test suite for example, so take a look at that for guidance :)

I've taken a quick scan and seen that your plot function saves a figure, simpler to just return the plot object (like the various similar methods in the `axelrod.Plot` class). You can then test the type of what that returns.

drvinceknight:
This for example https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_resultset.py#L590 tests a method that updates a particular attribute in the `ResultSet`.

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,107 @@
...
+              (0, 12),
+              (1, 13),
+              (0, 14),
+              (0, 15),
+              (1, 16),
+              (1, 17)]
+
+
+class TestFingerprint(unittest.TestCase):
+    """Some stuff"""
```
drvinceknight: You can just remove this docstring (or write something relevant).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,285 @@
...
+        from math import sqrt
+        granularity = int(sqrt(len(self.coordinates)))
+        if filename is None:
+            filename = self.strategy.name + ' and ' + self.probe.name + ".pdf"
+
+        ordered_data = [self.data[coord] for coord in self.coordinates]
+        plotting_data = np.reshape(ordered_data, (granularity, granularity))
+
+        plt.imshow(plotting_data, cmap=col_map, )
+        plt.savefig(filename)
```
drvinceknight: Return the plot object (see `axelrod.Plot` for similar methods).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,285 @@
...
+            The number of turns per match
+        repetitions : integer, optional
+            The number of times the round robin should be repeated
+        step : float, optional
+            The separation between each coordinate. Smaller steps will
+            produce more coordinates that will be closer together.
+        processes : integer, optional
+            The number of processes to be used for parallel processing
+        """
+        outputfile = NamedTemporaryFile(mode='w')
```
drvinceknight: Needs test because this won't work on Windows I believe. 

Recall our conversation from last Monday: have the same options as for the `axelrod.Tournament.play` method: make it so that it can do things in memory if need be, similarly also allow for the option to write to a named file (because why not?). Then include a relevant check if on Windows (basically just do the same as the `Tournament` does).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,285 @@
...
+            Defaults to the current directory with the name
+            `Strategy and Probe.pdf`
+        col_map : str, optional
+            A matplotlib colour map, full list can be found at
+            http://matplotlib.org/examples/color/colormaps_reference.html
+        """
+
+        # sns.heatmap(self.data, cmap=col_map)
+        # plt.show()
+        from math import sqrt
```
drvinceknight: Move this import.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,285 @@
...
+        filename : str, optional
+            The location and name that the resulting plot should be saved to.
+            Defaults to the current directory with the name
+            `Strategy and Probe.pdf`
+        col_map : str, optional
+            A matplotlib colour map, full list can be found at
+            http://matplotlib.org/examples/color/colormaps_reference.html
+        """
+
+        # sns.heatmap(self.data, cmap=col_map)
```
drvinceknight: Remove unneeded comments.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,285 @@
...
+        # plt.show()
+        from math import sqrt
+        granularity = int(sqrt(len(self.coordinates)))
+        if filename is None:
```
drvinceknight: Remove the filename as an option, just return the plot object, that can then be saved by the user.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,285 @@
...
+            original player, the second is the dual, the rest are the probes.
+
+        """
+        probe_coordinates = create_coordinates(step)
+        self.coordinates = probe_coordinates
+        edges = create_edges(probe_coordinates)
+
+        dual = DualTransformer()(self.strategy)()
+        probe_players = create_probes(self.probe, probe_coordinates)
+        # probes = probe_players.values()
```
drvinceknight: Remove

axelrod/fingerprint.py
drvinceknight: Also feels like you shouldn't need to recalculate granularity, perhaps just have `self.granularity` set on init?

theref: yeah I've been debating with myself about the best option

drvinceknight: I see why you're doing it like this now though. Leave it as it is for now perhaps and let's see what it all looks like once everything else is in place (will need tests for edge cases to make sure something strange can't happen).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,294 @@
+import axelrod as axl
+import numpy as np
+import matplotlib.pyplot as plt
+from axelrod.strategy_transformers import JossAnnTransformer, DualTransformer
+from axelrod.interaction_utils import compute_final_score_per_turn as cfspt
```
drvinceknight: Could we leave this as `compute_final_score_per_turn`, I had no idea what `cfspt` was when reading lower down.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,294 @@
...
+        A dictionary where the keys are coordinates of the form (x, y) and
+        the values are the mean score for the corresponding interactions.
+    """
+    edge_scores = [np.mean([cfspt(scores) for scores in interactions[edge]])
+                   for edge in edges]
+    coordinate_scores = dict(zip(coordinates, edge_scores))
+    return coordinate_scores
+
+
+def read_interactions(filename):
```
drvinceknight: This exists in `interactions_utils`? If you're doing something slightly different here I suggest you simply modify the output from what you would get from `interactions_utils`, (there is no need to duplicate the code for reading in the file).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,294 @@
...
+from axelrod.interaction_utils import compute_final_score_per_turn as cfspt
+from axelrod import on_windows
+from collections import namedtuple
+from tempfile import NamedTemporaryFile
+
+
+Point = namedtuple('Point', 'x y')
+
+
+def create_jossann(coordinate, probe):
```
drvinceknight: On reflection, I understand the merit of having these outside of the class, however some of these are quite particular to the Ashlock fingerprint and I can imagine that we will implement further fingerprints that don't make use of these functions **or** make use of modifications of them.

So I'm actually going to suggest that they (not all, will mark the ones I think this applies) get put back in the class (as static methods). They can still be 're-used' via inheritance if need be, and the test just requires a simple instance initialisation.

Before you make this change though @theref, I suggest you wait to hear what @meatballs and perhaps @marcharper have to say.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,294 @@
...
+    coordinates : list of tuples
+        Tuples of length 2 representing each coordinate, eg. (x, y)
+    """
+    coordinates = list(Point(j, k) for j in np.arange(0, 1, step)
+                       for k in np.arange(0, 1, step))
+
+    return coordinates
+
+
+def create_probes(probe, coordinates):
```
drvinceknight: Like `create_jossann` I would move this back to the `Ashlock` class.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,294 @@
...
+    probe_dict : ordered dictionary
+        An Ordered Dictionary where the keys are tuples representing each
+        coordinate, eg. (x, y). The value is a `JossAnnTransformer` with
+        parameters that correspond to (x, y).
+    """
+    probes = [create_jossann(coordinate, probe) for coordinate in coordinates]
+    return probes
+
+
+def create_edges(coordinates):
```
drvinceknight: Like `create_jossann` I would move this back to the `Ashlock` class.

At the very least this should be renamed as it's creating a very particular set of edges that correspond to rules that are particular to Ashlock's fingerprint.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,294 @@
...
+        #  and it's Dual
+        if sum(coordinate) > 1:
+            edge = (1, index + 2)
+        else:
+            edge = (0, index + 2)
+        edges.append(edge)
+    return edges
+
+
+def generate_data(interactions, coordinates, edges):
```
drvinceknight: This could also potentially go back to the class, mainly because of the expected structure of the edges which is particular to an `Ashlock` fingerprint.

```
+    edges : list of tuples
￼ +        A list containing tuples of length 2. All tuples will have either 0
￼ +        or 1 as the first element. The second element is the index of the
￼ +        corresponding probe (+2 to allow for including the Strategy and it's
￼ +        Dual).
```

axelrod/fingerprint.py
```diff
@@ -0,0 +1,294 @@
...
+        repetitions : integer, optional
+            The number of times the round robin should be repeated
+        step : float, optional
+            The separation between each coordinate. Smaller steps will
+            produce more coordinates that will be closer together.
+        processes : integer, optional
+            The number of processes to be used for parallel processing
+        """
+        in_memory = False
+        if on_windows and (filename is None):
```
drvinceknight: Appveyor (windows) tests are failing, I expect it has something to do with this part.

See: https://ci.appveyor.com/project/drvinceknight/axelrod/build/1.0.272

```
C:\projects\axelrod\axelrod\tests\unit\test_player.py:203: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
  argspec = inspect.getargspec(PlayerClass.__init__)
.................
ERROR: test_fingerprint (axelrod.tests.unit.test_fingerprint.TestFingerprint)
  File "C:\projects\axelrod\axelrod\tests\unit\test_fingerprint.py", line 96, in test_fingerprint
    af.fingerprint(turns=10, repetitions=2, step=0.25, processes=1)
  File "C:\projects\axelrod\axelrod\fingerprint.py", line 256, in fingerprint
    filename = outputfile.name
UnboundLocalError: local variable 'outputfile' referenced before assignment
ERROR: test_plot (axelrod.tests.unit.test_fingerprint.TestFingerprint)
  File "C:\projects\axelrod\axelrod\tests\unit\test_fingerprint.py", line 130, in test_plot
    af.fingerprint(turns=10, repetitions=2, step=0.25, processes=1)
  File "C:\projects\axelrod\axelrod\fingerprint.py", line 256, in fingerprint
    filename = outputfile.name
UnboundLocalError: local variable 'outputfile' referenced before assignment
Ran 1842 tests in 111.844s
FAILED (errors=2, skipped=8)
Starting new match

```

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,133 @@
...
+        self.assertEqual(tournament_players[0].__class__, af.strategy)
+
+    def test_fingerprint(self):
+        af = AshlockFingerprint(self.strategy, self.probe)
+        af.fingerprint(turns=10, repetitions=2, step=0.25, processes=1)
+        edge_keys = sorted(list(af.interactions.keys()))
+        coord_keys = sorted(list(af.data.keys()))
+        self.assertEqual(af.step, 0.25)
+        self.assertEqual(edge_keys, sorted(long_edges))
+        self.assertEqual(coord_keys, long_coordinates)
```
drvinceknight: Can we also test the generated data? (Basically anything that fingerprint creates, can we check it).

Could you also do a check that fingerprint works on all strategies (obviously for a huge size, tiny turns etc...). I suggest you do this in two ways:

- Write a hypothesis test (I can help) that picks two strategies, sets one to be the probe and checks that the data is of the correct form (this will also check that it ran). 
- Write a loop that just checks them all. I think that this would be my preferred option.

I think this is important to have, in the same way that we have tests on tournaments that the full tournament runs for example.

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,133 @@
...
+        self.assertEqual(edges, expected_edges)
+
+    def test_construct_tournament_elemets(self):
+        af = AshlockFingerprint(self.strategy, self.probe)
+        edges, tournament_players = af.construct_tournament_elements(0.5)
+        self.assertEqual(edges, self.expected_edges)
+        self.assertEqual(len(tournament_players), 6)
+        self.assertEqual(tournament_players[0].__class__, af.strategy)
+
+    def test_fingerprint(self):
```
drvinceknight: Can we test that fingerprint can run in parallel and in series (so check what happens when we call `fingerprint` without the processes option etc). (Take a look at the tests for tournaments)

Might need to skip the parallel test if on windows.

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,133 @@
...
+                                          keep_interactions=True)
+        data = generate_data(results.interactions, long_coordinates, edges)
+        keys = sorted(list(data.keys()))
+        values = [0 < score < 5 for score in data.values()]
+        self.assertEqual(sorted(keys), long_coordinates)
+        self.assertEqual(all(values), True)
+
+    def test_plot(self):
+        af = AshlockFingerprint(self.strategy, self.probe)
+        af.fingerprint(turns=10, repetitions=2, step=0.25, processes=1)
```
drvinceknight: I think you should run these without `processes` (when you're calling it with `processes=1`).

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,133 @@
...
+    def test_construct_tournament_elemets(self):
+        af = AshlockFingerprint(self.strategy, self.probe)
+        edges, tournament_players = af.construct_tournament_elements(0.5)
+        self.assertEqual(edges, self.expected_edges)
+        self.assertEqual(len(tournament_players), 6)
+        self.assertEqual(tournament_players[0].__class__, af.strategy)
+
+    def test_fingerprint(self):
+        af = AshlockFingerprint(self.strategy, self.probe)
+        af.fingerprint(turns=10, repetitions=2, step=0.25, processes=1)
```
drvinceknight: Also need to test that the `in_memory` option works. (Take a look at the `tournament.play` tests).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,294 @@
...
+        self.spatial_tourn = axl.SpatialTournament(tourn_players,
+                                                   turns=turns,
+                                                   repetitions=repetitions,
+                                                   edges=edges)
+        self.spatial_tourn.play(build_results=False,
+                                filename=filename,
+                                processes=processes,
+                                in_memory=in_memory)
+        self.interactions = read_interactions(filename)
+        self.data = generate_data(self.interactions, self.coordinates, edges)
```
drvinceknight: As well as setting `self.data` we could also return the data. This would be akin to all the `{match, tournament, moran}.play()` methods which not only set something but also return it...

axelrod/fingerprint.py
meatballs: If they are specific to the Ashlock fingerprint, I've no objection to them being static methods within the AshlockFingerprint class (note: static methods, not class methods).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,278 @@
...
+def create_jossann(coordinate, probe):
+    """Creates a JossAnn probe player that matches the coordinate.
+
+    If the coordinate sums to more than 1 the parameters are flipped and
+    subtracted from 1 to give meaningful probabilities. This is outlined further
+    in [Ashlock2010]_.
+
+    Parameters
+    ----------
+    coordinate : tuple of length 2
```
meatballs: Could these (and other similar parameters) now be instances of Point?

theref:
iterm2 1nv0ja wsls_large
iterm2 fh9g5r wsls

drvinceknight:
Plots look great. Are they the same orientation? Looks like it could be
flipped along the y=. 5 line. Hard to be sure, can you check to be sure?

drvinceknight:
@theref the failure is because of the ordering of the dict (system dependant). I suggest the following:

```
>>> data = af.fingerprint(turns=10, repetitions=2, step=0.2)
>>> data
{...
>>> data[(0, 0)]
3.0
```

That shows that the output is indeed a dict and we also check one given value.

axelrod/__init__.py
```diff
@@ -20,3 +20,4 @@
 from .tournament import Tournament, ProbEndTournament, SpatialTournament, ProbEndSpatialTournament
 from .result_set import ResultSet, ResultSetFromFile
 from .ecosystem import Ecosystem
+from .fingerprint import AshlockFingerprint, create_coordinates
```
drvinceknight: why do we need `create_coordinates` to be imported?

docs/tutorials/further_topics/fingerprinting.rst
```diff
@@ -0,0 +1,47 @@
+.. _fingerprinting:
+
+Fingerprinting
+==============
+
+In [Ashlock2010] a methodology for obtaining visual representation of a
+strategy's behaviour is given.
+The basic method is to play the strategy against a probe strategy with varying
+noise parameters.
+These noise parameters are implemented through the `JossAnnTransformer`.
```
drvinceknight: The correct rst syntax is:

```
:code:`JossAnnTransformer`
```

docs/tutorials/further_topics/fingerprinting.rst
```diff
@@ -0,0 +1,47 @@
...
+The Joss-Ann of a strategy is a new strategy which has a probability :code:`x` of
+choosing the move C, a probability :code:`y` of choosing the move D, and otherwise
```
drvinceknight: of cooperating, ... of defecting

drvinceknight: Can you add a detail here about the subtelty of what to do when x + 1 >= 1

drvinceknight:
I clicked on `approve` but that was a miss click, technically I'm asking for some very minor fixes.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+from axelrod import on_windows
+from collections import namedtuple
+from tempfile import NamedTemporaryFile
+
+
+Point = namedtuple('Point', 'x y')
+
+
+def create_points(step):
+    """Creates a set of coordinates over the unit square.
```
meatballs: Points rather than coordinates

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+
+    Parameters
+    ----------
+    step : float
+        The separation between each coordinate. Smaller steps will
+        produce more coordinates that will be closer together.
+
+    Returns
+    ----------
+    coordinates : list of Points
```
meatballs: Points rather than coordinates

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+        Named Tuples of length 2 representing each coordinate, eg. (x, y)
+    """
+    coordinates = list(Point(j, k) for j in np.arange(0, 1, step)
```
meatballs: Points rather than coordinates

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+                       for k in np.arange(0, 1, step))
+
+    return coordinates
```
meatballs: points rather than coordinates

drvinceknight:
Thanks for fc4d47b, I have opened https://github.com/theref/Axelrod/pull/4 with a PR to your branch that should fix the error and also tidies a couple of things.

axelrod/__init__.py
```diff
@@ -20,3 +20,4 @@
 from .tournament import Tournament, ProbEndTournament, SpatialTournament, ProbEndSpatialTournament
 from .result_set import ResultSet, ResultSetFromFile
 from .ecosystem import Ecosystem
+from .fingerprint import AshlockFingerprint, create_points
```
drvinceknight: Why do we need to import `create_points`?

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+        The separation between each coordinate. Smaller steps will
+        produce more coordinates that will be closer together.
+
+    Returns
+    ----------
+    points : list of Point
+        Named Tuples of length 2 representing each coordinate (x, y)
+    """
+    points = list(Point(j, k) for j in np.arange(0, 1, step)
+                       for k in np.arange(0, 1, step))
```
Nikoleta-v3: PEP-8 ?

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,161 @@
...
+        af = AshlockFingerprint(self.strategy, self.probe)
+        edges, players = af.construct_tournament_elements(0.5)
+        spatial_tournament = axl.SpatialTournament(players,
+                                                   turns=10,
+                                                   repetitions=2,
+                                                   edges=edges)
+        results = spatial_tournament.play(progress_bar=False,
+                                          keep_interactions=True)
+        data = af.generate_data(results.interactions, self.expected_points,
+                             self.expected_edges)
```
Nikoleta-v3: PEP 8?

drvinceknight:
As

drvinceknight:
> As

Tried commenting on my phone but the size of this page is too much for my phone!

So here goes again:

As well as answering my question about the import, could you throw some more example plots on this PR please? (pick out 2 or 3 of the analytical expressions at the end of Ashlock's paper).

It's kind of all we've got to be sure it's implemented correctly and the (possible) symmetry of the plot for WSLS is making slightly suspicious (could just be stochasticity of course).

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -0,0 +1,155 @@
...
+        af = AshlockFingerprint(self.strategy, self.probe)
+        af.fingerprint(turns=10, repetitions=2, step=0.25, progress_bar=False)
+        p = af.plot()
+        self.assertIsInstance(p, matplotlib.pyplot.Figure)
+        q = af.plot(col_map='jet')
+        self.assertIsInstance(q, matplotlib.pyplot.Figure)
+
+    def test_actual_data_fingerprint(self):
+        axl.seed(0)  # Fingerprinting is a random process
+        test_data = OrderedDict([(Point(x=0.5, y=0.75), 2.200),
```
drvinceknight: Can you change this to be a dict (my mistake, should have done that on my PR).

So:

```
test_data = {Point(x=0.5, y=0.75): 2.200,...
```

Then can you remove `from collections import OrderedDict` from line 8.

axelrod/fingerprint.py
meatballs: also, its not it's

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+        edges : list of tuples
+            A list containing tuples of length 2. All tuples will have either 0
+            or 1 as the first element. The second element is the index of the
+            corresponding probe (+2 to allow for including the Strategy and it's
+            Dual).
+        """
+        edges = []
+        for index, point in enumerate(points):
+            #  Add 2 to the index because we will have to allow for the Strategy
+            #  and it's Dual
```
meatballs: its not it's

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+        return probes
+
+    def construct_tournament_elements(self, step):
+        """Build the elements required for a spatial tournament
+
+        Parameters
+        ----------
+        step : float
+            The separation between each coordinate. Smaller steps will
+            produce more coordinates that will be closer together.
```
meatballs: Should this say points rather than coordinates?

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+
+        Returns
+        ----------
+        edges : list of tuples
+            A list containing tuples of length 2. All tuples will have either 0
+            or 1 as the first element. The second element is the index of the
+            corresponding probe (+2 to allow for including the Strategy and it's
```
meatballs: its

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+        Matches interactions from `results` to their corresponding Point in
+        `probe_points`.
+
+        Parameters
+        ----------
+        interactions : dictionary
+            A dictionary of the interactions of a tournament
+        points : list of Points
+            A list of tuples of length 2, where each tuple represents a
+            coordinate (x, y).
```
meatballs: point with coordinates

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+        edges : list of tuples
+            A list containing tuples of length 2. All tuples will have either 0
+            or 1 as the first element. The second element is the index of the
+            corresponding probe (+2 to allow for including the Strategy and it's
```
meatballs: its

axelrod/fingerprint.py
```diff
@@ -0,0 +1,277 @@
...
+
+        Parameters
+        ----------
+        turns : integer, optional
+            The number of turns per match
+        repetitions : integer, optional
+            The number of times the round robin should be repeated
+        step : float, optional
+            The separation between each coordinate. Smaller steps will
+            produce more coordinates that will be closer together.
```
meatballs: points rather than coordinates?

marcharper:
Sorry I haven't been following this thread, and now it seems to late to try to join. The images above look awesome! Once the PR is in I suggest we add a fingerprint image for each strategy to the axelrod examples repository.

Is it possible to have an interpolated version that's visually closer to the analytic version?

drvinceknight:
> The images above look awesome! Once the PR is in I suggest we add a fingerprint image for each strategy to the axelrod examples repository.

Something  along those lines will definitely happen. Each fingerprint (if done at a high enough granularity) can take quite a while so I'm not sure if it should be in the notebook repo or in it's own separate repo.

> 
> Is it possible to have an interpolated version that's visually closer to the analytic version?

Great idea, I think this can just be an argument passed to `fingerprint.plot` which would in turn pass it to `imshow`: http://matplotlib.org/examples/images_contours_and_fields/interpolation_methods.html

Could you make a note of that in the docs too @theref? Perhaps draw an interpolated version of the larger image that you have in there.

If it ends up being more complicated then I suggest it's a separate issue.

axelrod/__init__.py
theref: We don't, I've removed it now ✂️

meatballs:
> and now it seems too late to try to join.

@marcharper I don't think so!

theref:
I will be posting some plots on here soon, I'll also look into which interpolation looks the 'best'

drvinceknight:
> I'll also look into which interpolation looks the 'best'

Great will be neat to look at: but you don't need to make a choice for the purpose of the code right? You can just let the user pass the `interpolation` variable to `fingerprint.plot` which would pass it to `imshow`, so the user can use whichever method they want.

theref:
> don't need to make a choice for the purpose of the code right?

Yeah, I've put the functionality in and given it a default of `'none'`

theref:
WSLS
screen shot 2016-11-18 at 10 52 24
screen shot 2016-11-17 at 22 03 13

theref:
Majority
screen shot 2016-11-17 at 22 04 19
screen shot 2016-11-17 at 21 04 51

theref:
TFT
screen shot 2016-11-17 at 22 01 51
screen shot 2016-11-17 at 21 04 30

drvinceknight:
I'm happy.

To give some context for the Majority image: that's a screengrab from the original article, it's a strategy for which you cannot get an anaytical function and is plotted with a different orientation/ordering of coordinates but a flip in the y=x axis gives the same image which is what we expect. (The @theref can you just confirm I've got that right).

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+    """Creates a set of Points over the unit square.
+
+    A Point has coordinates (x, y). This function constructs points that are
+    separated by a step equal to `step`. The points are over the unit
+    square which implies that the number created will be 1/`step`^2.
+
+    Parameters
+    ----------
+    step : float
+        The separation between each coordinate. Smaller steps will
```
meatballs: point rather than coordinate

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+        produce more coordinates that will be closer together.
```
meatballs: points

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+
+    Returns
+    ----------
+    points : list of Point
+        Named Tuples of length 2 representing each coordinate (x, y)
```
meatballs:
```
points : list
     of Point objects
```

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+        """Creates a JossAnn probe player that matches the Point.
+
+        If the coordinates of point sums to more than 1 the parameters are
+        flipped and subtracted from 1 to give meaningful probabilities. We also
+        use the Dual of the probe. This is outlined further in [Ashlock2010]_.
+
+        Parameters
+        ----------
+        point : Point
+            has coordinates (x, y)
```
meatballs: get rid of the description line. It adds nothing and is a source of confustion

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+                                     progress_bar=progress_bar)
+        if in_memory:
+            self.interactions = self.spatial_tournament.interactions_dict
+        else:
+            self.interactions = read_interactions_from_file(filename)
+
+        self.data = self.generate_data(self.interactions, self.points, edges)
+        return self.data
+
+    def plot(self, col_map='seismic', interp_method='none'):
```
drvinceknight: The tiniest of things (not a blocker for me). Do we want `interp_method` to be `interpolation` so that's it's the same variable name as for `imshow`?

Not a blocker for me, happy to let @meatballs and/or @marcharper have the final say on that.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+        ----------
+        point : Point
+            has coordinates (x, y)
+        probe : class
+            A class that must be descended from axelrod.strategies
+
+        Returns
+        ----------
+        joss_ann: Joss-AnnTitForTat object
+            `JossAnnTransformer` with parameters that correspond to (x, y).
```
meatballs: point rather than (x, y)

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+    def create_edges(points):
+        """Creates a set of edges for a spatial tournament.
+
+        Constructs edges that correspond to `points`. All edges begin at 0, and
+        connect to the index +1 of the probe.
+
+        Parameters
+        ----------
+        points : list of Points
+            Tuples of length 2 representing each coordinate (x, y)
```
meatballs: same comment as previously

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Returns
+        ----------
+        edges : list of tuples
+            A list containing tuples of length 2. All tuples will have 0 as the
+            first element. The second element is the index of the
+            corresponding probe (+1 to allow for including the Strategy and it's
```
meatballs: its rather than it's

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+            Dual).
+        """
+        edges = [(0, index + 1) for index, point in enumerate(points)]
+        return edges
+
+    def create_probes(self, probe, points):
+        """Creates a set of probe strategies over the unit square.
+
+        Constructs probe strategies that correspond to (x, y) coordinates. The
```
meatballs: points rather than coordinates

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+        probes are created using the `JossAnnTransformer`.
+
+        Parameters
+        ----------
+        probe : class
+            A class that must be descended from axelrod.strategies.
+        points : list of Points
```
meatballs: same comment as previously

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+            Tuples of length 2 representing each coordinate (x, y)
+
+        Returns
+        ----------
+        probes : list
+            A list of `JossAnnTransformer` players with
+            parameters that correspond to (x, y).
```
meatballs: point rather than (x, y)

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+    def generate_data(interactions, points, edges):
+        """Generates useful data from a spatial tournament.
+
+        Matches interactions from `results` to their corresponding Point in
+        `probe_points`.
+
+        Parameters
+        ----------
+        interactions : dictionary
+            A dictionary of the interactions of a tournament
```
meatballs: a dictionary mapping what to what? This could do with some more clarity here.

axelrod/fingerprint.py
```diff
@@ -0,0 +1,267 @@
...
+        points : list of Points
```
meatballs: same comment as previously

docs/tutorials/further_topics/fingerprinting.rst
```diff
@@ -0,0 +1,60 @@
...
+
+In reality we would need much more detail to make this plot useful.
+Using pararemeters :code:`turns=50, repetitions=2, step=0.01` we get the plot:
+
+.. image:: _static/fingerprinting/WSLS_large.png
+     :width: 100%
+     :align: center
+
+We are also able to specify a matplotlib colour map and interpolation. Passing
+parameters :code:`col_map='PuOr', interp_method='bicubic'` to the :code:`plot()`
```
drvinceknight: This is no longer correct (`interp_method` has been changed to `interpolation`).

PR #761, opened 2016-11-10
Stop random_choice sampling when used with 0 or 1

### LINKED ISSUE #760: Do we want to skip random sampling for deterministic memory one (and other) strategies?
**EDIT**: This is a long issue but it actually suggests a very minor edit to the `random_choice` function. Thought documenting the long thing was worth it as it's kind of quirky behaviour. :)

Here's a gist (with the code I'm about to write below): https://gist.github.com/drvinceknight/eea3e3e9efd5170fa9f15cdc9cf283a1

Seeding play and creating a `Random` player. Against `p2` and `p3` (two instances of `WinStayLoseShift`) get the same sequences of play for the `Random` player: `['D', 'D', 'C', 'C', 'C']` but against: `p4`(`TitForTat`)  get `['D', 'D', 'C', 'C', 'D']` (different -1'th element).

```
>>> axl.seed(0)
>>> p1 = axl.Random()
>>> p2 = axl.WinStayLoseShift()
>>> match = axl.Match((p1, p2), turns=5)
>>> random_plays_v_p2 = [r[0] for r in match.play()]
>>> random_plays_v_p2
['D', 'D', 'C', 'C', 'C']
>>> axl.seed(0)
>>> p1 = axl.Random()
>>> p3 = axl.WinStayLoseShift()
>>> match = axl.Match((p1, p3), turns=5)
>>> random_plays_v_p2 = [r[0] for r in match.play()]
>>> random_plays_v_p2
['D', 'D', 'C', 'C', 'C']
>>> axl.seed(0)
>>> p1 = axl.Random()
>>> p4 = axl.TitForTat()
>>> match = axl.Match((p1, p4), turns=5)
>>> random_plays_v_p2 = [r[0] for r in match.play()]
>>> random_plays_v_p2
['D', 'D', 'C', 'C', 'D']
```

This is because `WinStayLoseShift` is implemented as a memory one strategy (<https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/memoryone.py#L60>). It's a deterministic memory one strategy (the values in the memory-one vector are 0 or 1) but nonetheless it samples a random number (which offsets the random path). As you can see above, seeding is working correctly: `Random` plays the same against `WinStayLoseShift` both times (so there's no issues about reproducibility) so I'm not sure this needs a fix **but**:

The simplest fix (does this need fixing?) to this would be to modify the `random_choice` function <https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/random_.py> so that if p=0 it returns `D` and if it is 1 it returns `C` without sampling:

```
def random_choice(p=0.5):
    """
    Return 'C' with probability `p`, else return 'D'
    Emulates Python's random.choice(['C', 'D']) since it is not consistent
    across Python 2.7 to Python 3.4"""
    if p == 0:
        return Actions.D
    if p == 1:
        return Actions.C

    r = random.random()
    if r < p:
        return Actions.C
    return Actions.D
```

This would have the added benefit of reducing some calls of the random library.

(Tagging @theref as he brought this to my attention.)

drvinceknight:
Closes #760

- Test included to check that random seed is not offset
- Adjusted some remaining tests (surprisingly, not that many!)

### DIFF
--- a/axelrod/random_.py
+++ b/axelrod/random_.py
@@ -2,6 +2,9 @@ import random
 import numpy
 from axelrod import Actions
 
+# A dictionary mapping 0 or 1 probability of cooperating to an action.
+choices = {0: Actions.D, 1: Actions.C}
+
 
 def random_choice(p=0.5):
     """
@@ -20,16 +23,11 @@ def random_choice(p=0.5):
     -------
     axelrod.Actions.C or axelrod.Actions.D
     """
-    if p == 0:
-        return Actions.D
-
-    if p == 1:
-        return Actions.C
-
-    r = random.random()
-    if r < p:
-        return Actions.C
-    return Actions.D
+    try:
+        return choices[p]
+    except KeyError:
+        r = random.random()
+        return choices[r < p]
 
 
 def randrange(a, b):
--- a/axelrod/random_.py
+++ b/axelrod/random_.py
@@ -2,9 +2,6 @@ import random
 import numpy
 from axelrod import Actions
 
-# A dictionary mapping 0 or 1 probability of cooperating to an action.
-choices = {0: Actions.D, 1: Actions.C}
-
 
 def random_choice(p=0.5):
     """
@@ -23,11 +20,16 @@ def random_choice(p=0.5):
     -------
     axelrod.Actions.C or axelrod.Actions.D
     """
-    try:
-        return choices[p]
-    except KeyError:
-        r = random.random()
-        return choices[r < p]
+    if p == 0:
+        return Actions.D
+
+    if p == 1:
+        return Actions.C
+
+    r = random.random()
+    if r < p:
+        return Actions.C
+    return Actions.D
 
 
 def randrange(a, b):
--- a/axelrod/random_.py
+++ b/axelrod/random_.py
@@ -8,7 +8,23 @@ def random_choice(p=0.5):
     Return 'C' with probability `p`, else return 'D'
 
     Emulates Python's random.choice(['C', 'D']) since it is not consistent
-    across Python 2.7 to Python 3.4"""
+    across Python 2.7 to Python 3.4
+
+    Parameters
+    ----------
+
+    p : float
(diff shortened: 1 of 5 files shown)


axelrod/random_.py
```diff
@@ -8,7 +8,23 @@ def random_choice(p=0.5):
...
 
     Emulates Python's random.choice(['C', 'D']) since it is not consistent
-    across Python 2.7 to Python 3.4"""
+    across Python 2.7 to Python 3.4
+
+    Parameters
+    ----------
+
+    p : float
+        The probability of picking 'C'
```
marcharper: `Axelord.Actions.C` maybe is better

marcharper:
Might be neat to profile before and after to see how many calls to `random.random()` are saved.

drvinceknight:
Cool idea. I'll give it a go at some point (we've got the branches).

axelrod/random_.py
```diff
@@ -8,7 +8,23 @@ def random_choice(p=0.5):
...
+
+    Returns
+    -------
+    axelrod.Actions.C or axelrod.Actions.D
+    """
+    if p == 0:
+        return Actions.D
+
+    if p == 1:
+        return Actions.C
```
meatballs: Guided by "Flat is better than nested," this might be nicer as
```
choices = {
    0: Actions.C,
    1: Actions.D
}

return choices[p] 
```

drvinceknight: That was my first approach but it failed tests in certain areas because of things like '0.0' etc. I also think that in this particular case it's perhaps a bit more readable/simple as is (there isn't any nesting).

drvinceknight: 36dc519 actually might be more up your street :)

drvinceknight:
@marcharper, @meatballs have potentially just over engineered this, let us know if you think it's become a bit too unreadable and I can revert a thing or two. (I quite like it as it is to be honest.)

PR #762, opened 2016-11-12
Refactor random_choice to use dict

meatballs:
Another attempt at using a dict in the random_choice function

### DIFF
--- a/axelrod/random_.py
+++ b/axelrod/random_.py
@@ -2,11 +2,8 @@ import random
 import numpy
 from axelrod import Actions
 
-
-choices = {
-    0: Actions.D,
-    1: Actions.C
-}
+# A dictionary mapping 0 or 1 probability of cooperating to an action.
+choices = {0: Actions.D, 1: Actions.C}
 
 
 def random_choice(p=0.5):
--- a/axelrod/random_.py
+++ b/axelrod/random_.py
@@ -3,6 +3,12 @@ import numpy
 from axelrod import Actions
 
 
+choices = {
+    0: Actions.D,
+    1: Actions.C
+}
+
+
 def random_choice(p=0.5):
     """
     Return 'C' with probability `p`, else return 'D'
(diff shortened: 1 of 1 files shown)


axelrod/random_.py
```diff
@@ -3,6 +3,12 @@
 from axelrod import Actions
 
 
+choices = {
```
drvinceknight: Again: completely stylistic preference so just ignore me but FWIW, this could also be done as a list:

choices = [Actions.D, Actions.C]
The lookup below just becomes indexing. (Interestingly, just checked and it still works with the boolean. Neat.)

(Whether or not it should is another question.)

meatballs:
> Any inkling about performance/profiling?
>
> Does this help with readability?
>
> I'm not sure it does but I also am at this point agnostic about this small function. I think we're in the domain of personal preferences here so I'm retracting my personal preferences, happy to be agnostic.

This entire PR is very definitely scratching a personal itch. It was irritating me that I couldn't get it to work yesterday.

Please feel free to dump it if you feel it's not adding anything!

axelrod/random_.py
```diff
@@ -20,16 +26,14 @@ def random_choice(p=0.5):
     -------
     axelrod.Actions.C or axelrod.Actions.D
     """
-    if p == 0:
-        return Actions.D
 
-    if p == 1:
-        return Actions.C
+    if p in [0, 1]:
+        choice = choices[p]
```
meatballs: It could, but I think multiple exits in a function are best avoided unless there's a good reason for them. Again, more personal preference!

axelrod/random_.py
meatballs: Don't really have a preference. I tend to use dicts out of habit and I prefer either to conditionals. Not entirely sure why, though.....

meatballs:
I'm closing this. It's producing heat rather than light, I suspect. My itch is now scratched and itches should mostly remain private, I suspect!

My apologies.

PR #767, opened 2016-11-19
Adjust the range for the coordinates for the fingerprint to include 1

### LINKED ISSUE #766: The fingerprint range for x and y doesn't go to 1.
Using `np.arange` for our coordinates means we don't necessarily get any coordinate at `1`:

```
>>> np.arange(0, 1, .2)
array([ 0. ,  0.2,  0.4,  0.6,  0.8])
```

Given the definition by Ashlock (which defines this as an analytical function on the unit cube) I think it would be better to have coordinates at 0 and 1 (inclusive) with a setting for the granulatiry. Numpy has this:

```
>>> np.linspace(0, 1, 5)
array([ 0.  ,  0.25,  0.5 ,  0.75,  1.  ])
```

I should have noticed this on the PR.

Just pushing a PR that implements this change, I'm suggesting a change of parameter name there which is a pity given that I've also just pushed a release but perhaps that's ok...

drvinceknight:
This ensures we get the full coordinate range for our fingerprints.

This closes #766

CC @theref


drvinceknight:
FYI, here are the outputs of this slight modification:

```
axl.seed(0)
af = axl.AshlockFingerprint(axl.WinStayLoseShift, axl.TitForTat)
_ = af.fingerprint(turns=50, repetitions=20, num=100, processes=0)
af.plot();
```

gives:

wsls

```
axl.seed(0)
af = axl.AshlockFingerprint(axl.GoByMajority, axl.TitForTat)
_ = af.fingerprint(turns=50, repetitions=20, num=100, processes=0)
af.plot();
```

gives 

majority

(So nothing is breaking here, just being slightly more consistent with Ashlock's definition.)

drvinceknight:
I have renamed this issue and force pushed a change that keeps all variable names the same as before.

PR #775, opened 2016-12-02
Remove memoized for py3 caching

drvinceknight:
This is possible thanks to us dropping py2 :)


marcharper:
The first of many such PRs I hope :)

PR #777, opened 2016-12-03
New meta strategies

marcharper:
* Several new meta strategies
* Reworked the meta tests significantly (much simpler now I hope)
* ~~Caught some other classification related issues, and a cache issue that I haven't yet addressed (see #779)~~ now fixed

### DIFF
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -99,7 +99,7 @@ class DeterministicCache(UserDict):
         if len(key) != 3:
             return False
 
-        # The triplet should be a pair of axelrod.Player sublclasses and an
+        # The triplet should be a pair of axelrod.Player subclasses and an
         # integer
         try:
             if not (
@@ -111,7 +111,7 @@ class DeterministicCache(UserDict):
         except TypeError:
             return False
 
-        # Each Player class should be deterministic
+        # Each Player should be deterministic
         if key[0].classifier['stochastic'] or key[1].classifier['stochastic']:
             return False
 
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -99,7 +99,7 @@ class DeterministicCache(UserDict):
         if len(key) != 3:
             return False
 
-        # The triplet should be a pair of axelrod.Player subclasses and an
+        # The triplet should be a pair of axelrod.Player instances and an
         # integer
         try:
             if not (
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -11,7 +11,7 @@ from .meta import (
     MetaMajorityMemoryOne, MetaMajorityFiniteMemory, MetaMajorityLongMemory,
     MetaMinority, MetaMixer, MetaWinner, MetaWinnerDeterministic,
     MetaWinnerEnsemble, MetaWinnerMemoryOne, MetaWinnerFiniteMemory,
-    MetaWinnerLongMemory, MetaWinnerStochastic, MWEDeterministic, MWEFast,
+    MetaWinnerLongMemory, MetaWinnerStochastic, MWEDeterministic,
     MWEFiniteMemory, MWELongMemory, MWEMemoryOne, MWERandom, MWEStochastic
     )
 
@@ -31,7 +31,6 @@ all_strategies += [MetaHunter,
                    MetaWinnerLongMemory,
                    MetaWinnerStochastic,
                    MWEDeterministic,
-                   MWEFast,
                    MWEFiniteMemory,
                    MWELongMemory,
                    MWEMemoryOne,
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -9,10 +9,10 @@ from ._filters import passes_filterset
 from .meta import (
     MetaHunter, MetaHunterAggressive, MetaPlayer, MetaMajority,
     MetaMajorityMemoryOne, MetaMajorityFiniteMemory, MetaMajorityLongMemory,
-    MetaMinority, MetaMixer, MetaWinner,
-    MetaWinnerMemoryOne, MetaWinnerFiniteMemory, MetaWinnerLongMemory,
-    MetaWinnerEnsemble, MWEDeterministic, MWEFast, MWEFiniteMemory,
-    MWELongMemory, MWEMemoryOne, MWERandom, MWEStochastic
+    MetaMinority, MetaMixer, MetaWinner, MetaWinnerDeterministic,
+    MetaWinnerEnsemble, MetaWinnerMemoryOne, MetaWinnerFiniteMemory,
+    MetaWinnerLongMemory, MetaWinnerStochastic, MWEDeterministic, MWEFast,
+    MWEFiniteMemory, MWELongMemory, MWEMemoryOne, MWERandom, MWEStochastic
     )
 
 all_strategies += [MetaHunter,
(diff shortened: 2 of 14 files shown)


drvinceknight:
There are a number of tests failing (that at a glance look genuine) but I took a quick look at why the docs weren't building. 

It seems to be because we mock `numpy` (we do this so that there are no install problems on readthedocs) in the `docs/conf.py` file but one of your strategies uses `numpy.random.choice`. This is failing at the `import axelrod` stage of `docs/conf.py`. 

I'm not entirely sure why this is breaking for this particular strategy (we have other strategies that make use of `numpy.random.choice`. Let me know if you need more of a hand with this and I can look in to it more.

marcharper:
I'm working on the tests... I need to rerun the strategy timing notebook to check the long run time values.

One of the new strategies, MWERandom, is going to be hard to test as fully since it picks a random team each time.

drvinceknight:
> One of the new strategies, MWERandom, is going to be hard to test as fully since it picks a random team each time.

Yeah I thought that. As long as the lines get hit and we have some cases covered I'm happy...

marcharper:
The issue with MWERandom is that it ends up in a classifier filter test and it's classifier changes based on the instantiation. So while the classification is correct, the test that compares the two strategy sets fails randomly if the teams cause different classifiers in the meta strategy.

marcharper:
Help on the last two tests is appreciated.

drvinceknight:
> Help on the last two tests is appreciated.

I've pushed some code to `new_meta-vk`. 

I've fixed the filterset test, not by dropping `MWERandom` out of the test but by seeding both filtering approaches (I made this a hypothesis test to randomly pick the seed):

- 4f99e28447e1eca3c167cd9417d183967ccf987b tweaks how the filterset was doing this. It was previously initialising a player for each property. Now it just initialises a player once.
- 2a8f1d99f1debcff84f35468470a78c32f015f8a seeds the two approaches in the test.

I'm just going to take a look at the other failing test now :)

drvinceknight:
> I suspect one is due to the MetaWinner reset bug I fixed

I didn't quite follow what the bug was and what the fix is. Looks like you've removed the `reset` method from `adaptive`. What's the reason?

marcharper:
Adaptive still has its reset method, I just thought it was bad form to call it in the constructor.

The MetaWinner bug was that the cumulative scores were being recorded in the attributes of the players in the team and not being reset. This tripped your test on repeated deterministic matches for MetaWinnerDeterministic (being the only deterministic meta strategy) and has gone unnoticed for the other MetaWinner variants until now.

The fix is reseting the scores in MetaWinner's reset function, and making the test reset method better to try to prevent this in the future.

marcharper:
The second commit doesn't want to cherry-pick, if you're able to do it feel free to push directly to this branch.

drvinceknight:
> The second commit doesn't want to cherry-pick, if you're able to do it feel free to push directly to this branch.

I've just pushed that. Looking at the second test now :)

drvinceknight:
Reverting 8128e72 didn't seem to affect that test.

I've fixed it with: b97d413, this just sets the seed before the play of the tournament. I'm not entirely sure why that is required here.

drvinceknight:
b157d91 fixes the doc build.

This is actually an interesting precedent. `MWERandom` is going to be the first strategy whose classification is random on some dimensions.

This triggered problems with the doc build:

- The doc build imports axelrod (to be able to autodocument the strategies);
- To be able to use readthedocs we have mocked a variety of modules, including numpy;
- During the import of axelrod a bunch of filtered strategy sets are created which requires strategies to be initialised (including `MWERandom`).
- This was breaking when hitting `numpy.choice` in `MWERandom`'s init function.

I swapped `numpy.choice` for `random.sample`.

marcharper:
> I've fixed it with: b97d413, this just sets the seed before the play of the tournament. I'm not entirely sure why that is required here.

Maybe because MWERandom uses random values in its initialization?

drvinceknight:
The docs are failing to build, I'm taking a look at that

> Maybe because MWERandom uses random values in its initialization?

I thought that **but** the way it was before was seeding _after_ creating the players... Although maybe something I'm forgetting in the tournament initialization was initializing `MWERandom` again... Although we were seeding before both tournaments initialised, not sure I'd need to think about that one but my head is pretty dead right now :)

drvinceknight:
@marcharper I've merged in #782 so perhaps that will let you tweak some of the fixes you have here?

marcharper:
Ok I removed the classification workaround and all the tests passed locally. I also rebased onto the new master (which required a force push).

axelrod/deterministic_cache.py
```diff
@@ -99,7 +99,7 @@ def _is_valid_key(self, key):
         if len(key) != 3:
             return False
 
-        # The triplet should be a pair of axelrod.Player sublclasses and an
+        # The triplet should be a pair of axelrod.Player subclasses and an
```
drvinceknight: This has been changed: should be `axelrod.Player instances`

axelrod/strategies/__init__.py
```diff
@@ -7,17 +7,36 @@
...
+    MetaWinnerLongMemory, MetaWinnerStochastic, MWEDeterministic,
+    MWEFiniteMemory, MWELongMemory, MWEMemoryOne, MWERandom, MWEStochastic
     )
 
-all_strategies.extend([MetaHunter, MetaMajority, MetaMinority, MetaWinner,
-                       MetaMajorityMemoryOne, MetaWinnerMemoryOne,
-                       MetaMajorityFiniteMemory, MetaWinnerFiniteMemory,
-                       MetaMajorityLongMemory, MetaWinnerLongMemory, MetaMixer,
-                       MetaWinnerEnsemble])
+all_strategies += [MetaHunter,
```
drvinceknight: Is this just a stylistic change?

(I don't mind at all, not requesting a change here, just curious.)

axelrod/strategies/meta.py
```diff
@@ -37,11 +45,12 @@ def __init__(self):
         self.team = [t() for t in self.team]
 
         # This player inherits the classifiers of its team.
+        # Note that memory_depth is not simply the max memory_depth of the team.
```
drvinceknight: Really, why not?

meatballs:
The filtering was being done at class level in order for the API to work - instances aren't available at the point it's used.

Have the changes here stopped the filtering working on a class?

drvinceknight:
> Have the changes here stopped the filtering working on a class?

No, they still work on a class, the main filtering function `passes_filterset` takes a class. The change is that the sub functions (that `passes_filterset` call) you had written which also took the class now take a player: in other once we're only creating a player instance once (as opposed to creating an instance for each property being checked).

FYI This is the commit that implemented it: https://github.com/Axelrod-Python/Axelrod/pull/777/commits/e9273ef434a646adc38b2600eea63b1d8a55d985 and no actual tests were changed for this.

drvinceknight:
This does point to another slightly sticky problem with `MWERandom` and the fact that when initiated it can be take a random classification (this strategy sets an interesting precedent for that so let's try to get it right).

I think I would prefer if the classification should just be hard coded to take extreme values: `memory_length=inf`, `makes_use_of['game']=True` etc... That's the potential `memory_length` it can have...

My main thinking is reproducibility... If we wanted `MWERandom` in an experiment and looked at scores vs memory length then we'd get different results.

drvinceknight:
> My main thinking is reproducibility... If we wanted MWERandom in an experiment and looked at scores vs memory length then we'd get different results.

We could of course seed, but we're then talking about seeding every single init of the strategy (for example inside the tournament play itself!). 

I think my current thoughts would suggest that as a general rule of thumb, classification should be 'set' by the input parameters (if any). So for example `Random(0)` is deterministic, and `MWERandom()` should have infinite memory length.

axelrod/strategies/__init__.py
marcharper: Yes

marcharper:
We can remove MWERandom from the default set of strategies, I'm not super attached to it. But it did help reveal some issues as I'm sure you've noticed.

drvinceknight:
> But it did help reveal some issues as I'm sure you've noticed.

Absolutely! Found some great stuff :)

Would still be nice to have in, just a question of classifying it. I'm going to remove it for now and push.

drvinceknight:
I've removed `MWERandom`. Giving this a :+1: from me.

PR #782, opened 2016-12-04
Fixes the cache bug.

### LINKED ISSUE #779: Cache Bug Lurking
We have to instantiate the strategy before checking if it's stochastic since various strategies have parameters that could change the classification. This could easily be true for the other classifiers such as memory depth and even say length but I'm not sure if there are instances of the latter.

We need to change the cache to use the strategy names as keys and to check the instances for stochasticity rather than the class classifiers. I don't know if this is yet causing problems since we tend to dynamically generate the cache in the Match class now but there is certainly the potential for a problem. Right now I think Matches are ok because  `_cache_update_required` is properly checking stochasticity (the players are already instantiated). But the DeterministicCache is not quite doing the same thing -- it's checking the lass classifier rather than the instance classifier. Hopefully the deterministic repetition tests are keeping us safe at the moment.

This is the [line](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/deterministic_cache.py#L97) that I'm worried about:

```
        if key[0].classifier['stochastic'] or key[1].classifier['stochastic']:
            return False
```

Right now `key[0]` and `key[1]` are player classes and should be instantiated player names. We could also have an issue with multiple instances of the same class (one stochastic, one not):
```
print (str(axl.GTFT(0)))
print (str(axl.GTFT(0.5)))
print (str(axl.GTFT(0).__class__))
print (str(axl.GTFT(0.5).__class__))
```

```
GTFT: 0
GTFT: 0.5
<class 'axelrod.strategies.memoryone.GTFT'>
<class 'axelrod.strategies.memoryone.GTFT'>
```
Since:
```
print (str(axl.GTFT(0).classifier['stochastic']))
print (str(axl.GTFT(0.5).classifier['stochastic']))
print (str(axl.GTFT(0).__class__.classifier['stochastic']))
print (str(axl.GTFT(0.5).__class__.classifier['stochastic']))
```
```
False
True
True
True
```

The other issue is key collision:

```
print (str(axl.GTFT(0)))
print (str(axl.GTFT(1)))
print (str(axl.GTFT(0).__class__))
print (str(axl.GTFT(1).__class__))
```

```
GTFT: 0
GTFT: 1
<class 'axelrod.strategies.memoryone.GTFT'>
<class 'axelrod.strategies.memoryone.GTFT'>
```

Those are both deterministic and different!

```
print (str(axl.GTFT(0).classifier['stochastic']))
print (str(axl.GTFT(1).classifier['stochastic']))
print (str(axl.GTFT(0).__class__.classifier['stochastic']))
print (str(axl.GTFT(1).__class__.classifier['stochastic']))
```

```
False
False
True
True
```

drvinceknight:
Closes #779

This makes the underlying read/write of the cache be the string repr.
The passed key is the player instances (as opposed to previously going
to have to go get the class).

### DIFF
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -46,6 +46,24 @@ class DeterministicCache(UserDict):
         if file_name is not None:
             self.load(file_name)
 
+    def _key_transform(self, key):
+        """
+        Parameters
+        ----------
+        key: tuple
+            A 3-tuple: (player instance, player instance, match length)
+        """
+        return key[0].name, key[1].name, key[2]
+
+    def __delitem__(self, key):
+        return UserDict.__delitem__(self, self._key_transform(key))
+
+    def __getitem__(self, key):
+        return UserDict.__getitem__(self, self._key_transform(key))
+
+    def __contains__(self, key):
+        return UserDict.__contains__(self, self._key_transform(key))
+
     def __setitem__(self, key, value):
         """Overrides the UserDict.__setitem__ method in order to validate
         the key/value and also to set the turns attribute"""
@@ -60,7 +78,7 @@ class DeterministicCache(UserDict):
             raise ValueError(
                 'Value must be a list with length equal to turns attribute')
 
-        UserDict.__setitem__(self, key, value)
+        UserDict.__setitem__(self, self._key_transform(key), value)
 
     def _is_valid_key(self, key):
         """Validate a proposed dictionary key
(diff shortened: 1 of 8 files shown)


axelrod/deterministic_cache.py
```diff
@@ -46,6 +46,24 @@ def __init__(self, file_name=None):
...
             self.load(file_name)
 
+    def _key_transform(self, key):
+        """
+        Parameters
+        ----------
+        key: tuple
+            A 3-tuple: (player instance, player instance, match length)
+        """
+        return key[0].name, key[1].name, key[2]
```
marcharper: Can we check if the player has been instantiated? This could still fail in the same way, no?

drvinceknight: This is assuming that `key[0]` and `key[1]` are instance of the Player class. I have modified the valid key check to check that that's indeed the case.

drvinceknight:
@meatballs I'm merging this under the bug fix rule.

PR #783, opened 2016-12-04
Reclassify Random player if p is 0 or 1

### LINKED ISSUE #781: Random strategy does not change it's classification
We need to include something similar to the setup in memory one strategies that overwrite the stochastic classifier upon initialisation:

```
>>> p1 = axl.Random(0)
>>> p2 = axl.Random(1)
>>> p1.classifier['stochastic'], p2.classifier['stochastic']
(True, True)
```

Mildly connected to #779 (found it whilst working on that).

drvinceknight:
- Add a test for stochastic classification of Random

Closes #781

### DIFF
--- a/axelrod/strategies/rand.py
+++ b/axelrod/strategies/rand.py
@@ -36,6 +36,8 @@ class Random(Player):
         """
         Player.__init__(self)
         self.p = p
+        if p in [0, 1]:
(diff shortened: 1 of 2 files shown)


drvinceknight:
Merging this under the "bug" rule.

PR #785, opened 2016-12-05
Allow fingerprinting of instances.

drvinceknight:
Currently fingerprinting takes classes as the target for the
fingerprint.

This was an oversight as it doesn't allow for the fingerprinting of
parametrized strategies. This allows for both.

### DIFF
--- a/axelrod/fingerprint.py
+++ b/axelrod/fingerprint.py
@@ -56,10 +56,12 @@ class AshlockFingerprint():
         """
         Parameters
         ----------
-        strategy : class
-            A class that must be descended from axelrod.Player
-        probe : class
-            A class that must be descended from axelrod.Player
+        strategy : class or instance
+            A class that must be descended from axelrod.Player or an instance of
+            axelrod.Player.
+        probe : class or instance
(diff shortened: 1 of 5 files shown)


axelrod/fingerprint.py
```diff
@@ -37,14 +37,15 @@ def create_points(step):
 
 
 class AshlockFingerprint():
-    def __init__(self, strategy, probe):
+    def __init__(self, strategy, probe=axl.TitForTat):
         """
         Parameters
         ----------
         strategy : class
```
marcharper: Docs should say strategy not class now?

axelrod/fingerprint.py
```diff
@@ -37,14 +37,15 @@ def create_points(step):
...
             A class that must be descended from axelrod.Player
         probe : class
             A class that must be descended from axelrod.Player
```
drvinceknight: Technically it could but we're using decorators (which require a class) down the line and that would make things slightly more complicated than I intended with this PR. I'll stop being lazy and do it...

drvinceknight:
> Just the comments about the doc strings and the probe

Done.

(Rebased on to master.)

PR #786, opened 2016-12-05
Add BetterAndBetter, WorseAndWorse2 and WorseAndWorse3 Strategies

varung97:


varung97:
This pull request is currently incomplete as I haven't yet implemented all the strategies.
I was hoping you could glance over it to see whether it meets the guidelines for this project.
Also, I was wondering whether, since the `BetterAndBetter` strategy is random, I should add a test that checks if say the number of defects out of the first 20 is very high, or a test which checks that after a 1000 turns the player always cooperates?
Thanks!

drvinceknight:
> This pull request is currently incomplete as I haven't yet implemented all the strategies.
> I was hoping you could glance over it to see whether it meets the guidelines for this project.

Sorry for taking a while to get back to you @varung97, this looks right in terms of guidelines. Please carry on implementing the other strategies and we'll review it properly :)

> 
> Also, I was wondering whether, since the BetterAndBetter strategy is random, I should add a test that checks if say given any seed, the number of defects out of the first 20 is very high, or a test which checks that after a 1000 turns the player always cooperates? I guess with the first there would be a certain probability that the test fails, which could be an issue.

The way we've done things so far is to include tests with seeds.

varung97:
Alright, thanks! I'm closing this pull request while I work on the other strategies, will reopen it when I'm done.

marcharper:
@varung97 You can leave it open if you want, just tag it with "ready for review" when you are ready. We're usually happy to help write tests as time allows.

axelrod/strategies/worse_and_worse.py
```diff
@@ -57,3 +57,72 @@ def strategy(self, opponent):
...
+    def strategy(self, opponent):
+        current_round = len(self.history) + 1
+
+        if current_round == 1:
+            return C
+        elif current_round <= 20:
+            return opponent.history[-1]
+        else:
+            probability = 20 / float(current_round)
+            return random_choice(probability)
```
varung97: Do you think I should make a `random_choice_defects` function or something of the sort, as it might be a little confusing that the stated probability is `1 - 20 / current turn`, but the probability in the code is just `20 / current turn`?
Or should I add a comment to that effect?

varung97:
I'm not able to add labels to the PR I think 😅 (unless I'm doing something wrong)

drvinceknight:
Yeah only admins can add labels but just ping us here when this is ready :) :+1:

axelrod/strategies/worse_and_worse.py
drvinceknight: Don't make a new function, I think it's fine as it is :)

varung97:
Oh I see. Could you add the ready for review label then?

drvinceknight:
Sure thing (on my phone right now but I'll do it as soon as I get on front of my machine) :+1:

varung97:
Alright, thanks!

axelrod/strategies/worse_and_worse.py
varung97: Okies

axelrod/strategies/better_and_better.py
```diff
@@ -0,0 +1,29 @@
...
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent):
+        current_round = len(self.history) + 1
+        probability = float(current_round) / 1000
```
drvinceknight: Axelrod doesn't support python 2 anymore so you don't need to force float division, you can have this simply as `current_round / 1000`.

axelrod/strategies/worse_and_worse.py
```diff
@@ -57,3 +57,72 @@ def strategy(self, opponent):
...
+
+    def strategy(self, opponent):
+        current_round = len(self.history) + 1
+
+        if current_round == 1:
+            return C
+        elif current_round <= 20:
+            return opponent.history[-1]
+        else:
+            probability = 20 / float(current_round)
```
drvinceknight: No need for force float division.

axelrod/strategies/worse_and_worse.py
```diff
@@ -57,3 +57,72 @@ def strategy(self, opponent):
...
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent):
+        current_round = len(self.history) + 1
+
+        if current_round == 1:
+            return C
+        else:
+            probability = 1 - opponent.defections / float(current_round - 1)
```
drvinceknight: No need for float.

axelrod/tests/unit/test_worse_and_worse.py
```diff
@@ -105,3 +105,87 @@ def test_strategy(self):
...
+        'memory_depth': float('inf'),
+        'stochastic': True,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    class TestWAW3vsCooperator(TestHeadsUp):
```
drvinceknight: You don't need to create another class here, you can simply use different methods for each other these.

axelrod/strategies/better_and_better.py
varung97: Oops right will fix that

axelrod/tests/unit/test_worse_and_worse.py
varung97: So should I do something like this?
`opponent = axelrod.Cooperator()`
`player = self.player()`
`match = axelrod.Match((opponent, player), turns=10)`
`self.assertEqual(match.play(), [('C', 'C')] * 10)`

drvinceknight: Yup you could do that, you could also use the inbuilt testing help function: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html

varung97:
Is this what you were looking for?

drvinceknight:
> Is this what you were looking for?

Yup, this looks good to me. Giving it a :+1:, fyi we have a two core reviewer policy: @marcharper and/or @meatballs might have some other requests :)

marcharper:
Looks good to me. Thanks for the contribution!

varung97:
Yay, thanks guys!

PR #787, opened 2016-12-06
Add plotting option to Fingerprint plot

theref:
Adds:
 - option to have a color bar
 - option to have the axis labels


axelrod/fingerprint.py
```diff
@@ -267,7 +267,7 @@ def fingerprint(self, turns=50, repetitions=10, step=0.01, processes=None,
         self.data = self.generate_data(self.interactions, self.points, edges)
         return self.data
 
-    def plot(self, col_map='seismic', interpolation='none', title=None):
+    def plot(self, col_map='seismic', interpolation='none', title=None, colorbar=True, labels=True):
```
meatballs: Surely this has to be colourbar?

meatballs: (I know it's called colorbar in matplotlib, but still.....)

axelrod/fingerprint.py
```diff
@@ -284,6 +284,8 @@ def plot(self, col_map='seismic', interpolation='none', title=None):
             http://matplotlib.org/examples/images_contours_and_fields/interpolation_methods.html
         title : str, optional
             A title for the plot
+        colorbar : bool, optional
+            Choose whether the colorbar should be included or not
```
drvinceknight: labels needs to be added here.

axelrod/fingerprint.py
```diff
@@ -293,9 +295,26 @@ def plot(self, col_map='seismic', interpolation='none', title=None):
...
+
+        plt.xlabel('y')
+        plt.ylabel('x', rotation=0)
+        ax.xaxis.tick_top()
+        ax.xaxis.set_label_position('top')
+        ax.tick_params(axis=u'both', which=u'both', length=0)
+        plt.xticks([0, len(plotting_data) - 1], ['0', '1'])
+        plt.yticks([0, len(plotting_data) - 1], ['0', '1'])
+
+        if not labels:
```
drvinceknight: Can you add a test with `labels=False` and `colorbar=False` (just to check the plot runs).

axelrod/fingerprint.py
meatballs: To be clear - I'm not actually asking for any change here.

drvinceknight:
> Also, as the defaults are True (which I like) could you update the images for the docs?

Not sure if you saw that comment @theref

axelrod/fingerprint.py
```diff
@@ -293,9 +297,26 @@ def plot(self, col_map='seismic', interpolation='none', title=None):
...
+        fig, ax = plt.subplots()
+        cax = ax.imshow(plotting_data, cmap=col_map, interpolation=interpolation)
+
+        if colorbar:
+            max_score = max(ordered_data)
+            min_score = min(ordered_data)
+            ticks = [min_score, (max_score + min_score) / 2, max_score]
+            fig.colorbar(cax, ticks=ticks)
+
+        plt.xlabel('y')
```
marcharper: `xlabel` is `'y'` and `ylabel` is `'x'`?

marcharper:
Just the one request for clarification from me.

theref:
A couple of examples:

screen shot 2016-12-07 at 09 14 00
screen shot 2016-12-07 at 09 14 13

drvinceknight:
This looks good, need to fix the conflicts with master and can you change `x` to `$x$` (same for `y`) in the axis label.

axelrod/fingerprint.py
drvinceknight: Yeah, I think it's worth the change.

axelrod/fingerprint.py
meatballs: also col_map and interpolation

axelrod/fingerprint.py
```diff
@@ -284,18 +312,38 @@ def plot(self, col_map='seismic', interpolation='none', title=None):
...
+        if colorbar:
+            max_score = max(self.data.values())
+            min_score = min(self.data.values())
+            ticks = [min_score, (max_score + min_score) / 2, max_score]
+            fig.colorbar(cax, ticks=ticks)
+
+        plt.xlabel('$x$')
+        plt.ylabel('$y$', rotation=0)
+        # ax.xaxis.tick_top()
+        # ax.xaxis.set_label_position('top')
```
meatballs: If these two lines aren't required, can you get rid of them, please?

axelrod/fingerprint.py
```diff
@@ -284,18 +312,38 @@ def plot(self, col_map='seismic', interpolation='none', title=None):
...
+        ax.tick_params(axis=u'both', which=u'both', length=0)
```
meatballs: There's no need for 'u' now as we no longer support Python 2

axelrod/fingerprint.py
theref: col_map and interpolation are already in there

meatballs: `filename` appears in the docstring but doesn't seem to be in the signature. Is that still required?

theref: nah, i've just got rid

drvinceknight:
I'm happy. 

@meatballs I suggest we wait till @marcharper wakes to merge, as all 3 of us have commented here now :)

drvinceknight:
@marcharper note that the examples in the conversation above are no longer correct, look at the files changed to see what they look like now :)

PR #791, opened 2016-12-12
Additions to Plots

### LINKED ISSUE #776: Add ability to pass more options to plot objects
axis labels for example (this came up on gitter).

Nikoleta-v3:
Closes #776
Passing matplotlib axes.

Giving access to the matplotlib axes of the plots
in the library. If user does not specify the axes
then it uses `plt.subplots`.

Some measures, such as number of players, players
etc were used in more than one plot so I initiliaze
them in the class. Also, I have not yet removed the
title argument because I am not sure if is needed
for when running our tournament and saving the plots.

Please let me know you suggestions.


.travis.yml
```diff
@@ -15,6 +15,7 @@ install:
   - pip install -r requirements.txt
   - pip install sphinx
   - pip install sphinx_rtd_theme
+  - pip install docutils==0.12
```
drvinceknight: What's this for?

Nikoleta-v3: Travis was failing because of this error: 
```
assert not self.context, 'len(context) = %s' % len(self.context)
AssertionError: len(context) = 1```

and i just followed the sphinx issues where they said you need to also install this.

drvinceknight: :+1:

drvinceknight:
> Also, I have not yet removed the
> title argument because I am not sure if is needed
> for when running our tournament and saving the plots.

I suggest leaving this as it is for the sake of backwards compatibility.

marcharper:
Looks good to me, I'll wait on @drvinceknight's title question to merge, or merge with my blessing as you wish.

drvinceknight:
If you're happy I'm happy. :+1:

PR #792, opened 2016-12-19
Add short run time strategy list.

### LINKED ISSUE #771: Short and Long Run-time strategies
I'd like to add a few more meta strategies to the library, such as MetaWinners for the collections of stochastic and deterministic strategies. I do not, however, want to add more long run time strategies, since we've already got several and they take up the vast majority of the run time (especially against each other).

So I'd like to propose that we leave out the long run time strategies from the default collection of strategies, or at least add a prominent collection of "short run time" strategies:
```
fast_strategies = [s for s in all_strategies if obey_axelrod(s())
                           and not s().classifier['long_run_time']]
```

This is because I don't think new comers to the library will necessarily be aware that there are long time strategies that take up a lot of run time.

drvinceknight:
Closes #771

Have included a very basic test and have also done some minor tidying
up.

### DIFF
--- a/axelrod/strategies/__init__.py
+++ b/axelrod/strategies/__init__.py
@@ -46,8 +46,8 @@ strategies = [s for s in all_strategies if obey_axelrod(s())]
 
 long_run_time_strategies = [s for s in all_strategies if
                             s().classifier['long_run_time']]
-short_run_time_strategies = [s for s in strategies
-                             if s not in long_run_time_strategies]
+short_run_time_strategies = [s for s in strategies if not
+                             s().classifier['long_run_time']]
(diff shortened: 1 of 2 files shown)


axelrod/strategies/__init__.py
```diff
@@ -46,6 +46,8 @@
 
 long_run_time_strategies = [s for s in all_strategies if
                             s().classifier['long_run_time']]
+short_run_time_strategies = [s for s in strategies
```
marcharper: Can we define this as:
```
short_run_time_strategies = [s for s in all_strategies if
                             not s().classifier['long_run_time']]
```

drvinceknight: Yeah good shout.

PR #794, opened 2016-12-23
Nice meta

marcharper:
Once upon a time I made all the MetaWinner strategies "nice" -- they never defect unless the opponent does first. This added a lot of nice strategies to the library. In this PR I add a transformer to make any strategy nice, and diversify the meta strategy collection a bit by making some not nice.

I recommend that we remove some of the variants from the standard collection in the near future-- there's not much difference in performance among the variants, and there are a number of better strategies incoming. But that's a conversation for another PR.


axelrod/strategy_transformers.py
```diff
@@ -219,6 +216,19 @@ def forgiver_wrapper(player, opponent, action, p):
     forgiver_wrapper, name_prefix="Forgiving")
 
 
+def nice_wrapper(player, opponent, action):
+    """If a strategy wants to defect, flip to cooperate with the given
```
drvinceknight: This docstring isn't correct right? There is no probability involved.

marcharper: Yep, lazy copy paste on my part.

marcharper: Done

drvinceknight:
One final thing, could you add a little something about this transformer to http://axelrod.readthedocs.io/en/latest/tutorials/advanced/strategy_transformers.html?

PR #796, opened 2016-12-29
Moran processes are always stochastic

marcharper:
The Matches in the process may still be deterministic but the process itself still uses random values for the fitness proportionate selection stages.


drvinceknight:
Merging under bug fix.

marcharper:
> Looks good to me. This is_stochastic property wasn't actually used anywhere right?

Not in the library anywhere.

PR #797, opened 2016-12-31
Add initial play summary to the results set.

### LINKED ISSUE #789: Add initial play rate to the result summary.
Would be cool to have the summary of the initial play rate for each player as a column in the results summary.

drvinceknight:
Keep track of the rate/count of which a
strategy cooperates in the initial round.

Closes #789.

### DIFF
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -383,6 +383,7 @@ class ResultSet(object):
         self.cooperation = [[0 for opponent in plist] for player in plist]
         self.normalised_cooperation = [[[] for opponent in plist]
                                        for player in plist]
+        self.initial_cooperation_count = [0 for player in plist]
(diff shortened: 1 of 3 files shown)


axelrod/result_set.py
```diff
@@ -532,6 +533,23 @@ def _update_cooperation(self, p1, p2, cooperations):
...
+        attribute
+
+        Parameters
+        ----------
+
+            p1, p2 : int
+                The indices of the first and second player
+            initial_cooperations : tuple
+                A 2 tuple with a 0 or 1 indicating if the initial move of each
+                player was cooperation (0).
```
marcharper: How about:
```
A 2 tuple with a 0 or 1 indicating if the initial move of each player was Cooperation (0) or Defection (1), e.g. (0, 1) for a round (C, D)
```

PR #799, opened 2017-01-01
Implemented the Moran process on graphs

marcharper:
[Slides here](http://www.nimbios.org/tutorials/gametheoryfiles/evolutionary_graph_theory_back_Broom.pdf) for context.

### DIFF
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -5,6 +5,7 @@ on_windows = os.name == 'nt'
 
 # The order of imports matters!
 from .version import __version__
+from . import graph
 from .actions import Actions, flip_action
 from .random_ import random_choice, seed
 from .plot import Plot
@@ -13,7 +14,7 @@ from .player import get_state_distribution_from_history, init_args, is_basic, \
     obey_axelrod, update_history, update_state_distribution, Player
 from .mock_player import MockPlayer, simulate_play
 from .match import Match
-from .moran import MoranProcess
+from .moran import MoranProcess, MoranProcessGraph
 from .strategies import *
 from .deterministic_cache import DeterministicCache
 from .match_generator import *
--- a/axelrod/graph.py
+++ b/axelrod/graph.py
@@ -18,33 +18,37 @@ class Graph(object):
         [[node1, node2, weights], ...]
     Weights can be omitted for an undirected graph.
 
-    For efficiency, neighbors are cached in dictionaries.
+    For efficiency, neighbors are cached in dictionaries. Undirected graphs
+    are implemented as directed graphs in which every edge (s, t) has the
+    opposite edge (t, s).
     """
 
     def __init__(self, edges=None, directed=False):
+        self.directed = directed
+        self.original_edges = edges
         self.out_mapping = defaultdict(lambda: defaultdict(float))
         self.in_mapping = defaultdict(lambda: defaultdict(float))
-        self.directed = directed
+        self._edges = []
         if edges:
             self.add_edges(edges)
 
-    def add_vertex(self, label):
-        self._vertices.add(label)
-
-    def add_edge(self, source, target, weight=1.):
-        self.out_mapping[source][target] = weight
-        self.in_mapping[target][source] = weight
-        if not self.directed:
+    def add_edge(self, source, target, weight=None):
+        if (source, target) not in self._edges:
+            self._edges.append((source, target))
+            self.out_mapping[source][target] = weight
+            self.in_mapping[target][source] = weight
+        if not self.directed and (source != target) and \
+                (target, source) not in self._edges:
+            self._edges.append((target, source))
             self.out_mapping[target][source] = weight
             self.in_mapping[source][target] = weight
 
     def add_edges(self, edges):
-        try:
-            for source, target, weight in edges:
-                self.add_edge(source, target, weight)
-        except ValueError:
-            for source, target in edges:
-                self.add_edge(source, target, 1.0)
+        for edge in edges:
+            self.add_edge(*edge)
+
+    def edges(self):
+        return self._edges
 
     def vertices(self):
         """Returns the set of vertices of the graph."""
@@ -62,20 +66,13 @@ class Graph(object):
         """Returns a dictionary of the incoming edges of source with weights."""
         return self.in_mapping[target]
 
-    def normalize_weights(self):
-        """Normalizes the weights coming out of each vertex to be probability
-        distributions."""
-        new_edges = []
-        for source in self.out_mapping.keys():
-            total = float(sum(out_mapping[source].values()))
-            for target, weight in self.out_mapping.items():
-                self.out_mapping[target] = weight / total
-        self._edges = new_edges
+    def in_vertices(self, source):
+        """Returns a list of the outgoing vertices."""
+        return list(self.in_mapping[source].keys())
 
-    def __getitem__(self, k):
-        """Returns the dictionary of outgoing edges. You can access the weight
-        of an edge with g[source][target]."""
-        return self.out_mapping[k]
+    def __repr__(self):
+        s = "<Graph: {}>".format(repr(self.original_edges))
+        return s
 
 
 ## Example Graphs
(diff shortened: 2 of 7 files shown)


axelrod/__init__.py
```diff
@@ -5,6 +5,7 @@
 
 # The order of imports matters!
 from .version import __version__
+from . import graph
```
drvinceknight: I would have thought this line wasn't needed? Am I wrong?

axelrod/graph.py
```diff
@@ -0,0 +1,126 @@
+"""
+Weighted undirected sparse graphs.
+
+Original source:
+https://github.com/marcharper/stationary/blob/master/stationary/utils/graph.py
```
drvinceknight: Not against this. A couple of initial thoughts:

1. As you have built this for your stationary library you no doubt have a reason but why we wouldn't just use networkx (which would have the quick lookups we want here)? (1 negative side is adding another dependency to the library) 
2. Can we add some tests please?
3. The spatial tournament does not necessarily need a graph object (it just loops through the edges once via the match generator) but if we're going to have an `axelrod.graph` class we should also use that there (for consistency). I see two approaches:
     1. We allow the spatial tournament to take both a list of edges of one of these graph objects (ensures backwards compatibility);
     2. We let the Moran process take a list of edges and obfuscate the building of this underlying graph object there. I think this would be my preferred approach.

axelrod/graph.py
```diff
@@ -0,0 +1,126 @@
...
+    graph = Graph(directed=directed)
+    edges = []
+    for i in range(length - 1):
+        edges.append((i, i+1))
+    edges.append((length - 1, 0))
+    graph.add_edges(edges)
+    return graph
+
+
+def complete_graph(length, directed=False):
```
drvinceknight: Do we need these extra functions? Networkx can be used to easily get a variety of graphs. (Just a thought)

axelrod/graph.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def vertices(self):
+        """Returns the set of vertices of the graph."""
+        return list(self.out_mapping.keys())
+
+    def out_dict(self, source):
+        """Returns a dictionary of the outgoing edges of source with weights."""
+        return self.out_mapping[source]
+
+    def out_vertices(self, source):
```
drvinceknight: If undirected, why not just have a `adjacent_vertices` method? (To replace both the `in` and `out` ones?).

axelrod/graph.py
drvinceknight: To be clear: If we go with my suggestion of `3.i` then that would be work for a different PR.

docs/tutorials/getting_started/moran.rst
```diff
@@ -70,6 +70,50 @@ function like :code:`takewhile` from :code:`itertools`):
...
     ...         break
     >>> mp.population_distribution()
-    Counter({'Tit For Tat': 4})
+    Counter({'Cooperator': 4})
+
+
+Moran Process on Graphs
+-----------------------
+
+The library also provides a graph-based Moran process with `MoranProcessGraph`.
```
drvinceknight: Would you be able to include a reference text/paper in the bibliography?

axelrod/graph.py
marcharper: I figured this would come up which is why I didn't add tests yet. We can switch this out for another library, I just used what I had laying around. It's a bit overkill since we don't need the weights (yet?), but it optimized in the right ways -- sparse implementation, precomputed neighbors. So we could also just simplify it to parts needed (relatively easy).

I'm not opposed to networkX but I'm not familiar with its API, what internal representation it uses (adjacency matrix? sparse matrix? etc.).

I agree re: consistent API with spatial tournaments -- we could also use `isinstance` to be flexible on the input, and also have the `__repr__` method just output the list of edges. There are cases however that a list of edges doesn't cover, like some types of random graphs.

axelrod/graph.py
marcharper: We may want to allow the graphs to be directed, see e.g. [here](http://www.usma.edu/nsc/siteassets/sitepages/publications/a%20novel%20analytical%20method%20for%20evolutionary%20graph.pdf).

axelrod/graph.py
marcharper: I needed them for testing...

docs/tutorials/getting_started/moran.rst
marcharper: Yes I can add a reference.

axelrod/__init__.py
drvinceknight: > It depends on how you access the graph class in code, axelrod.graph.Graph will fail

I thought it wouldn't fail but happy as it is :)

axelrod/graph.py
drvinceknight: Cool.

axelrod/graph.py
drvinceknight: > There are cases however that a list of edges doesn't cover, like some types of random graphs.

That's the killer point right there! I guess that also includes the possibility of graphs that change over time etc which also rules out networkX (which really is a data representation library coupled with graph theoretic algorithms as opposed to the flexibility we might need here).

I suggest going for flexibility on the input (so we can open an issue for my suggestion of `3.i`).

axelrod/graph.py
drvinceknight: Cool, we can also use these in some of the tests for Spatial Tournaments (not in this PR).

axelrod/graph.py
drvinceknight: In order to get the `__repr__` method to output the list of edges I expect we could use an `edges` attribute or otherwise? That could be a useful attribute to have anyway?

marcharper: Yeah, we can save the input edges, or output a sorted list.

axelrod/graph.py
marcharper: I started to add more, an N x M grid could be useful... we can just add them as we go.

axelrod/graph.py
marcharper: Turns out we need the graph to be weighted for some applications in the literature. If the reproduction graph is weighted then it changes the death probabilities in the birth-death case.

drvinceknight:
There's a py3.6 failure (on the merge with master).

docs/reference/bibliography.rst
```diff
@@ -26,6 +26,7 @@ documentation.
 .. [PRISON1998] LIFL (1998) PRISON. Available at: http://www.lifl.fr/IPD/ipd.frame.html (Accessed: 19 September 2016).
 .. [Robson1989] Robson, Arthur, (1989), EFFICIENCY IN EVOLUTIONARY GAMES: DARWIN, NASH AND SECRET HANDSHAKE, Working Papers, Michigan - Center for Research on Economic & Social Theory, http://EconPapers.repec.org/RePEc:fth:michet:89-22.
 .. [Singer-Clark2014] Singer-Clark, T. (2014). Morality Metrics On Iterated Prisoner’s Dilemma Players.
+.. [Shakarian2013] Shakarian, P., Roos, P. & Moores, G. A Novel Analytical Method for Evolutionary Graph Theory Problems.
```
drvinceknight: Thanks for this: could you cite it in the moran.rst file?

drvinceknight:
> There's a py3.6 failure (on the merge with master).

@marcharper I took a look at the tests and pushed a fix. I pushed directly to this PR branch but feel free to change/modify/delete if you'd rather not.

> Thanks for this: could you cite it in the moran.rst file?

I also took the opportunity to pop the citation where I thought it should go. Only trying to be helpful so if you're not a fan please throw it out :)

marcharper:
Thanks for the test fix!

axelrod/graph.py
```diff
@@ -0,0 +1,128 @@
...
```
meatballs: Might some of these methods be better as properties?

marcharper: Getters would make sense for e.g `vertices` not sure about the others. Maybe edges makes sense with a setter, essentially rewriting the graph. I'm open to suggestions.

meatballs: Let's leave it for now. If we fancy refactoring the graph, we can do so later.

PR #803, opened 2017-01-02
ML strategies

marcharper:
This PR adds several new strategies.

* Two strategies from the literature: Winner12 and Winner21
* A new class of strategies based on hidden Markov models
* Newly trained versions of players based on finite state machines, HMM, EvolvedLookerUp, PSOGambler, and Evolved ANN

Of particular note:
* EvolvedFSM16, a 16 node FSM player that is the new "best" strategy
* EvolvedFSM16Noise05, a 16 node FSM player that is the new best strategy for noisy tournaments (and quite good in the noise free tournaments)
* PSOGamblerMem1, a memory one strategy trained with the PSO algorithm
* A revised evolved looker up that performs well
* EvolvedANN5 (with a smaller inner layer)
* EvolvedHMM, a hidden markov model based strategy

It's likely that these strategy are not the best possible, and that better versions can be evolved in the future. I trained many strategies, some of which are not included, particularly:
* Preliminary models for the Moran process (not added to the standard list, and likely to be improved)
* Models trained to win rather than achieve a high net score

There are also other conceivable training modes, e.g. maximum total score (self + opponent). The point is: we may decide to add or remove trained strategies in the future.

Also in this PR:
* Some of the strategy classes were refactored to allow better training. See the code in the [axelrod-evolver](https://github.com/Axelrod-Python/axelrod-evolver) repository for reference and training code.
* Model data for the trained strategies is stored in a subdirectory allowing for easy addition of new strategies, and to prevent storing data directly in strategy files. The exception is HMM (for which I've only included one strategy).
* To the point, since there are a lot of e.g. lookerup models, the strategy classes are created at runtime.


drvinceknight:
Oooof this is a biggy!

axelrod/strategies/worse_and_worse.py
```diff
@@ -12,12 +13,11 @@ class WorseAndWorse(Player):
 
     Names:
         - Worse and Worse: [PRISON1998]_
-
     """
 
     name = 'Worse and Worse'
     classifier = {
-        'memory_depth': float('inf'),
```
drvinceknight: It needs to know the current turn so does that not equate to knowing how long the game has been? (Thus infinite memory).

axelrod/tests/unit/test_hmm.py
```diff
@@ -0,0 +1,123 @@
+"""Tests for Finite State Machine Strategies."""
```
drvinceknight: hidden markov model strategies

docs/reference/bibliography.rst
```diff
@@ -22,6 +22,10 @@ documentation.
 .. [Li2009] Li, J. & Kendall, G. (2009). A Strategy with Novel Evolutionary Features for the Iterated Prisoner’s Dilemma. Evolutionary Computation 17(2): 257–274.
 .. [Li2011] Li, J., Hingston, P., Member, S., & Kendall, G. (2011). Engineering Design of Strategies for Winning Iterated Prisoner ’ s Dilemma Competitions, 3(4), 348–360.
 .. [Li2014] Li, J. and Kendall, G. (2016). The Effect of Memory Size on the Evolutionary Stability of Strategies in Iterated Prisoner's Dilemma. IEEE Transactions on Evolutionary Computation, 18(6) 819-826
+.. [Mathieu2015] Mathieu, P. and Delahaye, J. New Winning Strategies
```
drvinceknight: When the docs build does this look right? (Just doesn't match the one line formatting used for all others).

setup.py
```diff
@@ -13,8 +13,12 @@
     install_requires=requirements,
     author='Vince Knight, Owen Campbell, Karol Langner, Marc Harper',
     author_email=('axelrod-python@googlegroups.com'),
-    packages=['axelrod', 'axelrod.strategies', 'axelrod.tests'],
+    packages=['axelrod', 'axelrod.strategies', 'axelrod.tests', 'axelrod.data'],
     url='http://axelrod.readthedocs.org/',
     license='The MIT License (MIT)',
     description='Reproduce the Axelrod iterated prisoners dilemma tournament',
+    include_package_data=True,
```
drvinceknight: If you haven't, could you check that this pip installs? (eg pip installing in to a virtual env from the local dir)

axelrod/load_data_.py
```diff
@@ -0,0 +1,45 @@
+import pkg_resources
```
drvinceknight: Could we have unit tests for these please.

axelrod/plot.py
```diff
@@ -279,6 +279,7 @@ def save_all_plots(self, prefix="axelrod", title_prefix="axelrod",
             f = getattr(self, method)(title="{} - {}".format(title_prefix,
                                                              name))
             f.savefig("{}_{}.{}".format(prefix, method, filetype))
+            plt.close(f)
```
drvinceknight: THANK YOU! (This error message has been like a really annoying pebble in a shoe on a long walk...)

axelrod/strategies/ann.py
```diff
@@ -169,62 +166,39 @@ class EvolvedANN(ANN):
...
+        ANN.__init__(self, weights, num_features, num_hidden)
+
+
+class EvolvedANN5(ANN):
+    """
+    A strategy based on a pre-trained neural network.
+
+    Names:
+
+     - EvolvedANN2: : Original name by Marc Harper.
```
drvinceknight: ANN5

axelrod/strategies/ann.py
```diff
@@ -169,62 +166,39 @@ class EvolvedANN(ANN):
...
-            hidden_layer_size
-        )
-        ANN.__init__(self, i2h, h2o, bias)
+        num_features, num_hidden, weights = nn_weights['1']
+        ANN.__init__(self, weights, num_features, num_hidden)
+
+
+class EvolvedANN5(ANN):
+    """
+    A strategy based on a pre-trained neural network.
```
drvinceknight: Perhaps add a line/comment describing the number of features/layers? (Same comment for the other ones).

axelrod/strategies/finite_state_machines.py
```diff
@@ -351,3 +351,138 @@ def __init__(self):
         )
 
         FSMPlayer.__init__(self, transitions, initial_state=1, initial_action=C)
+
+
+class EvolvedFSM4(FSMPlayer):
```
drvinceknight: Could these all have a docstring please (names etc)

axelrod/strategies/gambler.py
```diff
@@ -1,75 +1,95 @@
...
 C, D = Actions.C, Actions.D
+tables = load_pso_tables("pso_gambler.csv", directory="data")
 
 
-# End with two defections if tournament length is known
-@FinalTransformer((D, D), name_prefix=None)
 class Gambler(LookerUp):
     """
-    A LookerUp class player which will select randomly an action in some cases.
-    It will always defect the last 2 turns.
```
drvinceknight: Cool. Nicer to get rid of this I think.

axelrod/strategies/gambler.py
```diff
@@ -1,75 +1,95 @@
...
+        # action could be 'C', 'D', or a float
         if action in [C, D]:
             return action
         return random_choice(action)
 
 
-class PSOGambler(Gambler):
+class PSOGamblerMem1(Gambler):
+    """
+    A 1x1x0 PSOGambler trained with pyswarm.
```
drvinceknight: "This corresponds to a fully trained memory 1 strategy."

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,149 @@
+from numpy.random import choice
+
+from axelrod import Actions, Player, init_args, random_choice
+
+C, D = Actions.C, Actions.D
+
+
+def is_stochastic_matrix(m, ep=1e-8):
```
drvinceknight: Could we have a test for this.

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,149 @@
...
+        emission_probabilities is a vector of values in [0, 1]
+        initial_state is an element of range(0, len(emission_probabilities))
+        """
+        self.transitions_C = transitions_C
+        self.transitions_D = transitions_D
+        self.emission_probabilities = emission_probabilities
+        self.state = initial_state
+
+    def is_well_formed(self):
+        if not is_stochastic_matrix(self.transitions_C):
```
drvinceknight: Perhaps: `return is_stochastic_matrix(self.transitions_C)` and  `return is_stochastic_matrix(self.transitions_D)`

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,149 @@
...
+        if abs(1. - s) > ep:
+            return False
+    return True
+
+
+class SimpleHMM(object):
+    """Implementation of a basic Hidden Markov Model. We assume that the
+    transition matrix is conditioned on the opponent's last action, so there
+    are two transition matrices. Emission distributions are stored as Bernoulli
+    probabilities for each state. This is essentially a stochastic FSM.
```
drvinceknight:
```
Names

    - SimpleHMM: ...
```

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,149 @@
...
+        else:
+            next_state = choice(num_states, 1, p=self.transitions_D[self.state])
+        self.state = next_state[0]
+        p = self.emission_probabilities[self.state]
+        action = random_choice(p)
+        return action
+
+
+class HMMPlayer(Player):
+    """Abstract base class for Hidden Markov Model players."""
```
drvinceknight: Same comment `names ...`

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,149 @@
...
+            # for the strategy to function
+            self.state = self.hmm.state
+            return action
+
+    def reset(self):
+        Player.reset(self)
+        self.hmm.state = self.initial_state
+
+
+class EvolvedHMM5(HMMPlayer):
```
drvinceknight: ```names```

axelrod/tests/unit/test_ann.py
```diff
@@ -14,7 +14,7 @@ class TestEvolvedANN(TestPlayer):
     expected_classifier = {
         'memory_depth': float('inf'),
         'stochastic': False,
-        'makes_use_of': set(["length"]),
```
drvinceknight: Is the number of turns no longer a feature?

axelrod/tests/unit/test_hmm.py
```diff
@@ -0,0 +1,123 @@
...
+        opponent = axelrod.Alternator()
+        for i in range(6):
+            player.play(opponent)
+        self.assertEqual(opponent.history, [C, D] * 3)
+        self.assertEqual(player.history, [C, C, D, D, C, C])
+
+    def test_malformed_params(self):
+        # Test a malformed table
+        t_C = [[1, 0.5], [0, 1]]
+        self.assertFalse(is_stochastic_matrix(t_C))
```
drvinceknight: Re my request for a test for this function, could it be pulled out of here and tested independently? (With a True assertion as well).

axelrod/strategies/worse_and_worse.py
marcharper: I'm actually not sure about this one so I've changed it back to `float('inf')` for now. The question is whether using the round number counts as using history or not, let's discuss elsewhere.

axelrod/tests/unit/test_hmm.py
marcharper: done

axelrod/tests/unit/test_hmm.py
marcharper: done

docs/reference/bibliography.rst
marcharper: Citation updated

setup.py
marcharper: It does locally, and the tests would fail on travis otherwise.

axelrod/load_data_.py
marcharper: Not sure how to better test these than simply loading the data and testing the strategies. I added a few integrity checks in ANN and LookerUp to make sure the data is of the expected length.

axelrod/strategies/ann.py
marcharper: done

axelrod/strategies/ann.py
marcharper: done

axelrod/strategies/finite_state_machines.py
marcharper: done

axelrod/strategies/gambler.py
marcharper: done

axelrod/strategies/hmm.py
marcharper: done

axelrod/strategies/hmm.py
marcharper: This one isn't a strategy but I updated the HMM Players with names

axelrod/strategies/hmm.py
marcharper: There are multiple checks here so I left them separate. We can change it if you feel strongly about the form.

axelrod/strategies/hmm.py
marcharper: done

axelrod/strategies/hmm.py
marcharper: done

axelrod/tests/unit/test_ann.py
marcharper: Nope! Just the round number.

axelrod/strategies/hmm.py
drvinceknight: No I don't feel strongly. I'm fine with it as you have it.

axelrod/strategies/hmm.py
drvinceknight: Sorry I got over zealous :)

axelrod/tests/unit/test_ann.py
drvinceknight: Cool :)

setup.py
drvinceknight: I don't think travis specifically tests this (just fyi) but appveyor does test the setup install (because we had some windows problems with that at some point...).

axelrod/load_data_.py
drvinceknight: I was thinking that we could have a `tests/unit/test_load_data.py` file that just checks that these functions run and that the data is of the expected format separately.

marcharper: I'm not sure that we'd be testing anything further in that case -- if the format or data types are wrong the strategies will fail when constructed or played.

drvinceknight: I agree that this wouldn't test anything further, it just consolidates things: for example in the future these ml strategies could be changed to no longer read the data (hypothetically), their tests adjusted and an error creeping in to these reader functions. That's a weird case but I think my point holds?

marcharper: But if the players no longer read the data then the data and these functions are unnecessary (and their coverage will disappear). So wouldn't we just delete the data and these functions in that case, unless something else is using them? And if something else is using them then a bad change will still break those things.

drvinceknight: Very good point, I'd still say too many tests is better than too phew but I won't insist. :) :+1:

marcharper: Maybe type annotations are a good check here so when start annotating we'll get a little extra coverage.

axelrod/data/ann_weights.csv
```diff
@@ -0,0 +1,5 @@
+# name, features, hidden_layer_size, weights...
```
drvinceknight: Small suggestion: remove the `#` and list full header (potentially useful for other analysis?). I expect this would need to be done on the `axelrod-evolver` repo and not for this PR.

axelrod/load_data_.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def load_file(filename, directory):
```
drvinceknight: Docstring for completeness. Also numpy style for this and the rest?

axelrod/moran.py
```diff
@@ -30,32 +30,38 @@ def fitness_proportionate_selection(scores):
...
-        method yet emulates the common method in the literature.
+        An agent based Moran process class. In each round, each player plays a
+        Match with each other player. Players are assigned a fitness score by
+        their total score from all matches in the round. A player is chosen to
+        reproduce proportionally to fitness, possibly mutated, and is cloned.
+        The clone replaces a randomly chosen player.
+
+        If the mutation_rate is 0, the population will eventually fixate on
+        exactly one player type. In this case a StopIteration exception is
+        raised and the play stops. If mutation_rate is not zero, then the
```
drvinceknight: `the mutation_rate`

axelrod/strategies/ann.py
```diff
@@ -169,62 +172,40 @@ class EvolvedANN(ANN):
...
-                       -36.177108409134966, -0.3812547430698569, -2959.4921368963633, -1.8635802741029985,
-                       0.020513128847167047, -0.9179124323385958]
-
-        (i2h, h2o, bias) = split_weights(
-            weights,
-            input_values,
-            hidden_layer_size
-        )
-        ANN.__init__(self, i2h, h2o, bias)
+        num_features, num_hidden, weights = nn_weights['1']
```
drvinceknight: Could we change the `1` in the dataset and here to be `10`? This would be for readability and corresponding to the size of the hidden layer.

axelrod/strategies/ann.py
```diff
@@ -169,62 +172,40 @@ class EvolvedANN(ANN):
...
+
+
+class EvolvedANN5(ANN):
+    """
+    A strategy based on a pre-trained neural network with 17 features and a
+    hidden layer of size 5.
+
+    Names:
+
+     - EvolvedANN5: : Original name by Marc Harper.
```
drvinceknight: Extra `:`

axelrod/strategies/ann.py
```diff
@@ -158,7 +160,8 @@ def strategy(self, opponent):
 
 class EvolvedANN(ANN):
     """
-    A strategy based on a pre-trained neural network.
+    A strategy based on a pre-trained neural network with 17 features and a
```
drvinceknight: Would it be worth including a bullet point list of the 17 features here?

axelrod/strategies/ann.py
```diff
@@ -169,62 +172,40 @@ class EvolvedANN(ANN):
...
+
+    @init_args
+    def __init__(self):
+        num_features, num_hidden, weights = nn_weights['5']
+        ANN.__init__(self, weights, num_features, num_hidden)
+
+
+class EvolvedANNNoise05(ANN):
+    """
+    A strategy based on a pre-trained neural network, trained with noise=0.05.
```
drvinceknight: How many layers?

axelrod/strategies/ann.py
```diff
@@ -169,62 +172,40 @@ class EvolvedANN(ANN):
...
+    Names:
+
+     - EvolvedANNNoise05: Original name by Marc Harper.
+    """
+
+    name = "EvolvedANNNoise05"
+
+    @init_args
+    def __init__(self):
+        num_features, num_hidden, weights = nn_weights['05']
```
drvinceknight: My comment about the readability of the key (in the dataset) holds here too. But perhaps make it even more verbose than a single number. Could we perhaps have `10layers-noise-05` for here? And `10layers` for above? (Or similar).

axelrod/strategies/finite_state_machines.py
drvinceknight: Could you include the usual convention:

```
names:

    - EvolvedFSM4: Original name....
```

drvinceknight: (Same for the rest in this file/PR)

axelrod/tests/unit/test_ann.py
```diff
@@ -29,15 +39,15 @@ def test_strategy(self):
...
     def test_rounds(self):
         self.versus_test(axelrod.EvolvedANN(), axelrod.Defector(),
-                         [C, D, D, D, D], [D] * 5)
+                         [C, C, D, D, D], [D] * 5)
 
 class TestEvolvedANNvsTFT(TestHeadsUp):
     def test_rounds(self):
         self.versus_test(axelrod.EvolvedANN(), axelrod.TitForTat(),
-                         [C, D, D, C, C], [C, C, D, D, C] * 5)
+                         [C] * 5, [C] * 5)
```
drvinceknight: Could we have some specific tests for the other two trained ANN strategies please.

Also: can you include a test for the exception that gets raised when the weights in the dataset have incorrect size.

axelrod/strategies/finite_state_machines.py
```diff
@@ -351,3 +351,142 @@ def __init__(self):
...
+            (2, D, 1, C),
+            (3, C, 3, D),
+            (3, D, 1, D)
+        )
+
+        FSMPlayer.__init__(self, transitions, initial_state=0, initial_action=C)
+
+
+class EvolvedFSM16(FSMPlayer):
+    """A 16 state FSM player trained with an evolutionary algorithm."""
```
drvinceknight: `names...`

axelrod/strategies/finite_state_machines.py
```diff
@@ -351,3 +351,142 @@ def __init__(self):
...
+            (14, C, 13, D),
+            (14, D, 13, D),
+            (15, C, 15, D),
+            (15, D, 2, C)
+        )
+
+        FSMPlayer.__init__(self, transitions, initial_state=0, initial_action=C)
+
+
+class EvolvedFSM16Noise05(FSMPlayer):
```
drvinceknight: `names...`

axelrod/strategies/finite_state_machines.py
```diff
@@ -351,3 +351,142 @@ def __init__(self):
         )
 
         FSMPlayer.__init__(self, transitions, initial_state=1, initial_action=C)
+
```
drvinceknight: Could you add some specific tests for each of these strategies please.

axelrod/strategies/gambler.py
```diff
@@ -1,75 +1,100 @@
...
-        # action could be 'C' or a float
+        # action could be 'C', 'D', or a float
         if action in [C, D]:
             return action
         return random_choice(action)
 
 
-class PSOGambler(Gambler):
+class PSOGamblerMem1(Gambler):
+    """
```
drvinceknight: `names...`

axelrod/strategies/gambler.py
```diff
@@ -1,75 +1,100 @@
...
         'stochastic': True,
-        'makes_use_of': set(['length']),
+        'makes_use_of': set(),
         'long_run_time': False,
         'inspects_source': False,
         'manipulates_source': False,
         'manipulates_state': False
     }
 
-    @init_args
```
drvinceknight: I don't quite follow why we can get rid of this.

axelrod/strategies/gambler.py
```diff
@@ -1,75 +1,100 @@
...
+    def __init__(self):
+        pattern = tables[('2', 2, 2, 2)]
+        lookup_table = create_lookup_table_from_pattern(
+            plays=2, op_plays=2, op_start_plays=2,
+            pattern=pattern)
+        Gambler.__init__(self, lookup_table=lookup_table)
+
+
+class PSOGambler2_2_2_Noise05(Gambler):
+    """
```
drvinceknight: docstring (include `names...`).

axelrod/strategies/gambler.py
```diff
@@ -1,75 +1,100 @@
...
+        pattern = tables[('2', 2, 2, 2)]
+        lookup_table = create_lookup_table_from_pattern(
+            plays=2, op_plays=2, op_start_plays=2,
+            pattern=pattern)
+        Gambler.__init__(self, lookup_table=lookup_table)
+
+
+class PSOGambler2_2_2(Gambler):
+    """
+    A 2x2x2 PSOGambler trained with pyswarm.
```
drvinceknight: `names`

axelrod/strategies/gambler.py
```diff
@@ -1,75 +1,100 @@
...
+        lookup_table = create_lookup_table_from_pattern(
+            plays=1, op_plays=1, op_start_plays=0,
+            pattern=pattern)
+        Gambler.__init__(self, lookup_table=lookup_table)
+        self.classifier['memory_depth'] = 1
+
+
+class PSOGambler1_1_1(Gambler):
+    """
+    A 1x1x1 PSOGambler trained with pyswarm.
```
drvinceknight: `names`

axelrod/strategies/gambler.py
drvinceknight: Is it just because this is a simple ancestor class? Would it not be potentially useful to keep around though? (I'm sure I'm just missing something).

axelrod/tests/unit/test_gambler.py
```diff
@@ -90,39 +94,38 @@ class TestPSOGambler(TestPlayer):
...
 
 
-class PSOGamblervsAlternator(TestHeadsUp):
+class PSOGambler2_2_2vsAlternator(TestHeadsUp):
     def test_vs(self):
-        self.versus_test(axelrod.PSOGambler(), axelrod.Alternator(),
-                         [C, C, D, D, D, D], [C, D, C, D, C, D])
+        axelrod.seed(10)
+        self.versus_test(axelrod.PSOGambler2_2_2(), axelrod.Alternator(),
+                         [C, C, C, C, C, C, C], [C, D, C, D, C, D, C])
```
drvinceknight: Could we also have tests for the other new gamblers please (specific plays and potentially checks for structure of underlying table):

- `Mem1`
- `1_1_1`
- `2_2_2`
- `2_2_2_noise05`

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,162 @@
...
+            return False
+        for p in self.emission_probabilities:
+            if (p < 0) or (p > 1):
+                return False
+        if self.state not in range(0, len(self.emission_probabilities)):
+            return False
+        return True
+
+    def move(self, opponent_action):
+        """Changes state and computes the response action."""
```
drvinceknight: numpy style docstrings needed here (and above).

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,162 @@
...
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    @init_args
+    def __init__(self, transitions_C=None, transitions_D=None,
+                 emission_probabilities=None, initial_state=0,
+                 initial_action=C):
```
drvinceknight: docstrings

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,162 @@
...
+        self.classifier['stochastic'] = self.is_stochastic()
+
+    def is_stochastic(self):
+        # If the transitions matrices and emission_probabilities are all 0 or 1
+        # Then the player is stochastic
+        values = set(self.hmm.emission_probabilities)
+        for m in [self.hmm.transitions_C, self.hmm.transitions_D]:
+            for row in m:
+                values.update(row)
+        if not values.issubset({0, 1}):
```
drvinceknight: `return not values.issubset({0, 1})`.

(This is stylistic, I don't feel strongly about it.)

axelrod/strategies/hmm.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class EvolvedHMM5(HMMPlayer):
+    """
+    An HMM-based player with five hidden states trained with an evolutionary
+    algorithm.
+
+    Names
+
+        - EvolvedHMM5
```
drvinceknight: `: Original name...`

axelrod/tests/unit/test_hmm.py
```diff
@@ -0,0 +1,135 @@
...
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_reset(self):
+        player = self.player([[1]], [[1]], [0], initial_state=0)
+        player.hmm.state = -1
+        player.reset()
+        self.assertFalse(player.hmm.state == -1)
```
drvinceknight: Could we have a specific test for the `EvolvedHMM5`.

drvinceknight: I think `assertNotEqual` would be better here.

And/or even `assertEqual` with 0.

axelrod/strategies/meta.py
```diff
@@ -19,7 +19,7 @@ class MetaPlayer(Player):
     classifier = {
         'memory_depth': float('inf'),  # Long memory
         'stochastic': True,
-        'makes_use_of': set(),
+        'makes_use_of': {'game', 'length'},
```
drvinceknight: Could you walk me through this one please?

axelrod/load_data_.py
meatballs: Wouldn't these be better as a set of pandas dataframes? There would be far less code and it would be quicker too.

I know it's another dependency, but we're already dependent on numpy, so the precedent has been set.

axelrod/strategies/gambler.py
marcharper: We can keep it just in case it's needed in the future. Currently we never invoke the base class explicitly so it's not needed.

axelrod/load_data_.py
marcharper: Yep! I prefer using pandas actually if the extra dependency is ok with @drvinceknight .

axelrod/strategies/gambler.py
drvinceknight: Cool. I don't feel strongly but my hunch would be to keep it in (and add a test for it but that can be an issue for another PR).

axelrod/load_data_.py
drvinceknight: I'm not averse to the extra dependency. :+1: 

Could change the output of the `results_set.summarize` to be a data frame too (not for this PR, another issue :)).

marcharper: Cool. I think anyone that uses anaconda or can pip install numpy should have access to rest of the scientific stack (for sure at least scipy and pandas).

axelrod/strategies/gambler.py
marcharper: Actually there's no `__init__` method for gambler at all any more so it's unnecessary.

axelrod/strategies/meta.py
marcharper: The default player set has members that use both the game and the match length.

marcharper:
I believe that I addressed all the comments sufficiently, let me know if you agree!

axelrod/load_data_.py
drvinceknight: Fine by me.

I'm not entirely sure use dfs would make things simpler in this case, the data as is would need to be pivoted for the df to be advantageous or the data could be stored with rows corresponding to "genes" and columns to different strategies... Perhaps not a bad idea (but I don't think necessary for this PR).

axelrod/data/ann_weights.csv
drvinceknight: Fine to leave for later, you could have the max number of columns with headers and have NANs in the other ones? (Not suggesting that for now).

Issue #806, opened 2017-01-04
History class


marcharper:
I added a preliminary implementation of a history class [here](https://github.com/Axelrod-Python/Axelrod/blob/history_class/axelrod/history.py). I'm curious for feedback @meatballs @drvinceknight before I progress any further. The idea is that by abstracting the history into a class we can:

* compute features in the history class rather than having strategies duplicate it
* enforce some boundaries between the player class and the history (like having the history store its cooperation counts rather than as attributes in the player class)
* opens up experimentation for the history to be implemented as a string, bitvector, integer, etc. for computational and memory efficiency

I'm also thinking about a joint history class (storing the match history) so that the state distribution and other features can be encapsulated, scoring could act on the joint history, etc. We could make some of the strategy methods into staticmethods if each player has a copy of the joint history.

drvinceknight:
Sounds good to me.

marcharper:
I'm going to close this one. If anyone wants to attempt this in the future feel free to reopen. My implementations had a big performance hit for e.g. copying shared histories in the metaplayers.

Issue #726, opened 2016-09-26
Need to improve documentation for strategy test functions


drvinceknight:
I am not the biggest fan of our strategy test functions (`test_responses` etc...). I feel they make the tests too implicit and often find myself rewriting things explicitly to try and find errors etc... (For example see: https://github.com/Huaraz2/Axelrod/pull/1/commits/4b62612d692c82cf6259bc21bb946f5d27ade4cf).

One person from the PyConUK sprints mentioned that they found them hard to use and that they could be better documented. 

Could we improve the documentation: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html Perhaps including more "this is equivalent to ..." and/or say explicitly that you do not need to use them? 

On a similar vain is `markov_test` a good name? It implies stochasticity to me but it's just a referral to the states of the common markov chain right?

marcharper:
`test_markov` isn't a good name (that's on me). It's really just a test of how the player behaves in the four possible first contexts CC, CD, DC, and DD, and is the logical follow up to the first play test. Something like `second_plays_test` is a better name.

I think we could definitely improve the docs for using the `attrs` parameter for `responses_test` -- it's powerful and should be used more.

drvinceknight:
> Something like second_plays_test is a better name.
> 
> I think we could definitely improve the docs for using the attrs parameter for responses_test -- it's powerful and should be used more.

Sounds good to me.

drvinceknight:
This was closed by #820

Issue #706, opened 2016-09-12
init_args as dict


meatballs:
In the API, I think the strategies endpoint should indicate whether a strategy class takes init arguments and, if so, what those arguments should be.

In order to do so we'd need to re-work the way in which we currently handle init_args:
- A strategy class would have a dictionary whose keys give the names of the arguments (and, possibly, default values to be used if the arguments are omitted).
- On instantiation, the **init** method for the class would be passed keyword args (instead of args at present), which we'd need to compare against the class level dict.

drvinceknight:
Would inspecting the init method of the class work? Using something like https://docs.python.org/2/library/inspect.html: which lets you "[...] extract and format the argument list for a function, or [...]"

meatballs:
No - that's only going to work for instantiated objects. I want to get something from the class itself, so the API effectively says "here are all the strategy classes available to you and, for each one, everything you need to know in order to instantiate it"

meatballs:
That would then allow a UI to present a list of strategies with enough information for a tournament to be defined and passed back to the API for execution.

meatballs:
e.g. https://axelrod-api.herokuapp.com/strategies/random/

currently returns:
```
{
    "url": "https://axelrod-api.herokuapp.com/strategies/random/",
    "id": "random",
    "name": "Random",
    "description": "A player who randomly chooses between cooperating and defecting.",
    "classifier": {
        "makes_use_of": [],
        "manipulates_source": false,
        "stochastic": true,
        "manipulates_state": false,
        "long_run_time": false,
        "memory_depth": 0,
        "inspects_source": false
    }
```

which gives no indication that a probability should be passed. I'm suggesting that this should instead be:

```
{
    "url": "https://axelrod-api.herokuapp.com/strategies/random/",
    "id": "random",
    "name": "Random",
    "description": "A player who randomly chooses between cooperating and defecting.",
    "classifier": {
        "makes_use_of": [],
        "manipulates_source": false,
        "stochastic": true,
        "manipulates_state": false,
        "long_run_time": false,
        "memory_depth": 0,
        "inspects_source": false
    },
   "init_args": {
        "probability": 0
    }
```

meatballs:
Possibly, the value for each dict entry should be a tuple of default value and a text description.

meatballs:
or even a further dict containing:

* default
* description
* mandatory?

drvinceknight:
I understand the use case.

[...]

PR #807, opened 2017-01-08
History class

marcharper:
I've implemented two versions of a history class, with the underlying data structure being a list or a string. Both have similar performance, and there is definitely some overhead for using a history class rather than just a list -- the tests take a bit longer (about 25%) due to (1) more copying of lists and (2) overhead from additional function calls. I think the impact on tournaments will be less (testing now and will update).

Nevertheless there are advantages (as I mentioned at #806):
* We can treat history as a list or a string in all cases, e.g as `[C, C, D]`, "CCD", and even `C + C + D` and `(C * 2 + D * 3) * 2` for `[C, C, D, D, D, C, C, D, D, D]`
* The latter has some advantages since `"C" == C` and also acts as a container like `['C']`, so you can do thinks like `history[-1:] == C` and `history[-2:] == C + C` rather than `history[-1:] == [C]` and `history[-2:] == [C, C]` 
* No risk of failing to properly copy a history list, we can overload various methods (like the comparison to a list, str, etc.)
* We can add features and other boolean tests to the history class beyond what we have not with just cooperations and defections

I rewrote many tests to use the additive notation instead and made sure all the tests work with either the list or string implementations. Whether a bit vector or similar is ultimately better is an experiment for the future. However this notation is great because we can just refer to `C` and `D` universally, and if we decide to use `enum` or some other definition of the actions we should be able to handle it more easily.

Also in this PR (which could be separated):
* `markov_test([C, D, D, C])` is now `second_play_test(C, D, D, C)`. See #726
* `responses_test` got some upgrades -- you can pass `init_args` and `init_kwargs` so there's no need to construct players manually and use `test_responses`
* I changed the argument order to be `responses_test(responses, history1=None, history2=None, ...)` so you can test the first X actions with e.g. `responses_test(C + D + D + ...)`
* Some bugs around testing and I fixed some tests that were not actually testing anything or were not quite correct. For example `responses_test` was incorrect if `random_seed was 0`. In general we should avoid manually setting the history of a player for most testing purposes.
* Lots of PEP8 violations and typos were fixed, some tests consolidated
* Some Python 2 stuff (but I didn't go out of my way to look for specifics)
* `@init_args` now saves `kwargs` as well, which hopefully helps with #706 

What this PR still needs:
* documentation of the history class
* `JointHistory` is still WIP and can be moved out

If you are on board with these changes then I'll write up some docs and consolidate the remaining tests.

### DIFF
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -11,6 +11,7 @@ from .actions import Actions, flip_action
 from .random_ import random_choice, seed
 from .plot import Plot
 from .game import DefaultGame, Game
+from .history import History
 from .player import (
     get_state_distribution_from_history, init_args, is_basic, obey_axelrod,
     update_history, update_state_distribution, Player)
--- a/axelrod/_strategy_utils.py
+++ b/axelrod/_strategy_utils.py
@@ -10,7 +10,7 @@ from axelrod.strategies.cycler import Cycler
 C, D = Actions.C, Actions.D
 
 
-def detect_cycle(history, min_size=1, offset=0):
+def detect_cycle(history, min_size=1, max_size=12, offset=0):
     """Detects cycles in the sequence history.
 
     Mainly used by hunter strategies.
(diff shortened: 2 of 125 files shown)


drvinceknight:
> If you are on board with these changes then I'll write up some docs and consolidate the remaining tests.

I'm on board :)


For the documentation, I think this should go under the "Further capabilities" of the library section (http://axelrod.readthedocs.io/en/latest/tutorials/advanced/index.html) and could you look carefully through the contribution section in case there's anything there that might not be doctested that needs to be updated?

As far as the slow down, I'm not worried about the tests but hopefully it doesn't impact the tournaments *too* much. The gain in readability is probably worth it though. :+1: 

Could you also include some of the improvements to the helper test functions in the docs (I've seen you've made some of the changes already).

I'm probably going to have a couple of questions (30 hours of travelling and poor wifi are making it difficult to look through carefully right now) but this looks great :)

drvinceknight:
> markov_test([C, D, D, C]) is now second_play_test(C, D, D, C). See #726

How about `after_turn_test`?

marcharper:
I'm not attached to the function name but I think second player is appropriate since that what it literally is -- there are four possible first rounds and this is what the player does next.

drvinceknight:
Sure. :+1:

marcharper:
I'm going to close, I'll reopen with a new branch after rebasing and making some modifications.

PR #809, opened 2017-01-09
Update cycle detection to be more efficient

marcharper:
Profiling indicates that `detect_cycles` is one most expensive functions in the library, so I rewrote it and the players that use it to be more efficient (without altering their behaviors significantly).

This PR decreases tournament run time by approx 20%.

### DIFF
--- a/axelrod/strategies/hunter.py
+++ b/axelrod/strategies/hunter.py
@@ -44,6 +44,13 @@ class CooperatorHunter(Player):
         return C
 
 
+def is_alternator(history):
+    for i in range(len(history) - 1):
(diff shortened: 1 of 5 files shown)


axelrod/strategies/hunter.py
```diff
@@ -44,6 +44,13 @@ def strategy(self, opponent):
         return C
 
 
+def is_alternator(history):
+    for i in range(len(history) - 1):
+        if history[i] == history[i+1]:
```
drvinceknight: PEP8 `i + 1`

PR #814, opened 2017-01-13
Add type hints to eigen.py

juliahou8:
Hi! I'm new to open source projects, and I thought this project would be a great place to get started. I'd love any feedback/hints/corrections/advice, especially as I'm not entirely sure I did this correctly.

### DIFF
--- a/axelrod/eigen.py
+++ b/axelrod/eigen.py
@@ -6,11 +6,7 @@ eigenvectors.
 """
 
 import numpy
-from typing import NamedTuple
-
-
-Eigenvector = NamedTuple('Eigenvector',
-			[('vector', numpy.ndarray), ('eigenvalue', float)])
+from typing import Tuple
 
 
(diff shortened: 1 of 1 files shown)


marcharper:
Hi @juliahou8 ! Thanks for the contribution. I have one comment: rather than a namedtuple type I think it would be simpler to just use a [tuple type](https://docs.python.org/3/library/typing.html#typing.Tuple)  (also from the `typing` module) instead of the `Eigenvector` namedtuple type:

```
Tuple[numpy.ndarray, float]
```

This is because we're not actually using a namedtuple (which comes from `collections`).

Otherwise it looks good to me!

marcharper:
Thanks! We have a two review policy for PRs so we'll need a :+1: from one of the other core devs. Usually this is pretty quick (but one is travelling so it may be a little longer than usual).

PR #815, opened 2017-01-13
Add type hints for action script

souravsingh:


axelrod/actions.py
```diff
@@ -1,10 +1,14 @@
+from typing import NewType
+
+C = NewType('Action', str)
```
marcharper: I think you probably want `Action = NewType('Action', str)` and to declare C and D as actions. Unfortunately it's not easy to do so without the Py3.6 additions to type hints. But I think this would still work for `flip_action`.

marcharper:
@souravsingh This is a tough one actually -- I'm not sure of the best way to annotate this file since we'd like to use the Py3.6 method that I linked in the issue, and Action should probably be an enum now. 

By the way, you can run the tests with `./test` in the axelrod directory to see the output locally before you push the PR, it's a bit faster than waiting on travis.

souravsingh:
@marcharper I have made some fixes. Is the Patch ready for merging?

axelrod/actions.py
```diff
@@ -1,10 +1,14 @@
+from typing import NewType
+
+Action = NewType('Action', str)
+Action = NewType('Action', str)
```
marcharper: Let's remove the duplicate line.

marcharper:
I think so, let's wait for another review and squash the commits on merge.

PR #817, opened 2017-01-15
Refactor Player.__new__ to automatically apply init_args

marcharper:
I've eliminated the need for `@init_args` by modifying `Player.__new__` and eliminated most usage of `init_args` in the library. `Player.__new__` now caches the `args` and `kwargs` passed to `__init__`. There's no need to explicitly use the decorator except for one case in `LookerUp` where the classes are created dynamically at runtime, but it may be possible to eliminate that as well.

@meatballs This may solve your issue on #706, and if not, it should be easier to fix now.

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -77,6 +77,7 @@ def init_args(func):
     def wrapper(self, *args, **kwargs):
         r = func(self, *args, **kwargs)
         self.init_args = args
+        self.init_kwargs = kwargs
         return r
     return wrapper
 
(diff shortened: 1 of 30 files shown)


drvinceknight:
Can you modify the contribution docs to reflect this please: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_the_new_strategy.html

(There's a paragraph at the end about adding the `init_args` decorator.)

marcharper:
I did change the docs :) I'll rebase.

drvinceknight:
My mistake! I did a quick scan and thought it wasn't there. Will review
properly when I get a chance :)

meatballs:
Nice! 👍

Issue #801, opened 2017-01-01
Sweep of Python 2 compatibility features


drvinceknight:
There are various things in the library that can now be simplified as we no longer support py 2.

For example various computations had to be explicitly done as float division. These can be removed now :)

rkty13:
Hi! I'm pretty new to open source projects even though I've been programming for a while now. I'm interested in contributing to this project and I thought this issue might be a good place to start. Do you have any suggestions for how I should tackle this?

drvinceknight:
> Hi! I'm pretty new to open source projects even though I've been programming for a while now. I'm interested in contributing to this project and I thought this issue might be a good place to start. Do you have any suggestions for how I should tackle this?

Hi @rkty13! Welcome!

I think this issue would be a good place to start indeed :)

To give a bit of background (and apologies if this is all known), this library used to support python 2 but now only supports python 3. In the past, to ensure compatibility we had to do certain little things that aren't needed anymore. For example, in python 2, division was always integer division so `5/2` would give `2`, to ensure that we always got floats when we wanted to we had a lot of things like `float(5) / 2` which gives `2.5` in both python 2 and 3.

So this issue is to refactor/clean up code to get rid of all the extra things we were doing to ensure the library worked fine with python 2. The idea is to simply go through the source code and remove them. For example in `axelrod/result_set.py` you can see this line <https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/result_set.py#L357>:

```
                counters.append(Counter({key: float(value) / total for
                                         key, value in counter.items()}))
```

That can be replaced with:

```
                counters.append(Counter({key: value / total for
                                         key, value in counter.items()}))
```

So you could simply remove that and all the tests should still pass (there's information on running the tests here: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/running_tests.html).

That's just one example where I'm talking about float division but there will be other little things here and there (including in the strategy files: `axelrod/strategies/*py`).

I hope that's helpful, but don't hesitate to ask if not. You can usually get one of us on the gitter channel if you want to have a quick chat: https://gitter.im/Axelrod-Python/Axelrod We take pride in being helpful to newcomers (to programmer and/or open source) so please let us know how we can help :)

The full documentation might also be helpful: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/index.html

rkty13:
[...]

PR #818, opened 2017-01-15
Dropping support for Python 2

rkty13:
This PR addresses [#801](https://github.com/Axelrod-Python/Axelrod/issues/801). I have gone through all of the files and removed any support for Python 2 that I could find.

Overview of the changes I made:

- Changed all floating point divisions that used a `float()` cast or had one of the operands as a float (e.g. `1.0/...` or `1./...`.
- Unicode strings are no longer prefixed as such: `u''`
- Changed all references to superclasses to use the preferred zero-argument format `super()`.
- Removed all imports from the `__future__` module.
- Removed some import logic to check if Python's version was 2 or 3.

Some things I could not change:

- Dropping the use of `random_choice()` and `randrange()` in [https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/random_.py](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/random_.py) as noted in [#801](https://github.com/Axelrod-Python/Axelrod/issues/801#issuecomment-272677267).
- Due to Multiple Inheritance, it is preferred to use the old-style superclass/direct reference in [https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/match_generator.py#L263](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/match_generator.py#L263).

Overall I think I have caught most of the differences between Python 2 and Python 3. Please let me know if I have missed anything or if I need to add anything else to this PR.

### DIFF
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -1,4 +1,3 @@
-from __future__ import absolute_import
 import os
 
 on_windows = os.name == 'nt'
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -1,9 +1,4 @@
-try:
-    # Python 2.x
-    from UserDict import UserDict
-except ImportError:
-    # Python 3.x
-    from collections import UserDict
+from collections import UserDict
 import pickle
 
 from axelrod import Player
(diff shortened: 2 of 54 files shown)


marcharper:
@rkty13 Can you also update the issue with what is left to do? Even if it's just the last few lines above.

Thanks!

rkty13:
Done!

Issue #808, opened 2017-01-08
Type Hints and Annotations


marcharper:
Now that we're Python 3.5+ exclusive, it's time for [type hints](https://docs.python.org/3/library/typing.html)! 

Note to contributors: please run mypy on any newly annotated files. After installing with `pip install mypy`, please run:

`mypy --ignore-missing-imports --follow-imports skip axelrod/<file.py>`

for whichever files you annotate. You should receive no errors. We'll get this added to the automated checks soon. Thanks!

Finally, note that we're officially supporting Python 3.5 and 3.6, so please don't use any of the newer typing features in Python 3.6 yet.

souravsingh:
I am interested in working on the issue.

marcharper:
@souravsingh Great! It's a huge undertaking so feel free to tackle it in chunks, maybe file by file, and open multiple PRs if you'd like.

A few notes:
* Start with the easy cases, things that are clearly ints or floats, things that are strings and unlikely to be anything else (like player names)
* Go inside -> out, meaning start with the basic things like actions and players rather than the derived things like actual strategies, to keep from rewriting when you change your mind
* Consider using type variables or `NewType` for things like `axelrod.Actions.C` (rather than say `str`) and then import that type as needed. This is so if we decide to e.g. change the actions to 0 and 1 rather than `'C'` and `'D'` then we'll get the rest of the types for free. E.g. the history would be a list of whatever type the action is, and `List[T]` where `T` is the type variable for action is better than `List[str]` as the type for history.

We're in uncharted territory here so if you are not sure how to annotate something, either leave it alone or open an issue and we'll discuss it.

marcharper:
@meatballs @drvinceknight Have you seen [PEP 526](https://www.python.org/dev/peps/pep-0526/)? It's Python 3.6 only and relevant to this issue. Unfortunately Python 3.7 won't be released for 18 months...

drvinceknight:
Just read it :)

marcharper:
@drvinceknight we should look into running a type checker (with travis?) once we have a some annotations in.

drvinceknight:
Yeah: great idea! Do you know of a type checker for Python? A quick google has lead me to give up as the wifi is a bit slow in my hotel right now...

souravsingh:
@drvinceknight @marcharper We can make use of [mypy](http://www.mypy-lang.org/) for type checks. Guido likes the tool himself.

marcharper:
[...]

PR #821, opened 2017-01-16
Add type hints to random

souravsingh:
Fixes a part of #808

### DIFF
--- a/axelrod/random_.py
+++ b/axelrod/random_.py
@@ -3,7 +3,7 @@ import numpy
 from axelrod import Actions
 
 
-def random_choice(p=0.5):
+def random_choice(p=0.5: float):
     """
     Return 'C' with probability `p`, else return 'D'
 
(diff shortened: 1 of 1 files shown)


axelrod/random_.py
```diff
@@ -3,7 +3,7 @@
 from axelrod import Actions
 
 
-def random_choice(p=0.5):
+def random_choice(p=0.5: float):
```
Nikoleta-v3: Hello @souravsingh I am a member of the Axelrod team :smile:!
I believe you should drop `:float` cause `p` is determined. That should be the syntax error

marcharper: I think the syntax parser wants something like `(p: float = 0.5)`

souravsingh: @Nikoleta-v3 Nice to meet you! 
I will look into the review.

@marcharper What should be the return type set? Should it be imported as Action.

marcharper: Yeah importing the type from `actions.py` is the way to go.

souravsingh:
The PR is now ready for merging.

PR #824, opened 2017-01-17
Add Type hints to Game

### LINKED ISSUE #808: Type Hints and Annotations

souravsingh:
Fixes a part of #808

### DIFF
--- a/axelrod/game.py
+++ b/axelrod/game.py
@@ -1,7 +1,6 @@
 from axelrod import Actions
 from typing import Tuple
 
-pair = Tuple[int, int]
 C, D = Actions.C, Actions.D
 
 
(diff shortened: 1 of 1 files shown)


axelrod/game.py
```diff
@@ -23,7 +25,7 @@ def RPST(self):
         T = self.scores[(D, C)][0]
         return (R, P, S, T)
 
-    def score(self, pair):
+    def score(self, pair) -> pair:
```
drvinceknight: I think it would be more explicit to have:

```
def score(self, pair) -> Tuple[int, int]
```

PR #828, opened 2017-01-23
Add Type Hints in Deterministic Cache

souravsingh:
Fixes a Part of #808

### DIFF
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -1,5 +1,6 @@
 from collections import UserDict
 import pickle
+from typing import Tuple
 
 from axelrod import Player
 
@@ -42,7 +43,7 @@ class DeterministicCache(UserDict):
             self.load(file_name)
 
     @staticmethod
-    def _key_transform(key):
+    def _key_transform(key: Tuple[str, str, int]):
         """
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


souravsingh:
@drvinceknight @marcharper Is there a type for the variable ```key```?

It is mentioned as Tuple, but I am not sure about the definition.

marcharper:
Take a look at the `is_valid_key` function -- a lot of it is type checking.

axelrod/deterministic_cache.py
```diff
@@ -78,7 +79,7 @@ def __setitem__(self, key, value):
         super().__setitem__(self._key_transform(key), value)
 
     @staticmethod
-    def _is_valid_key(key):
+    def _is_valid_key(key) -> boolean:
```
Nikoleta-v3: `bool` instead of `boolean`

axelrod/deterministic_cache.py
```diff
@@ -145,7 +146,7 @@ def save(self, file_name: str):
             pickle.dump(self.data, io)
         return True
 
-    def load(self, file_name: str):
+    def load(self, file_name: str) -> boolean:
```
Nikoleta-v3: same comment

axelrod/deterministic_cache.py
```diff
@@ -145,7 +146,7 @@ def save(self, file_name):
             pickle.dump(self.data, io)
         return True
 
-    def load(self, file_name):
+    def load(self, file_name: str) -> boolean:
```
meatballs: This needs to `bool` rather than `boolean`

souravsingh: Done

PR #835, opened 2017-01-30
Add type hints to adaptive

janga1997:
#808 
Please review @drvinceknight @marcharper . I'm half certain I made a mistake.


marcharper:
Just one error: `axelrod/strategies/adaptive.py:30: error: The return type of "__init__" must be None`.

All you need to do is add `-> None` to `__init__` as the return type.

PR #836, opened 2017-01-30
Add type hints to alternator

janga1997:
#808 

@drvinceknight  @meatballs. Please review.

### DIFF
--- a/axelrod/strategies/alternator.py
+++ b/axelrod/strategies/alternator.py
@@ -1,4 +1,5 @@
 from axelrod import Actions, Player
+import axelrod
 
 C, D = Actions.C, Actions.D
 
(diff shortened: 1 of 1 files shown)


marcharper:
I wonder if it's enough to type hint the parent class (`Player`) `strategy` method, or if overridding the method ignores the base class type hints. Perhaps this depends on the type checker?

janga1997:
@marcharper that is what I thought too. Seems like the issue we face for `ResultSet` and `ResultSetFromFile`?

marcharper:
I think testing for `ResultSet` is better in this case.

axelrod/strategies/alternator.py
```diff
@@ -17,7 +18,7 @@ class Alternator(Player):
         'manipulates_state': False
     }
 
-    def strategy(self, opponent):
+    def strategy(self, opponent: axelrod.player.Player) -> axelrod.actions.Action:
```
marcharper: Just `Player` is fine here since you imported it; you can import `Action` from `axelrod.actions` and then there's no need to import axelrod in line 2.

PR #837, opened 2017-01-31
Fix Type Annotations and add type test shell script

marcharper:
Some of the type annotations already merged are not correct. Please be sure to run mypy on any new files with type annotations (see the included shell script). You will need to `pip install mypy`.

I was not able to get `mypy -m axelrod` to work -- it doesn't like our dynamically generated classes and I didn't track down a fix yet. (I've asked the mypy authors for advice in their gitter room.)

### DIFF
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -1,6 +1,6 @@
-from typing import NewType
 
-Action = NewType('Action', str)
+# Type alias for actions.
+Action = str
 
 
 class Actions(object):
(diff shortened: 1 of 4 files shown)


souravsingh:
@marcharper Thanks for fixing the type annotations. Would it be possible to add an entry to documentation on testing the types annotations using ``mypy``?

marcharper:
@souravsingh yes we'll add it to the docs and to travis so that mypy runs automatically for all PRs.

PR #846, opened 2017-02-02
Add type hints to apavlov

janga1997:
#808 @drvinceknight @marcharper 
mypy produces two errors. 
`axelrod/strategies/apavlov.py:31: error: Need type annotation for variable
axelrod/strategies/apavlov.py:98: error: Need type annotation for variable
`
It seems like an [issue](https://github.com/python/mypy/issues/1164) with mypy itself. I'm not sure.

And the local test produces errors which I don't think has anything to do with type hinting.
`
/home/janga/Axelrod/axelrod/mock_player.py:61: UserWarning: Simulated play mismatch with expected history: Round was (C, C) but (C, D) was expected for player: Adaptive Pavlov 2006
  s1, s2, action1, action2, str(player1))
`
and more of the same error.


drvinceknight:
> mypy produces two errors.
> axelrod/strategies/apavlov.py:31: error: Need type annotation for variable axelrod/strategies/apavlov.py:98: error: Need type annotation for variable
> It seems like an issue with mypy itself. I'm not sure.

It does look like a mypy issue. I'm not sure how to deal with it. Any suggestion @marcharper?

> 
> And the local test produces errors which I don't think has anything to do with type hinting.
> /home/janga/Axelrod/axelrod/mock_player.py:61: UserWarning: Simulated play mismatch with expected history: Round was (C, C) but (C, D) was expected for player: Adaptive Pavlov 2006 s1, s2, action1, action2, str(player1))
> and more of the same error.

Don't worry about this, it's just a warning and not a failure. Travis triggered a failure but that was for a timeout on some of our property based tests, I've reset that so we'll see if that fixes itself.

marcharper:
@janga1997 please add the strategy file to type_tests.sh

In Python 3.5 we cannot annotate variables, that's a [Python 3.6 feature](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep526). If you are using Python 3.6 for testing with mypy you may get different results. Travis will check both once you add the file to type_tests.

janga1997:
@marcharper I am using Python 3.5. 
But I am sure a similar situation has occured in previous commits, but it never raised an issue there. 
I don't understand what is unique about this situation?

marcharper:
I'm not sure really -- can you rebase/merge and run mypy again? I'll take a closer look at the errors.

marcharper:
mypy is unhappy because we're using both None and strings for the same variable without declaring its type. Just change both lines to: `self.opponent_class = ""`.

drvinceknight:
Since #852 has been merged in there are now merge conflicts @janga1997 :)

Issue #845, opened 2017-02-01
Discussing the dynamically created classes


drvinceknight:
The dynamically created classes are creating very minor issues with type checking (and also causing slight issues with hashing in the fingerprinting repo but that's not a big deal). Is it worth perhaps revisiting whether or not they are necessary?

Perhaps we could include the data in the corresponding modules themselves (as a class attribute for each given class). This has the negative side effect of a bit more manual work to update the classes but it simplifies things (like with type checking and also removes the need for the `read_data` module).

Any thoughts?

marcharper:
Dynamic generation isn't necessary, though it could be useful in some situation. We can leave the data loading as is and just name the classes explicitly.

drvinceknight:
> Dynamic generation isn't necessary, though it could be useful in some situation. 

Is there a particular situation you have in mind? 

> We can leave the data loading as is and just name the classes explicitly.

Good idea, it will keep things a bit tidier. I'm happy to have a go at this.

marcharper:
If you wanted to study a tournament of randomly generated FSM strategies, for instance, it would be a huge pain to do if the classes were not dynamically generated. So if we remove that example from the main library, an advanced tutorial would be nice.

Also, before you remove those guys I'd check that the strategy transformers don't cause the same problem

drvinceknight:
> If you wanted to study a tournament of randomly generated FSM strategies, for instance, it would be a huge pain to do if the classes were not dynamically generated. So if we remove that example from the main library, an advanced tutorial would be nice.
> 

I might be missing something but could this not be taken care of using inheritance from a parent FSM class? (like for the simpler memory one strategies for example)

> Also, before you remove those guys I'd check that the strategy transformers don't cause the same problem

Yeah, I was thinking that the PR should change travis to run `mypy` on everything anyway so it would see what this catches (or doesn't).

marcharper:
Yeah maybe there is an easier way if the base strategies are parameterized, so maybe my example isn't very good.

PR #852, opened 2017-02-05
Remove dynamic classes and tidy imports

drvinceknight:
This gets rid of the dynamic `lookerup` classes.

I also tided a number of things in our imports to make type checking happier:

This is the remaining output of running mypy on the full library:

```
Axelrod(845) ✗: mypy --ignore-missing-imports --follow-imports skip axelrod
axelrod/player.py:92: error: Need type annotation for variable
axelrod/strategies/meta.py:388: error: Invalid type "axelrod.strategies.meta.NiceMetaWinnerEnsemble"
axelrod/strategies/meta.py:388: error: Invalid base class
axelrod/strategies/meta.py:400: error: Invalid type "axelrod.strategies.meta.NiceMetaWinnerEnsemble"
axelrod/strategies/meta.py:400: error: Invalid base class
axelrod/strategies/meta.py:411: error: Invalid type "axelrod.strategies.meta.NiceMetaWinnerEnsemble"
axelrod/strategies/meta.py:411: error: Invalid base class
axelrod/strategies/meta.py:422: error: Invalid type "axelrod.strategies.meta.NiceMetaWinnerEnsemble"
axelrod/strategies/meta.py:422: error: Invalid base class
axelrod/strategies/meta.py:433: error: Invalid type "axelrod.strategies.meta.NiceMetaWinnerEnsemble"
axelrod/strategies/meta.py:433: error: Invalid base class
axelrod/__init__.py:19: error: Name 'strategies' already defined
axelrod/tests/unit/test_windows_detection.py:9: error: 'overload' decorator expected
axelrod/tests/unit/test_windows_detection.py:15: error: 'overload' decorator expected
axelrod/tests/unit/test_player.py:170: error: Need type annotation for variable
axelrod/tests/unit/test_interaction_utils.py:26: error: List item 0 has incompatible type "Tuple[Tuple[str, str], float]"
axelrod/tests/unit/test_interaction_utils.py:26: error: List item 1 has incompatible type "Tuple[Tuple[str, str], float]"
axelrod/tests/unit/test_interaction_utils.py:27: error: List item 0 has incompatible type "Tuple[Tuple[str, str], float]"
axelrod/tests/unit/test_interaction_utils.py:28: error: List item 0 has incompatible type "Tuple[Tuple[str, str], float]"
axelrod/tests/unit/test_interaction_utils.py:28: error: List item 1 has incompatible type "Tuple[Tuple[str, str], float]"

```

I suggest that once this PR is merged we take that list and open another issue aiming to get rid of those lasting items.


marcharper:
Can this be fixed without removing strategy transformers?
```
axelrod/strategies/meta.py:388: error: Invalid type "axelrod.strategies.meta.NiceMetaWinnerEnsemble"
axelrod/strategies/meta.py:388: error: Invalid base class
```

axelrod/strategies/averagecopier.py
```diff
@@ -21,8 +22,7 @@ class AverageCopier(Player):
         'manipulates_state': False
     }
 
-    @staticmethod
-    def strategy(opponent: Player) -> Action:
```
marcharper: because of the random?

axelrod/strategies/lookerup.py
```diff
@@ -165,23 +167,310 @@ def strategy(self, opponent):
...
-        name, plays, op_plays, op_start_plays)
-    # Dynamically create the class
-    new_class = type(class_name, (LookerUp,), {})
-    new_class.__init__ = init_args(partial(
-        new_class.__init__, lookup_table=table, initial_actions=initial))
-    new_class.name = class_name
-    # Add the generated class to the module dictionary so it can be
-    # imported into other modules
-    setattr(module, class_name, new_class)
+class EvolvedLookerUp0_0_1(LookerUp):
```
marcharper: I don't know that we actually want to add all of these to the library. I only included a few in the axelrod.strategies list -- maybe just add those? Some of the small ones shadow strategies in the library already, and the ones with larger parameters don't seem to outperform 2_2_2 for the most part.

drvinceknight:
> Can this be fixed without removing strategy transformers?

I have no idea.

It took me quite a while to get this far so I suggest that gets looked in to after this PR. Either way we're not going to remove strategy transformers right?

marcharper:
> It took me quite a while to get this far so I suggest that gets looked in to after this PR. Either way we're not going to remove strategy transformers right?

I don't know? If not then why are we removing the dynamic class generation for LookerUp? For readability? I'm not opposed per se I'm just not sure what problem it solves.

axelrod/strategies/averagecopier.py
drvinceknight: This was being pointed out as an error by `mypy` something to do with the method not being in line with the parent class.

axelrod/strategies/lookerup.py
drvinceknight: Sure, you only include 1-1-1 and 2-2-2 in the whole list. Should we go ahead and just include those?

drvinceknight:
> I don't know?

We certainly should not remove the strategy transformers.

> If not then why are we removing the dynamic class generation for LookerUp? For readability? I'm not opposed per se I'm just not sure what problem it solves.

Readability would be a big reason for me, I'd almost turn this around the other way, going back to my original question on #845: what problem does having dynamically generated classes solve?

I actually found this quite hard to untangle putting this PR so would be strongly in favour of verbose class definitions (in line with the rest of the library). I'd also suggest that if we are only going to have 1-1-1 and 2-2-2 (which I'm fine with) then we could get rid of the data file for lookerups.

I'd actually suggest that the whole reading in of data from the external source doesn't help with readability. For example in this particular case there was an error in the header which caused me a 5 minute labyrinth run around but that's perhaps a discussion for another PR.

marcharper:
Ok, I understand. The reason I did it the way it currently is was so that if someone else wants to play with the other variants they can easily, and they wouldn't be tempted to add new variants that had already been created. It was also easy to add a bunch of variants to the library by just adding new lines in the data file rather than write new classes each time, which is pretty tedious and involves a lot of repeated boilerplate that's easy to mess up -- class name, data, player name, parent class... it was much easier to just comment out a line in the data file and add a new one. The dynamic code is ~20 lines, all the individuals classes written out are ~300.

I also didn't want to remove all the variants, but I didn't want to add 15 very similar strategies to the standard tournament -- we already have a number of players that are quite similar, like the gobymajorities and retaliators. 

So I don't think there is a great solution, but if we want to say readability takes precedent then I'll go along with that.

Regardless I think it's better no having large static data chunks in the strategy classes themselves, so I would prefer to keep the data loading as is.

drvinceknight:
> The dynamic code is ~20 lines, all the individuals classes written out are ~300.

As I said, I'm not sure it's self evident that that's a good thing.

> So I don't think there is a great solution, but if we want to say readability takes precedent then I'll go along with that.

Yes my vote goes to readability. It would be a minor refactor to handle all this with inheritance which would appease your previous concern (and keep in line with the rest of the library with other such cases).

> 
> Regardless I think it's better no having large static data chunks in the strategy classes themselves, so I would prefer to keep the data loading as is.

I would personally suggest keeping things in one spot where it's not tedious to do is better with the argument of removing complexity (I went on a bit of a merrygoround ride when working on this).

I've just pushed removing the various ones that were not added to the list, so only have `2_2_2` and `1_1_1`. I have removed the data loading for this case as it was in essence there for reading two short strings which felt silly.

drvinceknight:
09daa29 refactors the code a bit more so that it takes use of inheritance.

drvinceknight:
> Regardless I think it's better no having large static data chunks in the strategy classes themselves, so I would prefer to keep the data loading as is.

I've also just pushed some changes that removed unused code (mainly relevant to reading in the two patterns). None of my main changes (including the inheritance) are incompatible with reading in the data if we really want that to stay.

axelrod/strategies/lookerup.py
```diff
@@ -165,60 +158,62 @@ def strategy(self, opponent):
...
+    Names:
+        - Evolved Lookerup 2 2 2: Original name by Marc Harper
+    """
+    name = "EvolvedLookerUp2_2_2"
+    pattern = 'CDCCDCCCDCDDDCCCDCDDDDDDDCDDDCDCDDDDCCDCCCCDDDDCCDDDDCCDCDDDDDDD'
+    parameters = (2, 2, 2)
+    initial_actions = (C, C)
 
 
-class Winner12(LookerUp):
```
marcharper: I prefer inheriting from LookerUp to reduce the number of inherited classes. It seems odd to me that Winner12 would derive from EvolvedLookerUp1_1_1.

axelrod/data/lookup_tables.csv
```diff
@@ -1,19 +0,0 @@
-# Name (string), plays (int), opp_plays(int), starting_plays(int), pattern (str)
```
marcharper: Can we find somewhere else to save this data? (the evolver repository, perhaps)

axelrod/strategies/lookerup.py
```diff
@@ -165,60 +158,62 @@ def strategy(self, opponent):
...
 
 
-# Create several classes at runtime based on the data in the look up tables
-# loaded above, one for each key.
+class EvolvedLookerUp1_1_1(LookerUp):
+    """
+    A 1_1_1 Lookerup trained with an evolutionary algorithm.
+
+    Names:
+        - Evolved Lookerup 1 1 1: Original name by Marc Harper
```
marcharper: The names are inconsistent -- let's pick underscores or no underscores.

axelrod/data/lookup_tables.csv
drvinceknight: Good call: https://github.com/Axelrod-Python/axelrod-evolver/pull/20

axelrod/strategies/lookerup.py
drvinceknight: I've gone with an intermediate class. Let me know what you think.

axelrod/strategies/lookerup.py
drvinceknight: :+1:

marcharper:
I think kwarg arguments for the pattern make more sense then the intermediate class having member variables. If I wanted to instantiate a new lookerup (say a 4, 4, 2) I have to make a derived classed; with arguments I could just use the existing abstract base class.

Otherwise I'm happy with the PR.

drvinceknight:
> I think kwarg arguments for the pattern make more sense then the intermediate class having member variables. If I wanted to instantiate a new lookerup (say a 4, 4, 2) I have to make a derived classed; with arguments I could just use the existing abstract base class.

I'm not sure I understand. Are you talking about using kwargs in the `init` method?

drvinceknight:
> I'm not sure I understand. Are you talking about using kwargs in the init method?

I might have it, just working on a commit now :)

drvinceknight:
Let me know if e7b02d6 is what you had in mind.

marcharper:
Thanks, that's what I wanted :) I approved the changes, looks like we need a rebase and then we're good.

drvinceknight:
> Thanks, that's what I wanted :) I approved the changes, looks like we need a rebase and then we're good.

Cool, just pushed that :) I also added another test for the pattern init.

marcharper:
@drvinceknight can you rebase/merge?

drvinceknight:
> @drvinceknight can you rebase/merge?

Just seen this :) Will have dinner and do it after.

drvinceknight:
> @drvinceknight can you rebase/merge?

Done. :+1:

PR #855, opened 2017-02-06
Add type hints to gobymajority, doubler and gambler

janga1997:
#808 
The file `gobymajority` produces some errors on `mypy`.
```
axelrod/strategies/gobymajority.py:48: error: Unsupported operand types for > ("float" and "object")
axelrod/strategies/gobymajority.py:54: error: Unsupported operand types for < ("int" and "object")
axelrod/strategies/gobymajority.py:54: error: Incompatible types in string interpolation (expression has type "object", placeholder has type "Union[int, float]")
axelrod/strategies/gobymajority.py:68: error: Unsupported operand type for unary - ("object")
```

It interprets `self.*` as objects instead of the types of the respective parameters.
Any thoughts? @drvinceknight @marcharper


marcharper:
Can you merge the new master and rerun mypy? I don't see any obvious reason why the types are failing (may be a mypy issue).

marcharper:
The mypy errors here are because we declared the type of one but not both of the variables in the comparison.

janga1997:
I'm not sure as to how to resolve that issue. How can I declare the type of the of `self.*`?

marcharper:
I think you just need a variable annotation on the first usage (hopefully)

janga1997:
This branch is beyond repair right now. I will add these type hints to #871 .

PR #857, opened 2017-02-08
Add new strategy ShortMem

janga1997:
From the suggested list of strategies by @marcharper from [this paper](http://abricom.org.br/wp-content/uploads/2016/03/bricsccicbic2013_submission_202.pdf).

I'm sure there a few things I missed. @drvinceknight Do take a look whenever free.
I will work on the tests if this is okay so far.


axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,36 @@
+from axelrod import Actions, Player
+from axelrod.actions import Action
+
+C, D = Actions.C, Actions.D
+
+class ShortMem(Player):
+    """A player who only ever defects."""
```
drvinceknight: I don't think this docstring is accurate right?

Also note that we require a few more details in the docstring, you can see an example here: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_the_new_strategy.html

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,36 @@
...
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    @staticmethod
+    def strategy(opponent: Player) -> Action:
+        
+        memoryDepth = self.classifier['memoryDepth']
+        cooperateRatio = array.count('C')/memoryDepth
```
drvinceknight: You haven't defined `array` so I expect this strategy would fail to play.

Note that we're going to need some tests for the strategy too, you can see some information on the tests here: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html

We're happy to give a hand with tests if you'd like one. Essentially we want code that checks that all the logic you write below works :)

Also, note that a `Player` has a `cooperations` attribute that counts the number of cooperations.

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,36 @@
...
+        
+        if len(self.history) <= memoryDepth:
+            return C
+            
+        if cooperateRatio - defectRatio > 0.3:
+            return C
+        elif defectRatio - cooperateRatio > 0.3:
+            return D
+        else:
+            return TitForTat().strategy(opponent)
```
drvinceknight: Could you code this explicity please. Something like:

```
else:
    return opponent.history[-1]
```

janga1997: Sorry about that. I copied over the code from defector and made the necessary changes.

axelrod/strategies/shortmem.py
janga1997: I though about that, but we need the number of cooperations and defections in the last 10 steps only, and to the extent that I understand it, it records the `C`s and `D`s for the whole play.

janga1997:
@drvinceknight I've been meaning to get a little feedback. In this case, is FIFO really relevant to the what this strategy does, because I don't think it is.

drvinceknight:
> I've been meaning to get a little feedback. In this case, is FIFO really relevant to the what this strategy does, because I don't think it is.

Sorry, what do you mean by FIFO?

janga1997:
It's an acronym for `first in, first out`, where the oldest (first) entry, or 'head' of the queue, is processed first. This term was specifically mentioned in the description of this strategy in the paper.

drvinceknight:
> It's an acronym for first in, first out, where the oldest (first) entry, or 'head' of the queue, is processed first. This term was specifically mentioned in the description of this strategy in the paper.

I think it's ok to simply say "the last ten moves".

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,43 @@
...
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    @staticmethod
+    def strategy(opponent: Player) -> Action:
+        
+        memoryDepth = self.classifier['memoryDepth']
+        array = self.history[:-11:-1]
```
drvinceknight: As order doesn't matter you can just take: `self.history[-10:]`.

```
>>> a = list(range(20))
>>> a[-10:]
[10,11,12,13,14,15,16,17,18,19]
```

However note that I think you mean for this to be `opponent.history`.

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,43 @@
...
+        cooperateRatio = array.count('C')/memoryDepth
+        defectRatio = array.count('D')/memoryDepth
+        
+        if len(self.history) <= memoryDepth:
```
drvinceknight: You could move this to before the counting (a tiny efficiency improvement).

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,43 @@
+from axelrod import Actions, Player
+from axelrod.actions import Action
+
+C, D = Actions.C, Actions.D
+
+class ShortMem(Player):
+    """
+    A player starts by always cooperating for the first 10 moves.
+    
+    The opponent answers are stored in the memory. The memory is FIFO and the maximum size of the memory is 10 results. From the tenth round on, the program analyzes the memory, and compare the number of defects and cooperates of the opponent, based in percentage. If cooperation occurs 30% more than defection, it will cooperate.
```
drvinceknight: Could you stick to PEP8 here, so keep the characters to less than 80 per line please :)

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,43 @@
...
+
+C, D = Actions.C, Actions.D
+
+class ShortMem(Player):
+    """
+    A player starts by always cooperating for the first 10 moves.
+    
+    The opponent answers are stored in the memory, whose maximum size is 10 results. From the tenth round on, the program analyzes the memory, and compare the number of defects and cooperates of the opponent, based in percentage. If cooperation occurs 30% more than defection, it will cooperate.
+    If defection occurs 30% more than cooperation, the program will defect. Otherwise, the program follows the TitForTat algorithm.
+
```
drvinceknight: Can you add something like:

```
    Names

    - Short Mem: [Axelrod1980]_
    """
```

Except replace `Axelrod1980` by a key that you'd put in https://github.com/Axelrod-Python/Axelrod/blob/master/docs/reference/bibliography.rst

janga1997:
@drvinceknight I hope its okay now?

Nikoleta-v3:
@janga1997 Hello I am also a contributor to the library! Did you remember to add the reference like suggested. 

>Except replace Axelrod1980 by a key that you'd put in https://github.com/Axelrod-Python/Axelrod/blob/master/docs/reference/bibliography.rst

janga1997:
@Nikoleta-v3 no I didn't yet. I will do it tomorrow along with the tests, as it late night here. Thanks .

janga1997:
@drvinceknight @Nikoleta-v3 Do review whenever you're free. I believe I've covered the necessary test cases.

docs/reference/bibliography.rst
```diff
@@ -37,3 +37,4 @@ documentation.
 .. [Stewart2012] Stewart, a. J., & Plotkin, J. B. (2012). Extortion and cooperation in the Prisoner’s Dilemma. Proceedings of the National Academy of Sciences, 109(26), 10134–10135. http://doi.org/10.1073/pnas.1208087109
 .. [Szabó1992] Szabó, G., & Fáth, G. (2007). Evolutionary games on graphs. Physics Reports, 446(4-6), 97–216. http://doi.org/10.1016/j.physrep.2007.04.004
 .. [Tzafestas2000] Tzafestas, E. (2000). Toward adaptive cooperative behavior. From Animals to Animals: Proceedings of the 6th International Conference on the Simulation of Adaptive Behavior {(SAB-2000)}, 2, 334–340.
+.. [Andre2013] Andre L. C., Honovan P., Felipe T. and Frederico G. (2013). Iterated Prisoner’s Dilemma - An extended analysis, http://abricom.org.br/wp-content/uploads/2016/03/bricsccicbic2013_submission_202.pdf
```
Nikoleta-v3: I am not sure if the reference is generated correctly because there is an extra `:`  in `shortmem.py ` line 15. Could you please check. :smile: 
>  ShortMem: [Andre:2013]_

janga1997:
@Nikoleta-v3 Is everything fine with the tests and the strategy itself?

Nikoleta-v3:
Everything else looks good to me! Happy to be working with you :smile: 
The core contributors need to give a :+1: now.

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,45 @@
+from axelrod import Actions, Player
+from axelrod.actions import Action
+
+C, D = Actions.C, Actions.D
+
+class ShortMem(Player):
+    """
+    A player starts by always cooperating for the first 10 moves.
+    
+    The opponent answers are stored in the memory, whose maximum size is 10 results. From the tenth round on, the program analyzes the memory, and compare the number of defects and cooperates of the opponent, based in percentage. If cooperation occurs 30% more than defection, it will cooperate.
```
drvinceknight: This all looks good to me, could you keep the number of characters on each line to less than 80 please? So it just needs this docstring to line break a bit earlier.

janga1997:
@drvinceknight Just a silly question, when I use the word wrap function in the editor, it doesn't show up in `git diff`. How exactly do you make a word wrap at the the 80th column.

drvinceknight:
> How exactly do you make a word wrap at the the 80th column.

It would depend on your editor. There are ways to set it up so it does it automatically but you can also create a visual queue that just shows where the 80 character limit is and then you can line break manually.

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,50 @@
...
+        'manipulates_state': False
+    }
+
+    @staticmethod
+    def strategy(opponent: Player) -> Action:
+        
+        if len(opponent.history) <= 10:
+            return C
+        
+        array = opponent.history[:-11:-1]
```
marcharper: Take a look at this line, the arrays are off by one.
```
>>> l = list(range(100))
>>> l[:-11:-1]
[99, 98, 97, 96, 95, 94, 93, 92, 91, 90]
>>> l[-11:]
[89, 90, 91, 92, 93, 94, 95, 96, 97, 98, 99]
```

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,50 @@
+from axelrod import Actions, Player
+from axelrod.actions import Action
+
+C, D = Actions.C, Actions.D
+
+class ShortMem(Player):
+    """
+    A player starts by always cooperating for the first 10 moves.
+    
+    The opponent answers are stored in the memory, whose maximum size is 10
```
marcharper: Technically all the history is saved in memory in our library. I would just say that the player analyzes the last ten rounds of play without any details about what's in memory.

answers -> actions

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @staticmethod
+    def strategy(opponent: Player) -> Action:
+        
+        if len(opponent.history) <= 10:
+            return C
+        
+        array = opponent.history[:-11:-1]
+        cooperateRatio = array.count('C')
```
marcharper: These aren't ratios, `cooperation_counts` and `defection_counts` would be better names, or simply `C_counts`.

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,50 @@
...
+class ShortMem(Player):
+    """
+    A player starts by always cooperating for the first 10 moves.
+    
+    The opponent answers are stored in the memory, whose maximum size is 10
+    results. From the tenth round on, the program analyzes the memory, and
+    compare the number of defects and cooperates of the opponent, based in
+    percentage. If cooperation occurs 30% more than defection, it will
+    cooperate.
+    If defection occurs 30% more than cooperation, the program will defect.
```
marcharper: program --> player

axelrod/tests/unit/test_shortmem.py
```diff
@@ -0,0 +1,34 @@
+"""Tests for the Cooperator strategy."""
```
marcharper: Cooperator -> shortmem

Please add a test to catch the array length issue above (array is of length 9 instead of 10 but none of the tests were tripped).

axelrod/tests/unit/test_shortmem.py
```diff
@@ -0,0 +1,34 @@
...
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_strategy(self):
+        # Starts by cooperating for the first ten moves.
+        self.responses_test([C], [C] * 4, [D] * 4)
+        self.responses_test([C], [C] * 4 + [D] *6, [D] * 4 + [C] * 6)
+        
+        #Cooperate if in the last ten moves, Cooperations - Defections >= 30
```
marcharper: 30% and see below

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,50 @@
...
+
+C, D = Actions.C, Actions.D
+
+class ShortMem(Player):
+    """
+    A player starts by always cooperating for the first 10 moves.
+    
+    The opponent answers are stored in the memory, whose maximum size is 10
+    results. From the tenth round on, the program analyzes the memory, and
+    compare the number of defects and cooperates of the opponent, based in
```
marcharper: defections and cooperations

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,50 @@
...
+    def strategy(opponent: Player) -> Action:
+        
+        if len(opponent.history) <= 10:
+            return C
+        
+        array = opponent.history[:-11:-1]
+        cooperateRatio = array.count('C')
+        defectRatio = array.count('D')
+            
+        if cooperateRatio - defectRatio >= 3:
```
marcharper: FYI standard Python style is `camel_case` instead of `CamelCase` for variables, search for PEP8 (the style guide).

axelrod/strategies/shortmem.py
janga1997: Correct me if I am wrong, since we are talking about the last ten rounds here, doesn't `l[-11:]` produce a slice with 11 elements?

axelrod/tests/unit/test_shortmem.py
janga1997: I didn't understand what you mean.

axelrod/tests/unit/test_shortmem.py
janga1997: I replied about the array length issue above.

axelrod/strategies/shortmem.py
drvinceknight: I think I mentioned this at some point above, I think you can just use:

```
>>> a = range(20)
>>> list(a[-10:])
[10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
```

janga1997: I understand about the readability/ambiguity issue, but what exactly is a test that checks this?

Is it    `self.responses_test([C], [C] * 4 + [D] *6, [D] * 4 + [C] * 6)` ?

marcharper: Sorry, @drvinceknight's suggestion is better. A test to check would fail on 9 trailing elements but not 10,

axelrod/tests/unit/test_shortmem.py
marcharper: We're not calculating percentages here. I get that it's pointless to divide by 10 on both sides but it's not an accurate description, and it might be confusing to look at this code in six months. So I'm asking that the comments elaborate more about what the code's intention.

axelrod/strategies/shortmem.py
janga1997: Sorry for being difficult, but I really don't see the reason for an additional test.

The strategy is to return `C` from 1-9 moves, and from the 10th move, the rest of the logic kicks in.
I don't understand what you and @drvinceknight mean by 

> A test to check would fail on 9 trailing elements but not 10

drvinceknight: > Sorry for being difficult, but I really don't see the reason for an additional test.

Don't worry you're not being difficult, it's good to talk these things over :)

I would definitely suggest you change:

```
opponent.history[:-11:-1]
```

to be 

```
opponent.history[-10:]
```

This just helps with readability.

In terms of the test, I suggest you just add the following one:

```
self.responses_test([C], [D] *9, [D]  * 9)  # This checks that when there are 9 previous plays you cooperate
self.responses_test([D], [D] *10, [D]  * 10)  # This checks that when there are 10 previous plays the logic kicks in
```

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,49 @@
...
+    def strategy(opponent: Player) -> Action:
+        
+        if len(opponent.history) <= 10:
+            return C
+        
+        array = opponent.history[:-11:-1]
+        C_counts = array.count('C')
+        D_counts = array.count('D')
+            
+        if C_counts - D_counts >= 3:
```
drvinceknight: Another question (not sure how we missed this), is the logic correct here?

You're looking at the difference in counts (subtraction) but the description is saying that there should be 30% more cooperations than defections? So I'm not sure the logic is actually correct.

I think this test should be:

```
if C_counts / D_counts >= 1.3:
```

That checks that there are 30% more cooperations.

janga1997: I am starting to doubt it a bit myself.
But doesn't the term `more than` imply subtraction?

janga1997: After thinking about it, it seems your logic makes more sense w.r.t. the given statement.

When I first thought about it, I made the assumption that percentages are to be calculated w.r.t the sum of C and D (10). My mistake. I would have never found it myself.

drvinceknight: > But doesn't the term more than imply subtraction?

My suggestion is incorrect, because we're only considering 10 turns this is actually correct. Apologies.

drvinceknight: @janga1997 I actually think I'm wrong. I think you logic was correct before.

axelrod/strategies/shortmem.py
```diff
@@ -0,0 +1,57 @@
...
+        C_counts = array.count('C')
+        D_counts = array.count('D')
+        
+        if C_counts == 0:
+            return D
+            
+        if D_counts == 0:
+            return C
+            
+        if C_counts / (D_counts * 1.0) >= 1.3:
```
drvinceknight: I think the logic was actually correct before (my mistake) I suggest you revert: a7b8d9f

janga1997:
@drvinceknight Is that everything?

drvinceknight:
> @drvinceknight Is that everything?

This looks good to me. I'll leave @marcharper to merge it :) :+1:

PR #858, opened 2017-02-09
Add type hints to handshake, inverse, mathematicalconstants

janga1997:
#808 
Only one error in mypy, regarding the file `mathematicalconstants.py`
We are importing `e` from the `math` module and defining a `class e` too, which is raising some errors. 
What would you like the `class e(CotoDeRatio):` to be renamed to?
@marcharper @drvinceknight @Nikoleta-v3


drvinceknight:
> What would you like the class e(CotoDeRatio): to be renamed to?

Don't rename the `e` class simply change the imports to be:

```
import math
```

and then use `math.e`, `math.pi` etc...

drvinceknight:
> @marcharper @drvinceknight @Nikoleta-v3

Also: there's no need to ping us all. The core developers get notifications.

janga1997:
Sorry about the pinging. Won't happen again.

marcharper:
You can ping us if you need to just know that we already get notifications for every PR and every comment/update. We're also in very different time zones.

drvinceknight:
@janga1997 it looks like this is forked from your short mem player (https://github.com/Axelrod-Python/Axelrod/pull/858/files#diff-a964005e2620fada6accc62bb679e816)? So merging this would include that player too but we're still discussing that on #857.

PR #860, opened 2017-02-11
Add type hints to geller, hunter, memorytwo

janga1997:
#808
Only one error on mypy for `hunter.py`.
`axelrod/strategies/hunter.py:107: error: Need type annotation for variable`
I think this is an error with mypy where it can't infer type for empty containers. See [here](https://github.com/python/mypy/issues/1534).

### DIFF
--- a/axelrod/strategies/geller.py
+++ b/axelrod/strategies/geller.py
@@ -1,6 +1,6 @@
 import inspect
 
-from axelrod.actions import Actions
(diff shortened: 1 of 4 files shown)


janga1997:
@meatballs Could you help out with this one?

janga1997:
@drvinceknight I've fixed the variable type annotation for this PR too.

PR #863, opened 2017-02-14
Add type hints to mindcontrol, mindreader, mutual, negation

janga1997:
#808 
No errors on mypy.


drvinceknight:
Looks great. :+1:

PR #839, opened 2017-01-31
Add type for Moran

souravsingh:
Fixes a part of #808

PR #866, opened 2017-02-17
Add new Strategy SelfSteem

janga1997:
#379 
I've tried to make sure everything was done, but I might have left something. Do take a look.

To see the [original paper](http://abricom.org.br/wp-content/uploads/2016/03/bricsccicbic2013_submission_202.pdf), Page 3


axelrod/strategies/selfsteem.py
```diff
@@ -0,0 +1,49 @@
...
+class SelfSteem(Player):
+    """
+    This strategy is based on the feeling with the same name.
+    It is modeled on the sine curve(f = sin( 2* pi * n / 10 )), which varies
+    with the current iteration.
+    
+    If f > 0.95, 'ego' of the algorithm is inflated; always defects.
+    If 0.95 > |f| > 0.3, rational behavior; follows TitForTat algortithm.
+    If 0.3 > f > -0.3; random behavior.
+    If f < -0.95, algorithm is at rock bottom; always cooperates.
```
drvinceknight: Can you add `Names ...` see: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_the_new_strategy.html This will also require you to add the reference to `docs/reference/bibliography.rst`.

janga1997:
One doubt I have is to whether to check for cases where the strategy returns a random `Action`?

axelrod/strategies/selfsteem.py
drvinceknight: Can you also add this strategy to `docs/reference/all_strategies.rst`

axelrod/strategies/selfsteem.py
```diff
@@ -0,0 +1,49 @@
...
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        turns_number = len(self.history)
+        sine_value =  sin(2*pi*turns_number/10)
```
drvinceknight: PEP8: `2 * pi * ...`

axelrod/strategies/selfsteem.py
drvinceknight: More info here: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/adding_the_new_strategy.html

axelrod/tests/unit/test_selfesteem.py
```diff
@@ -0,0 +1,37 @@
...
+        # Check for f < -0.95
+        self.responses_test([C], [C] * 7, [C] * 7)
+        self.responses_test([C], [C] * 18, [D] * 18)
+        
+        # Check for 0.95 > abs(f) > 0.3
+        self.responses_test([C], [C], [C])
+        self.responses_test([D], [C] * 16, [D] * 16)
+        self.responses_test([C], [D] * 9, [C] * 9)
+        
+
```
drvinceknight: Can you get rid of the blank lines but also include a test for the random behaviour, you can do this by passing a `seed=` to the `responses_test` function: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html

janga1997: What exactly is happening here with `seed` ? 
I saw in `seed =1` and `seed = 2` being passed. What does it do?

janga1997:
@drvinceknight Does the build error have something to do with my PR? 
I saw a lot of errors about documentation and `moran.py`

docs/reference/all_strategies.rst
```diff
@@ -135,6 +135,9 @@ Here are the docstrings of all the strategies in the library.
 .. automodule:: axelrod.strategies.shortmem
    :members:
    :undoc-members:
+.. automodule:: axelrod.strategies.selfsteem
+  :members:
```
drvinceknight: Your indentation levels aren't correct here. Compare them to the other strategies.

drvinceknight:
> @drvinceknight Does the build error have something to do with my PR?
> I saw a lot of errors about documentation and moran.py

The error is in the building of the docs:

```
Warning, treated as error:
/home/travis/build/Axelrod-Python/Axelrod/axelrod/strategies/selfsteem.py:docstring of axelrod.strategies.selfsteem.SelfSteem:5: ERROR: Undefined substitution referenced: "f".
```

It might be due to the indentation I pointed out but it could be something else. If you run `make html` in the `/docs` dir you can make the docs locally and see what the error is (you might need to pip install a couple of sphinx dependencies).

drvinceknight:
There are other errors, I'm taking a look now.

axelrod/strategies/selfsteem.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class SelfSteem(Player):
+    """
+    This strategy is based on the feeling with the same name.
+    It is modeled on the sine curve(f = sin( 2* pi * n / 10 )), which varies
+    with the current iteration.
+    
+    If f > 0.95, 'ego' of the algorithm is inflated; always defects.
+    If 0.95 > |f| > 0.3, rational behavior; follows TitForTat algortithm.
```
drvinceknight: This line is creating an error as sphinx is confused by `|f|`. I suggest replacing it with `abs(f)`.

axelrod/strategies/selfsteem.py
```diff
@@ -0,0 +1,53 @@
...
+        sine_value =  sin( 2 * pi * turns_number / 10 )
+          
+        if sine_value > 0.95:
+            return D
+        
+        elif abs(sine_value) < 0.95 and abs(sine_value) > 0.3:
+            return opponent.history[-1]
+        
+        elif sine_value < 0.3 and sine_value > -0.3:
+            return random.choice((C, D))
```
drvinceknight: Better to use the inbuilt `random_choice()` from `axelrod.random_`.

axelrod/strategies/selfsteem.py
```diff
@@ -0,0 +1,53 @@
...
+    If 0.95 > |f| > 0.3, rational behavior; follows TitForTat algortithm.
+    If 0.3 > f > -0.3; random behavior.
+    If f < -0.95, algorithm is at rock bottom; always cooperates.
+    
+    Names:
+    
+    - SelfSteem: [Andre2013]_
+    """
+
+    name = 'SelfSteem'
```
drvinceknight: What is the name of this strategy?

`SelfSteem` or `SelfEsteem`? Across module names etc you seem to be combining these.

drvinceknight:
My suggestions there fix some of the errors **but** not all of them. My hunch is that it's something offsetting the random seed somewhere but I'd need more time to look through it. I'll try and get to it over the weekend if you don't find the fix but I'm travelling so it might be delayed :)

I'm sure we'll figure it out :)

axelrod/tests/unit/test_selfesteem.py
drvinceknight: So a random process starts with a random seed. This just ensures the "randomness" can be repeated:

```
>>> import random
>>> random.seed(0)
>>> random.random()
0.8444218515250481
>>> random.seed(0)
0.8444218515250481
```

The `seed=1` and `seed=2` just set those seeds for the strategies.

axelrod/strategies/selfsteem.py
janga1997: It is `SelfSteem`. I mistakenly named the test `test_selfesteem` instead of `test_selfsteem`.

janga1997:
@drvinceknight One question keeps coming up to me.
If we know the value generated is random, why do we need to test it?
What are we testing for here? that whether the value generated is indeed random?

drvinceknight:
@janga1997 I've opened a PR on your fork: https://github.com/janga1997/Axelrod/pull/1

If you merge that in, it will automatically update the PR with my suggested commits.

The main error you were getting was caused because of a mis classification: you were saying the strategy was not stochastic and so it was being used in other tests (but it is stochastic).

janga1997:
@drvinceknight I've been seeing this error since the beginning.
```
File "./docs/tutorials/advanced/classification_of_strategies.rst", line 49, in classification_of_strategies.rst

Failed example:

    len(strategies)

Expected:

    64

Got:

    65
```

janga1997:
@drvinceknight And why did we merge #839 ?
It seems to throw a lot of errors on mypy.

drvinceknight:
> @drvinceknight I've been seeing this error since the beginning.

Yup, that will be the doctest for this page: http://axelrod.readthedocs.io/en/latest/tutorials/advanced/classification_of_strategies.html

Just bump the number.

drvinceknight:
> @drvinceknight And why did we merge #839 ?
> It seems to throw a lot of errors on mypy.

You're completely correct! I'm not sure how that got by. I'm looking at it now. I might revert it.

janga1997:
@drvinceknight I guess that was the main error. The tests seem to be passing now.

drvinceknight:
Yup, the only thing left is those errors about the type annotations in the moran process and as I've written on #868 that's not being picked up by travis :)

The strategy looks good :+1: I noted there's one point in the paper:

> Furthermore, the algorithms implements a retaliation policy, if the opponent defects; the sin curve is shifted (changing the sin phase).

As far as I can tell there's no indication as to how that is done right?

Could you also tweak in `docs/reference/bibliography.rst` and move `[Andre2013] ...` to be in alphabetical order?

janga1997:
@drvinceknight It didn't specify anything, not even how much the phase change is supposed to be, and in which direction.

drvinceknight:
> @drvinceknight It didn't specify anything, not even how much the phase change is supposed to be, and in which direction.

Yeah I see that. Not sure if this means it's worth rewording the docstring to make a note of this. Something like "This strategy is based on ..., however due to a limited number of details in the original description the sin function is not shifted as in the original work."

I'm happy with everything else.

marcharper:
I agree with @drvinceknight -- let's make a note in a docstring and email the authors for a clarification. Unfortunately this is a common issue and one of the goals of the library is to enable reproducible results going forward. It's clear that journals are not requiring source code disclosures and peer reviewers are not demanding sufficient descriptions :/

axelrod/strategies/selfsteem.py
```diff
@@ -0,0 +1,50 @@
...
+from axelrod.random_ import random_choice
+
+from math import pi, sin
+
+C, D = Actions.C, Actions.D
+
+
+class SelfSteem(Player):
+    """
+    This strategy is based on the feeling with the same name.
```
marcharper: self esteem?

axelrod/strategies/selfsteem.py
```diff
@@ -0,0 +1,50 @@
...
+    It is modeled on the sine curve(f = sin( 2* pi * n / 10 )), which varies
+    with the current iteration.
```
marcharper: current iteration `n` is the round number?

axelrod/strategies/selfsteem.py
```diff
@@ -0,0 +1,50 @@
...
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        turns_number = len(self.history)
+        sine_value =  sin(2 * pi * turns_number / 10)
```
marcharper: Just checking -- is this integer or float division in the reference?

axelrod/strategies/selfsteem.py
janga1997: It means self esteem, that's how its named in the paper. I decided not to change it.

axelrod/strategies/selfsteem.py
janga1997: In python3, doesn't it return a float by default?

axelrod/strategies/selfsteem.py
janga1997: Yes.

janga1997:
@marcharper I've emailed the authors of the paper regarding what I'm trying to do here, and what information I require. Yet to receive a reply.

axelrod/strategies/selfsteem.py
marcharper: It does, I'm just making sure that's correct.

marcharper:
Great, let's give them a day or two to reply. If we don't hear anything then we'll go with our best guess.

axelrod/strategies/selfsteem.py
janga1997: Yes, we do need a float division.

PR #869, opened 2017-02-19
Add strategy VeryBad

janga1997:
#379 
Here is the [paper](http://abricom.org.br/wp-content/uploads/2016/03/bricsccicbic2013_submission_202.pdf) . Page 3.

### DIFF
--- a/axelrod/strategies/verybad.py
+++ b/axelrod/strategies/verybad.py
@@ -1,0 +1,46 @@
+from axelrod.actions import Actions, Action
+from axelrod.player import Player
+
+C, D = Actions.C, Actions.D
+
+class VeryBad(Player):
+    """
+    It cooperates in the first three rounds, and uses probability
+    (it implements a memory, which stores the opponent’s moves) to decide for
+    cooperating or defecting.
+        
+    Names:
+    
+    - VeryBad: [Andre2013]_
+    """
+
+    name = 'VeryBad'
+    classifier = {
+        'memory_depth': float('inf'),
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    @staticmethod
+    def strategy(opponent: Player) -> Action:
+        
+        if len(opponent.history) < 4:
+            return C
+        
+        total_moves = len(opponent.history)
+        cooperations = opponent.cooperations
+        
+        cooperation_probability = cooperations / (total_moves * 1.0)
+        
+        if cooperation_probability > 0.5:
+            return C
+            
+        elif cooperation_probability < 0.5:
+            return D
+            
+        else:
+            return opponent.history[-1]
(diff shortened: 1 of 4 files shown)


axelrod/strategies/verybad.py
```diff
@@ -0,0 +1,46 @@
...
+    @staticmethod
+    def strategy(opponent: Player) -> Action:
+        
+        if len(opponent.history) < 4:
+            return C
+        
+        total_moves = len(opponent.history)
+        cooperations = opponent.cooperations
+        
+        cooperation_probability = cooperations / (total_moves * 1.0)
```
drvinceknight: There is no need for `* 1.0` here (this is all Python 3).

janga1997: Just though it couldn't hurt. Changing it now.

drvinceknight:
I've taken a look at the paper and I'm really not sure there's enough details there for us to assume this is an implementation of that strategy:

> It cooperates in the first three rounds, and uses probability (it implements a memory, which stores the opponent’s moves) to decide for cooperating or defecting.

The fact that they say "uses probability" makes me think that there should be some randomness somewhere...

That doesn't mean this strategy shouldn't be added to the library but I'm not sure we can say it's an implementation of the one from that paper... It might need to change name. Will be good to get the others' thoughts on this.

janga1997:
@drvinceknight I was confused by the lack of details too, but I assumed the most general meaning.
I could definitely be wrong about the intentions of the paper.

drvinceknight:
> I was confused by the lack of details too, but I assumed the most general meaning.
> I could definitely be wrong about the intentions of the paper.

Yeah it's really confusing and one of the problems with modern research where papers are written that are not possible to replicate... It might be that like I said on #866 we might need to say "losely based on a description given in...". Another option might be getting in touch with the authors and explaining what you're doing and asking for more details?

janga1997:
@drvinceknight I will try to contact one of the authors to clarify both.

marcharper:
If we don't hear back from the others just use your best judgement and update the docstrings to make it clear that a complete description could not be obtained.

janga1997:
It has been almost a week, and I haven't heard back from any of the authors.
I'm updating the docstrings for both the strategies.

marcharper:
Thanks @janga1997 . Please check the docstring formatting, that's why the build "failed".

axelrod/strategies/verybad.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    It cooperates in the first three rounds, and uses probability
+    (it implements a memory, which stores the opponent’s moves) to decide for
+    cooperating or defecting.
+    Due to a lack of information as to what that probability refers to in this
+    context, probability(P(X)) refers to (Count(X)/Total_Moves) in this
+    implementation
+    P(C) = Cooperations / Total_Moves
+    P(D) = Defections / Total_Moves = 1 - P(C)
+    Names:
```
drvinceknight: Could you add a blank line before `Names` please.

PR #870, opened 2017-02-19
Add state to action analysis.

drvinceknight:
Add two functions to interaction_utils:

- `interaction_utils.compute_state_to_action_distribution`: computes
  list of counter objects of mapping a `(state, action)` to a count.
- `interaction_utils.compute_normalised_state_to_action_distribution`: computes
  list of counter objects of mapping a `(state, action)` to a normalised
  count (for a given state).

There have then been incorporated in `axelrod.result_set` and in particular in
the `result_set.summarise()` output. Which now includes the average
rates for each strategy in a tournament. So for memory 1 strategies this
means we can measure their parameters and for non memory 1 strategies we
can infer what the parameters correspond to the data.

### DIFF
--- a/axelrod/interaction_utils.py
+++ b/axelrod/interaction_utils.py
@@ -175,10 +175,7 @@ def compute_state_to_action_distribution(interactions):
 
 def compute_normalised_state_to_action_distribution(interactions):
     """
-    Returns the normalised count of each state to action
-    for a set of interactions.
-
-    Returns a list (for each player) of normlised counts of each state to action
+    Returns a list (for each player) of normalised counts of each state to action
     pair for a set of interactions. A state to action pair is of the form:
 
     ((C, D), C)
--- a/axelrod/interaction_utils.py
+++ b/axelrod/interaction_utils.py
@@ -2,7 +2,7 @@
 Functions to calculate results from interactions. Interactions are lists of the
 form:
 
-    [('C', 'D'), ('D', 'C'),...]
+    [(C, D), (D, C),...]
 
 This is used by both the Match class and the ResultSet class which analyse
 interactions.
@@ -145,7 +145,7 @@ def compute_state_to_action_distribution(interactions):
     Implying that from a state of (C, D) (the first player having played C and
     the second playing D) the player in question then played C.
 
-    The following counter object, implies that the player in question was in
+    The following counter object implies that the player in question was in
     state (C, D) for a total of 12 times, subsequently cooperating 4 times and
     defecting 8 times.
 
@@ -167,8 +167,8 @@ def compute_state_to_action_distribution(interactions):
     if not interactions:
         return None
 
-    distributions = [Counter([(state, outcome[j]) for state, outcome in zip(interactions,
-                                                              interactions[1:])])
+    distributions = [Counter([(state, outcome[j])
+                     for state, outcome in zip(interactions, interactions[1:])])
                      for j in range(2)]
     return distributions
 
@@ -180,13 +180,12 @@ def compute_normalised_state_to_action_distribution(interactions):
 
     ((C, D), C)
 
-    Implying that from a state of (C, D) (the first player having played C and
+    implying that from a state of (C, D) (the first player having played C and
     the second playing D) the player in question then played C.
 
     The following counter object, implies that the player in question was only
-    ever in
-    state (C, D), subsequently cooperating 1/3 of the time and
-    defecting 2/3 times.
+    ever in state (C, D), subsequently cooperating 1/3 of the time and defecting
+    2/3 times.
 
     Counter({((C, D), C): 0.333333, ((C, D), D): 0.66666667})
 
@@ -197,11 +196,11 @@ def compute_normalised_state_to_action_distribution(interactions):
         this file.
 
     Returns
-    ----------
+    -------
     normalised_state_to_C_distributions : List of Counter Object
         List of Counter objects where the keys are the states and actions and
-        the values the normalized counts.. The
-        first/second Counter corresponds to the first/second player.
+        the values the normalized counts. The first/second Counter corresponds
+        to the first/second player.
     """
     if not interactions:
         return None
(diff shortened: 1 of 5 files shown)


axelrod/interaction_utils.py
```diff
@@ -135,6 +135,71 @@ def compute_normalised_state_distribution(interactions):
     return normalized_count
 
 
+def compute_state_to_action_distribution(interactions):
+    """
+    Returns the count of each state to action
+    for a set of interactions.
```
meatballs: I think this needs a little more explanation as to what the function does. It wasn't clear to me until I looked at how it's being used in the result set. Perhaps an example would be useful?

axelrod/interaction_utils.py
```diff
@@ -135,6 +135,71 @@ def compute_normalised_state_distribution(interactions):
...
+                                                              interactions[1:])])
+                     for j in range(2)]
+    return distributions
+
+
+def compute_normalised_state_to_action_distribution(interactions):
+    """
+    Returns the normalised count of each state to action
+    for a set of interactions.
+
```
meatballs: Same here. A little more explanation of what this is doing might be in order.

axelrod/interaction_utils.py
drvinceknight: Yeah good shout. I'll clear that up and find the others. I think I got lazy :)

drvinceknight:
Let me know if you think it needs more than those last two commits :) Just jumping in the plane, will pick this up later :D

axelrod/interaction_utils.py
```diff
@@ -135,6 +135,94 @@ def compute_normalised_state_distribution(interactions):
...
+    """
+    Returns a list (for each player) of counts of each state to action pair
+    for a set of interactions. A state to action pair is of the form:
+
+    ((C, D), C)
+
+    Implying that from a state of (C, D) (the first player having played C and
+    the second playing D) the player in question then played C.
+
+    The following counter object, implies that the player in question was in
```
marcharper: Pedantic: "The following Counter object implies that"

axelrod/interaction_utils.py
```diff
@@ -135,6 +135,94 @@ def compute_normalised_state_distribution(interactions):
...
+    state_to_C_distributions : List of Counter Object
+        List of Counter objects where the keys are the states and actions and
+        the values the counts. The
+        first/second Counter corresponds to the first/second player.
+    """
+    if not interactions:
+        return None
+
+    distributions = [Counter([(state, outcome[j]) for state, outcome in zip(interactions,
+                                                              interactions[1:])])
```
marcharper: Can we make the spacing here more aesthetically pleasing?

axelrod/interaction_utils.py
```diff
@@ -135,6 +135,94 @@ def compute_normalised_state_distribution(interactions):
...
+
+
+def compute_normalised_state_to_action_distribution(interactions):
+    """
+    Returns a list (for each player) of normalised counts of each state to action
+    pair for a set of interactions. A state to action pair is of the form:
+
+    ((C, D), C)
+
+    Implying that from a state of (C, D) (the first player having played C and
```
marcharper: Pedantic: lowercase "impying"

axelrod/interaction_utils.py
```diff
@@ -135,6 +135,94 @@ def compute_normalised_state_distribution(interactions):
...
+    the second playing D) the player in question then played C.
+
+    The following counter object, implies that the player in question was only
```
marcharper: formatting

axelrod/interaction_utils.py
```diff
@@ -135,6 +135,94 @@ def compute_normalised_state_distribution(interactions):
...
+    ----------
+    interactions : list of tuples
+        A list containing the interactions of the match as shown at the top of
+        this file.
+
+    Returns
+    ----------
+    normalised_state_to_C_distributions : List of Counter Object
+        List of Counter objects where the keys are the states and actions and
+        the values the normalized counts.. The
```
marcharper: double period

axelrod/interaction_utils.py
```diff
@@ -135,6 +135,94 @@ def compute_normalised_state_distribution(interactions):
...
+        first/second Counter corresponds to the first/second player.
+    """
+    if not interactions:
+        return None
+
+    distribution = compute_state_to_action_distribution(interactions)
+    normalized_distribution = []
+    for player in range(2):
+        counter = {}
+        for state in [('C', 'C'), ('C', 'D'), ('D', 'C'), ('D', 'D')]:
```
marcharper: `C`, `D` as in `Actions.C`, `Actions.D` rather than `'C'` and `'D'`

axelrod/result_set.py
```diff
@@ -358,6 +358,33 @@ def _build_normalised_state_distribution(self):
             norm.append(counters)
         return norm
 
+    @update_progress_bar
+    def _build_normalised_state_to_action_distribution(self):
+        """
+        Returns
+        ----------
+
```
marcharper: return var name / spacing

axelrod/result_set.py
```diff
@@ -358,6 +358,33 @@ def _build_normalised_state_distribution(self):
...
+            action.
+        """
+        norm = []
+        for player in self.state_to_action_distribution:
+            counters = []
+            for counter in player:
+                norm_counter = Counter()
+                for state in [(C, C), (C, D), (D, C), (D, D)]:
+                    total = counter[(state, C)] + counter[(state, D)]
+                    if total > 0:
```
marcharper: `if total == 0` then we add nothing to the list? just checking that is the intention rather than None or an empty dict.

axelrod/tests/unit/test_interaction_utils.py
```diff
@@ -22,11 +22,27 @@ class TestMatch(unittest.TestCase):
                           Counter({('D', 'C'): 2}),
                           Counter({('C', 'C'): 1, ('C', 'D'): 1}),
                           None]
+    state_to_action_distribution = [[Counter({(('C', 'D'), 'D'): 1}),
```
marcharper: Can we use `C` and `D` instead of `'C'` and `'D'`?

axelrod/tests/unit/test_interaction_utils.py
```diff
@@ -63,6 +79,37 @@ def test_compute_normalised_state_distribution(self):
         for inter, dist in zip(self.interactions, self.normalised_state_distribution):
             self.assertEqual(dist, iu.compute_normalised_state_distribution(inter))
 
+    def test_compute_state_to_action_distribution(self):
+        for inter, dist in zip(self.interactions,
+                               self.state_to_action_distribution):
+            self.assertEqual(dist,
+                             iu.compute_state_to_action_distribution(inter))
+        inter = [(C, D), (D, C), (C, D), (D, C), (D, D), (C, C), (C, D)]
```
marcharper: mixing `C` and `'C'`

marcharper:
Looks fine overall, I added some minor comments about docstrings and `C` versus `'C'`

axelrod/result_set.py
drvinceknight: If `total == 0` we still add the empty `norm_counter` so the behaviour is consistant :)

The test cases catch this occurrence :)

drvinceknight:
> Looks fine overall, I added some minor comments about docstrings and C versus 'C'

I think I've caught all of them and answered your question above, happy to further improve it though if you think of anything else :) :+1:

PR #871, opened 2017-02-20
Add type hints

janga1997:
#808
Only one error I am not able to resolve.
`axelrod_first.py:442: error: "type" has no attribute "memory_depth"`

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -3,9 +3,10 @@ from functools import wraps
 import random
 import copy
 
-from axelrod.actions import Actions, flip_action
+from axelrod.actions import Actions, flip_action, Action
 from .game import DefaultGame
 
+from typing import Dict, Any
 
 C, D = Actions.C, Actions.D
 
@@ -89,7 +90,7 @@ class Player(object):
     """
 
     name = "Player"
-    classifier = {}
+    classifier = {} # type: Dict[str, Any]
     default_classifier = {
         'stochastic': False,
         'memory_depth': float('inf'),
--- a/axelrod/strategies/axelrod_first.py
+++ b/axelrod/strategies/axelrod_first.py
@@ -439,7 +439,7 @@ class Tullock(Player):
         """
         super().__init__()
         self._rounds_to_cooperate = rounds_to_cooperate
-        self.__class__.memory_depth = rounds_to_cooperate
+        self.memory_depth = rounds_to_cooperate
 
     def strategy(self, opponent: Player) -> Action:
         rounds = self._rounds_to_cooperate
--- a/axelrod/strategies/axelrod_first.py
+++ b/axelrod/strategies/axelrod_first.py
@@ -4,10 +4,12 @@ Additional strategies from Axelrod's first tournament.
 
 import random
 
-from axelrod.actions import Actions, flip_action
+from axelrod.actions import Actions, flip_action, Action
 from axelrod.player import Player
 from axelrod.random_ import random_choice
 from.memoryone import MemoryOnePlayer
+
+from typing import List, Dict, Tuple
 
 C, D = Actions.C, Actions.D
 
@@ -35,7 +37,7 @@ class Davis(Player):
         'manipulates_state': False
     }
 
-    def __init__(self, rounds_to_cooperate=10):
+    def __init__(self, rounds_to_cooperate: int =10) -> None:
         """
         Parameters
         ----------
@@ -45,7 +47,7 @@ class Davis(Player):
         super().__init__()
         self._rounds_to_cooperate = rounds_to_cooperate
 
-    def strategy(self, opponent):
+    def strategy(self, opponent: Player) -> Action:
         """Begins by playing C, then plays D for the remaining rounds if the
         opponent ever plays D."""
         if len(self.history) < self._rounds_to_cooperate:
@@ -77,7 +79,7 @@ class RevisedDowning(Player):
         'manipulates_state': False
     }
 
-    def __init__(self, revised=True):
+    def __init__(self, revised: bool =True) -> None:
         super().__init__()
         self.revised = revised
         self.good = 1.0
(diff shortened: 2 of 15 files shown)


axelrod/player.py
```diff
@@ -3,9 +3,10 @@
 import random
 import copy
 
-from axelrod.actions import Actions, flip_action
+from axelrod.actions import Actions, flip_action, Action
 from .game import DefaultGame
 
+from typing import Dict, Any
```
marcharper: pedantic: alphabetize imports, move typing import above to the section with the standard library imports (this is in PEP8 IIRC)

axelrod/strategies/axelrod_first.py
```diff
@@ -77,7 +79,7 @@ class RevisedDowning(Player):
         'manipulates_state': False
     }
 
-    def __init__(self, revised=True):
+    def __init__(self, revised: bool =True) -> None:
```
marcharper: space after `=`?

axelrod/strategies/gobymajority.py
```diff
@@ -19,7 +19,7 @@ class GoByMajority(Player):
     default this is 0)
     """
 
-    name = 'Go By Marjority'
+    name = 'Go By Majority'
```
marcharper: jiminy cricket! are we not catching this with a test?

axelrod/strategies/memoryone.py
```diff
@@ -1,7 +1,11 @@
-from axelrod.actions import Actions
+from axelrod.actions import Actions, Action
 from axelrod.player import Player
 from axelrod.random_ import random_choice
 
+from typing import List, Tuple, Union
+
+type_four_vector = Union[List[float], Tuple[float,float,float,float]]
```
marcharper: pedantic: spaces after commas

axelrod/strategies/titfortat.py
```diff
@@ -487,7 +487,7 @@ class AdaptiveTitForTat(Player):
 
     Algorithm
 
-    if (opponent played C in the last cycle) then
+    if (opponent: Player  -> Actiplayed C in the last cycle) then
```
marcharper: something is weird with this line

axelrod/player.py
janga1997: I thought the typing imports should be kept separate from the library imports?

janga1997:
@marcharper the only workaround I could find for the type error of `self.classifier['memory_depth']` is the declare the type of `classifier` as `Dict[str, Any]`, since the values in `classifier` are too varied to pin down to specific types.

janga1997:
I didn't see the purpose of `__class__` in `self.__class__.memory_depth = rounds_to_cooperate` in `Tullock` in `axelrod_first`.
Changing it did not trip the unit test, but made mypy happy.

janga1997:
Only one error in `sequence_player`.
`sequence_player.py:91: error: Too many arguments for "__init__" of "ThueMorse"`
I can't understand why it raises an error, since the it has the exact number of args the parent class
`SequencePlayer` requires.

axelrod/strategies/gobymajority.py
drvinceknight: Just taken a look: we kind of are catching it. The init method of the strategy renames the instance (which is tested):

```
 53         self.name = (                                                                                                                             
 54             'Go By Majority' + (self.memory > 0) * (": %i" % self.memory))                                                                        
 55         if self.soft:                                                                                                                             
 56             self.name = "Soft " + self.name                                                                                                       
 57         else:                                                                                                                                     
 58             self.name = "Hard " + self.name                                                                                                       
 59                                                  
```

 but yeah we can make that better:

```
 53         self.name = (                                                                                                                             
 54           self.name + (self.memory > 0) * (": %i" % self.memory))                                                                        
 55         if self.soft:                                                                                                                             
 56             self.name = "Soft " + self.name                                                                                                       
 57         else:                                                                                                                                     
 58             self.name = "Hard " + self.name                                                                                                       
 59                                                  
```

I think that would have tripped a test...

janga1997:
Can someone have a look at this PR?
The rest of the files have too obscure variables for me to do in a sprint. I may have to tackle them file by file later.

axelrod/player.py
marcharper: It's not a big deal but usually standard library imports come before others. I guess we haven't discussed our official preference for typing imports.

axelrod/strategies/axelrod_first.py
```diff
@@ -310,15 +312,16 @@ def __init__(self):
         super().__init__()
 
     @staticmethod
-    def score_history(my_history, opponent_history, score_map):
+    def score_history(my_history: List[Action], opponent_history: List[Action], score_map: Dict[Tuple[Action, Action], int]) -> int:
```
marcharper: Can you split into multiple lines:
```
def score_history(my_history: List[Action], opponent_history: List[Action],
                             score_map: Dict[Tuple[Action, Action], int]) -> int:
```

type_tests.sh
```diff
@@ -38,3 +38,8 @@ mypy --ignore-missing-imports --follow-imports skip axelrod/strategies/oncebitte
 mypy --ignore-missing-imports --follow-imports skip axelrod/strategies/prober.py
 mypy --ignore-missing-imports --follow-imports skip axelrod/strategies/punisher.py
 mypy --ignore-missing-imports --follow-imports skip axelrod/strategies/rand.py
+mypy --ignore-missing-imports --follow-imports skip axelrod/strategies/memoryone.py
```
marcharper: Let's alphabetize these

marcharper:
Can you check this out:

```
sh ./type_tests.sh

mypy: can't read file 'axelrod/strategies/worseandworse.py': No such file or directory
```

Should be "worse_and_worse.py". I'm not seeing the other error that you mentioned.

axelrod/strategies/axelrod_first.py
janga1997: As for the imports, I guess I will have to make a change to all the files which have typehints.
Do you want me to tackle all of them, or just in any PRs from now?

drvinceknight:
This looks good to me now.

meatballs:
Default Value Syntax:

As @marcharper suggested, the guidance at http://mypy.readthedocs.io/en/latest/basics.html suggests that there should be spaces both before and after the '=' in arguments with default values. There are several instances here where one of those spaces is missing.

janga1997:
@meatballs Made the changes.

meatballs:
I'm happy with this one now but I'll leave it to @marcharper to decide if his requested changes are now ok.

PR #873, opened 2017-03-03
Stalker

janga1997:
#379


drvinceknight:
You need to import the `FinalTransformer`:

```
from axelrod.strategy_transformers import FinalTransformer
```

axelrod/strategies/stalker.py
```diff
@@ -41,38 +43,41 @@ class Stalker(Player):
...
+    def strategy(self, opponent: Player) -> Action:
+
+        if len(self.history) == 0:
+            return C
+
         self.score_last_round(opponent)
 
-        current_average_score = sum(scores) / len(scores)
+        current_average_score = sum(self.scores) / len(self.scores)
+        very_good_score = self.R
```
marcharper: I think moving these three lines to `__init__` is better than saving R, P, S, and T to `self` since they won't change and this calculation will only be done once. It's logically ok as is.

axelrod/strategies/stalker.py
```diff
@@ -41,38 +43,41 @@ class Stalker(Player):
...
-        # Update scores from the last play
+    def strategy(self, opponent: Player) -> Action:
+
+        if len(self.history) == 0:
+            return C
+
         self.score_last_round(opponent)
 
-        current_average_score = sum(scores) / len(scores)
+        current_average_score = sum(self.scores) / len(self.scores)
```
marcharper: It's better to keep a cumulative score total then to save the scores for each round unless you need the individual scores. Summing the list each time is O(n^2) in the length of the match, keeping a cumulative sum is O(n).

axelrod/tests/unit/test_stalker.py
```diff
@@ -0,0 +1,53 @@
+"""Tests for the SelfSteem strategy."""
```
marcharper: Stalker

axelrod/strategies/stalker.py
marcharper: There is also a special [method](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/player.py#L125) of the Player class intended for this situation.

axelrod/strategies/stalker.py
```diff
@@ -1,38 +1,40 @@
...
-    If f > 0.95, 'ego' of the algorithm is inflated; always defects.
-    If 0.95 > abs(f) > 0.3, rational behavior; follows TitForTat algortithm.
-    If 0.3 > f > -0.3; random behavior.
-    If f < -0.95, algorithm is at rock bottom; always cooperates.
-
-    Futhermore, the algorithm implements a retaliation policy, if the opponent
-    defects; the sin curve is shifted. But due to lack of further information,
-    this implementation does not include a sin phase change.
+    This is a strategy which is only influenced by the score.
+    Its behavior is based in three scores:
```
drvinceknight: `on three values:`

marcharper:
Looks good to me, can you check on the merge conflict?

janga1997:
Where?
I don't see a merge conflict...

axelrod/tests/unit/test_stalker.py
```diff
@@ -0,0 +1,53 @@
...
+        self.responses_test([D], [C] * 7 + [C] * 5, [C] * 7 + [D] * 5)
+
+        # current_average_score < 1
+        self.responses_test([D], [D] * 7 + [C] * 5, [D] * 7 + [D] * 5, seed = 6)
+        self.responses_test([C], [D] * 7 + [C] * 5, [D] * 7 + [D] * 5, seed = 7)
+
+        # defect in last round
+        self.responses_test([C, D], [C] * 198, [C] * 198, length=200)
+
+    def test_reset(self):
```
drvinceknight: Can you call this something different: `test_reset_current_score`. That way it will still run the base `test_reset` (which checks that the history is reset).

marcharper:
Hmm it says "This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. " -- strange that we can merge but not rebase. I guess we can squash the commits.

axelrod/tests/unit/test_stalker.py
marcharper: > Can you call this something different: test_reset_current_score. That way it will still run the base test_reset (which checks that the history is reset).

That test is called `test_reset_history` [now](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_player.py#L210) and is always run. There isn't a `test_history` anymore to override. (Relatively recent change IIRC.)

marcharper: In fact, that test also checks that internal attributes are reset, though an explicit test is probably still a good idea.

drvinceknight: Awesome. :+1: Ignore this @janga1997

Issue #874, opened 2017-03-03
Document passing a match length attribute to `responses_test`


drvinceknight:
This is currently not documented at http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html

but it is implemented: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/unit/test_backstabber.py#L36 :+1:

marcharper:
I thought about this a while ago. I think we should try to move to testing direct matches whenever possible rather than enforcing potentially impossible histories. In this case we could test that the last action is a defection by simply running a short match and checking the last action.

Mock histories are ok in some cases, especially when finding just the right random seed would be very difficult, but as all the build warnings show now we're overusing this trick.

drvinceknight:
That sounds good to me. Would involve going through all the tests and structuring the docs a bit differently (just reordering what's there).

drvinceknight:
I was thinking of going through and sorting out those warnings at some point so I can change over to matches too. It'll probably be a slow burn but if no one else wants to do it I'm happy to assign this to myself...

drvinceknight:
Although thinking about it, we should do this carefully and from a review point of view it's probably easier to do it small set of strategies at a time. Perhaps I can do a few and then we can open a nice descriptive issue.

drvinceknight:
@marcharper I've just taken a quick look but the docs for `TestMatch.versus_test` are incorrect. http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html currently reads:

```
class TestTF2TvsBully(TestMatch):
    """Test Tit for Two Tats vs Bully"""
    def test_rounds(self):
        outcomes = [[C, D], [C, D], [D, D], [D, C], [C, C], [C, D], [C, D], [D, D]]
        self.versus_test(axelrod.TitFor2Tats, axelrod.Bully, outcomes)
```

However the `versus_test` method of that class is:

```
>>370     def versus_test(self, player1, player2, expected_actions1,                                                                                  
  371                     expected_actions2, noise=None, seed=None):                                                                                  
  372         """Tests a sequence of outcomes for two given players."""                                                                               
  373         if len(expected_actions1) != len(expected_actions2):                                                                                    
  374             raise ValueError("Mismatched History lengths.")                                                                                     
  375         if seed:                                                                                                                                
[...]

PR #875, opened 2017-03-05
Add a versus_test method to TestPlayer

drvinceknight:
Addresses #874 with a tweaked testing framework.

The `versus_test` method removes the option to set player histories and can be used to test strategies (and player attributes during matches) by creating a match between the player in question and an opponent.

The opponent can either be an actual player from the library or is defined as a cycle by a passed sequence of actions (there are examples of this in the docs and `test_titfortat.py`).

I've refactored all the tests in `test_titfortat.py` (which is now one module that raises no warnings).

When/if we're happy with this as a suggestion, we should open an issue about refactoring all the tests to use this approach. (This would be a big/important piece of work so we should ask for small contributions at a time to make sure we don't miss anything with the reviews).

Once that refactor is complete we could potentially remove unused methods in the `TestPlayer` class.

### DIFF
--- a/axelrod/strategies/finite_state_machines.py
+++ b/axelrod/strategies/finite_state_machines.py
@@ -35,8 +35,8 @@ class SimpleFSM(object):
     def __eq__(self, other):
         """Equality of two FSMs"""
         check = True
-        check += self.state == other.state
-        check += self.state_transitions == other.state_transitions
+        for attr in ["state", "state_transitions"]:
+            check = check and getattr(self, attr) == getattr(other, attr)
         return check
 
 
--- a/axelrod/strategies/finite_state_machines.py
+++ b/axelrod/strategies/finite_state_machines.py
@@ -32,6 +32,13 @@ class SimpleFSM(object):
         self.state = next_state
         return next_action
 
+    def __eq__(self, other):
+        """Equality of two FSMs"""
+        check = True
+        check += self.state == other.state
+        check += self.state_transitions == other.state_transitions
+        return check
+
 
 class FSMPlayer(Player):
     """Abstract base class for finite state machine players."""
@@ -57,6 +64,7 @@ class FSMPlayer(Player):
             initial_action = C
         super().__init__()
         self.initial_state = initial_state
+        self.state = initial_state
         self.initial_action = initial_action
         self.fsm = SimpleFSM(transitions, initial_state)
 
@@ -73,6 +81,7 @@ class FSMPlayer(Player):
     def reset(self):
         super().reset()
         self.fsm.state = self.initial_state
+        self.state = self.initial_state
 
 
 class Fortress3(FSMPlayer):
--- a/axelrod/tests/unit/test_grumpy.py
+++ b/axelrod/tests/unit/test_grumpy.py
@@ -45,7 +45,7 @@ class TestGrumpy(TestPlayer):
                             init_kwargs={"grumpy_threshold": 3,
                                          "nice_threshold": 0})
 
-    def test_reset(self):
+    def test_reset_state(self):
         P1 = axelrod.Grumpy(starting_state='Grumpy')
         P1.state = 'Nice'
         P1.reset()
--- a/axelrod/tests/unit/test_grumpy.py
+++ b/axelrod/tests/unit/test_grumpy.py
@@ -45,7 +45,7 @@ class TestGrumpy(TestPlayer):
                             init_kwargs={"grumpy_threshold": 3,
                                          "nice_threshold": 0})
 
-    def test_reset_state(self):
+    def test_reset_state_with_non_default_init(self):
         P1 = axelrod.Grumpy(starting_state='Grumpy')
         P1.state = 'Nice'
         P1.reset()
--- a/axelrod/tests/unit/test_human.py
+++ b/axelrod/tests/unit/test_human.py
@@ -91,3 +91,7 @@ class TestHumanClass(TestPlayer):
         expected_action = C
         actual_action = human.strategy(Player(), lambda: C)
         self.assertEqual(actual_action, expected_action)
+
+    def test_reset_history_and_attributes(self):
+        """Overwrite the reset method for this strategy."""
+        pass
--- a/axelrod/tests/unit/test_meta.py
+++ b/axelrod/tests/unit/test_meta.py
@@ -46,16 +46,11 @@ class TestMetaPlayer(TestPlayer):
                              msg="%s - Behaviour: %s != Expected Behaviour: %s" %
                                  (key, player.classifier[key], classifier[key]))
 
-    def attribute_test(self, player, clone):
+    def attribute_equality_test(self, player, clone):
         """Overwriting this specific method to check team."""
-        opponent = axelrod.Cooperator()
-        player.play(opponent)
-        player.play(opponent)
-        player.play(opponent)
-        player.reset()
-        for i, p in enumerate(player.team):
-            self.assertEqual(len(p.history), 0)
-
+        for p1, p2 in zip(player.team, clone.team):
+            self.assertEqual(len(p1.history), 0)
+            self.assertEqual(len(p2.history), 0)
 
         team_player_names = [p.__repr__() for p in player.team]
         team_clone_names = [p.__repr__() for p in clone.team]
--- a/axelrod/tests/unit/test_meta.py
+++ b/axelrod/tests/unit/test_meta.py
@@ -46,16 +46,21 @@ class TestMetaPlayer(TestPlayer):
                              msg="%s - Behaviour: %s != Expected Behaviour: %s" %
                                  (key, player.classifier[key], classifier[key]))
 
-    def test_reset(self):
-        p1 = self.player()
-        p2 = axelrod.Cooperator()
-        p1.play(p2)
-        p1.play(p2)
-        p1.play(p2)
-        p1.reset()
-        for player in p1.team:
-            self.assertEqual(len(player.history), 0)
+    def test_reset_history_and_attributes(self):
+        player = self.player()
+        clone = player.clone()
+        opponent = axelrod.Cooperator()
+        player.play(opponent)
+        player.play(opponent)
+        player.play(opponent)
+        player.reset()
+        for i, p in enumerate(player.team):
+            self.assertEqual(len(p.history), 0)
 
+
+        team_player_names = [p.__repr__() for p in player.team]
+        team_clone_names = [p.__repr__() for p in clone.team]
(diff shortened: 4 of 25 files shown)


marcharper:
I like most of this. Some quick thoughts:

* Having opponent be a player or a sequence is a bit confusing -- using a MockPlayer would be better IMO, that's its purpose.
* We might be able to get away with only having init_kwargs, or it may be better to tackle that issue before we rewrite a bunch of tests.

drvinceknight:
> Having opponent be a player or a sequence is a bit confusing -- using a MockPlayer would be better IMO, that's its purpose.

Yeah I wasn't completely sure about having two input possibilities. I'll make this tweak.

> We might be able to get away with only having init_kwargs, or it may be better to tackle that issue before we rewrite a bunch of tests.

Can you expand on this a bit more? I'm not too sure I follow which issue you mean.

drvinceknight:
I've also just pushed a refactor of the `Random` strategy. I've used `init_kwargs` there.

marcharper:
Issue #706 --  maybe we could make all the Player args into kwargs, which I think would help with #706 as well.

drvinceknight:
> Issue #706 -- maybe we could make all the Player args into kwargs, which I think would help with #706 as well.

Sure thing, I'll remove the `init_args` option for now so that it will push us in that direction.

marcharper:
It's ok to use `test_reset`.

drvinceknight:
> It's ok to use test_reset.

Did you mean to type that here or in the other issue? Surely that's overwriting the `test_reset` method of the parent class?

axelrod/tests/unit/test_grumpy.py
```diff
@@ -45,7 +45,7 @@ def test_strategy(self):
                             init_kwargs={"grumpy_threshold": 3,
                                          "nice_threshold": 0})
 
-    def test_reset(self):
+    def test_reset_state(self):
```
marcharper: I think this test is unnecessary.

axelrod/tests/unit/test_player.py
```diff
@@ -288,8 +288,61 @@ def second_play_test(self, rCC, rCD, rDC, rDD, seed=None):
...
+
+        Parameters:
+        -----------
+
+        opponent: Player or list
+            An instance of a player OR a sequence of actions. If a sequence of
+            actions is passed, a Mock Player is created that cycles over that
+            sequence.
+        expected_outcomes: List
+            the expected outcomes of the match (list of tuples of actions).
```
marcharper: T

axelrod/tests/unit/test_player.py
```diff
@@ -288,8 +288,61 @@ def second_play_test(self, rCC, rCD, rDC, rDD, seed=None):
...
+            The random seed to be used
+        length: int
+            The length of the game. If `opponent` is a sequence of actions then
+            the length is taken to be the length of the sequence.
+        match_attributes: dict
+            The match attributes to be passed to the players.  For example,
+            `{length:-1}` implies that the players do not know the length of the
+            match.
+        attrs: dict
+            dictionary of internal attributes to check at the end of all plays
```
marcharper: D

axelrod/tests/unit/test_rand.py
```diff
@@ -24,7 +24,24 @@ def test_strategy(self):
         """Test that strategy is randomly picked (not affected by history)."""
         self.first_play_test(C, seed=1)
         self.first_play_test(D, seed=2)
-        self.responses_test([C], [C, D, C], [C, C, D], seed=1)
+
+        opponent = axelrod.MockPlayer()
+        outcomes = [(C, C), (D, C), (D, C)]
```
marcharper: This tests the histories above but not the next move, so it's not exactly equivalent. Doesn't matter in this case of course.

axelrod/tests/unit/test_titfortat.py
```diff
@@ -30,12 +30,42 @@ class TestTitForTat(TestPlayer):
...
-        # Starts by cooperating.
         self.first_play_test(C)
-        # Repeats last action of opponent history.
-        self.second_play_test(C, D, C, D)
-        self.responses_test([C], [C] * 4, [C, C, C, C])
-        self.responses_test([D], [C] * 5, [C, C, C, C, D])
+        self.second_play_test(rCC=C, rCD=D, rDC=C, rDD=D)
+
+        # Play against opponents
+        outcomes = [(C, C), (C, D), (D, C), (C, D)]
```
marcharper: Same comment as before re: not checking the last move anymore.

drvinceknight:
@marcharper the particular test in `test_grumpy` was not covered by the default case as it's checking a non default value **but** your comment got me to find a bug in our default test case. It was in effect not testing anything (it was simply comparing the attributes of the clone to the attributes of the clone and not the reset player).

I've fixed this (cff2e53 and a refactor here: b7ce33a) and ended up having to fix some strategies that were then failing. 

This is worth a slow and thorough review in case I've missed anything so no rush :)

axelrod/tests/unit/test_human.py
```diff
@@ -91,3 +91,7 @@ def test_strategy(self):
         expected_action = C
         actual_action = human.strategy(Player(), lambda: C)
         self.assertEqual(actual_action, expected_action)
+
+    def test_reset_history_and_attributes(self):
```
marcharper: Is this necessary? I though there was a catch in TestPlayer for this strategy. If you've removed that then ignore.

axelrod/tests/unit/test_player.py
```diff
@@ -207,37 +208,48 @@ def test_match_attributes(self):
...
 
     def test_reset_clone(self):
         """Make sure history resetting with cloning works correctly, regardless
         if self.test_reset() is overwritten."""
         player = self.player()
         clone = player.clone()
-        for k, v in clone.__dict__.items():
-            if isinstance(v, np.ndarray):
-                self.assertTrue(np.array_equal(v, getattr(clone, k)))
+        for attribute, value in player.__dict__.items():
```
marcharper: Would `player == clone` work here now?

axelrod/tests/unit/test_meta.py
```diff
@@ -46,16 +46,21 @@ def classifier_test(self, expected_class_classifier=None):
...
+    def test_reset_history_and_attributes(self):
+        player = self.player()
+        clone = player.clone()
+        opponent = axelrod.Cooperator()
+        player.play(opponent)
+        player.play(opponent)
+        player.play(opponent)
+        player.reset()
+        for i, p in enumerate(player.team):
+            self.assertEqual(len(p.history), 0)
```
marcharper: Can we use `player == clone` here?

axelrod/tests/unit/test_player.py
```diff
@@ -207,37 +208,48 @@ def test_match_attributes(self):
...
-                self.assertTrue(np.array_equal(v, getattr(p_clone, k)))
+        player = self.player()
+        clone = player.clone()
+        opponent = axelrod.Random()
+
+        for seed in range(10):
+            axelrod.seed(seed)
+            player.play(opponent)
+
+        player.reset()
```
marcharper: `player == clone`?

marcharper:
That's interesting. I remember catching a subtle bug with MetaPlayer with that test so that's surprising that it worked at all! I like the new additions, looks like maybe we could remove some redundant code by using `player == clone` in a few places.

We might consider updating MetaPlayer's clone method to clone all the team members, or just take a closer look at it.

drvinceknight:
> That's interesting. I remember catching a subtle bug with MetaPlayer with that test so that's surprising that it worked at all! I like the new additions, looks like maybe we could remove some redundant code by using player == clone in a few places.

You might be able to tell from the commit history that I implemented a player `__eq__` but felt that that was stepping a bit too far for this PR so I reverted it (we should squash these commits for with the merge). Without the `__eq__` method `player == clone` would just look at instance equality (so be equivalent to `player is clone`). That could be added at another date.

> 
> We might consider updating MetaPlayer's clone method to clone all the team members, or just take a closer look at it.

Perhaps for another PR?

drvinceknight:
I've just rebased on to `master` and caught a misclassification of `stalker` :)

I'm not sure if this means I'm going to miss any comments you wrote but I think all of them were suggesting `player == clone` which (unless I'm missing something) won't work because of the fact that we haven't got a `__eq__` method implemented but that could be another issue to raise/discuss?

marcharper:
OK. `__eq__` was still there when I was commenting. I would vote to add it back in, but it can be another PR.

axelrod/strategies/finite_state_machines.py
```diff
@@ -57,6 +64,7 @@ def __init__(self, transitions=None, initial_state=None,
             initial_action = C
         super().__init__()
         self.initial_state = initial_state
+        self.state = initial_state
```
marcharper: What's the purpose of this line?

marcharper:
Do we want to change `expected_outcomes` to `expected_actions` or something else more inline with the terminology elsewhere in the library?

drvinceknight:
> OK. __eq__ was still there when I was commenting. I would vote to add it back in, but it can be another PR.

Cool. Once this is in I'll open a couple of issues:

1. To refactor the tests;
2. To add an `__eq__` method in to the players and refactor some of the base tests;
3. Once 1. is complete to refactor the base tests to remove anything that is not needed anymore.

> Do we want to change expected_outcomes to expected_actions or something else more inline with the terminology elsewhere in the library?

Good call. I've pushed that.

axelrod/strategies/finite_state_machines.py
drvinceknight: Needed for the reset test. The cloned player did not have a `self.state` attribute if it had not played yet.

drvinceknight:
@marcharper  in bc398f2 I've gone a step further and removed `first_play_test` and `second_play_test` from the docs and `test_tit_for_tat, test_rand`. (I also added a couple more `versus_tests` in `test_tit_for_tat` to make sure all states were being tested.)

My thinking is that we can test those properties using matches and it moves us away from the potential history mistmatches that can still occur with `second_play_test`. 

Can always revert if we don't think it's a good idea.

marcharper:
I'd prefer to keep those tests -- the warnings for those are suppressed already. Or -- let's take that discussion to a new issue / PR?

drvinceknight:
> I'd prefer to keep those tests -- the warnings for those are suppressed already. Or -- let's take that discussion to a new issue / PR?

Fine by me, I've reverted the commit.

meatballs:
This one looks ready to roll to me. Anyone object?

drvinceknight:
> This one looks ready to roll to me. Anyone object?

Not from me :)

PR #876, opened 2017-03-07
Add Double Resurrection Strategy

souravsingh:
Adds a new strategy in Axelrod, adapted from Coopsim.

### DIFF
--- a/axelrod/strategies/resurrection.py
+++ b/axelrod/strategies/resurrection.py
@@ -34,14 +34,14 @@ class Resurrection(Player):
         else:
             return opponent.history[-1]
 
-class Resurrection(Player):
+class DoubleResurrection(Player):
     """
     A player starts by cooperating and defects if the number of rounds 
     played by the player is greater than five and the last five rounds are cooperations.
     If the last file rounds were defections, the player cooperates.
     
     Names:
-    - Resurrection: Name from CoopSim https://github.com/jecki/CoopSim     
+    - DoubleResurrection: Name from CoopSim https://github.com/jecki/CoopSim     
     """
 
     # These are various properties for the strategy
--- a/axelrod/strategies/resurrection.py
+++ b/axelrod/strategies/resurrection.py
@@ -38,7 +38,7 @@ class DoubleResurrection(Player):
     """
     A player starts by cooperating and defects if the number of rounds 
     played by the player is greater than five and the last five rounds are cooperations.
-    If the last file rounds were defections, the player cooperates.
+    If the last five rounds were defections, the player cooperates.
     
     Names:
     - DoubleResurrection: Name from CoopSim https://github.com/jecki/CoopSim     
@@ -56,11 +56,16 @@ class DoubleResurrection(Player):
     }
 
     def strategy(self, opponent: Player) -> Action:
-        if len(self.history) == 0:
+        if not self.history:
             return C
-        if len(self.history) >= 5 and self.history[-5:] == [C, C, C, C, C]:
-            return D
-        elif len(self.history) >= 5 and self.history[-5:] == [D, D, D, D, D]:
-            return C
-        else:
+        
+        is_history_sufficient = len(self.history) >= 5    
+        actions = {
+            (True, 'CCCCC'): 'D',
+            (True, 'DDDDD'): 'C'
+        }
+
+        try:
+            return actions[(is_history_sufficient, self.history[-5:])]
+        except KeyError:
             return opponent.history[-1]
--- a/axelrod/strategies/resurrection.py
+++ b/axelrod/strategies/resurrection.py
@@ -44,8 +44,7 @@ class DoubleResurrection(Player):
     - DoubleResurrection: Name from CoopSim https://github.com/jecki/CoopSim     
     """
 
-    # These are various properties for the strategy
-    name = 'Double Resurrection'
+    name = 'DoubleResurrection'
     classifier = {
         'memory_depth': 1,  
         'stochastic': False,
--- a/axelrod/strategies/resurrection.py
+++ b/axelrod/strategies/resurrection.py
@@ -34,3 +34,34 @@ class Resurrection(Player):
         else:
             return opponent.history[-1]
 
+class Resurrection(Player):
+    """
+    A player starts by cooperating and defects if the number of rounds 
+    played by the player is greater than five and the last five rounds are cooperations.
+    If the last file rounds were defections, the player cooperates.
+    
+    Names:
+    - Resurrection: Name from CoopSim https://github.com/jecki/CoopSim     
+    """
+
+    # These are various properties for the strategy
+    name = 'Double Resurrection'
+    classifier = {
+        'memory_depth': 1,  
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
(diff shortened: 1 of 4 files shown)


drvinceknight:
Tests are failing @souravsingh, also this is going to need tests for the strategy itself. :+1:

drvinceknight:
There are still a couple of steps missing, you need to add the strategy to the ones being imported by the library. Perhaps take a look through the steps here: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/index.html

Let us know if we can assist.

souravsingh:
Oh,I forgot about that step! Thanks for reminding @drvinceknight

souravsingh:
I have added a test for the Double Resurrection

drvinceknight:
The tests are failing @souravsingh.

souravsingh:
The test error is arising from types in Punisher, which is unrelated to DoubleResurrection.

drvinceknight:
I'm afraid that's not correct, there are errors in the unit test:

```
  s1, s2, action1, action2, str(player1))
........
FAIL: test_repr (axelrod.tests.unit.test_resurrection.TestDoubleResurrection)
Test that the representation is correct.
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tests/unit/test_player.py", line 187, in test_repr
    self.assertEqual(str(self.player()), self.name)
AssertionError: 'Resurrection' != 'DoubleResurrection'
- Resurrection
+ DoubleResurrection
? ++++++
FAIL: test_strategy (axelrod.tests.unit.test_resurrection.TestDoubleResurrection)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tests/unit/test_resurrection.py", line 57, in test_strategy
    self.responses_test([C], [D, C, C, D, D, D, D, D] , [C] * 8)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tests/unit/test_player.py", line 327, in responses_test
    opponent_history, seed=seed, attrs=attrs)
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tests/unit/test_player.py", line 143, in test_responses
    test_class.assertEqual(s1, response)
AssertionError: 'D' != 'C'
- D
+ C
Ran 2377 tests in 428.758s
FAILED (failures=2, skipped=1)
```

There's information about running the unit tests locally here: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/running_tests.html

So you should run:

```
$python -m unittest axelrod.tests.unit.test_resurrection
```

There is also an error in the typing for Punisher but that is not what travis is complaining about, I think it just needs an import. If you could add that here as well it would be appreciated (but don't worry if not).

souravsingh:
My apologies, I misread the Travis logs.

souravsingh:
Okay, The problem was with the import naming for tests. The tests now pass. Are there any more test cases that need to be added for DoubleResurrection?

axelrod/strategies/resurrection.py
```diff
@@ -34,3 +34,33 @@ def strategy(self, opponent: Player) -> Action:
         else:
             return opponent.history[-1]
 
+class DoubleResurrection(Player):
+    """
+    A player starts by cooperating and defects if the number of rounds 
+    played by the player is greater than five and the last five rounds are cooperations.
+    If the last file rounds were defections, the player cooperates.
```
meatballs: typo: five

axelrod/strategies/resurrection.py
```diff
@@ -34,3 +34,33 @@ def strategy(self, opponent: Player) -> Action:
...
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if len(self.history) == 0:
```
meatballs: this might be better as `if not self.history:`

axelrod/strategies/resurrection.py
```diff
@@ -34,3 +34,33 @@ def strategy(self, opponent: Player) -> Action:
...
+            return C
+        if len(self.history) >= 5 and self.history[-5:] == [C, C, C, C, C]:
+            return D
+        elif len(self.history) >= 5 and self.history[-5:] == [D, D, D, D, D]:
+            return C
+        else:
+            return opponent.history[-1]
```
meatballs: Another way to do this would be to use an 'actions' dictionary rather than conditionals. Feel free to ignore this suggestion, but I always prefers using dicts to contain logic whenever possible!
```
    if not self.history:
        return C

    is_history_sufficient = len(self.history) >= 5    
    actions = {
        (True, 'CCCCC'): 'D',
        (True, 'DDDDD'): 'C'
    }

    try:
        return actions[(is_history_sufficient, self.history[-5:])]
    except KeyError:
            return opponent.history[-1]
```

axelrod/strategies/resurrection.py
```diff
@@ -34,3 +34,38 @@ def strategy(self, opponent: Player) -> Action:
...
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if not self.history:
+            return C
+        
+        is_history_sufficient = len(self.history) >= 5    
+        actions = {
+            (True, 'CCCCC'): 'D',
+            (True, 'DDDDD'): 'C'
```
meatballs: Sorry, this is my fault. These shouldn't be strings The dict ought to be:
```
actions = {
    (True, CCCCC): D,
    (True, DDDDD): C
}
```
or perhaps, even better:
```
actions = {
    (True, C * 5): D,
    (True, D * 5): C
}
```
You could even use a variable in place of the hard coded 5 here and in line 62

meatballs: And I've forgotten that you can't use lists as keys in a dictionary. I do apologise for causing such a mess for you!

axelrod/strategies/resurrection.py
```diff
@@ -34,3 +34,38 @@ def strategy(self, opponent: Player) -> Action:
...
+        }
+
+        try:
+            return actions[(is_history_sufficient, self.history[-5:])]
```
meatballs: this would have to be `return actions[(is_history_sufficient, tuple(self.history[-5:]))]`

Feel free to revert this logic to your original using conditionals. It was a suggestion rather than a request!

axelrod/strategies/resurrection.py
souravsingh: @meatballs The logic of using a dict to check for defection or cooperation looked really elegant to me (I hate writing multiple conditionals). So, I made the changes accordingly. I will revert back to the original logic.

meatballs: I think so too, but I didn't check the code of my suggestion thoroughly enough.

I think it really just needs the list in `actions[(is_history_sufficient, self.history[-5:])]` converting to a tuple in order to work properly - but I'm going to leave it entirely up to you and stop interfering!

drvinceknight:
@souravsingh there'll now be a merge conflict with `master`. We've move all the `test_strategy.py` files to a different folder: they're now in `axelrod/tests/strategies/test_...`.

souravsingh:
I will be sending a PR again with the updates to the tests.

PR #878, opened 2017-03-09
Add proper exit code check to mypy script.

drvinceknight:
Have included an error in worse_and_worse so this should cause travis to fail.


drvinceknight:
Have moved the error to Actions.

drvinceknight:
This is ready to go in now. You can see on the commit history that 744bf44 failed properly on travis (here's the builds: https://travis-ci.org/Axelrod-Python/Axelrod/jobs/209291076). This should make things easier :)

Issue #884, opened 2017-03-09
Refactor all strategy tests


drvinceknight:
With the merge of #875 we have a new method for testing strategies called `versus_test` (http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html). The goal is for this to eventually replace the currently used `responses_test`.

This will take a slow refactor of all strategy tests. Ideally any PRs towards this issue should only do a small number of modules at a time to ensure that nothing is missed during the review process.

EDIT: Once all of these are complete, we can refactor the `TestPlayer` class to remove unused methods.

alajara:
Hi I would like to help with this. Its my first contribution, any guidance how to start, I will much appreciate it.

drvinceknight:
Hi @alajara, welcome!

We have introduced a new testing framework for strategies, you can read about this here: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html

So this issue is about rewriting the various tests that have already been written for strategies using the new framework.

Here are some PRs that I've written that do this:

- https://github.com/Axelrod-Python/Axelrod/pull/893/files
- https://github.com/Axelrod-Python/Axelrod/pull/909/files
- https://github.com/Axelrod-Python/Axelrod/pull/914/files

So the idea is to look at the files in `axelrod/tests/strategies/` and modify them as you can see in those PRs. 

I've assumed there that you're ok with the notion of automated testing but please let me know if you're not and I'd be happy to walk you through anything/everything :)

To help coordinate (I'm slowly working on these myself) perhaps once you're ready to go, comment on this issue when you start working on a particular test file (just so we don't work on the same one). Also, to help keep the reviews simple: let's stick to just one file per PR.

Let me know if anything I've said is unclear :) Looking forward to your first contribution!

alajara:
Hey Mr. Knight,


Thank you for the email response explaining how to get started. A little background on me, I recently got out of the military, 6 years , and this past year I have been attending college for Computer Systems and have been learning Python and Java.  I have been doing different types of projects to teach myself more and i have gotten to the point where I can read and write my own code. I still have a lot of work and things to learn, but I am motivated.  Again sir thank you for the opportunity to help, as soon as I have a chance today, I will get started.



Best

Armando, Lajara

________________________________

eric-s-s:
Hello Dr Knight,

[...]

PR #886, opened 2017-03-10
Refactor adaptive test for #884.

drvinceknight:


marcharper:
Looks fine though perhaps the first test is redundant - MockPlayer is a cooperator in that case?

drvinceknight:
Good shout. Have fixed. :+1: 

I was lazily converting but #884 is a good opportunity to catch things like this :+1:

PR #888, opened 2017-03-10
Add missing tests for match.py

meatballs:
As picked up from coverage report


axelrod.sublime-project
```diff
@@ -0,0 +1,34 @@
+{
```
drvinceknight: This got pulled in by mistake?

meatballs: Agggghhh!!! That's because the change to .gitignore is on another branch and I forgot about it

drvinceknight:
> f5af602

We can squash the commits on this PR to get rid of that file from the history. :+1:

PR #891, opened 2017-03-10
Coverage ignore instructions

meatballs:
Tell coverage to ignore lines that we know cannot be tested.

### DIFF
--- a/axelrod/tournament.py
+++ b/axelrod/tournament.py
@@ -101,7 +101,7 @@ class Tournament(object):
             self.progress_bar = tqdm.tqdm(total=len(self.match_generator),
                                           desc="Playing matches")
 
-        if on_windows and (filename is None):
+        if on_windows and (filename is None):  # pragma: no cover
             in_memory = True
 
         self.setup_output(filename, in_memory)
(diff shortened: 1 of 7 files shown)


axelrod/tournament.py
```diff
@@ -101,7 +101,7 @@ def play(self, build_results=True, filename=None,
             self.progress_bar = tqdm.tqdm(total=len(self.match_generator),
                                           desc="Playing matches")
 
-        if on_windows and (filename is None):
+        if on_windows and (filename is None):  # pragma: no cover
```
drvinceknight: Some of these, we can test for (on appveyor for example using skips based on the OS) but we're not running coveralls through them (we get our coverage report sent to coveralls from travis). Is this the best way to deal with it? (I think it might be but worth making sure).

meatballs: I think this is the best way. There is no function to make the skip pragma conditional, so we just need to pick whether coverage is called from travis or appveyor. I reckon we stick as we currently have it and use the pragma to skip windows stuff.

drvinceknight: Yeah I agree.

drvinceknight:
This will also help with #884 as it'll make it more visible if somehow we miss a test case when refactoring.

PR #892, opened 2017-03-10
Fix bug in fingerprint with filename.

drvinceknight:
This wasn't working correctly and had been missed as we were not
covering a line.

This should also improve the coverage score.


axelrod/tests/unit/test_tournament.py
```diff
@@ -89,7 +89,7 @@ def test_warning(self):
             repetitions=1)
         with warnings.catch_warnings(record=True) as w:
             # Check that a warning is raised if no results set is built and no
```
marcharper: "Check that a warning is raised if **no results set is built** and **no** filename **not** given".

drvinceknight: Fixed.

PR #893, opened 2017-03-10
Refactor tests for alternator for #884

drvinceknight:


axelrod/tests/strategies/test_alternator.py
```diff
@@ -23,7 +23,15 @@ class TestAlternator(TestPlayer):
...
-        for i in range(10):
-            self.responses_test([C, D] * i)
-        self.responses_test([C], [C, D, C, D], [C, C, C, C])
-        self.responses_test([D], [C, D, C, D, C], [C, C, C, C, C])
+
+        opponent = axelrod.MockPlayer()
+        actions = [(C, C), (D, C)] * 5
+        self.versus_test(opponent, expected_actions=actions)
+
+        opponent = axelrod.MockPlayer(D)
```
marcharper: Can we use `[D]`?. I know that `D='D'` is technically a sequence but this could cause problems if we change to an enum.

drvinceknight: Good call.

drvinceknight:
In two cases I changed the `opponent` to `Cooperator` and `Defector`.

PR #895, opened 2017-03-11
Use distutils version comparison.

drvinceknight:
I also reverted back to use if statements, which felt simpler in this (2
option) case.

### DIFF
--- a/axelrod/plot.py
+++ b/axelrod/plot.py
@@ -3,6 +3,7 @@ from numpy import arange, median, nan_to_num
 import tqdm
 import warnings
 
+from distutils.version import LooseVersion
 from typing import List, Union
 
 matplotlib_installed = True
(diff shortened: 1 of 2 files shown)


marcharper:
Looks fine. If it's not for master you can do whatever you want, so feel free to merge if you both agree.

meatballs:
No problem using distutils. I was just trying to clean up the existing function.

However, passing a mutable string as a default is an anti-pattern. I  think it's better to pass None as the default.

drvinceknight:
> However, passing a mutable string as a default is an anti-pattern. I think it's better to pass None as the default.

Cool. Have changed that.

marcharper:
Strings are immutable so I think it's ok to use a string as a default.

drvinceknight:
I suggest you pull this in to your branch @meatballs and possibly modify it further if you so wish, or just close it :) Any discussion can happen on the eventual PR to master 👍

PR #897, opened 2017-03-12
Fix bug in plot.py and add missing test

### LINKED ISSUE #894: Bug and Testing of plot.py
There is no unit test for the `default_cmap` function and, as a result, there's a bug in the case of matplotlib not being installed.

The coverage report is also showing several other lines in `plot.py` which are not covered by tests.

meatballs:
Fixes #894

### DIFF
--- a/axelrod/fingerprint.py
+++ b/axelrod/fingerprint.py
@@ -258,7 +258,7 @@ class AshlockFingerprint():
 
         if on_windows and (filename is None):  # pragma: no cover
             in_memory = True
-        elif filename is not None:  # pragma: no cover
+        elif filename is not None:
             outputfile = open(filename, 'w')
             filename = outputfile.name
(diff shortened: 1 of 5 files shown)


drvinceknight:
Just pushed another commit which ensures the colour bar is correctly placed too.

marcharper:
Looks fine to me. I don't want to squash all these commits... You might consider an interactive rebase `git rebase -i` onto upstream, removing the revert commit and the original commit by omitting them from the list. Then force push to the branch. If you are both fine with the unnecessary commits then just merge.

drvinceknight:
> If you are both fine with the unnecessary commits then just merge.

I'm agnostic, whatever Owen prefers to do.

meatballs:
I'd go with just merging with the reversal et al

PR #898, opened 2017-03-12
Add missing tests for graph.py

meatballs:


axelrod/tests/unit/test_graph.py
```diff
@@ -86,6 +86,87 @@ def test_init(self):
...
 
+    def test_out_dict(self):
+        # Undirected graph with vertices and unweighted edges
+        g = graph.Graph(edges=[[1, 2], [2, 3]])
+        expected_out_mapping = {
+            1: {2: None},
+            2: {1: None, 3: None},
+            3: {2: None}
+        }
+        for key in expected_out_mapping:
```
marcharper: `assertDictEqual` is better since this loop just tests that one dict is a subset of the other.

meatballs: Yep. I forgot all about that assertion. Will do...

drvinceknight: Using `assertEqual` work just fine too for comparing dictionaries.

drvinceknight: From https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertDictEqual

> Test that two dictionaries are equal. If not, an error message is constructed that shows the differences in the dictionaries. This method will be used by default to compare dictionaries in calls to assertEqual().

drvinceknight:
FYI `def out_dict` and `def in_dict` are still not covered (just mentioning it in case you wanted to grab them in this PR too: it can also wait for another one). :)

axelrod/tests/unit/test_graph.py
meatballs: Cut and paste job from the previous code - but that had to cope with comparing a dict with a defaultdict. Not necessary here!

meatballs:
> FYI def out_dict and def in_dict are still not covered (just mentioning it in case you wanted to grab them in this PR too: it can also wait for another one). :)

Whoops. That's exactly what these two tests were supposed to be for but I've got the assertion wrong. Hold on.....

meatballs:
It wasn't the loop that shouldn't have been there - it was the pasted content inside the loop that was wrong!

drvinceknight:
This also looks good to me.

PR #899, opened 2017-03-12
Add missing tests for filters.py

meatballs:


axelrod/tests/unit/test_filters.py
```diff
@@ -16,6 +16,14 @@ class TestStrategy(object):
             'makes_use_of': ['game', 'length']
         }
 
+    class TestStrategy2(object):
+        classifier = {
+            'stochastic': True,
+            'inspects_source': False,
+            'memory_depth': 'infinity',
```
marcharper: `float('inf')`?

marcharper: I'm not sure how the test passes as is.

meatballs: The scenario where a strategy has `memory_depth': 'infinity'` is precisely what this test is for. There's a line in the filter that catches it and converts it to a float and it's that line that comes up on the coverage report as not covered.

I think I had to add the line to cope with one or two strategies that have this entry and were causing the API to fall over.

drvinceknight: > I think I had to add the line to cope with one or two strategies that have this entry and were causing the API to fall over.

I don't think we have any strategies with that entry (or at least we shouldn't):

```python
>>> import axelrod as axl
>>> players = [s() for s in axl.strategies]
>>> mem_lengths = [p.classifier['memory_depth'] for p in players]
>>> set(mem_lengths)
 {0, 1, 2, 3, 4, 5, 6, 9, 10, 11, 12, 16, 20, 40, 200, inf}

```

Did you pergaps put this in so that the API could pass `infinity` to a filter? (Should that be done at the API level perhaps? So that the API converts `infinity` to a `float`?)

meatballs: possibly! I'll swap the test around

meatballs: nope. looking at the code, it's definitely there to cope with strategies having 'infinity' but I can't see any that do. Perhaps there were at some point.

We can either get rid of the line or add a test for it. Which do we prefer?

drvinceknight: > We can either get rid of the line or add a test for it. Which do we prefer?

I'd vote for getting rid of the line.

drvinceknight:
Looks good to me (subject to travis). I'm applying the `ready-to-merge` label as a signal that if @marcharper's happy he should merge, otherwise just remove it :)

drvinceknight:
This also looks good to me.

marcharper:
@meatballs Is "infinity" there for API purposes? If not then it makes sense to do so, since float('inf') isn't a thing in javascript / JSON. If not maybe it's a good idea.

drvinceknight:
Could we let the API convert `infinity` to `float(inf)`?

marcharper:
Yes, I suppose it could.

PR #900, opened 2017-03-13
Add missing tests for lookerup.py

meatballs:


drvinceknight:
FYI, line 157 is apparently still not covered: https://coveralls.io/builds/10562402/source?filename=axelrod%2Fstrategies%2Flookerup.py#L157

(Can wait for another PR of course.)

meatballs:
I missed line 157 - I thought it had gone to 100%, but I was looking at test_lookerup.py :-(

I'll sort it on this PR...

meatballs:
Looks like covering line 157 is more tricky than I thought...

meatballs:
Ah. I thought I could just pass a zero value for op_plays in the init_args, but there's line 122 which calcs is from the lookup table. This could take a while...

axelrod/tests/strategies/test_lookerup.py
```diff
@@ -28,6 +27,31 @@ class TestLookerUp(TestPlayer):
     expected_class_classifier = copy.copy(expected_classifier)
     expected_class_classifier['memory_depth'] = float('inf')
 
+    def test_create_lookup_table_keys(self):
+        expected = [
+            ('C', 'C', 'C'), ('C', 'C', 'D'), ('C', 'D', 'C'), ('C', 'D', 'D'),
```
marcharper: We should use `C, D = Actions.C, Actions.D` throughout

meatballs: I went with what was already there in the existing test, but yes, you're right. I'll fix the whole lot.

PR #903, opened 2017-03-14
Fix coverage of test_player.py

meatballs:
I've added another 'no cover' pragma to fix this one or else we end up writing tests for our tests, which didn't seem entirely sensible. Happy to discuss!

### DIFF
--- a/axelrod/tests/strategies/test_player.py
+++ b/axelrod/tests/strategies/test_player.py
@@ -439,7 +439,7 @@ class TestMatch(unittest.TestCase):
     def versus_test(self, player1, player2, expected_actions1,
                     expected_actions2, noise=None, seed=None):
         """Tests a sequence of outcomes for two given players."""
-        if len(expected_actions1) != len(expected_actions2):  # pragma: no cover
+        if len(expected_actions1) != len(expected_actions2):
             raise ValueError("Mismatched History lengths.")
         if seed:
             axelrod.seed(seed)
@@ -453,6 +453,15 @@ class TestMatch(unittest.TestCase):
             player1.play(player2)
             self.assertEqual(player1.history[i], outcome1)
             self.assertEqual(player2.history[i], outcome2)
+
+    def test_versus_with_incorrect_history_lengths(self):
+        """Test the error raised by versus_test if expected actions do not
+        match up"""
+        with self.assertRaises(ValueError):
+            p1, p2 = axelrod.Cooperator(), axelrod.Cooperator()
+            actions1 = [C, C]
+            actions2 = [C]
+            self.versus_test(p1, p2, actions1, actions2)
 
 
 def test_four_vector(test_class, expected_dictionary):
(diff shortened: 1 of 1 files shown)


drvinceknight:
I think in this particular case we should have a test (this is a function used in a test and not a test itself). It would be good to know that if someone were to write a test with different history lengths that the test would complain (if it didn't complain we could theoretically let pass some incorrect tests). Something like:

```
  455                                                                                                                                                 
+ 456     def test_versus_with_incorrect_history_lengths(self):                                                                                       
+ 457         """Test the error raised by versus_test if expected actions do not match                                                                
+ 458         up"""                                                                                                                                   
+ 459         with self.assertRaises(ValueError):                                                                                                     
+ 460             p1, p2 = axelrod.Cooperator(), axelrod.Cooperator()                                                                                 
+ 461             actions1 = [C, C]                                                                                                                   
  462             actions2 = [C]                                                                                                                      
  463             self.versus_test(p1, p2, actions1, actions2)    
```

(I've just checked and that does cover the lines properly)

I can't say I feel too strongly though, the hope being after #884 that we'll refactor this class. (But I think it's correct to test it if it's there.)

marcharper:
I thought about this today and I'm inclined to have one layer of tests for our generic test functions. That protect us against breaking the test function and then losing specific tests to entropy. It's sort of like how a cell has one gene that protects against cancer and you have to lose that gene and also something else to cause the cancer.

Let's not get crazy with tests for our tests for our tests.

drvinceknight:
Sure, in a way that one layer is the fact that the tests are used right?

That is not the case for this particular function because the whole point
of the error it raises is that it should not be used.

marcharper:
Yeah I guess the test would be two mock players with known histories -- pass those in and the test passes, pass in wrong histories and the test fails. Does that make sense or am I misunderstanding?

drvinceknight:
> Yeah I guess the test would be two mock players with known histories -- pass those in and the test passes, pass in wrong histories and the test fails. Does that make sense or am I misunderstanding?

I suggested a test in my first comment on this thread. 

I think in this case we just need to test that the error is raised. We have a variety of tests that pass so there's no need to test that in my opinion, the main thing we need to test here is that that error works as we expect (so that we know there are no tests with unbalanced histories that have crept in).

drvinceknight:
Could of course change the cooperators I suggest in my example to Mockplayer and also add a line where we pass the same history and check that it runs:

```
  455                                                                                                                                                 
+ 456     def test_versus_with_incorrect_history_lengths(self):                                                                                       
+ 457         """Test the error raised by versus_test if expected actions do not match                                                                
+ 458         up"""                                                                                                                                   
+ 459         with self.assertRaises(ValueError):                                                                                                     
+ 460             p1, p2 = axelrod.MockPlayer(), axelrod.MockPlayer()                                                                                 
+ 461             actions1 = [C, C]                                                                                                                   
  462             actions2 = [C]                                                                                                                      
  463             self.versus_test(p1, p2, actions1, actions2)  
  464         self.self.vesus_test(p1, p2, actions1, actions2 + [C])
```

marcharper:
I agree with this specific case and test. This example made me think about what would happen if`versus_test` were somehow changed to always pass. How would we know if all our tests usually pass? That's why I'm suggesting that we have tests for `versus_test` that we expect to fail.

drvinceknight:
Sure, it would do no harm.

How about for this PR we add my suggestion (or similar) and then open another issue for testing `versus_test`?

marcharper:
sounds good

PR #908, opened 2017-03-14
Refactor apavlov for #884.

drvinceknight:
Also: pushed coverage to 100%.

Have added more test cases and also removed a special case that is now
tested with the versus test.

### DIFF
--- a/axelrod/tests/strategies/test_apavlov.py
+++ b/axelrod/tests/strategies/test_apavlov.py
@@ -22,27 +22,47 @@ class TestAPavlov2006(TestPlayer):
 
     def test_strategy(self):
         self.first_play_test(C)
-        self.responses_test([C], [C] * 6, [C] * 6,
-                            attrs={"opponent_class": "Cooperative"})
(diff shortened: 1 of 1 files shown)


axelrod/tests/strategies/test_apavlov.py
```diff
@@ -22,27 +22,47 @@ class TestAPavlov2006(TestPlayer):
...
+        actions = [(C, C)] * 6 + [(C, D), (D, C)]
+        self.versus_test(opponent, expected_actions=actions,
+                         attrs={"opponent_class": "Cooperative"})
+
+        actions = [(C, D)] + [(D, D)] * 6
+        self.versus_test(axelrod.Defector(), expected_actions=actions,
+                         attrs={"opponent_class": "ALLD"})
+
+        opponent = axelrod.MockPlayer([D, C, D, C, D, C])
+        actions = [(C, D), (D, C), (C, D), (D, C),
```
marcharper: wrap to two lines?

Issue #916, opened 2017-03-20
Some typos in titfortat.OmegaTFT


Chadys:
While looking at the code of titfortats strategies, I found some mistakes in the implementation of OmegaTFT.

titfortat L319:
```python
            # If the opponent's last move differed from mine,
            # increase the counter
            if self.history[-1] == opponent.history[-1]:
                self.randomness_counter += 1
```
As the comment implies, the `==` should be `!=`

titfortat L323:
```python
            # Compare counts to thresholds
            # If randomness_counter exceeds Y, Defect for the remainder
            if self.randomness_counter >= 8:
```
`8` should be replaced by `self.randomness_threshold` (whose default value is 8, typo must come from this)

The reference to this strategy can be found here : <https://arxiv.org/pdf/cs/0609017.pdf>. The description of the strategy itself starts on section 1.3.1.4 (page 14) and you can also see the source code (page 16). The line we are interested in are those (L 29-31):
```
    if( opponentMove != myPreviousMove )        randomnessMeasure++; 

    if( randomnessMeasure >= RANDOMNESS_THRESHOLD ) 
```
I've already done the corrections, I'll just check what I need to adjust in the tests and make a pull request.

Chadys:
Also, but it's just a personal opinion and not a mistake, instead of
titfortat L303:
```python
return D if opponent.history[-1:] == [D] else C
```
and
titfortat L329:
```python
move = D if opponent.history[-1:] == [D] else C
```
I would rather have
```python
return opponent.history[-1]
```
and
```python
move =  opponent.history[-1]
```
for performance sake

PR #917, opened 2017-03-20
916

Chadys:
Issue #916

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -300,7 +300,7 @@ class OmegaTFT(Player):
             return C
         # TFT on round 2
         if len(self.history) == 1:
-            return D if opponent.history[-1:] == [D] else C
+            return opponent.history[-1]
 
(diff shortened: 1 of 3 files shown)


axelrod/strategies/titfortat.py
```diff
@@ -300,7 +300,7 @@ def strategy(self, opponent: Player) -> Action:
             return C
         # TFT on round 2
         if len(self.history) == 1:
-            return D if opponent.history[-1:] == [D] else C
```
marcharper: I think we use this because that's how TFT is actually implemented in the library, where we don't know that the history length is greater than zero. I don't have a preference here and I think your line might be more readable, but we might prefer consistency. So I'm approving and I'll let @drvinceknight and @meatballs weigh in if they have a preference.

drvinceknight: No I prefer this too. This is how TFT __was__ implemented a while ago but no longer :)

drvinceknight:
Closes #917 (so that github will close the issue for us).

Issue #921, opened 2017-03-20
Delete parameters from GoByMajority variants


Chadys:
For all the GoByMajority variants, I don't think we should have parameters. Because, for example, if you call `HardGoByMajority5(10, True)`, it's not a `HardGoByMajority5` anymore but a `GoByMajority10`. I suggest that only GoByMajority should have parameters, and the variants should only call GoByMajority with their defining parameters. If that's ok, I'll make the modification.

drvinceknight:
Seems sensible to me. Please open a PR :) :+1:

drvinceknight:
Closed by #922

PR #922, opened 2017-03-20
921

Chadys:
Axelrod-Python/Axelrod#921
I took more time than I though to correct the tests, since `TestHardGoByMajority` uses `TestGoByMajority`. Tell me if there is a problem with how I changed things.
And I also corrected the type hint of memory_depth which indicated `float` even though, as the docstring says on line 37 :
```
Parameters
        ----------
        memory_depth, int >= 0
```
(default value `float('inf')` must have confused the contributor)

### DIFF
--- a/axelrod/tests/strategies/test_gobymajority.py
+++ b/axelrod/tests/strategies/test_gobymajority.py
@@ -6,11 +6,12 @@ from .test_player import TestPlayer
 C, D = axelrod.Actions.C, axelrod.Actions.D
 
 
-class TestGoByMajority(TestPlayer):
+class TestHardGoByMajority(TestPlayer):
 
-    name = "Soft Go By Majority"
-    player = axelrod.GoByMajority
-    default_soft = True
+    name = "Hard Go By Majority"
+    player = axelrod.HardGoByMajority
+    default_soft = False
+    eq_play = D
 
     expected_classifier = {
         'stochastic': False,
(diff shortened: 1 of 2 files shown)


axelrod/tests/strategies/test_gobymajority.py
```diff
@@ -74,10 +74,15 @@ def test_strategy(self):
         self.responses_test([D], [C, D, D, D], [D, D, C, C])
         self.responses_test([D], [C, C, D, D, C], [D, D, C, C, D])
 
-        # Test tie break rule for soft=True
-        player = self.player(soft=True)
-        opponent = axelrod.Cooperator()
-        self.assertEqual('C', player.strategy(opponent))
+    def test_soft(self):
+        pass
```
meatballs: Why is this here? Does it need a test to be added?

Chadys: No, it's just to override the test from TestGoByMajority, that checks the soft argument works correctly. Since HardGoByMajority doesn't have a soft argument anymore, that test isn't needed and was causing an exception if I didn't override it to remove it. I thought about not making TestHardGoByMajority inherite from TestGoByMajority at all, but there are still enough common points between the two for it to be meaningful.

meatballs: 👍  Could you add a comment in to explain all that?

Chadys: I could have used delattr instead.

meatballs: 👍  Could you add a comment to explain the need for line 78? (I had to ask this again to keep github's review process happy).

Chadys: Sure, will do tonight !

Chadys:
So, maybe I got a little over-enthusiastic here, but I modified more things than planned. Instead of adding a comment, I finally concluded that, since `TestHardGoByMajority` proceeded to less tests than `TestGoByMajority`, it would be more logical to inverse the genealogy. So that's what I did, making `TestHardGoByMajority` much clearer by not having any function dealing with the `soft` parameter and making only `TestGoByMajority` implements them. I also re-factorized `test_strategy` a little, to avoid code repetition. And finally I added a test that was missing (testing that `GoByMajority` cooperates when opponent cooperated more often than defected, as only the 'equality' or 'more defect' cases were tested). But if I overstepped my bounds please tell me, and I'll just add the comment instead.

meatballs:
> But if I overstepped my bounds please tell me

I don't think we have any to be overstepped!

I'll have a proper look at this in the morning.

marcharper:
Looks good to me!

drvinceknight:
:+1:

PR #923, opened 2017-03-22
New Strategy: MetaWinnerTopTypes

meatballs:


drvinceknight:
Looks good. 

I believe we had a strategy like this in the past but it got removed (I might be remembering incorrectly, I think @marcharper did this?). 

(As this is a parametrized version of a strategy another option would be to include it in the overall tournament run. So adding to the `parametrized_players` list here https://github.com/Axelrod-Python/tournament/blob/gh-pages/players.py#L5)

meatballs:
> As this is a parametrized version of a strategy another option would be to include it in the overall tournament run.

Yep. I like the sound of that. I'll ditch this.

marcharper:
I actually just added a new strategy type to the dojo to select the optimal team and weights for an adaptive meta player (could also be used for a meta mixer). It's on [this branch](https://github.com/Axelrod-Python/axelrod-dojo/tree/mwua). I haven't done a long training run yet.

drvinceknight:
Ok, @meatballs has closed this PR and added the strategy to the parametrized run over on the tournament repo. We can change things if/when you've trained the strategy?

Issue #816, opened 2017-01-15
Actions should probably be an enum now.


drvinceknight:
@marcharper you've mentioned this a couple of times, opening this issue to discuss :)

marcharper:
It causes a lot of errors since `C` becomes `Actions.C` but I'm sure we can figure it out...

marcharper:
Also note that `enum.auto()` is Python 3.6+.

drvinceknight:
> It causes a lot of errors since C becomes Actions.C but I'm sure we can figure it out...

And perhaps from a readability point of view just having `C` is still preferable?

I am not actually familiar with `enum` at all. What problem would moving to them solve?

marcharper:
> What problem would moving to them solve?
None that I have in mind at the moment but it is an Enum technically.

eric-s-s:
here's my 10 minute demo of what it would look like.  for convenience, I built it in another project, but you get the idea.
```
from enum import Enum


class Action(Enum):
    C = 'C'
    D = 'D'

    def flip(self):
        flipped = {self.C: self.D, self.D: self.C}
        return flipped[self]

    def __repr__(self):
        return 'Action.{}'.format(self.name)
```

as far as i can see, the advantages are that 

- it's impossible to get an error when you flip.  
- It can't, in any way be mixed up with strings.  
- It now works in the type hints.  (i'll put the type hints for current action below.)  
- there is no need for Action and Actions to take care of type checking.  
- flip_action no longer needs a specific import.  It's just there for any C or D whenever you need it.

here's what it did in my interperter.
```
In[29]: from battle.scrap import Action
In[30]: C, D = Action.C, Action.D

In[32]: C
Out[32]:
Action.C

In[33]: C == C
Out[33]:
True

In[34]: C != D
Out[34]:
True

In[35]: C == "C"
Out[35]:
False

In[36]: C.flip()
Out[36]:
Action.D

In[37]: D.flip()
Out[37]:
Action.C

In[38]: def blah(num: int) -> Action:
   ...:     if num < 0:
   ...:         return C
   ...:     return D
   ...: 
In[39]: help(blah)
Help on function blah in module __main__:

blah(num:int) -> battle.scrap.Action

In[40]: blah(2)
Out[40]: 
Action.D

In[41]: x = 5 * [C]
In[42]: x
Out[42]: 
[Action.C, Action.C, Action.C, Action.C, Action.C]  <- not represented as strings
```

[...]

PR #925, opened 2017-03-23
844 b and c

eric-s-s:
#884 

Cycler depended on knowing that Actions.C and Actions.D were 'C' and 'D'. I wrote a few lines to refactor that and can remove them with ease.

### DIFF
--- a/axelrod/_strategy_utils.py
+++ b/axelrod/_strategy_utils.py
@@ -26,18 +26,15 @@ def detect_cycle(history, min_size=1, max_size=12, offset=0):
         The amount of history to skip initially
     """
     history_tail = history[offset:]
-    max_ = min(len(history_tail) // 2, max_size)
-    for i in range(min_size, max_):
+    new_max_size = min(len(history_tail) // 2, max_size)
+    for i in range(min_size, new_max_size + 1):
         has_cycle = True
         cycle = tuple(history_tail[:i])
         for j, elem in enumerate(history_tail):
             if elem != cycle[j % len(cycle)]:
                 has_cycle = False
                 break
-        if has_cycle and (j == len(history_tail) - 1):
-            # We made it to the end, is the cycle itself a cycle?
-            # E.G. CCC is not ok as cycle if min_size is really 2
-            # Since this is the same as C
+        if has_cycle:
             return cycle
     return None
 
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -1,3 +1,5 @@
+from types import GeneratorType
+
 # Type alias for actions.
 Action = str
 
@@ -16,13 +18,16 @@ def flip_action(action: Action) -> Action:
         raise ValueError("Encountered a invalid action.")
 
 
-def str_to_actions(actions: str):
+def str_to_actions(actions: str) -> GeneratorType:
     """Takes a string like 'cCdD' and returns a generator that yields the appropriate actions."""
-    for character in actions:
-        as_upper = character.upper()
-        if as_upper == 'C':
-            yield Actions.C
-        elif as_upper == 'D':
-            yield Actions.D
-        else:
-            raise ValueError('The characters of action_str may only be "C", "c", "D" and "d"')
+    def make_generator(actions_str):
+        for character in actions_str:
+            as_upper = character.upper()
+            if as_upper == 'C':
+                yield Actions.C
+            elif as_upper == 'D':
+                yield Actions.D
+            else:
+                raise ValueError('The characters of action_str may only be "C", "c", "D" and "d"')
+
+    return make_generator(actions)
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -1,4 +1,4 @@
-
+from types import GeneratorType
 # Type alias for actions.
 Action = str
 
@@ -15,3 +15,15 @@ def flip_action(action: Action) -> Action:
         return Actions.C
     else:
         raise ValueError("Encountered a invalid action.")
+
+
+def str_to_actions(actions: str) -> GeneratorType:
+    """Takes a string like 'cCdD' and returns a generator that yields the appropriate actions."""
+    for character in actions:
+        as_upper = character.upper()
+        if as_upper == 'C':
+            yield Actions.C
+        elif as_upper == 'D':
+            yield Actions.D
+        else:
+            raise ValueError('The characters of action_str may only be "C", "c", "D" and "d"')
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -1,5 +1,3 @@
-from types import GeneratorType
-
 # Type alias for actions.
 Action = str
 
@@ -18,16 +16,11 @@ def flip_action(action: Action) -> Action:
         raise ValueError("Encountered a invalid action.")
 
 
-def str_to_actions(actions: str) -> GeneratorType:
-    """Takes a string like 'cCdD' and returns a generator that yields the appropriate actions."""
-    def make_generator(actions_str):
-        for character in actions_str:
-            as_upper = character.upper()
-            if as_upper == 'C':
-                yield Actions.C
-            elif as_upper == 'D':
-                yield Actions.D
-            else:
-                raise ValueError('The characters of action_str may only be "C", "c", "D" and "d"')
-
-    return make_generator(actions)
+def str_to_actions(actions: str) -> tuple:
+    """Takes a string like 'CCDD' and returns a tuple of the appropriate actions."""
+    action_dict = {'C': Actions.C,
+                   'D': Actions.D}
+    try:
+        return tuple(action_dict[action] for action in actions)
+    except KeyError:
+        raise ValueError('The characters of "actions" str may only be "C" or "D"')
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -1,4 +1,3 @@
-from types import GeneratorType
 # Type alias for actions.
 Action = str
 
@@ -17,7 +16,7 @@ def flip_action(action: Action) -> Action:
         raise ValueError("Encountered a invalid action.")
 
 
-def str_to_actions(actions: str) -> GeneratorType:
+def str_to_actions(actions: str):
     """Takes a string like 'cCdD' and returns a generator that yields the appropriate actions."""
     for character in actions:
         as_upper = character.upper()
--- a/axelrod/strategies/cooperator.py
+++ b/axelrod/strategies/cooperator.py
@@ -37,13 +37,23 @@ class TrickyCooperator(Player):
         'manipulates_state': False
     }
 
-    @staticmethod
-    def strategy(opponent: Player) -> Action:
+    _min_history_required_to_try_trickiness = 3
+    _max_history_depth_for_trickiness = -10
+
+    def strategy(self, opponent: Player) -> Action:
         """Almost always cooperates, but will try to trick the opponent by defecting.
 
-        Defect once in a while in order to get a better payout, when the opponent
-        has not defected in the last ten turns and only cooperated during last 3 turns.
+        Defect once in a while in order to get a better payout.
+        After 3 rounds, if opponent has not defected to a max history depth of 10, Defect.
         """
-        if D not in opponent.history[-10:] and opponent.history[-3:] == [C]*3:
+        if (self._has_played_enough_rounds_to_be_tricky() and
+                self._opponents_has_cooperated_enough_to_be_tricky(opponent)):
             return D
         return C
+
+    def _has_played_enough_rounds_to_be_tricky(self):
+        return len(self.history) >= self._min_history_required_to_try_trickiness
+
+    def _opponents_has_cooperated_enough_to_be_tricky(self, opponent):
+        rounds_to_be_checked = opponent.history[self._max_history_depth_for_trickiness:]
+        return D not in rounds_to_be_checked
--- a/axelrod/strategies/cycler.py
+++ b/axelrod/strategies/cycler.py
@@ -3,11 +3,12 @@ from axelrod.player import Player
 import itertools
 import copy
 
+C, D = Actions.C, Actions.D
 
 class AntiCycler(Player):
     """
     A player that follows a sequence of plays that contains no cycles:
-    C CD CCD CCCD CCCCD CCCCCD ...
+    CDD  CD  CCD CCCD CCCCD ...
     """
 
     name = 'AntiCycler'
@@ -25,20 +26,28 @@ class AntiCycler(Player):
         super().__init__()
         self.cycle_length = 1
         self.cycle_counter = 0
+        self.first_three = self._get_first_three()
+
+    @staticmethod
+    def _get_first_three():
+        return [C, D, D]
 
     def strategy(self, opponent: Player) -> Action:
+        while self.first_three:
+            return self.first_three.pop(0)
         if self.cycle_counter < self.cycle_length:
             self.cycle_counter += 1
-            return Actions.C
+            return C
         else:
             self.cycle_length += 1
             self.cycle_counter = 0
-            return Actions.D
+            return D
 
     def reset(self):
         super().reset()
         self.cycle_length = 1
         self.cycle_counter = 0
+        self.first_three = self._get_first_three()
 
 
 class Cycler(Player):
--- a/axelrod/strategies/cycler.py
+++ b/axelrod/strategies/cycler.py
@@ -5,6 +5,7 @@ import copy
 
 C, D = Actions.C, Actions.D
 
+
 class AntiCycler(Player):
     """
     A player that follows a sequence of plays that contains no cycles:
--- a/axelrod/strategies/cycler.py
+++ b/axelrod/strategies/cycler.py
@@ -1,6 +1,6 @@
-from axelrod.actions import Actions, Action
+from axelrod.actions import Actions, Action, str_to_actions
 from axelrod.player import Player
-
+import itertools
 import copy
 
 
@@ -55,7 +55,7 @@ class Cycler(Player):
         'manipulates_state': False
     }
 
-    def __init__(self, cycle="CCD") -> None:
+    def __init__(self, cycle_str="CCD") -> None:
         """This strategy will repeat the parameter `cycle` endlessly,
         e.g. C C D C C D C C D ...
 
@@ -67,16 +67,20 @@ class Cycler(Player):
 
         """
         super().__init__()
-        self.cycle = cycle
-        self.name = "Cycler {}".format(cycle)
-        self.classifier['memory_depth'] = len(cycle) - 1
+        self.cycle_str = cycle_str
+        self.cycle = self.get_new_itertools_cycle()
+        self.name = "Cycler {}".format(cycle_str)
+        self.classifier['memory_depth'] = len(cycle_str) - 1
+
+    def get_new_itertools_cycle(self):
+        return itertools.cycle(str_to_actions(self.cycle_str))
 
     def strategy(self, opponent: Player) -> Action:
-        actions = {'C': Actions.C, 'D': Actions.D}
-        curent_round = len(self.history)
-        index = curent_round % len(self.cycle)
-        action_str = self.cycle[index]
-        return actions[action_str]
+        return next(self.cycle)
+
+    def reset(self):
+        super(Cycler, self).reset()
+        self.cycle = self.get_new_itertools_cycle()
 
 
 class CyclerDC(Cycler):
@@ -85,8 +89,8 @@ class CyclerDC(Cycler):
     classifier = copy.copy(Cycler.classifier)
     classifier['memory_depth'] = 1
 
-    def __init__(self, cycle="DC") -> None:
-        super().__init__(cycle=cycle)
+    def __init__(self, cycle_str="DC") -> None:
+        super().__init__(cycle_str=cycle_str)
 
 
 class CyclerCCD(Cycler):
@@ -95,8 +99,8 @@ class CyclerCCD(Cycler):
     classifier = copy.copy(Cycler.classifier)
     classifier['memory_depth'] = 2
 
-    def __init__(self, cycle="CCD") -> None:
-        super().__init__(cycle=cycle)
+    def __init__(self, cycle_str="CCD") -> None:
+        super().__init__(cycle_str=cycle_str)
 
 
 class CyclerDDC(Cycler):
@@ -105,8 +109,8 @@ class CyclerDDC(Cycler):
     classifier = copy.copy(Cycler.classifier)
     classifier['memory_depth'] = 2
 
-    def __init__(self, cycle="DDC") -> None:
-        super().__init__(cycle=cycle)
+    def __init__(self, cycle_str="DDC") -> None:
+        super().__init__(cycle_str=cycle_str)
 
 
 class CyclerCCCD(Cycler):
@@ -115,8 +119,8 @@ class CyclerCCCD(Cycler):
     classifier = copy.copy(Cycler.classifier)
     classifier['memory_depth'] = 3
 
-    def __init__(self, cycle="CCCD") -> None:
-        super().__init__(cycle=cycle)
+    def __init__(self, cycle_str="CCCD") -> None:
+        super().__init__(cycle_str=cycle_str)
 
 
 class CyclerCCCCCD(Cycler):
@@ -125,8 +129,8 @@ class CyclerCCCCCD(Cycler):
     classifier = copy.copy(Cycler.classifier)
     classifier['memory_depth'] = 5
 
-    def __init__(self, cycle="CCCCCD") -> None:
-        super().__init__(cycle=cycle)
+    def __init__(self, cycle_str="CCCCCD") -> None:
+        super().__init__(cycle_str=cycle_str)
 
 
 class CyclerCCCDCD(Cycler):
@@ -135,5 +139,5 @@ class CyclerCCCDCD(Cycler):
     classifier = copy.copy(Cycler.classifier)
     classifier['memory_depth'] = 5
 
-    def __init__(self, cycle="CCCDCD") -> None:
-        super().__init__(cycle=cycle)
+    def __init__(self, cycle_str="CCCDCD") -> None:
+        super().__init__(cycle_str=cycle_str)
--- a/axelrod/strategies/cycler.py
+++ b/axelrod/strategies/cycler.py
@@ -46,7 +46,7 @@ class Cycler(Player):
 
     name = 'Cycler'
     classifier = {
-        'memory_depth': 1,
+        'memory_depth': 2,
         'stochastic': False,
         'makes_use_of': set(),
         'long_run_time': False,
--- a/axelrod/strategies/cycler.py
+++ b/axelrod/strategies/cycler.py
@@ -55,7 +55,7 @@ class Cycler(Player):
         'manipulates_state': False
     }
 
-    def __init__(self, cycle_str="CCD") -> None:
+    def __init__(self, cycle: str = "CCD") -> None:
         """This strategy will repeat the parameter `cycle` endlessly,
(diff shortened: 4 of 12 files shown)


axelrod/strategies/cycler.py
```diff
@@ -72,9 +72,11 @@ def __init__(self, cycle="CCD") -> None:
         self.classifier['memory_depth'] = len(cycle) - 1
 
     def strategy(self, opponent: Player) -> Action:
+        actions = {'C': Actions.C, 'D': Actions.D}
```
eric-s-s: this is where i changed Cycler so that it no longer assumes Actions.C = 'C' and Actions.D = 'D'

drvinceknight: Yes, I think this is a good move, @marcharper has spoken about potentially changing the actions to be another class so things like this give us the flexibility to do so if we end up doing that.

drvinceknight: I'll do a proper review when I have more time (probably tomorrow).

eric-s-s: yes.  i saw the issue where he mentioned making Actions an Enum, and had that in mind.

marcharper:
Perhaps a string to actions function would be useful for the library, something like:
```
def str_to_actions(action_str):
    for c in action_str:
         if c.lower() not in ('c', 'd'):
              raise SomeError
         yield actions[c]
```

eric-s-s:
@marcharper I second that.  see comment in #816

axelrod/strategies/cycler.py
```diff
@@ -72,9 +72,11 @@ def __init__(self, cycle="CCD") -> None:
...
         curent_round = len(self.history)
         index = curent_round % len(self.cycle)
-        return self.cycle[index]
+        action_str = self.cycle[index]
+        return actions[action_str]
```
drvinceknight: This could all be refactored used `itertools.cycle` (as in the `axelrod.MockPlayer`). Since you've refactored the strategy I suggest doing it that way (removes the need for index etc).

axelrod/tests/strategies/test_better_and_better.py
```diff
@@ -23,5 +23,14 @@ class TestBetterAndBetter(TestPlayer):
     def test_strategy(self):
         """Tests that the strategy gives expected behaviour."""
         self.first_play_test(D, seed=3)  # D is very unlikely
```
drvinceknight: Could you add another `first_play_test` with a different seed that shows that this is indeed random (so sometimes we will get `C`).

drvinceknight: Similar to this: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/strategies/test_averagecopier.py#L25

axelrod/tests/strategies/test_calculator.py
```diff
@@ -35,16 +36,65 @@ def test_strategy(self):
...
+        twenty_test_actions = get_joss_strategy_actions(twenty_alternator_actions, seed_two_flip_indices)
+
+        expected_actions = twenty_test_actions + [(D, C), (D, D), (D, C), (D, D)]
+        self.versus_test(axelrod.Alternator(), twenty_test_actions, seed=seed)
+        self.versus_test(axelrod.Alternator(), expected_actions, seed=seed)
+
+    def test_twenty_rounds_joss_then_tit_for_tat_for_non_cyclers(self):
+        """uses axelrod.strategies.axelrod_first.Joss strategy for first 20 rounds"""
+        seed = 2
+        seed_two_flip_indices = [1, 2]
```
drvinceknight: could you just rename this to `indices_to_flip` (which matches the function var).

No need for the seed detail.

axelrod/tests/strategies/test_cycler.py
```diff
@@ -45,13 +46,31 @@ class TestCycler(TestPlayer):
...
             """Starts by cooperating"""
-            for i in range(20):
-                responses = itertools.islice(itertools.cycle(cycle), i)
-            self.responses_test(responses)
+            match_len = 20
+            actions_cycle = _get_actions_cycle_against_cooperator(cycle)
+            test_actions = list(itertools.islice(itertools.cycle(actions_cycle), match_len))
+            self.versus_test(axelrod.Cooperator(), test_actions)
 
     return TestCycler
```
drvinceknight: Should this line be here? (I know you didn't add it @eric-s-s)

axelrod/tests/strategies/test_better_and_better.py
eric-s-s: that's actually a bit tough.  on the first round, the chance is 0.001.  here's the problem.
```
def test_for_first_round_cooperations(self):
        for seed in range(2000):
            try:
                self.first_play_test(D, seed=seed)
            except AssertionError:
                print('failed at {}'.format(seed))
```
the output is 
```
......failed at 1514
failed at 1651
.......
Ran 13 tests in 0.119s

```
note the time it takes.  without this it runs in 0.008s .  

that's why i included the following test.  I thought it nicely visualized getting better and better on the scale of probability (which increases 0.001 per round so is 0.2 by round 200)
```
        for index in range(200):
            if index in [64, 79, 91, 99, 100, 107, 111, 119, 124, 127, 137, 141, 144, 154, 192, 196]:
                actions.append((C, D))
            else:
                actions.append((D, D))
        self.versus_test(axelrod.Defector(), expected_actions=actions, seed=8)
```

marcharper: We just want one seed where it's C and one seed where it's D to show that both behaviors are encountered.

axelrod/tests/strategies/test_cycler.py
eric-s-s: do you mean the blank before return TestCycler (line 53)? I didn't add it, but i do like it.  I think it nicely separates the instructions to construct the class from the act of returning that class.  it made it easy for me to read.  

if you meant the two blank lines after the factoryclass.  my IDE tries to enforce two blank lines between a class and whatever comes after.  I assume it's convention

eric-s-s: i have no problem removing it, though.

axelrod/tests/strategies/test_better_and_better.py
eric-s-s: got it.  will do.

axelrod/strategies/cycler.py
eric-s-s: i just got a very cool problem.  
```
FAIL: test_reset_history_and_attributes (axelrod.tests.strategies.test_cycler.test_cycler_factory.<locals>.TestCycler)
Make sure resetting works correctly.
  File "E:\work\Axelrod\axelrod\tests\strategies\test_player.py", line 239, in test_reset_history_and_attributes
    self.attribute_equality_test(player, clone)
  File "E:\work\Axelrod\axelrod\tests\strategies\test_player.py", line 268, in attribute_equality_test
    self.assertEqual(reset_value, original_value, msg=attribute)
AssertionError: <itertools.cycle object at 0x0000000006069288> != <itertools.cycle object at 0x0000000006088F48> : cycle

FAIL: test_reset_clone (axelrod.tests.strategies.test_cycler.test_cycler_factory.<locals>.TestCycler)
Make sure history resetting with cloning works correctly, regardless
  File "E:\work\Axelrod\axelrod\tests\strategies\test_player.py", line 246, in test_reset_clone
    self.attribute_equality_test(player, clone)
  File "E:\work\Axelrod\axelrod\tests\strategies\test_player.py", line 268, in attribute_equality_test
    self.assertEqual(reset_value, original_value, msg=attribute)
AssertionError: <itertools.cycle object at 0x00000000060750C8> != <itertools.cycle object at 0x0000000006088C08> : cycle
```
i'll state conclusions below.

axelrod/tests/strategies/test_cycler.py
drvinceknight: No, sorry ignore me, I meant `return TestCycler` but that was a stupid question.

axelrod/strategies/cycler.py
eric-s-s: there is a funky thing going on with inheritance.  Cycler has several child classes.  I'm pretty sure this error triggered on the children but not the parent.  there are 6 children and 12 errors.  so i wrote a new cycler that inherits from MockPlayer and got the same issue.  here's the code for child of MockPlayer.

```
class Cycler(MockPlayer):
    name = 'Cycler'
    classifier = {
        'memory_depth': 1,
        'stochastic': False,
        'makes_use_of': set(),
        'long_run_time': False,
        'inspects_source': False,
        'manipulates_source': False,
        'manipulates_state': False
    }
    
    def __init__(self, cycle):
        for_mock = []
        for char in cycle:
            if char == 'C':
                for_mock.append(Actions.C)
            else:
                for_mock.append(Actions.D)
        super(Cycler, self).__init__(for_mock)
```

do you have any idea what's going on?  i'll bet anything there's some issue with the way the children inherit but i don't know what it is.

drvinceknight: This is due to the automatic reset method which is not recognising the cycle as a genorator:

https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/strategies/test_player.py#L263

```python
>>> import itertools
>>> import types
>>> isinstance((k for k in [1, 2, 3]), types.GeneratorType)
True
>>> isinstance(itertools.cycle([1, 2, 3]), types.GeneratorType)
False
>>> isinstance(itertools.cycle([1, 2, 3]), itertools.cycle)
True

```

I suggest you add another check for cycles at this line: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/strategies/test_player.py#L263

```
-  elif isinstance(reset_value, types.GeneratorType) 
+ elif isinstance(reset_value, types.GeneratorType) or isinstance(reset_value, itertools.cycle):
```

That should check those things properly :)

eric-s-s: thanks!  that almost did it.  now there's a reset bug.  should i push that change in test_player?

eric-s-s: fortunately, the reset bug is within cycler itself.

drvinceknight: Cool and yup, include the patch to `test_player` in this PR please :)

axelrod/tests/strategies/test_cycler.py
eric-s-s: ok

axelrod/actions.py
```diff
@@ -15,3 +15,15 @@ def flip_action(action: Action) -> Action:
...
+def str_to_actions(actions: str) -> GeneratorType:
+    """Takes a string like 'cCdD' and returns a generator that yields the appropriate actions."""
+    for character in actions:
+        as_upper = character.upper()
+        if as_upper == 'C':
+            yield Actions.C
+        elif as_upper == 'D':
+            yield Actions.D
+        else:
+            raise ValueError('The characters of action_str may only be "C", "c", "D" and "d"')
```
eric-s-s: @marcharper mentioned this might be a thing to have and so i used the code he posted.  I found myself repeating code in cycler so I put this here.  it can easlily be pushed down to the two modules it's used in (cycler.py and test_cycler.py).  

it might be useful here if Actions ever gets refactored as a useful thing to go with it.  

it might be better as a list for fail-fast purposes.  this only raises an error once you iterate to offending character and not at instantiation (see tests).  if it returned a list, it would always fail immediately.

axelrod/strategies/cooperator.py
```diff
@@ -37,13 +37,23 @@ class TrickyCooperator(Player):
...
+
+    def strategy(self, opponent: Player) -> Action:
         """Almost always cooperates, but will try to trick the opponent by defecting.
 
-        Defect once in a while in order to get a better payout, when the opponent
-        has not defected in the last ten turns and only cooperated during last 3 turns.
+        Defect once in a while in order to get a better payout.
+        After 3 rounds, if opponent has not defected to a max history depth of 10, Defect.
         """
-        if D not in opponent.history[-10:] and opponent.history[-3:] == [C]*3:
```
eric-s-s: i had trouble understanding the original intent and wanted to state explicitly what it does.  both ways pass all original tests and tests that i wrote.  was this supposed to be like TrickyDefector, something like `if D in opponent.history and opponent.history[-3:] == [C] * 3` ?

axelrod/tests/strategies/test_cooperator.py
```diff
@@ -42,8 +42,50 @@ def test_strategy(self):
...
+        self.versus_test(axelrod.Cooperator(), [(C, C), (C, C), (C, C), (D, C), (D, C)])
+
+        opponent_actions = [C, C, C, C, D, D]
+        expected_actions = [(C, C), (C, C), (C, C), (D, C), (D, D), (C, D)]
+        self.versus_test(axelrod.MockPlayer(opponent_actions), expected_actions=expected_actions)
+
+        opponent_actions = [C, C, C, C] + [D, D] + [C] * 10
+        expected_actions = [(C, C), (C, C), (C, C), (D, C)] + [(D, D), (C, D)] + [(C, C)] * 10
+        self.versus_test(axelrod.MockPlayer(opponent_actions), expected_actions=expected_actions)
+
```
eric-s-s: these two tests are according to this.  i would like to get rid of this set of tests labeled """according to docstring""" or the ones following it labeled """according to code meaning"""
[https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/cooperator.py#L44-L45](url)

> Defect once in a while in order to get a better payout, when the opponent
        has not defected in the last ten turns and only cooperated during last 3 turns.

axelrod/tests/strategies/test_player.py
```diff
@@ -260,7 +260,7 @@ def attribute_equality_test(self, player, clone):
                 self.assertTrue(np.array_equal(reset_value, original_value),
                                 msg=attribute)
 
-            elif isinstance(reset_value, types.GeneratorType):
+            elif isinstance(reset_value, types.GeneratorType) or isinstance(reset_value, itertools.cycle):
```
eric-s-s: @drvinceknight how do i reference your previous comment about this?

eric-s-s:
Cycler has an oddity for testing.  

it's class variable has "memory_depth": 1, but on __init__ it changes this value.  The default, Cycler() creates "memory_depth": 2.  I could not get TestPlayer to run successfully on Cycler, whether I ran it as memory_depth: 1 or 2 I kept getting errors.

axelrod/strategies/cooperator.py
drvinceknight: Yes I think the original docstring was unclear and I think you've got it right here.

axelrod/tests/strategies/test_calculator.py
```diff
@@ -23,28 +24,77 @@ class TestCalculator(TestPlayer):
...
-        P1.history = [C] * 20
-        P2 = axelrod.Player()
-        P2.history = [C, D] * 10
+        p1 = axelrod.Calculator()
+        p1.history = [C] * 20
+        p2 = axelrod.Player()
+        p2.history = [C, D] * 10
         # Defects on cycle detection
-        self.assertEqual(D, P1.strategy(P2))
+        self.assertEqual(D, p1.strategy(p2))
```
drvinceknight: Could you replace these with `versus_test` versions?

axelrod/tests/strategies/test_cooperator.py
drvinceknight: Assuming that "according to docstring" you mean the original docstring (before your fix) then I agree. Remove them and let's see if coverage is affected (it it's not I think that'll be good to go).

axelrod/tests/strategies/test_cooperator.py
```diff
@@ -42,8 +42,50 @@ def test_strategy(self):
...
+        self.versus_test(axelrod.Cooperator(), against_cooperator)
+        self.versus_test(axelrod.Alternator(), against_alternator)
+
+    def test_defects_after_three_rounds_if_opponent_only_cooperated_in_max_history_depth_ten(self):
+        """according to code meaning"""
+        against_cooperator = [(C, C)] * 3 + [(D, C)] * 20
+        self.versus_test(axelrod.Cooperator(), against_cooperator)
+
+    def test_defects_when_opponent_has_no_defections_to_history_depth_ten(self):
+        """according to code meaning"""
```
drvinceknight: Once you've removed `"""according to docstring"""` tests you can remove the label `"""according to code meaning"""`.

axelrod/tests/strategies/test_player.py
drvinceknight: > how do i reference your previous comment about this?

Sorry, I'm not sure what you mean. I think this is fine as it is. No need to reference anything.

axelrod/actions.py
```diff
@@ -15,3 +16,18 @@ def flip_action(action: Action) -> Action:
         return Actions.C
     else:
         raise ValueError("Encountered a invalid action.")
+
+
+def str_to_actions(actions: str) -> GeneratorType:
+    """Takes a string like 'cCdD' and returns a generator that yields the appropriate actions."""
+    def make_generator(actions_str):
+        for character in actions_str:
+            as_upper = character.upper()
```
drvinceknight: Should we insist on uppercase as inputs? So `c` would raise a `ValueError`?

axelrod/tests/strategies/test_calculator.py
eric-s-s: will do.

axelrod/actions.py
eric-s-s: `str_to_actions('ccddd')`  and   
`str_to_actions('CCDDD')`   
the lower one looks more natural to all the current axelrod code and evokes the actions clearly.  so it would seem to make sense to enforce that in code.  i like that suggestion.  i'll await a decision.

axelrod/tests/strategies/test_calculator.py
eric-s-s: on second look.  are these tests necessary?  i think i covered them in the other two test functions.  i suspect rewriting them would look exactly like the next two tests.

eric-s-s:
on fiddling on a test for calculator i discovered this.

```>>>detect_cycle([C, C, D] * 2]) 
None
>>>detect_cycle([C, C, D] * 2 + [C, C])
(C, C, D)
```
was this by design or error?  if error, the fix is to change this line from
[https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/_strategy_utils.py#L32](url)
to ` cycle = tuple(history_tail[:i + 1])`

axelrod/actions.py
```diff
@@ -15,3 +16,18 @@ def flip_action(action: Action) -> Action:
...
+            if as_upper == 'C':
+                yield Actions.C
+            elif as_upper == 'D':
+                yield Actions.D
+            else:
+                raise ValueError('The characters of action_str may only be "C", "c", "D" and "d"')
+
```
meatballs: How about using a dict to do this instead? e.g.
```
actions = {
    'C': Actions.C,
    'D': Actions.D
}

try:
    return actions[character.upper()]
except KeyError:
    raise ValueError('The characters of action_str may only be "C", "c", "D" and "d"')
```

axelrod/tests/strategies/test_calculator.py
drvinceknight: Yeah,  I thought the same, I think you can remove them.

drvinceknight:
> was this by design or error? if error, the fix is to change this line from
> https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/_strategy_utils.py#L32
> cycle = tuple(history_tail[:i])
> to cycle = tuple(history_tail[:i + 1])

That looks like an error to me. Does that fix trigger a lot of test failures?

eric-s-s:
> That looks like an error to me. Does that fix trigger a lot of test failures?

the fix triggered no test failures.

drvinceknight:
Ok, cool, can you add the fix and your example as a test?

drvinceknight:
Looks like this has triggered a failure in the `CycleHunter`:

```
FAIL: test_strategy (axelrod.tests.strategies.test_hunter.TestCycleHunter)
  File "/home/vince/src/Axelrod/axelrod/tests/strategies/test_hunter.py", line 139, in test_strategy
    self.assertEqual(player.history[-1], C)
AssertionError: 'D' != 'C'
- D
+ C

```

eric-s-s:
> Looks like this has triggered a failure in the CycleHunter:

FAIL: test_strategy (axelrod.tests.strategies.test_hunter.TestCycleHunter)
  File "/home/vince/src/Axelrod/axelrod/tests/strategies/test_hunter.py", line 139, in test_strategy
    self.assertEqual(player.history[-1], C)
AssertionError: 'D' != 'C'
- D
+ C


@drvinceknight 
excellent.  i'll look into that and refine or retract.

eric-s-s:
@drvinceknight 
oooh.  found it.  AntiCycler has a 3 length cycle when it has 6 history.  it plays [C] + [C, D] + [C, C, D]
CycleHunter detects this as it has a minimum cycle len of 3.  
so this is the part of the test that triggers failure
[https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/strategies/test_hunter.py#L132-L135](url)

how should i proceed?

eric-s-s:
perhaps detect_cycle needs to have a cycle fully repeat twice or CycleHunter needs to search for longer cycles or ...

drvinceknight:
> perhaps detect_cycle needs to have a cycle fully repeat twice or CycleHunter needs to search for longer cycles or ...

I'm just about to push my suggested fix and PR to your branch :)

I think that you've found a bug in `AntiCycler`, it's first six moves clearly have a cycle so I'm going to suggest we bump it to not have that initial `C`.

drvinceknight:
@eric-s-s https://github.com/eric-s-s/Axelrod/pull/1 but as I said there I ended up with the same failure anyway...

drvinceknight:
> @eric-s-s eric-s-s#1 but as I said there I ended up with the same failure anyway...

Ok I've figured out where the problem now occurs:

```python
>>> player = axl.CycleHunter()
>>> opponent = axl.AntiCycler()
>>> match = axl.Match([player, opponent], turns=7)
>>> match.play()
[('C', 'C'),
 ('C', 'D'),
 ('C', 'C'),
 ('C', 'C'),
 ('C', 'D'),
 ('C', 'C'),
 ('D', 'C')]
```

Now CycleHunter is recognising `C, D, C, C, D, C` as a 2 cycle of `C, D, C`. Having a play, will push a suggested fix to my PR to your branch.

drvinceknight:
I've just closed that, I suggest you simply remove `axl.AntiCycler`:
```
- 134         for opponent in [axelrod.Random(), axelrod.AntiCycler(),      
+ 134         for opponent in [axelrod.Random(),                                                                            
135                          axelrod.Cooperator(), axelrod.Defector()]: 
```

eric-s-s:
i'm a little confused.  do you have a fix for anticycler?  if not, i'll work on one now.  i also noticed that i completely failed to refactor the anticycler test.

eric-s-s:
well, i have a fix.  it's not the prettiest, but it works.

eric-s-s:
if you have a more elegant fix for anticycler, by all means.  the new tests i wrote should be quite effective for any version of it.

axelrod/_strategy_utils.py
```diff
@@ -26,18 +26,15 @@ def detect_cycle(history, min_size=1, max_size=12, offset=0):
...
-    for i in range(min_size, max_):
+    new_max_size = min(len(history_tail) // 2, max_size)
+    for i in range(min_size, new_max_size + 1):
         has_cycle = True
         cycle = tuple(history_tail[:i])
         for j, elem in enumerate(history_tail):
             if elem != cycle[j % len(cycle)]:
                 has_cycle = False
                 break
-        if has_cycle and (j == len(history_tail) - 1):
```
eric-s-s: i couldn't figure out what j==len(history_tail) -1 did, and i couldn't write a test that needed it.  i'm not sure if there's some test that i missed.

axelrod/actions.py
```diff
@@ -15,3 +14,13 @@ def flip_action(action: Action) -> Action:
         return Actions.C
     else:
         raise ValueError("Encountered a invalid action.")
+
+
```
eric-s-s: i made this return a tuple instead of a generator so that it would fail when called, and not fail when the generator reached the offending character.  This makes something like cycler much safer in that it will immediately fail on instantiation and not be a time-bomb waiting to fail when a bad character comes up.

axelrod/_strategy_utils.py
drvinceknight: It looks like coverage has stayed up so I think we're ok here.

axelrod/actions.py
drvinceknight: Nice thinking.

eric-s-s:
- actions.py
    - actions_to_str uses a dictionary and returns a tuple instead of generator.
    - test_actions changed to reflect this
- _strategy_utils.detect_strategy
    - added a +1 to range so if min=2 and max=5 range(2, 5+1) uses 2,3, 4, 5
    - removed (if j==len(history_tail) -1 and comment.  see tests
    - test_strategy_utils added tests for different variables
- test_calculator
    - added tests for edge case at len(cycle) = 10 / 11
- cycler
    - minor: changed Cycler.memory_depth: 2 so that it can use a TestPlayer class
    - major: change AntiCycler.  it deviates from original strategy which contained a cycle
- test_cycler
    - TestAntiCycler has added methods for confirming no cycle
    - TestBasicCycler is new. before it was impossible to directly test Cycler.

axelrod/strategies/cycler.py
```diff
@@ -1,13 +1,15 @@
...
 import copy
 
+C, D = Actions.C, Actions.D
+
 
 class AntiCycler(Player):
     """
     A player that follows a sequence of plays that contains no cycles:
-    C CD CCD CCCD CCCCD CCCCCD ...
+    CDD  CD  CCD CCCD CCCCD ...
```
marcharper: Was this wrong originally? It's supposed to be n-Cs and 1-D for n in 0, 1, 2, ...

axelrod/tests/strategies/test_calculator.py
```diff
@@ -20,31 +21,90 @@ class TestCalculator(TestPlayer):
...
-        # Test post 20 rounds responses
-        self.responses_test([D], [C] * 21, [C] * 21)
-        history = [C, C, D, C, C, D, C, C, C, D, C, C, C, C, D, C, C, C, C, C,
-                   D]
-        self.responses_test([D], [C] * 21, history)
-        history = [C, C, D, C, C, D, C, C, C, D, C, C, C, C, D, C, C, C, C, C,
-                   D, C]
-        self.responses_test([C], [C] * 22, history)
+    def test_twenty_rounds_joss_then_defects_for_cyclers(self):
+        """uses axelrod.strategies.axelrod_first.Joss strategy for first 20 rounds"""
```
marcharper: nit: Capitalize the docstings (others below as well)

axelrod/strategies/cycler.py
eric-s-s: the first six to eight moves are a cycle.  CCDCCD(CC).  

are you saying it should have beem D CD CCD CCCD CCCCD?  the original was what the old docstring said. 

D CD CCD ... would still trigger detect_cycle on round 4.

eric-s-s: this test in the link would fail on the original code.
[https://github.com/eric-s-s/Axelrod/blob/64ec1819521550cdeb4c7902313d3068da7f3a8e/axelrod/tests/strategies/test_cycler.py#L27-L35](url)

axelrod/tests/strategies/test_calculator.py
eric-s-s: Thank you.  I missed that one. 
 
> (others below as well)

were there more?

eric-s-s: and, of course, when you go to check for one, you discover five more.

PR #926, opened 2017-03-24
Add SlowTitForTwoTats2 Strategy

jtsmith2:
Strategy from PRISON(1998) that plays C twice for opening. If opponent plays same move twice in a row, the player mimics that move, otherwise just plays it's previous move.

'slow_tft' from PRISON list in #379

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -75,7 +75,7 @@ from .titfortat import (
     TitForTat, TitFor2Tats, TwoTitsForTat, Bully, SneakyTitForTat,
     SuspiciousTitForTat, AntiTitForTat, HardTitForTat, HardTitFor2Tats,
     OmegaTFT, Gradual, ContriteTitForTat, SlowTitForTwoTats, AdaptiveTitForTat,
-    SpitefulTitForTat)
+    SpitefulTitForTat, SlowTitForTwoTats2)
 from .verybad import VeryBad
 from .worse_and_worse import (WorseAndWorse, KnowledgeableWorseAndWorse,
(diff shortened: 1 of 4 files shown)


axelrod/strategies/titfortat.py
```diff
@@ -602,3 +602,37 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        # Start with two cooperations
+        if len(self.history) < 2:
+            return C
+
+        # Mimic if opponent plays the same move twice
+        if opponent.history[-2] == opponent.history[-1]:
+            return opponent.history[-1]
+
+        # Otherwise cooperate
```
drvinceknight: `Otherwise play own previous move`

axelrod/strategies/_strategies.py
```diff
@@ -75,7 +75,7 @@
     TitForTat, TitFor2Tats, TwoTitsForTat, Bully, SneakyTitForTat,
     SuspiciousTitForTat, AntiTitForTat, HardTitForTat, HardTitFor2Tats,
     OmegaTFT, Gradual, ContriteTitForTat, SlowTitForTwoTats, AdaptiveTitForTat,
-    SpitefulTitForTat)
+    SpitefulTitForTat, SlowTitForTwoTats2)
```
drvinceknight: You also need to add this strategy to the `all_strategies` list in this file.

axelrod/strategies/titfortat.py
jtsmith2: Doh! Fixed in newest push.

axelrod/strategies/_strategies.py
jtsmith2: Should be fixed now.

PR #931, opened 2017-03-25
test_selfsteem updated

alajara:


alajara:
Its giving error indentation problem, I believe the actions were to long.

drvinceknight:
> Its giving error indentation problem, I believe the actions were to long.

Hi @alajara, no the actions are fine, the reason this is failing is because the test is not in the correct place. You needed to make the changes to the file that is already there: `Axelrod/axelrod/tests/strategies/test_selfsteem.py`. You have created a new file: `Axelrod/test_selfsteem.py`.

Once you put the changes in the correct place we'll be able to see if things are right or not :)

alajara:
I was trying to push it thru CLI using git on my raspberry pit. I was trying to learn how to do it, looks like I am not doing it properly, back on my other machine, ill fix it right away.

Issue #941, opened 2017-03-28
a way to sidestep inspecting source code.


eric-s-s:
I was refactoring the tests for Darwin and came across the use of the "inspect" module.  This slows down the strategy by about a factor of 20ish.  same goes for Geller and MindReader.  

I came up with a workaround.  
I have a PR that, while not great, clearly demonstrates the idea and will submit a PR that references this as soon as this issue is up and running.

here's the gist of my solution.

Player has a new method

```
def strategy_for_simulation(self, opponent):
    return self.strategy(opponent)
```
in _strategy_utils, limited_simulated_play now calls strategy_for_simulation.  all other simulation methods in _strategy_utils rely on that method.

Darwin, Geller and MindReader override the method so that they return some default when another strategy tries to run a simulation on them.  

Geller's strategy method calls strategy_for_simulation for its opponent.   
MindReader's strategy method calls _strategy_utils.look_ahead which call strategy_for_simulation.

all regular testing is unaffected as that relies on mock_player.py which uses the "strategy" method

on my computer, this cut testing time by over a factor of ten for each strategy.
Darwin: 0.209s -> 0.004s
Geller:  3.167s -> 0.011s
MindReader: 3.451s -> 0.209s

drvinceknight:
I'm not sure I like the sound of this idea: these strategies are really niche and artefacts of the start of their library. They're labelled as `cheaters` so don't get used often. I don't think it's worth spending too much time on this and also I'm not a fan of adding a method to the `Player` class just to deal with this weird strategies.

drvinceknight:
A more succinct version: I'm hesitant to add a (minor) layer of abstraction/complexity for the sake of performance in this (niche) case.

But let's wait to hear what @marcharper and @meatballs have to say.

eric-s-s:
got it.  
so are these strategies basically dead-ends, then?  could/should they be deleted?

drvinceknight:
> so are these strategies basically dead-ends, then? could/should they be deleted?

That's certainly a valid/logical pov. I would suggest they're a bit of fun so would personally like to see them stay in the library but I don't think we should worry about optimising them.

eric-s-s:
yes, that all makes a lot of sense.  i do have the tendency to get caught up on small details, in case the previous PR did not make that obvious.

meatballs:
[...]

PR #942, opened 2017-03-28
No inspect source

eric-s-s:
this is actually more meant as a demonstration of the idea for #941 .  i did this quick-and-dirty as a proof of concept.


drvinceknight:
Have responded on #941.

eric-s-s:
when it comes time to kill this PR, i hit the "close" button, correct?

meatballs:
> when it comes time to kill this PR, i hit the "close" button, correct?

Yep

PR #943, opened 2017-03-29
Update test_stalker.py

alajara:


marcharper:
Hi @alajara . Thanks for the contribution. According to the coveralls we're not covering line 69 anymore. Can you take a look and add or modify a test to make sure it's covered?

Please also take a look at all the tests to make sure they are equivalent to the former tests.

alajara:
that one test is giving me a hard time, once the current average score is higher than Cooperation it should defect. Can I get some help on this one.

axelrod/tests/strategies/test_stalker.py
```diff
@@ -21,26 +21,39 @@ class TestStalker(TestPlayer):
     }
 
     def test_strategy(self):
+        
         # Start with cooperation
         self.first_play_test(C)
 
         # current_average_score > very_good_score
-        self.responses_test([D], [C] * 2 + [D] * 4, [D] * 2 + [C] * 4)
```
marcharper: There are two tests here but one replacement. I think if we transcribe the tests exactly we should have the coverage we had before.

This line:
```
self.responses_test([D], [C] * 2 + [D] * 4, [D] * 2 + [C] * 4)
```

corresponds to

```
actions = [(C, D)] * 2 + [(D, C)] * 4 + [(D, *)]
```
where * indicates a missing value from the original test (since the original test only looked at the player's next move and not the opponent's).

alajara: @marcharper I try changing it to its original state and its still not testing the current_average > very_good_score. Any guidance on how to fix it. Thanks

drvinceknight: @alajara I've opened https://github.com/alajara/Axelrod/pull/4 to this branch.

It include the test that's missing here. It was a bit tricky to figure out how to get the `current_average > very_good_score` and I do that by tricking the player to defecting by lying to it about the length of the Match. 

Note that that PR also include a merge from master, if you click merge on the PR, it'll automatically appear here and should be good to go.

drvinceknight:
@alajara there was a slight error in my PR (apologies).

The first import line needs to be:

```
import axelrod as axl
```

It is currently:

```
import axl as axl
```

drvinceknight:
I've pushed a fix to your branch on this @alajara. :+1: Assuming tests pass I believe this is good to go.

marcharper:
What happened to the commits here?

drvinceknight:
> What happened to the commits here?

What do you mean? Are you asking why my user name is appearing on them? I don't know. I think this is just a github thing because I pushed to the branch.

marcharper:
Let's squash when we merge this time

drvinceknight:
> Let's squash when we merge this time

I agree. This looks good to me: feel free to merge @marcharper.

Issue #802, opened 2017-01-01
Automatically set strategy classes (and transformers) __repr__ to include parameters.


drvinceknight:
As an example:

Should:

```
>>> NoisyCooperator = NoisyTransformer(0.5)(axl.Cooperator)
>>> player = NoisyCooperator()
```

which currently gives:

```
>>> player
Noisy Cooperator
```

give:

```
>>> player
Noisy Cooperator (0.5)
```

?

marcharper:
Yes, it should. Do we want to change the name, or the `__repr__` function?

drvinceknight:
We're currently adding a prefix to the name so I suggest doing that?

marcharper:
I think name is fine for now although we do have some redundancy / overlap with the class name and the `__repr__` method, and it might be better for the transformer to wrap the `__repr__` method.

Chadys:
Instead of making it the strategy responsibility to change its \_\_repr\_\_ method according to its parameters like Random does, wouldn't it be more efficient to change Player.\_\_repr\_\_ to something like this : 
```python
def __repr__(self):
     name = self.name
     sig = inspect.signature(self.__init__)
     prefix = ': '
     for param in sig.parameters:
         if param != 'self':
             name += prefix+str(getattr(self,param, ''))
             prefix = ', '
     return name
```
Tested on Random, returns `Random: 0.5`

marcharper:
Interesting idea! For some strategies this could produce very long names depending on the parameters. I think we need to check that this produces good results for all strategies. We'd also need to document the automatic name generation and give instructions on how to override.

I recommend rewriting the function a bit to use the stored `__init__` parameters that `Player.__new__` places in `self.init_args` and `self.kwargs`, which I think eliminates the need for `inspect` in this case. The parameter extraction should have already been done before `__repr__` is called. We should also probably check that `class.name` makes sense for unsubstantiated players.

drvinceknight:
> Interesting idea! For some strategies this could produce very long names depending on the parameters. I think we need to check that this produces good results for all strategies. We'd also need to document the automatic name generation and give instructions on how to override.

I really like this idea and @marcharper's suggestions too. :+1:

Chadys:
[...]

PR #946, opened 2017-03-29
Remove init_args from Player

### LINKED ISSUE #945: Remove init_args
Now that all strategies are using kwargs instead of args, we can remove init_args from the Player class.

meatballs:
Fixes #945

### DIFF
--- a/axelrod/tests/strategies/test_player.py
+++ b/axelrod/tests/strategies/test_player.py
@@ -115,7 +115,7 @@ class TestPlayerClass(unittest.TestCase):
 
     def test_clone(self):
         """Tests player cloning."""
-        player1 = axelrod.Random(0.75)  # 0.5 is the default
+        player1 = axelrod.Random(p=0.75)  # 0.5 is the default
         player2 = player1.clone()
         turns = 50
         for op in [axelrod.Cooperator(), axelrod.Defector(),
@@ -284,7 +284,7 @@ class TestPlayer(unittest.TestCase):
         turns = 50
         r = random.random()
         for op in [axelrod.Cooperator(), axelrod.Defector(),
-                   axelrod.TitForTat(), axelrod.Random(r)]:
+                   axelrod.TitForTat(), axelrod.Random(p=r)]:
             player1.reset()
             player2.reset()
             seed = random.randint(0, 10 ** 6)
--- a/axelrod/tests/strategies/test_player.py
+++ b/axelrod/tests/strategies/test_player.py
@@ -62,8 +62,7 @@ class TestPlayerParameters(unittest.TestCase):
         # Test that passing a keyword argument successfully sets the
         # init_kwargs dict
         player = ParameterisedPlayer(parameter1='test')
-        assert 'parameter1' in player.init_kwargs
-        assert player.init_kwargs['parameter1'] == 'test'
+        assert player.init_kwargs == {'parameter1': 'test'}
 
         # Test that passing a positional argument raises an error
         with self.assertRaises(TypeError):
--- a/axelrod/tests/strategies/test_player.py
+++ b/axelrod/tests/strategies/test_player.py
@@ -27,6 +27,49 @@ def randomize(*args):
     return random.choice([C, D])
 
 
+class ParameterisedPlayer(Player):
+    """A simple Player class for testing init parameters"""
+
+    def __init__(self, parameter1=None):
+        super().__init__()
+
+
+class TestPlayerParameters(unittest.TestCase):
+    """A class to test the initialisation parameters of Players"""
+
+    def test_player(self):
+        """Tests for Players with no init parameters"""
+
+        # Test that the init_kwargs dict exists and is empty
+        player = Player()
+        assert player.init_kwargs == {}
+
+        # Test that passing a postional argument raises an error
+        with self.assertRaises(TypeError):
+            player = Player('test')
+
+        # Test that passing a keyword argument raises an error
+        with self.assertRaises(TypeError):
+            player = Player(parameter1='test')
+
+    def test_parameterised_player(self):
+        """Tests for Players with init parameters"""
+
+        # Test that the init_kwargs dict exists and is empty
+        player = ParameterisedPlayer()
+        assert player.init_kwargs == {}
+
+        # Test that passing a keyword argument successfully sets the
+        # init_kwargs dict
+        player = ParameterisedPlayer(parameter1='test')
+        assert 'parameter1' in player.init_kwargs
+        assert player.init_kwargs['parameter1'] == 'test'
+
+        # Test that passing a positional argument raises an error
+        with self.assertRaises(TypeError):
+            player = ParameterisedPlayer('test')
+
+
 class TestPlayerClass(unittest.TestCase):
 
     name = "Player"
--- a/axelrod/tests/strategies/test_player.py
+++ b/axelrod/tests/strategies/test_player.py
@@ -64,6 +64,10 @@ class TestPlayerParameters(unittest.TestCase):
         player = ParameterisedPlayer(parameter1='test')
         assert player.init_kwargs == {'parameter1': 'test'}
 
+        # Test that passing an unknown keyword argument raises an error
+        with self.assertRaises(TypeError):
+            player = ParameterisedPlayer(parameter2='test')
+
         # Test that passing a positional argument raises an error
         with self.assertRaises(TypeError):
             player = ParameterisedPlayer('test')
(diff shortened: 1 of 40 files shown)


meatballs:
Agggh!!! I ran the unit tests, but forgot to run the strategy tests. Back later......

marcharper:
Looks good to me. We might consider renaming some parameters in another PR -- `epsilon` instead of `ep`, replace `p` in `Random(p)` with something more informative.

axelrod/tests/strategies/test_worse_and_worse.py
```diff
@@ -92,18 +92,18 @@ def test_strategy(self):
...
+        self.versus_test(opponent=axelrod.MockPlayer(actions=[C,C,D,C]), expected_actions = actions)
 
         actions = [(C, C)] * 18 + [(C, D), (D, C)]
-        self.versus_test(opponent=axelrod.MockPlayer([C] * 18 + [D, C]), expected_actions = actions)
+        self.versus_test(opponent=axelrod.MockPlayer(actions=[C] * 18 + [D, C]), expected_actions = actions)
 
         # After round 20, strategy follows stochastic behavior given a seed
         actions = [(C, C)] * 20 + [(C, D), (D, C), (C, C), (C, D)] 
-        self.versus_test(opponent=axelrod.MockPlayer([C] * 20 + [D, C, C, D]), expected_actions = actions, 
+        self.versus_test(opponent=axelrod.MockPlayer(actions=[C] * 20 + [D, C, C, D]), expected_actions = actions,
```
drvinceknight: Is it worth tweaking the `MockPlayer` so that it can be passed positional arguments?

If not, the docs for the tests need to include `actions=` http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html

drvinceknight:
> Looks good to me. We might consider renaming some parameters in another PR -- epsilon instead of ep, replace p in Random(p) with something more informative.

This suggestion and this PR introduce some backwards incompatibility right?

I'm not necessarily against it but I just want to make sure we think it through. 

I understand one of the motivating factors: but I'd just like to make sure we'd make these decisions even if the API wasn't in consideration. 

- we'll be breaking any code written with verbose parameter input (whether or not this is what people "should" be doing is not what I'm concerned about);
- changing parameter names would break that as well (perhaps instead of changing parameter names we just make sure they're well described in the docstring which is also accessible to the API).

Again: not against this (I'd certainly prefer better variable names and like the verbose calls), just want to make sure we think it through.

axelrod/tests/strategies/test_worse_and_worse.py
drvinceknight: (There are some ongoing work on #884 that might fall over if this goes in as is: that's not a huge deal but something to consider.)

meatballs:
Yes, we would be introducing some backward incompatibility - any code that uses the library and has positional arguments in Player instantiation would fail.

axelrod/tests/strategies/test_player.py
```diff
@@ -27,6 +27,52 @@ def randomize(*args):
...
+
+class TestPlayerParameters(unittest.TestCase):
+    """A class to test the initialisation parameters of Players"""
+
+    def test_player(self):
+        """Tests for Players with no init parameters"""
+
+        # Test that the init_kwargs dict exists and is empty
+        player = Player()
+        assert player.init_kwargs == {}
```
drvinceknight: Could this be: `self.assertEquals(player.init_kwards, {})`.

I believe that `assert` would not run in the same fashion: a failure would give an error as opposed to a test failure (which are reported differently by unittest).

axelrod/tests/strategies/test_player.py
```diff
@@ -27,6 +27,52 @@ def randomize(*args):
...
+        # Test that passing a keyword argument raises an error
+        with self.assertRaises(TypeError):
+            player = Player(parameter1='test')
+
+    def test_parameterised_player(self):
+        """Tests for Players with init parameters"""
+
+        # Test that the init_kwargs dict exists and is empty
+        player = ParameterisedPlayer()
+        assert player.init_kwargs == {}
```
drvinceknight: `self.assertEquals(player.init_kwards, {})`

axelrod/tests/strategies/test_player.py
```diff
@@ -27,6 +27,52 @@ def randomize(*args):
...
+
+        # Test that passing a keyword argument successfully sets the
+        # init_kwargs dict
+        player = ParameterisedPlayer(parameter1='test')
+        assert player.init_kwargs == {'parameter1': 'test'}
```
drvinceknight: self.assertEquals(player.init_kwards, {''parameter1': 'test'})

drvinceknight:
I'm marking this as good to go but will leave @marcharper to merge just in case anyone wants to talk about the backwards incompatibility some more. My point of view:

- This enables a fix of a problem in the library: #802
- This makes no change to how strategies are written (so good for newcomers)
- The backwards compatibility is minor and arguably enforces good practice.
- Bonus: it helps the API.

meatballs:
> Bonus: it helps the API.

Stronger than that. The API can't handle parameterised strategies without it.

drvinceknight:
> The API can't handle parameterised strategies without it.

(Although you could do the inspect "thing" there right? But anyway, this works towards #802 so :+1:)

axelrod/tests/strategies/test_player.py
meatballs: Sorry - too used to using py.test on other projects. Should be sorted now.

meatballs:
> (Although you could do the inspect "thing" there right? 

No - that would still rely on having only keyword args

drvinceknight:
> No - that would still rely on having only keyword args

Ah yes, I'm confusing this with your proposed fix to #802 yes?

meatballs:
The API needs to present 'something' to a user interface to define the parameters that are necessary for a strategy. Positional args have no information to present whereas keyword args have names and default values.

drvinceknight:
Yup. With you. :+1:

Chadys:
Python has no positional arguments in definition, only in call, so the inspect in the API would have worked anyway and we could have had "The Random strategy has a 'p' argument with a default value of 0" by using `inspect.signature(Random.__init__).bind_partial().applydefault()` independently of this PR. But no matter, `init_params` and `__repr__` will be easier for me this way and I will make use of your TestPlayerParametized for #802 :)

meatballs:
@Chadys Yep, you're right. I take back most of what's written above.

drvinceknight:
> Yep, you're right. I take back most of what's written above.

So to confirm, the reasons in favour of this are independent of the API and we should make our decision about the negative aspects of the backwards incompatibility independent of the API.

My thoughts remain the same:

> This enables a fix of a problem in the library: #802
> This makes no change to how strategies are written (so good for newcomers)
> The backwards incompatibility is minor and arguably enforces good practice.
> Bonus: it helps the API.

meatballs:
/me goes to crawl back under a rock

Chadys:
Well I don't **need** it for #802, it makes it a little bit easier is all, I'm only waiting for this PR to avoid merge conflicts and because it changes how I implement it. Either I use `*args` and `**kwargs` to populate `init_kwargs` in `__new__` or I only use `**kwargs` if  `*args` is disabled, but both are doable.

drvinceknight:
It sounds like this is ultimately a decision to enforce the use of `kwargs` when initialising a player or not, which make me less excited.

drvinceknight:
@Chadys am I missing something? I've tried your `inspect` call and don't seem to get the expected output:

```python
>>> import axelrod as axl
>>> import inspect
>>> inspect.signature(axl.Random.__init__).bind_partial().bind_partial()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-a7f3f3aa97c5> in <module>()
      1 import axelrod as axl
      2 import inspect
----> 3 inspect.signature(axl.Random.__init__).bind_partial().bind_partial()

AttributeError: 'BoundArguments' object has no attribute 'bind_partial'
```

I've tried to have a (lazy) quick look around and don't seem to find anything.

drvinceknight:
Although this seems to give something:

```python
>>> inspect.signature(axl.Random.__init__).bind_partial().signature
<Signature (self, p:float=0.5) -> None>
```

Chadys:
you called `bind_partial` twice. try this : 
```python
>>> import inspect, axelrod as axl
>>> b = inspect.signature(axl.Random.__init__).bind_partial()
>>> b.apply_defaults()
>>> b.arguments
OrderedDict([('p', 0.5)])
```
(not that you can't chain `bind_partial` and `apply_defaults` as `apply_defaults` return None

drvinceknight:
Got it thanks. :+1:

drvinceknight:
My preference would be to close this PR. I don't feel enforcing the use of `kwargs` alone is a good thing which if I've understood everything correctly is now what this PR would achieve.

If anyone **does** feel that only allowing `kwargs` input is perhaps worth enforcing then I suggest we open an issue about that and discuss it there.

PR #948, opened 2017-03-29
884 d

eric-s-s:
#884


axelrod/_strategy_utils.py
```diff
@@ -68,13 +69,13 @@ def calculate_scores(p1, p2, game):
...
     """Looks ahead for `rounds` and selects the next strategy appropriately."""
     results = []
-
     # Simulate plays for `rounds` rounds
+    players = {C: Cooperator(), D: Defector()}
     strategies = [C, D]
     for strategy in strategies:
         # Instead of a deepcopy, create a new opponent and play out the history
         opponent_ = player_2.clone()
-        player_ = Cycler(strategy)  # Either cooperator or defector
```
eric-s-s: this used knowledge that Actions.C = 'C' and Actions.D = 'D'.

axelrod/strategies/darwin.py
```diff
@@ -44,16 +43,18 @@ class Darwin(Player):
...
         self.response = Darwin.genome[0]
+        super().__init__()
 
     def receive_match_attributes(self):
         self.outcomes = self.match_attributes["game"].scores
 
     def strategy(self, opponent: Player) -> Action:
         # Frustrate psychics and ensure that simulated rounds
         # do not influence genome.
-        if inspect.stack()[1][3] not in Darwin.valid_callers:
```
eric-s-s: oops.  didn't notice that i changed this.  i had been trying to understand inspect.stack().  Change it back, or no need.

marcharper: two lines is fine

axelrod/tests/strategies/test_darwin.py
```diff
@@ -15,25 +15,41 @@ class TestDarwin(TestPlayer):
         'stochastic': False,
         'makes_use_of': set(),
         'long_run_time': False,
-        'inspects_source': False,
+        'inspects_source': True,
         'manipulates_source': False,
         'manipulates_state': True
     }
 
+    def setUp(self):
```
marcharper: Can we use `setup` or `set_up`? We usually reserve camelCase for classes.

axelrod/tests/strategies/test_defector.py
```diff
@@ -44,9 +44,18 @@ def test_strategy(self):
...
-        opponent_history = [C, C, C, C, D] + [D] + [C] * 10
-        self.responses_test([D], history, opponent_history)
+
+    def test_cooperates_if_opponent_history_has_C_and_last_three_are_D(self):
+        opponent_actions = [D, C, D, D, D] + [D, D]
+        actions = [(D, D), (D, C), (D, D), (D, D), (D, D)] + [(C, D), (C, D)]
+        self.versus_test(axelrod.MockPlayer(opponent_actions), actions)
+
+    def test_defects_if_opponent_never_cooperated(self):
+        opponent_actions = [D, D, D, D, D] + [D, D]
```
marcharper: Let's use `[D]*7` or a single list

axelrod/tests/strategies/test_defector.py
```diff
@@ -44,9 +44,18 @@ def test_strategy(self):
...
+        actions = [(D, D), (D, D), (D, D), (D, D), (D, D)] + [(D, D), (D, D)]
```
marcharper: same

axelrod/tests/strategies/test_defector.py
```diff
@@ -44,9 +44,18 @@ def test_strategy(self):
...
+        self.versus_test(axelrod.MockPlayer(opponent_actions), actions)
+
+    def test_defects_if_opponent_last_three_are_not_D(self):
+        opponent_actions = [D, C, D, D, D] + [C, D]
```
marcharper: single list

axelrod/tests/strategies/test_defector.py
```diff
@@ -44,9 +44,18 @@ def test_strategy(self):
...
+        actions = [(D, D), (D, C), (D, D), (D, D), (D, D)] + [(C, C), (D, D)]
```
marcharper: same

axelrod/tests/strategies/test_defector.py
```diff
@@ -44,9 +44,18 @@ def test_strategy(self):
         # Starts by defecting.
         self.first_play_test(D)
         self.second_play_test(D, D, D, D)
-        # Test if strategy tries to trick opponent.
```
marcharper: Are we covering all of the previous behaviors with the new tests?

axelrod/tests/strategies/test_darwin.py
eric-s-s: unfortunately, no.  this is overriding the method of unittest.TestCase.setUp which gets called before every test.

axelrod/tests/strategies/test_defector.py
eric-s-s: i think i got them.  i tried to cover all the behavior more explicitly here.  that was why i separated out the lists.  i was trying to show where the deviations would occur (or mirror it and fail to occur) without making the reader scan a list of random c's and d's.

axelrod/tests/strategies/test_defector.py
```diff
@@ -44,9 +44,18 @@ def test_strategy(self):
...
-        self.responses_test([D], [C, C, C], [C, C, C])
-        self.responses_test([D], [C, C, C, D, D], [C, C, C, C, D])
-        history = [C, C, C, D, D] + [C] * 11
-        opponent_history = [C, C, C, C, D] + [D] + [C] * 10
-        self.responses_test([D], history, opponent_history)
+
+    def test_cooperates_if_opponent_history_has_C_and_last_three_are_D(self):
```
eric-s-s: is this way ok?  i prefer this to a long list (to me, it more clearly shows the expected behavior), but i'll do long list.

drvinceknight: This is completely fine :)

axelrod/tests/strategies/test_darwin.py
```diff
@@ -15,29 +15,45 @@ class TestDarwin(TestPlayer):
...
+
     def test_strategy(self):
         p1 = self.player()
-        p2 = axelrod.Cooperator()
-        self.assertEqual(p1.strategy(p2), C)  # Always cooperate first.
-        for i in range(10):
-            p1.play(p2)
-        self.assertEqual(p1.strategy(p2), C)
+        p1.reset()
+        self.versus_test(axelrod.Cooperator(), [(C, C)] * 5)
```
drvinceknight: could you write in `expected_actions=[(C, C)] * 5` (and elsewhere: verbosity :+1: )

axelrod/tests/strategies/test_darwin.py
```diff
@@ -15,29 +15,45 @@ class TestDarwin(TestPlayer):
...
+        self.assertEqual(p1.genome, [C] * 5)
```
drvinceknight: You don't need this extra line, it can be done in the `versus_test`:

From http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html:

> The function versus_test also accepts a dictionary parameter of attributes to check at the end of the match. For example this test checks if the player’s internal variable opponent_class is set to "Cooperative":

This would look like:

```
self.versus_test(axelrod.Cooperator(), expected_actions=[(C, C)] * 5, attrs={"genome": [5] * 5})
```

I've got to go to class now so will review the rest later but if you catch things like this that'd be great :)

axelrod/tests/strategies/test_defector.py
eric-s-s: as far as i can tell, the old test never showed TrickyDefector being tricky.  the new tests include both normal and tricky behavior.

axelrod/tests/strategies/test_darwin.py
eric-s-s: well that's cool! :)

axelrod/tests/strategies/test_darwin.py
eric-s-s: done in test_darwin.

i'm assuming things like the code below, where the variable actions is clearly named, do not need this?

```
def test_some_guy(self):
    actions = [(C, C)] * 10
    self.versus_test(Cooperator(), actions)
```

or is it still better to do like this?

```
def test_some_guy(self):
    actions = [(C, C)] * 10
    self.versus_test(Cooperator(), expected_actions=actions)
```

i'll await an answer before next push.

drvinceknight: We have been going with `self.versus_test(Cooperator(), expected_actions=actions)` which (sorry) I'd prefer :) :+1:

eric-s-s: done!

eric-s-s:
@drvinceknight 

> We have been going with `self.versus_test(Cooperator(), expected_actions=actions)`

righto.  there were some of those missing in the old tests i pushed.  so, while i'm hitting ctrl-v, i can do it a few more times.

drvinceknight:
Thanks, this looks good to me now.

PR #950, opened 2017-03-31
an idea for Geller, Darwin, MindReader

eric-s-s:
#941 

okay.  an idea flashed into my head on how to solve this without affecting anyone but the three cheaters.  i spent all of 10 minutes implementing it (as you'll see by the naming and commented-out code.)

if the general idea is ok, i think it's worth spending the time to make this a presentable PR.  aside from getting that same feeling as when i try to walk past an unfinished jigsaw puzzle, using inspect.stack() is just asking to be broken.  if anything changes in play, it may or may not require some weird maintenance for these guys.  this should solve that.

the changes only involve the strategy_utils function written specifically for the cheaters, so no one else should know or ever care about the code change there.  

the other changes are adding a specific method to the cheaters to deal with cheaters that, once again, the whole rest of the code can remain blissfully ignorant of.


axelrod/_strategy_utils.py
```diff
@@ -45,11 +45,19 @@ def limited_simulate_play(player_1, player_2, h1):
     player_2's strategy method to set any internal variables as needed. If you
     need a more complete simulation, see `simulate_play` in player.py. This
     function is specifically designed for the needs of MindReader."""
-    h2 = player_2.strategy(player_1)
+    h2 = cheating_bastard(player_1, player_2)
     update_history(player_1, h1)
     update_history(player_2, h2)
 
 
+def cheating_bastard(cheater, opponent):
```
drvinceknight: This function name will need to change. Similarly `cheater` needs to be something more verbose, I've been using the word `cheater` as shorthand, but technically it's just that it does not obey Axelrod's original rules.

I suggest:

```python
def simulated_strategy(player, opponent):
    ```
    Returns the simulated strategy of a player against an opponent. If the strategy has an implemented `simulation_strategy` then that gets called. 
    ```
```

drvinceknight: Can you also write one or two tests for this please.

eric-s-s: no problem.  i'll put some work in making this more presentable, make tests for the simulated_strategies, and some tests for cheating_bastard (are you sure that name needs changing? ;) ).  

names:

- `inspect_strategy(inspector, opponent):` 
- `foil_strategy_inspections(self, opponent):`

how do those sound?

eric-s-s: > I noticed one of your commit messages about mypy, any problems running that locally? (It's pip installable).

i had some weird issues that took working out.  for some reason, the mypy command doesn't make doesn't add to any kind of PATH variable.  so, with all this installed in a virtualenv i had to write a shell script pointing directly to a batch file the environment's Scripts folder.  all in all, the workaround needed 

- a python script to get names from run_mypy.py into a csv file
- the csv file
- a shell script to run the first one, load the second one and then run each variable through a path to the mypy.bat file.  

not pretty, but i've barely worked with bash before, so that was useful.

drvinceknight: Eeeesh 

> names:
> 
> inspect_strategy(inspector, opponent):
> foil_strategy_inspections(self, opponent):
> how do those sound?

That sounds good to me :+1:

> i also had issues understanding what the problem was. i was not aware of the variable declaration as an inline comment.

Eeeesh that sounds difficult, I can't offer any assistance I'm afraid...

axelrod/strategies/geller.py
```diff
@@ -47,27 +46,25 @@ class Geller(Player):
...
+
+        return inspect_strategy(self, opponent)
 
 
 class GellerCooperator(Geller):
     """Observes what the payer will do (like :code:`Geller`) but if unable to
     will cooperate.
     """
     name = 'Geller Cooperator'
-    default = lambda self: C
```
eric-s-s: the IDE balked at making a variable a lambda and recommended changing to a function.  so i did.

axelrod/tests/strategies/test_darwin.py
```diff
@@ -47,23 +51,6 @@ def test_against_geller_and_mindreader(self):
 
         self.versus_test(axelrod.MindReader(), expected_actions=[(C, D)] * 2, attrs={'genome': [D, C]})
 
-    def test_play(self):
```
eric-s-s: when i refactored test_darwin, i forgot to recommend removing test_play(self).  because versus_test calls "play" these are now redundant.

eric-s-s:
for the test_<strategy> modules, i did not refactor the modules.  i only added the test for foil_strategy_inspection.  i'll be refactoring test_geller in the next few days in a PR to #884.  test_darwin just got refactored and test_mindreader will be handled by someone (possibly me).

axelrod/strategies/darwin.py
```diff
@@ -44,18 +41,18 @@ class Darwin(Player):
...
     def receive_match_attributes(self):
         self.outcomes = self.match_attributes["game"].scores
 
-    def strategy(self, opponent: Player) -> Action:
-        # Frustrate psychics and ensure that simulated rounds
-        # do not influence genome.
-        if inspect.stack()[1][3] not in Darwin.valid_callers:
-            return C
+    @staticmethod
+    def foil_strategy_inspection() -> Action:
```
drvinceknight: Could you add a docstring here please.

axelrod/strategies/geller.py
```diff
@@ -47,27 +46,25 @@ class Geller(Player):
         'manipulates_state': False
     }
 
+    @staticmethod
+    def foil_strategy_inspection() -> Action:
```
drvinceknight: Docstring.

axelrod/strategies/mindreader.py
```diff
@@ -90,18 +83,10 @@ class MirrorMindReader(ProtectedMindReader):
...
     }
 
-    def strategy(self, opponent: Player) -> Action:
-        """Will read the mind of the opponent and play the opponent's strategy.
-
-        Also avoid infinite recursion when called by itself or another mind
-        reader or bender by cooperating.
-        """
+    @staticmethod
+    def foil_strategy_inspection() -> Action:
```
drvinceknight: Docstring please.

PR #951, opened 2017-03-31
More type hints

MariosZoulias:
#808


drvinceknight:
This looks good @MariosZoulias 

(To Marc or Owen: I have checked and Marios has added `finite_state_machines.py` and `human.py` to the `run_mypy.py` script the other two strategies were already there: `hmm.py` and `prober.py`).

PR #952, opened 2017-03-31
Use kwargs for Player initialisation throughout library

### LINKED ISSUE #945: Remove init_args

meatballs:
Fixes #945


marcharper:
I approve -- not sure about the merge order so I'll let you all push the button since I've not been as involved in the discussions.

PR #955, opened 2017-04-01
842 - Approximate Moran Process

### LINKED ISSUE #842: Merge in Approximate Moran Process
From the Moran repository.

drvinceknight:
Closes #842 

Adds an approximate Moran process that can take a sample of match outcomes (so that these are not rerun every time).

This also makes a `random_.Pdf` class available, which is a generic probability distribution function with a `sample` method (could be useful elsewhere).

### DIFF
--- a/axelrod/tests/unit/test_random_.py
+++ b/axelrod/tests/unit/test_random_.py
@@ -1,8 +1,9 @@
 """Tests for the random functions."""
+from collections import Counter
 import random
 import unittest
 import numpy
-from axelrod import random_choice, seed, Actions
+from axelrod import random_choice, seed, Actions, Pdf
 
 C, D = Actions.C, Actions.D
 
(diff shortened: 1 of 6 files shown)


axelrod/tests/unit/test_random_.py
```diff
@@ -39,3 +40,37 @@ def test_seed_not_offset_by_deterministic_call(self):
             seed(0)
             random_choice(p)
             self.assertEqual(r, random.random())
+
+
+class TestPdf(unittest.TestCase):
+    """A suite of tests for the Pdf class"""
+    observations = [('C', 'D')] * 4 + [('C', 'C')] * 12 + \
```
marcharper: `C` and `D` instead of `'C'` and `'D'`

drvinceknight: Good catch. Bad habits die hard. Done :+1:

PR #959, opened 2017-04-02
Override the repr for the Human strategy.

### LINKED ISSUE #958: The `__repr__` for the human strategy includes an extra `C` and `D`.
This is a result of #953 which now automatically add the various attributes to the players so it looks like:

```python
>>>vince = axl.Human()
>>> vince
Human: human, C, D
```

You can pass different symbols for the `C` and `D`:

```python
>>> vince = axl.Human(name="Vince", c_symbol=":)", d_symbol=":(")
>>> vince
Human: Vince, :), :(
```

I suggest that in this case we can override the `repr` to not display the `c_symbol` and the `d_symbol`. I think that as the Human player is only going to be used at a repl so it's ok to clear it up a bit (whereas normally we want the `__repr__` to be precise).

drvinceknight:
Closes #958.

### DIFF
--- a/axelrod/strategies/human.py
+++ b/axelrod/strategies/human.py
@@ -157,3 +157,9 @@ class Human(Player):
             action = input_function()
 
         return action
+
+    def __repr__(self):
+        """
+        Override the default __repr__ of the class
+        """
+        return "Human: {}".format(self.human_name)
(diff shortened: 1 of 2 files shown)


marcharper:
This is fine but perhaps there's a better place for c_symbol and d_symbol -- maybe in the Actions class. There are similar parameters for sparklines (in Match).

drvinceknight:
:+1: I'm throwing the `ready to merge` label on this (and have done on #955) as a visual prompt to Owen but please throw it off if you had left it off on purpose because you wanted to look at it again.

P.S. Open to moving the symbol commands (although they feel natural to me here from a UI pov, and perhaps could interface to the `Actions` class on #816 when we get to that).

marcharper:
I'm fine with merging this.

PR #960, opened 2017-04-03
A very minor fix to moran docs formatting.

drvinceknight:


drvinceknight:
Closing (taken care of in #962, can always reopen if we don't want that).

PR #966, opened 2017-04-08
#884 lookerup and gambler

eric-s-s:
#884 

I did a rather involved re-write of the re-write to Lookerup.  

0. made keys to LookupTable a namedtuple, ActionKeys (could probably use a renaming)
1. moved a lot of functionality to new object LookupTable
2. made create_lookup_table_keys follow API
4. changed the "paramaters" parameter to require an ActionKeys and not a tuple.
5. refactored __init__ and created helper functions for readablitiy
6. made a new function called lookup_table_display() to easily visualize information while writing tests.

0) ActionKeys.:     
The keys to the lookup table are now the namedtuple, ActionKeys(self_plays, op_plays, op_openings).  This will solve problems when moving to Actions(Enum) and even if the move never happens, removes reliance on knowledge of the value of Actions.C, Actions.D

1) LookupTable class:
All the functionality for creating a lookup table and getting information from it really was a class different from dictionary.  This simplifies a lot of LookerUp __init__, error checking and strategy.  It is also an immutable object (unless you want to muck around in private variables, but that's not my problem).  

2) create_lookup_keys and API:  
Every piece of API, uses (self_plays, op_plays, op_openings), but the original lookup_table keys used (op_openings, self_plays, op_plays). Having functions and inputs that relied on a different ordering for keys made this confusing. So i rewrote ActionKeys, LookupTable and create_lookup_keys to follow the API of everything else, which seemed like a more intuitive way to input information anyways.  The nice advantage of a namedtuple and LookupTable obj is that this will now be easy to change, willynilly in the future.  

3) Old data (in case you were wondering what happened to "3" above):   
Once I changed how lookup_table keys were made, I had to change the order of all the set patterns.  The original patterns are all hard-coded tested for in the tests.  They are currently still in comment lines for ease of undoing the change that I did (and I have the original .csv file saved for easy reversal).  Since create_action_keys has the same change, any newly created patterns from the axelrod dojo should work with the new API just fine.

4) The "parameters" parameter:  
I changed "parameters" to using an ActionKeys instead of a tuple.  This forces the explicit naming of each action.  since ActionKeys is still treated as a tuple, old code using a tuple CAN be passed in without breaking anything (mypy should complain, but the code will still run)

5) refactored __init__:   
the init function has a LOT of different things that needed to happen.  It required a lot of comments, and took a while to read, and get the gist of everything it was doing.  So i took all that functionality and made it pirvate helper functions.  This helped expose some situations where bugs could crop up. Each of these helper functions is explicitly tested for, now. This, hopefully, made the init function easier to read.

6) lookup_table_display():   
Using a named tuple of tuples for the lookup_table makes the repr hopelessly messy to read (but easy to change).  i gathered that the point of changing the order of the inputs (see pt 2) was to make it easy to look up the next action on the lookup table.  so i wrote this to make writing my tests easier.  I then cleaned it up and wrote tests for it.  And now moved it into LookupTable where it belongs. Even if the ActionKeys is renamed or re-ordered, the repr of a LookupTable.dictionary is now hopelessly noisy.  [the horrible dictionary repr](https://github.com/eric-s-s/Axelrod/blob/884-lookerup-gambler/axelrod/tests/strategies/test\_lookerup.py#L33-L42).  [and the display function](https://github.com/eric-s-s/Axelrod/blob/884-lookerup-gambler/axelrod/tests/strategies/test\_lookerup.py#L74-L100)So this was written to address that disadvantage.

NOTE:  lookup_table_display() relies on being able to sort actions. This could be an issue or it might be worthwhile anyways to create Actions.C < Actions.D

I hope these changes improve readability and readiness for change.


eric-s-s:
you all might want to wait a day before looking at this.  it occurs to me that this is slightly messy because LookupTable is most definitely a class by itself.  i will push up some changes withing 24hrs.

drvinceknight:
This has some implications in other places. For example https://github.com/Axelrod-Python/axelrod-dojo has some code that trains strategies so that would need to be adjusted.

Am I correct that the main motivation for this is the potential move to a ("real") class for actions?

>  i will push up some changes withing 24hrs.

Will wait to review properly :) :+1:

eric-s-s:
@drvinceknight thanks.  good to know about the dojo.  at a glance, it seems that this shouldn't affect that too much.  it will create the keys in a different order now, but it will create the correct output (just in the new order).  i was going to put "create_lookup_table_keys" inside the LookupTable class, but now see why i shouldn't.

eric-s-s:
edited the opening description to reflect the changes i just wrote.

@drvinceknight 

> Am I correct that the main motivation for this is the potential move to a ("real") class for actions?

That was a big part of the motivation.  I also wanted to make this more ready for change of any kind.  with the extra classes, different functionality is hidden from different parts and future changes should be easier.  

and then, of course, i enjoy sussing out the hidden structure in something to find what it's trying to say.  and writing the tests to show that. i teach one of my students some programming for fun, and i'm probably one of the few people to ever utter the sentence, "Today, I'm going to show you something really cool, unit tests!"

marcharper:
On a quick look this all looks fine in principle. Since we don't comprehensively test the patterns I'd like to see a full tournament run on this PR for a sanity check comparison on the pattern translation.

The dojo should be fine but the saved patterns from our previous runs won't be valid anymore. (Not a big deal IMO.)

drvinceknight:
> On a quick look this all looks fine in principle. Since we don't comprehensively test the patterns I'd like to see a full tournament run on this PR for a sanity check comparison on the pattern translation.

Would you be able to do this @eric-s-s or would you like one of us to do it?

Here is the code with the tournament: https://github.com/Axelrod-Python/tournament but I'm happy to grab this branch and run a tournament on our University cluster tomorrow.

> 
> The dojo should be fine but the saved patterns from our previous runs won't be valid anymore. (Not a big deal IMO.)

I'm happy with this if you are @marcharper :+1:

eric-s-s:
@drvinceknight thank you for that offer.  I would very much appreciate you running this on your university cluster.  A single repetition of each took about 12 minutes on the family computer.  This means about 20hrs of whirring CPU fan that would drive me up the wall (it's a loud fan.)

drvinceknight:
> I would very much appreciate you running this on your university cluster.

I'll get that running later today :+1: 

In the meantime here are two Ashlock fingerprints for `EvolvedLookerUp_2_2_2`:

download

download 1

marcharper:
Is there a standard difference measure for the fingerprints? maybe mean squared difference on the pixel values?

drvinceknight:
> Is there a standard difference measure for the fingerprints? maybe mean squared difference on the pixel values?

We have the underlying data in the fingerprint class so yeah, we can just compare those. I'll do that and post back here later. :+1:

eric-s-s:
@drvinceknight that fingerprint module is pretty cool.  i noticed some stuff as i was reading through.  any objections to my taking a crack at some refactoring?

drvinceknight:
> i noticed some stuff as i was reading through. any objections to my taking a crack at some refactoring?

No objections, if it can be cleared up/improved that's always good. Feel free to open an issue if you're thinking of big changes in case it's worth discussing before you spend a lot of time on things.

drvinceknight:
Sorry for taking a while to get this done but I've run the following on both `master`  and this branch:

```python
import axelrod as axl
import csv

def write_data_to_file(fp, filename):
    """
    Write the fingerprint data to a file.
    """
    columns = ['x', 'y', 'score']

    with open(filename, 'w') as f:
        w = csv.writer(f)
        w.writerow(columns)
        for key, value in fp.data.items():
            w.writerow([key.x, key.y, value])

lookerups = [axl.EvolvedLookerUp1_1_1, axl.EvolvedLookerUp2_2_2, 
             axl.Winner12, axl.Winner21]

axl.seed(0)
for lookerup in lookerups:
    af = axl.AshlockFingerprint(strategy=lookerup)
    _ = af.fingerprint()
    write_data_to_file(fp=af, filename="{}.csv".format(lookerup))
```
that generated the data available in the following tar file:

[Comparing Lookerups.tar.gz](https://github.com/Axelrod-Python/Axelrod/files/913157/Comparing.Lookerups.tar.gz)

I then ran:

```python
def make_dict(filename):
    """Make a dictionary from a data file"""

    with open(filename, "r") as f:
        reader = csv.reader(f)
        return {(x, y): score for x, y, score in reader}  

for lookerup in lookerups:
    fingerprint_on_master = make_dict("master/{}.csv".format(lookerup))
    fingerprint_on_884 = make_dict("884/{}.csv".format(lookerup))
    print(fingerprint_on_master == fingerprint_on_884)
```

The output is:

```python
True
True
True
True
```

So I think we're ok to merge this in.

drvinceknight:
> So I think we're ok to merge this in.

If perhaps someone else could run the above to confirm (because of the branch change there's a manual step that I could have messed up). It doesn't take long.

eric-s-s:
@drvinceknight i'll do those tests today.

eric-s-s:
i've confirmed your fingerprint results, and i also included all the PSOGambler classes.

drvinceknight:
> On a quick look this all looks fine in principle. Since we don't comprehensively test the patterns I'd like to see a full tournament run on this PR for a sanity check comparison on the pattern translation.

Great. I think this is good to go @marcharper?

marcharper:
How are the dictionaries equal? The plots are slightly different.

drvinceknight:
The plots weren't done with the same seed. I seeded the dictionaries though.

axelrod/strategies/lookerup.py
```diff
@@ -232,10 +233,12 @@ class LookerUp(Player):
 
     The dictionary must contain all possible permutations of C's and D's.
 
-    LookerUp can also be instantiated with pattern=str/tuple of actions, and
-    parameters=Plays(self_plays=player_depth: int,
-                     op_plays=op_depth: int,
-                     op_openings=op_openings_depth: int).
+    | LookerUp can also be instantiated with pattern=str/tuple of actions, and
```
drvinceknight: What's the pipe for here?

drvinceknight:
Here are the plots with the same seed:

This branch:



Current master:

master

axelrod/strategies/lookerup.py
eric-s-s: what is a pipe? you mean like in bash?

the docstrings use rst notation, correct?  I thought it would be useful to preserve the line-breaks there.  so i used that.  without the "hey rst! preserve linebreaks"-notation, it threw an error with one of the auto checks for having naughty indentation.  should i change to "::" notation?  

while i think of it,  can i go ahead and delete all the commented-out original data?

drvinceknight: I hadn't seen the `|` in rst before. I'm not sure it looks like it should though. Here's what the docs build looks like on this branch:

screenshot from 2017-04-13 09-46-20

Is that what you were going for?

This tweak:

```
  235                                                                                                                                                 
~ 236     LookerUp can also be instantiated with :code:`pattern=str/tuple` of actions                                                                 
~ 237     , and::                                                                                                                                     
~ 238                                                                                                                                                 
~ 239         parameters=Plays(                                                                                                                       
~ 240             self_plays=player_depth: int,                                                                                                       
+ 241             op_plays=op_depth: int,                                                                                                             
+ 242             op_openings=op_openings_depth: int).                                                                                                
  243                                                                                                                                                 
  244     It will create keys of len=2 ** (sum(parameters)) and map the pattern to                                                                    
  245     the keys.   
```
looks like:

screenshot from 2017-04-13 09-50-35

> while i think of it, can i go ahead and delete all the commented-out original data?

Sounds good to me. :+1:

meatballs:
> If perhaps someone else could run the above to confirm (because of the branch change there's a manual step that I could have messed up). It doesn't take long.

I've repeated this and can confirm the result

eric-s-s:
@drvinceknight 
> Thanks @eric-s-s!
:)

PR #967, opened 2017-04-11
Add an extra test for the FSM players

drvinceknight:
I found a bug working with v2.8.0, but #956 seems to have cleared it up.

I'm just adding a test to confirm it.

I've also cleared up some minor PEP8 stuff.

### DIFF
--- a/axelrod/tests/strategies/test_finite_state_machines.py
+++ b/axelrod/tests/strategies/test_finite_state_machines.py
@@ -11,9 +11,11 @@ C, D = axelrod.Actions.C, axelrod.Actions.D
 
 class TestSimpleFSM(unittest.TestCase):
     def setUp(self):
-        self.two_state_transition = ((1, C, 0, C), (1, D, 0, D), (0, C, 1, D), (0, D, 1, C))
+        self.two_state_transition = ((1, C, 0, C), (1, D, 0, D),
+                                     (0, C, 1, D), (0, D, 1, C))
 
-        self.two_state = SimpleFSM(transitions=self.two_state_transition, initial_state=1)
+        self.two_state = SimpleFSM(transitions=self.two_state_transition,
+                                   initial_state=1)
 
     def test__eq__true(self):
         new_two_state = SimpleFSM(transitions=self.two_state_transition,
@@ -29,8 +31,10 @@ class TestSimpleFSM(unittest.TestCase):
         self.assertFalse(new_two_state.__eq__(self.two_state))
 
     def test__eq__false_by_transition(self):
-        different_transitions = ((1, C, 0, D), (1, D, 0, D), (0, C, 1, D), (0, D, 1, C))
-        new_two_state = SimpleFSM(transitions=different_transitions, initial_state=1)
+        different_transitions = ((1, C, 0, D), (1, D, 0, D),
+                                 (0, C, 1, D), (0, D, 1, C))
+        new_two_state = SimpleFSM(transitions=different_transitions,
+                                  initial_state=1)
 
         self.assertFalse(new_two_state.__eq__(self.two_state))
 
@@ -56,11 +60,13 @@ class TestSimpleFSM(unittest.TestCase):
         self.assertEqual(self.two_state.state, 1)
 
     def test_bad_transitions_raise_error(self):
-        bad_transitions = [(1, C, 0, D), (1, D, 0, D), (0, C, 1, D)]
-        self.assertRaises(ValueError, SimpleFSM, transitions=bad_transitions, initial_state=1)
+        bad_transitions = ((1, C, 0, D), (1, D, 0, D), (0, C, 1, D))
+        self.assertRaises(ValueError, SimpleFSM, transitions=bad_transitions,
+                          initial_state=1)
 
     def test_bad_initial_state_raises_error(self):
-        self.assertRaises(ValueError, SimpleFSM, transitions=self.two_state_transition, initial_state=5)
+        self.assertRaises(ValueError, SimpleFSM,
+                          transitions=self.two_state_transition, initial_state=5)
 
     def test_state_setter_raises_error_for_bad_input(self):
         with self.assertRaises(ValueError) as cm:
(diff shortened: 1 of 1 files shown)


axelrod/tests/strategies/test_finite_state_machines.py
```diff
@@ -622,17 +643,22 @@ def test_strategy(self):
         self.transitions_test(state_and_actions)
 
         # finished: 0, 5, 10
-        state_and_actions = [(0, D), (12, D), (11, D)] + [(5, D), (10, C), (11, D), (5, D), (10, D), (8, C)] * 3
+        state_and_actions = [(0, D), (12, D), (11, D)] + [(5, D), (10, C),
+                                                          (11, D), (5, D),
```
marcharper: interesting spacing...

axelrod/tests/strategies/test_finite_state_machines.py
```diff
@@ -622,17 +643,22 @@ def test_strategy(self):
...
+                              (10, C), (11, C), (15, C), (15, C), (15, D)] +
                              [(2, C)] * 3 + [(2, D), (14, C), (13, C)])
         self.transitions_test(state_and_actions)
 
         # finished: 0, 2, 3, 5, 10, 11, 12, 13, 14, 15
         to_state_fourteen = [(0, D), (12, D), (11, C), (15, D), (2, D)]
-        state_and_actions = to_state_fourteen + [(14, D), (13, C), (13, C), (13, D), (7, C)] + [(3, D), (3, C)] * 3
+        state_and_actions = to_state_fourteen + [(14, D), (13, C), (13, C),
+                                                 (13, D), (7, C)] + [(3, D),
+                                                                     (3, C)] * 3
```
marcharper: and here

drvinceknight: Yeah, there were loads of 80 character PEP8 rules so I let my auto linter fix it. I can clear it up if you'd prefer?

eric-s-s: ha!  sorry.  (-_-*;)  i've been using what was set by pycharm, which is 120.  (apparently, there is some debate.  my brother was near frothing at the mouth giving me shit for sticking to 80 chars about a year back.)  after using 120 for a while, i agree with him, but i can absolutely make the effort to keep it at 80 here.

drvinceknight: I'm generally easy but do feel that calls for longer than 80 are not usually (I obviously don't know your brother's reasons) for great reasons: not everyone **does** have a wide monitor and actually think the 80 character limit is more inclusive. Some people work on small laptops and 80 still lets them have two windows up without wrapping :)

Either way: we've generally stuck to 80 in the library so would be great to stick to it :+1:

axelrod/tests/strategies/test_finite_state_machines.py
```diff
@@ -187,6 +193,19 @@ def test_all_states_reachable(self):
         extra_info = 'The following states are un-reachable: {}'.format(un_callable_states)
         self.assertEqual(un_callable_states, set(), msg=extra_info)
 
+    def test_strategy(self):
+        """A particular FSM test"""
+        transitions = [(0, C, 0, C), (0, D, 3, C), (1, C, 5, D),
```
eric-s-s: very very minor.  all the transitions are tuple[tuple].  mypy gave me no end of trouble when they were either list or tuple.

drvinceknight: Good catch! This is more of my lazy bad habits :)

axelrod/tests/strategies/test_finite_state_machines.py
drvinceknight: @marcharper I'm fixing the spacing now :+1:

axelrod/tests/strategies/test_finite_state_machines.py
```diff
@@ -187,6 +203,19 @@ def test_all_states_reachable(self):
         extra_info = 'The following states are un-reachable: {}'.format(un_callable_states)
         self.assertEqual(un_callable_states, set(), msg=extra_info)
 
+    def test_strategy(self):
```
eric-s-s: should you mark this somehow as a regression test so that it doesn't get deleted in some later iteration?  or is there no need?

eric-s-s:
@drvinceknight thanks for catching all those errors on the test transitions (and cleaning up 120 chars)! 👍

axelrod/tests/strategies/test_finite_state_machines.py
drvinceknight: Good call, I'll add that now!

drvinceknight:
> @drvinceknight thanks for catching all those errors on the test transitions (and cleaning up 120 chars)!

No no: thank you for the great code contribution, that's the hard part, style refactoring is simple enough :)

eric-s-s:
i see where the old bug was.  here's the new code

```
    def __init__(self, transitions: tuple = ((1, C, 1, C), (1, D, 1, D)),
                 initial_state: int = 1,
                 initial_action: Action = C) -> None:

        super().__init__()
        self.initial_state = initial_state
        self.initial_action = initial_action
        self.fsm = SimpleFSM(transitions, initial_state)
```

and here's the old code.

```
    def __init__(self, transitions=None, initial_state=None,
                 initial_action=None) -> None:
        if not transitions:
            # Tit For Tat
            transitions = [(1, C, 1, C), (1, D, 1, D)]
            initial_state = 1
            initial_action = C
        super().__init__()
        self.initial_state = initial_state
        self.state = initial_state
        self.initial_action = initial_action
        self.fsm = SimpleFSM(transitions, initial_state)
```

so it used to be that if there was a transition provided, but no initial_state or initial_action, they'd be assigned to `None` . The tests i wrote never tested that the __init__ defaults worked independently of each other.

so that's what your test is catching.

eric-s-s:
it looks good to me.  👍

Issue #969, opened 2017-04-11
some ideas on refactoring fingerprint


eric-s-s:
@drvinceknight 

here are some thoughts on refactoring fingerprint.py

- move most of the staticmethods to private functions.  it seems that their main point is to aid in the fingerprint or plot methods, and not be of use by themselves.
- use those private methods to make a set of instance variables.  thus, at instantiation, call methods to create self._probe_list, self._points, self._cheap_to_generate_variables.
- use property decorator and setter decorator so that when a new step or probe is created, can reset all the instance variable that rely on them (nothing that requires playing a tournament, just the edges/points/probes)
- move the data reshaping and plotting to a separate class that can be easily re-used if/when other fingerprinting methods come along.  also, will aid in adding functionality of comparing two data sets to see if fingerprints are a match.  (no idea HOW to do that, but with a separate DataManipulation class it would be easier to write/change).  The fingerprint class can easily instantiate and call the plot method.  (also, more freedom to create different plot methods)

hope these are useful ideas.

bonus: type hinting

drvinceknight:
> move most of the staticmethods to private functions. it seems that their main point is to aid in the fingerprint or plot methods, and not be of use by themselves.
> use those private methods to make a set of instance variables. thus, at instantiation, call methods to create self._probe_list, self._points, self._cheap_to_generate_variables.

This was something we discussed quite a lot and that @meatballs is in favour of. So :+1: from me.

> use property decorator and setter decorator so that when a new step or probe is created, can reset all the instance variable that rely on them (nothing that requires playing a tournament, just the edges/points/probes)

Sounds good :+1: 

> move the data reshaping and plotting to a separate class that can be easily re-used if/when other fingerprinting methods come along. also, will aid in adding functionality of comparing two data sets to see if fingerprints are a match. (no idea HOW to do that, but with a separate DataManipulation class it would be easier to write/change). The fingerprint class can easily instantiate and call the plot method. (also, more freedom to create different plot methods)

I'm not sure about this one. Comparing fingerprint data is actually straightforward the fp class has a dictionary attribute so you can just compare those directly (the main reason it was awkward on #966 is because I had to checkout different branches so needed to write the data to file). 

I suggest leaving this one on hold until /if there is a need for it?

eric-s-s:
thank you.  will get to work on it.

drvinceknight:
> will get to work on it.

[...]

PR #970, opened 2017-04-13
refactor for fingerprint

eric-s-s:
#969

### DIFF
--- a/axelrod/fingerprint.py
+++ b/axelrod/fingerprint.py
@@ -229,14 +229,6 @@ class JossAnnProbeCreator(object):
             point_list = tqdm.tqdm(point_list, desc="Generating probes")
         return {point: self.get_probe(point) for point in point_list}
 
-    def get_probe_dict_from_step(self, step: float) -> PointsPlayers:
-        """
-
-        :param step: 0.0 < step <= 1.0
-        """
-        point_list = create_points(step)
-        return self.get_probe_dict(point_list)
-
 
 def get_class_and_kwargs(strategy_or_player: Any) -> tuple:
     """If strategy, returns a class and kwargs to create default player.
--- a/axelrod/fingerprint.py
+++ b/axelrod/fingerprint.py
@@ -156,10 +156,15 @@ class SpatialTournamentCreator(object):
                 for index, point in enumerate(self._points)}
 
     def get_tournament(self, **tournament_kwargs) -> axl.SpatialTournament:
+
         players = [self._player]
         edges = []
         edge_map = self.get_points_to_edges()
-        probe_map = self._probe_gen.get_probe_dict(self._points)
+
+        progress_bar = False
+        if 'progress_bar' in tournament_kwargs:
+            progress_bar = tournament_kwargs['progress_bar']
+        probe_map = self._probe_gen.get_probe_dict(self._points, progress_bar)
 
         for point in self._points:
             players.append(probe_map[point])
@@ -187,12 +192,31 @@ class JossAnnProbeCreator(object):
             )(**self._probe_kwargs)
         return joss_ann
 
-    def get_probe_dict(self, point_list: list) -> PointsPlayers:
+    def get_probe_dict(self, point_list: list,
+                       progress_bar: bool) -> PointsPlayers:
+        if progress_bar:
+            point_list = tqdm.tqdm(point_list, desc="Generating probes")
         return {point: self.get_probe(point) for point in point_list}
 
-    def get_probe_dict_from_step(self, step: float) -> PointsPlayers:
+    def get_probe_dict_from_step(self, step: float,
+                                 progress_bar: bool) -> PointsPlayers:
         point_list = create_points(step)
-        return self.get_probe_dict(point_list)
+        return self.get_probe_dict(point_list, progress_bar)
+
+
+def get_class_and_kwargs(strategy_or_player: Any) -> tuple:
+    """If strategy, returns a class and kwargs to create default player.
+    If Player, returns a class and kwargs to re-create that instance."""
+    probe_player = create_player(strategy_or_player)
+    return probe_player.__class__, probe_player.init_kwargs
+
+
+def create_player(strategy_or_player: Any) -> Player:
+    """If strategy, returns the default Player of that strategy. Else
+    returns the Player."""
+    if isinstance(strategy_or_player, axl.Player):
+        return strategy_or_player.clone()
+    return strategy_or_player()
 
 
 def create_points(step: float) -> List[Point]:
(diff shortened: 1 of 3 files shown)


axelrod/fingerprint.py
```diff
@@ -65,168 +28,91 @@ def __init__(self, strategy, probe=axl.TitForTat):
...
+            tournament.
         """
-        self.strategy = strategy
-        self.probe = probe
+        self.player = create_player(strategy)
+        self._probe_class, self._probe_kwargs = get_class_and_kwargs(probe)
+        self._step = step
+        self.progress_bar = progress_bar
+        self.points = self._create_points()
+        self._probe_list = self._create_probes()
```
drvinceknight: So this is a bit of a change of how things are currently done. 

Having the probes created only when we choose to fingerprint was a design choice at the time. Is there a reason why you're moving it here?

From an API point of view it's different to the rest of the library where `.play()` does all the work. That is what is currently happening on master (which is why the `progress_bar` and `steps` argument is in the `fingerprint` method).

Unless there's a reason I'm missing I would prefer not to change the API too much and "just" move the static methods to private functions.

axelrod/fingerprint.py
```diff
@@ -65,168 +28,91 @@ def __init__(self, strategy, probe=axl.TitForTat):
...
-            for scores in interactions[edge]])
-            for edge in edges
-        ]
-        point_scores = dict(zip(points, edge_scores))
-        return point_scores
-
-    def fingerprint(self, turns=50, repetitions=10, step=0.01, processes=None,
-                    filename=None, in_memory=False, progress_bar=True):
+    def fingerprint(self, turns: int = 50, repetitions: int = 10,
+                    new_step: float = None, processes: int = None,
```
drvinceknight: we could keep this variable as `step` so that it doesn't introduce any backwards compatibility.

axelrod/fingerprint.py
eric-s-s: i'll have to think on it some, but here's roughly my thought process.

i thought of AshlockFingerprint as a kind of auto-generated SpacialTournament .  It has the same kinds of characteristics as a Tournament where it owns some basic data about who will play who and how and then can generate data depending on how much you want those players to play.

It owns player and probes and points (instead of edges).  it has interactions (and data).  (now that i think about it, turns, and repetitions should be in __init__ then and new_step should not be in fingerprint(except for backwards compatibility. saw that while typing))

fingerprint() method then plays the "tournament"  like tournament.play()

I found it an easier way to think about AshlockFingerprint and an easier way thoroughly test.  It split up the responsibility for creating everything needed to play and then playing/processing data.

drvinceknight: Yup, there's nothing wrong with that thought process. Another way of thinking about it though is that a fingerprint is an abstract thing (technically it's an analytical function) and to `fingerprint` is to create a tournament (ie probes which is also why the `turns, steps` etc are in there) and play the tournament.

Given that the status quo is already implemented I suggest it stays put (otherwise someone else might come along at another point with a different way of thinking about it and change it to that etc...). If there's a reason other than your interpretation that I'm missing though that's worth discussing.

The other argument in favour of the status quo is that `.fingerprint()` is where all the work happens (for a small step size, creating probes can actually take a long time). This is similar to the rest of the library where `Match.play()`, `Tournament.play()` and `Moran.play()` does all the work.

drvinceknight:
> It split up the responsibility for creating everything needed to play and then playing/processing data.

This split can still happen with private functions and testing. It's just that the private functions are used in the `fingerprint` method as opposed to `init`.

drvinceknight:
(FYI, here's the original PR that implemented the fp: https://github.com/Axelrod-Python/Axelrod/pull/759

It'll take a while to load as it's the PR with the most comments to date: we talked a lot about how to implement this. The one discussion point that didn't entirely end up satisfying everone is the private functions vs methods so I suggest going for that :))

axelrod/fingerprint.py
eric-s-s: on more thought, either way of thinking about it, i absolutely agree that probes creation belongs in fingerprint and, frankly, will be easy enough to move.  i have 1/2 baked thoughts on the matter, but first, some light reading :), #759

drvinceknight: Haha enjoy! :D :+1:

eric-s-s:
@drvinceknight there's something tangential that is driving me up the wall.  maybe you can clear it up.

```
is_rand = axl.strategy_transformers.JossAnnTransformer((0.5, 0.5))(axl.Cooperator)()

y = axl.Tournament([is_rand, axl.Cooperator()], turns=5, repetitions=5)
y.play()

z =  axl.Tournament([axl.Random(), axl.Cooperator()], turns=5, repetitions=5)
z.play()
```
i run this and here's the results
```
>>> y.interactions_dict[(0, 1)]
[[('D', 'C'), ('C', 'C'), ('D', 'C'), ('C', 'C'), ('D', 'C')],
 [('D', 'C'), ('C', 'C'), ('D', 'C'), ('C', 'C'), ('D', 'C')],
 [('D', 'C'), ('C', 'C'), ('D', 'C'), ('C', 'C'), ('D', 'C')],
 [('D', 'C'), ('C', 'C'), ('D', 'C'), ('C', 'C'), ('D', 'C')],
 [('D', 'C'), ('C', 'C'), ('D', 'C'), ('C', 'C'), ('D', 'C')]]

>>>z.interactions_dict[(0, 1)]
[[('D', 'C'), ('C', 'C'), ('C', 'C'), ('C', 'C'), ('D', 'C')],
 [('D', 'C'), ('C', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C')],
 [('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C')],
 [('C', 'C'), ('D', 'C'), ('C', 'C'), ('D', 'C'), ('C', 'C')],
 [('C', 'C'), ('C', 'C'), ('D', 'C'), ('D', 'C'), ('C', 'C')]]
```

JossAnn(0.5, 0.5) consistently gives a random result, but every repetition is random in exactly the same way.  any idea why?  it's driving me batty?

drvinceknight:
> JossAnn(0.5, 0.5) consistently gives a random result, but every repetition is random in exactly the same way. any idea why? it's driving me batty?

I'm not able to reproduce I'm afraid, here's my code:

```python
>>> axl.seed(0)
>>> is_rand = axl.strategy_transformers.JossAnnTransformer((0.5, 0.5))(axl.Cooperator)()
>>> y = axl.Tournament([is_rand, axl.Cooperator()], turns=5, repetitions=5)
>>> y.play(in_memory=True)   # On linux so have to specifically ask for the interactions to be kept in mem
>>> z =  axl.Tournament([axl.Random(), axl.Cooperator()], turns=5, repetitions=5)
>>> z.play(in_memory=True)
>>> y.interactions_dict[(0, 1)] == z.interactions_dict[(0, 1)]
False
>>> y.interactions_dict[(0, 1)]
[[('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C'), ('C', 'C')],
 [('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C'), ('C', 'C')],
 [('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C'), ('C', 'C')],
 [('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C'), ('C', 'C')],
 [('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C'), ('C', 'C')]]
>>> z.interactions_dict[(0, 1)]
[[('D', 'C'), ('C', 'C'), ('C', 'C'), ('C', 'C'), ('D', 'C')],
 [('D', 'C'), ('C', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C')],
 [('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C'), ('D', 'C')],
 [('C', 'C'), ('D', 'C'), ('C', 'C'), ('D', 'C'), ('C', 'C')],
 [('C', 'C'), ('C', 'C'), ('D', 'C'), ('D', 'C'), ('C', 'C')]]
```

drvinceknight:
Ignore my comment @eric-s-s. That was my brain half asleep. I see what you're pointing out. Looking at it now.

drvinceknight:
> JossAnn(0.5, 0.5) consistently gives a random result, but every repetition is random in exactly the same way. any idea why? it's driving me batty?

Yuck. As far as I can tell it's a bug with the transformers. I'm opening another issue now.

drvinceknight:
#971, it's a bug, have figured out what's wrong and working on a fix.

eric-s-s:
ok.  i'm about to push up a change of a half-baked idea, because i'd like opinions before i move it farther.  I tried creating a series of smaller objects that are really responsible for one thing.  (not sure how well i did, but well enough for you to get the gist) there is no documentation or tests for them. 

I think this may help with separation and testing.  and hopefully create some clearer boundaries.  

i'm sure it will fail the coverage test.

drvinceknight:
> ok. i'm about to push up a change of a half-baked idea, because i'd like opinions before i move it farther. I tried creating a series of smaller objects that are really responsible for one thing. (not sure how well i did, but well enough for you to get the gist) there is no documentation or tests for them.

A quick flick through: it looks fine to me.

eric-s-s:
okay.  this rewrite is much closer to the original.  because of changes coming up in #972 this will need to wait for that to be merged into master.  I'll then have to resolve some merge issues before this is ready.  there are some issues that i'll mark.

axelrod/fingerprint.py
```diff
@@ -14,219 +16,355 @@
...
+
+
+class AshlockFingerprint(object):
+    def __init__(self, strategy: PlayerOrStrategy,
+                 probe: PlayerOrStrategy) -> None:
+        self._strategy = strategy
+        self._probe = probe
+        self._data = None  # type: DataOrganizer
+
+    @property
```
eric-s-s: I changed the af.interactions here to return {Point: matches} instead of {Edge: matches}  because the edges have no meaning beyond the tournament which, by this point has already been discarded.  it seemed more useful to have the tournament results map to the appropriate Points.

axelrod/fingerprint.py
```diff
@@ -14,219 +16,355 @@
...
+                    step: float = 0.01, processes: int = None,
+                    filename: str = None, in_memory: bool = False,
+                    progress_bar: bool = True) -> PointsFloats:
+
+        tournament_creator = SpatialTournamentCreator(self._strategy,
+                                                      self._probe,
+                                                      step)
+        tournament_kwargs = {'turns': turns, 'repetitions': repetitions}
+
+        play_kwargs = _get_play_kwargs(filename=filename, in_memory=in_memory,
```
eric-s-s: i don't like this, but i don't know why.  it feels clunky.  i'm not sure if anything is actually bad here, but would appreciate some input.

axelrod/fingerprint.py
```diff
@@ -14,219 +16,355 @@
...
+
+def _get_play_kwargs(filename: Any, in_memory: bool,
+                     processes: Union[int, None], progress_bar: bool
+                     ) -> Dict[str, Any]:
+    """Adjust and build the kwargs for tournament.play() based on issues
+    with Windows OS."""
+    play_kwargs = {'build_results': False,
+                   'processes': processes,
+                   'progress_bar': progress_bar}
+
```
eric-s-s: how is this different from what Tournament already does?  How does Tournament usually handle `in_memory=None, file_name=None` ?

drvinceknight: The tournament does things in the same way but we need to specifically pass these things to the tournament (for example if we don't pass a `filename` to the fingerprint we still need to know what the temp filename would be called to be able to analyse it with the `interaction_utils`). I might be misunderstanding your question.

I'm about to log off for the night: will go through this properly tomorrow :)

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -1,6 +1,11 @@
 import unittest
```
drvinceknight: Reviewing this diff is going to take a bit of work because it looks like a big rewrite (chunks of addition/deletion). If you could write a comment with some pointers that would be helpful.

drvinceknight:
Tests are failing.

axelrod/fingerprint.py
drvinceknight: > in_memory=None, file_name=None ?

To answer that specific question: this is the default (and preferred way), a temporary file is created that interactions are written to and nothing is kept in memory.

For large tournaments the memory consumption can be huge (>16GB). Writing to files has hardly any memory footprint. We have the `in_memory` option only to deal with Windows really (although anyone can do it) as Python's tempfile module doesn't really work on Windows (this is a known upstream thing).

drvinceknight:
> Tests are failing.

Last comment before I head off: fixing this might well be trivial but if not my main advice on this PR is to not mess with the internal functioning much (I haven't read through this yet). By all means pull out methods into private functions but that shouldn't require much changing elsewhere.

eric-s-s:
fixing the test fail.  that was just a bit of stupid when i pulled out a method.

axelrod/tests/unit/test_fingerprint.py
eric-s-s: will do

eric-s-s: i'll have to get to that tomorrow.  i'm chock full of stupid right now.

eric-s-s:
- AshlockFingerprint
    - fingerprint
        - calls TournamentCreator
            - calls JossAnnProbeCreator
        - calls update_according_to_os
        - calls DataOrganizer and saves as state       
    - plot
        - calls DataOrganizer.get_plotting_data
- update_according_to_os
- SpatialTournamentCreator
    - init
        - calls JossAnnProbeCreator
        - calls create_points
        - calls create_player
    - get_tournament
        - calls JossAnnProbeCreator.get_prob_dict
- JossAnnProbeCreator
    - calls get_class_and_kwargs
- get_class_and_kwargs
    - calls create_player
- create_player
- create_points
-DataOrganizer
    - get_plotting_data
        -call get_mean_score
-get_mean_score

eric-s-s:
i checked the coverage report.  it is impossible for the test to be covered.  There is a piece of code that should be tested but that relies on which OS is being used.

axelrod/fingerprint.py
```diff
@@ -361,3 +149,152 @@ def plot(self, col_map='seismic', interpolation='none', title=None,
         if title is not None:
             plt.title(title)
         return fig
+
+
+def update_according_to_os(filename: Union[str, None],
```
eric-s-s: what does this do that Tournament doesn't already do?

axelrod/fingerprint.py
```diff
@@ -246,73 +68,39 @@ def fingerprint(self, turns=50, repetitions=10, step=0.01, processes=None,
...
+        filename, in_memory = update_according_to_os(filename, in_memory)
+        tournament.play(filename=filename, in_memory=in_memory,
+                        processes=processes, progress_bar=progress_bar,
+                        build_results=False)
+
+        point_edge_map = tournament_creator.get_points_to_edges()
         if in_memory:
-            self.interactions = self.spatial_tournament.interactions_dict
+            interactions = tournament.interactions_dict
         else:
```
eric-s-s: when i run a SpatialTournament (on windows) with the non-windows parameters it writes to automatic file just fine.  it also has the interactions_dict in memory.  is this not the case on linux/mac?

```
>>> x = axl.SpatialTournament(players=[axl.Cooperator(), axl.Cooperator()], edges=[(0, 1)],turns=1, repetitions=1)
>>> x.play(build_results=False,filename=None, in_memory=False)
Playing matches:   0%|          | 0/1 [00:00<?, ?it/s]E:\work\environments\tester\lib\site-packages\axelrod\tournament.py:113: UserWarning: Tournament results will not be accessible since build_results=False and no filename was supplied.
  "Tournament results will not be accessible since "
                                                      
>>> x.interactions_dict 
{(0, 1): [[('C', 'C')]]}
```

axelrod/fingerprint.py
drvinceknight: I don't understand your question.

axelrod/fingerprint.py
drvinceknight: No, by default everything is written to file and no memory is used. The memory cost for large tournaments can be huge so it is best to be avoided.

drvinceknight:
@eric-s-s this really feels like a complete re write of a module. If we're going to include PRs that rewrite a module from scratch it must be for a good reason (and not just personal preference). I'm hesitant to invest a lot of time understanding your re write of this module given that it was essentially working fine and there's no big improvement (that I can see) as well as a bunch of little testing errors that you've had to pick up (so I'd need to be super careful to make sure nothing is missed). I'm also hesitant to do so given the amount of time that was spent reviewing the original PR.

Personally I feel that this rewrite actually makes things look a bit more complicated than they were (but this is where personal preference comes in). It introduces multiple abstractions which unless they're warranted I personally feel make the code base more complex. I'm not a fan of over engineering things (for example: the data manipulation class: we can always write that if/when we need it, at present we don't).

The premise for this refactor was to move some static methods to private functions. That in itself should not be a complete re write: picking out the static methods, write functions and then adjust the tests. 

If I'm missing something please let me know but if you think this complete re write is really warranted note that it's  going to take me a while to come to review it. I'm around most of today so feel free to ping me in the gitter channel to talk about it https://gitter.im/Axelrod-Python/Research.

@marcharper and @meatballs might have a differing opinion to mine of course but as a precedent I'm cautious about PRs for which the reviewer time to impact ratio is large.

axelrod/fingerprint.py
drvinceknight: > non-windows parameters it writes to automatic file just fine

The tournament will overwrite that on windows. Here's the relevant line in `tournament.py`:

```python
106         if on_windows and (filename is None):  # pragma: no cover                                                                                 
107             in_memory = True  
```

On windows named temp files disappear as soon as they're closed so can't be used.

drvinceknight: Here's the PR that fixed this bug: https://github.com/Axelrod-Python/Axelrod/pull/721

meatballs:
I agree with @drvinceknight here. I'm failing to see the benefit of many of these changes.

It might be better to create several, much smaller PRs which make one improvement each - that way, the intention is clear and the review process is much simpler.

drvinceknight:
> It might be better to create several, much smaller PRs which make one improvement each - that way, the intention is clear and the review process is much simpler.

Perhaps as a starting point we can just pull out the following static methods in to their own functions:

- `create_jossann`
- `create_edges`
- `generate_data`
- `reshape_data`

This has the clear benefit of being able to properly isolate the unit tests.

eric-s-s:
@drvinceknight 

> but as a precedent I'm cautious about PRs for which the reviewer time to impact ratio is large.

yes.  absolutely!  i am sorry about that.   
 as i was writing and pulling things out and trying to understand, things got away from me and i couldn't leave well enough alone (which has totally never ever happened before).  the final product was the  attempt to pull back into something reasonable.  I'm absolutely fine with this getting closed and scrapped. 

the re-write did lay some things bare that might need to be addressed later.  (and that is why i don't at all mind scrapping this particular PR.  the writing and organizing itself are what showed where some potential cracks are.  i do not know how/when/if these should be addressed or if they even need addressing. but they certainly were NOT part of the original issue I raised.)

- there is an implicit reliance on ordered edges being mapped to ordered Points.  This is never spelled out and might wreak havoc at some later point.  
- there is a somewhat stated but implicit reliance on a perfect unit square of points.  my best address of this was in the contract of DataOrganizer.  (i had written and scrapped some `raise_error_for_...` as i figured that, in a support obj, the explicit docstring statement was enough). This reliance leads to a possibly surprising side-effect that step input is not necessarily the actual step. 
- self.step and self.spatial_tournament should either be part of the __init__ or not part of the class. (those two were yellow-carded by my linter.  that and the staticmethods jumped out at first viewing). 
 - if both the data and the interactions_dict are held in memory, is there any reason to allow writing to file or assigning `in_memory=False`?

<sigh>.  several different things.  all or some of which might be nothing at all.

drvinceknight:
Thanks @eric-s-s :) :+1: 

> so several different things. all or some of which might be nothing at all.

They all looks like great things to discuss and indeed awesome to have had you look through this so carefully to identify them. Perhaps they could each be opened as separate issues? If you're ok to open the issues to do this that'd be great but let me know and I'd be happy to open an issue to get the discussion going (appreciate you have put a boat load of work in to this already).

I'm going to go ahead and close this PR. I'm going to open an issue for the static methods -> functions.

eric-s-s:
@drvinceknight I'll be happy to take the lead on those issues. since they're either unrelated to each other or sequential, i'll wait for the first issue to close before opening the next.

Since you've opened the new issue, then perhaps it's a good idea to close #969 ?

PR #972, opened 2017-04-14
Fix bug in classification of transformers

### LINKED ISSUE #971: Bug with strategy transformers
On #970 @eric-s-s has pointed out what I believe is a bug with how the strategy transformers are cloned and random seeding (somehow?).

Here's 5 matches between a noisy transform of `Cooperator` and `Cooperator` itself:

```python
>>> import axelrod as axl
>>> for seed in range(5):
...    axl.seed(seed)                                                                                                                                                      
...    p1 = axl.strategy_transformers.NoisyTransformer(.7)(axl.Cooperator)()                                                                                                                                              
...    p2 = axl.Cooperator()                                                                                                                                                                               
...    match = axl.Match([p1, p2], turns=3)
...    print(match.play())
[('C', 'C'), ('C', 'C'), ('D', 'C')]
[('D', 'C'), ('C', 'C'), ('C', 'C')]
[('C', 'C'), ('C', 'C'), ('D', 'C')]
[('D', 'C'), ('D', 'C'), ('D', 'C')]
[('D', 'C'), ('D', 'C'), ('D', 'C')]
```

Each row corresponds to a match and the first action corresponds to the Cooperator. We see there that the transformer is acting randomly in each match.

Similarly if I clone the strategy:

```python
>>> transformer = axl.strategy_transformers.NoisyTransformer(.7)(axl.Cooperator)()                                                                                                                                              

>>> for seed in range(5):
...    axl.seed(seed)
...    p1 = transformer.clone()
...    p2 = axl.Cooperator()                                                                                                                                                                               
...    match = axl.Match([p1, p2], turns=3)
...    print(match.play())
[('C', 'C'), ('C', 'C'), ('D', 'C')]
[('D', 'C'), ('C', 'C'), ('C', 'C')]
[('C', 'C'), ('C', 'C'), ('D', 'C')]
[('D', 'C'), ('D', 'C'), ('D', 'C')]
[('D', 'C'), ('D', 'C'), ('D', 'C')]
```

**However** if we create a tournament with repetitions (doing this with a `Random` player for comparison):

```python
>>> axl.seed(0)
>>> transformer = axl.strategy_transformers.NoisyTransformer(noise=.7)(axl.Cooperator)()
>>> t_with_transformer = axl.Tournament([is_rand, axl.Cooperator()], turns=3, repetitions=5)
>>> transformer_results = t_with_transformer.play(keep_interactions=True)
>>> t_with_random =  axl.Tournament([axl.Random(), axl.Cooperator()], turns=3, repetitions=5)
>>> random_results = t_with_random.play(keep_interactions=True)
```

all the transformers plays are the same across the repetitions:

```python
>>> transformer_results.interactions[0, 1]
[[('C', 'C'), ('D', 'C'), ('D', 'C')],
 [('C', 'C'), ('D', 'C'), ('D', 'C')],
 [('C', 'C'), ('D', 'C'), ('D', 'C')],
 [('C', 'C'), ('D', 'C'), ('D', 'C')],
 [('C', 'C'), ('D', 'C'), ('D', 'C')]]
```

whereas we have the expected behaviour for our `Random` player:

```python
>>> random_results.interactions[0, 1]
[[('D', 'C'), ('C', 'C'), ('C', 'C')],
 [('D', 'C'), ('C', 'C'), ('C', 'C')],
 [('D', 'C'), ('C', 'C'), ('D', 'C')],
 [('C', 'C'), ('D', 'C'), ('D', 'C')],
 [('C', 'C'), ('C', 'C'), ('C', 'C')]]
```

drvinceknight:
Our transformers were not reclassifying the strategies (mostly we weren't trying but in one case we were but it was not tested and not working). See #971 

This implements the ability to pass a `reclassifier` function to the strategy transformer factory so that we can reclassify based on input parameters (for example if using the Noisy transformer but passing noise of `0` or `1` then the classifier does not change). 

I've also written such reclassifiers for all relevant transformers EDIT: ~~although technically the Mixed strategy transformer isn't as smart as it could be but I suggest this gets merged under the bug fix rule and an issue gets opened for that transformer. (I've set it to blankly say that it make strategies stochastic which is what we want for any sensible use case).~~ I've done that one too now.

Closes #971

### DIFF
--- a/axelrod/strategy_transformers.py
+++ b/axelrod/strategy_transformers.py
@@ -6,6 +6,7 @@ See the various Meta strategies for another type of transformation.
 """
 
 import collections
+import copy
 import inspect
 import random
 from numpy.random import choice
@@ -21,7 +22,8 @@ C, D = Actions.C, Actions.D
 # Alternator.
 
 
-def StrategyTransformerFactory(strategy_wrapper, name_prefix=None):
+def StrategyTransformerFactory(strategy_wrapper, name_prefix=None,
+                               reclassifier=None):
     """Modify an existing strategy dynamically by wrapping the strategy
     method with the argument `strategy_wrapper`.
 
@@ -37,6 +39,8 @@ def StrategyTransformerFactory(strategy_wrapper, name_prefix=None):
         Any keyword arguments to pass to the wrapper
     name_prefix: string, "Transformed "
         A string to prepend to the strategy and class name
+    reclassifier: function,
+        A function to use to update the class's classifier.
     """
 
     # Create a class that applies a wrapper function to the strategy method
@@ -52,14 +56,17 @@ def StrategyTransformerFactory(strategy_wrapper, name_prefix=None):
             else:
                 self.name_prefix = name_prefix
 
+            if "reclassifier" in kwargs:
+                self.reclassifier = kwargs["reclassifier"]
+            else:
+                self.reclassifier = reclassifier
+
         def __call__(self, PlayerClass):
             """
             Parameters
             ----------
             PlayerClass: A subclass of axelrod.Player, e.g. Cooperator
                 The Player Class to modify
-            name_prefix: str
-                A string to prepend to the Player and Class name
 
             Returns
             -------
@@ -74,6 +81,10 @@ def StrategyTransformerFactory(strategy_wrapper, name_prefix=None):
                 # If "name_prefix" in kwargs remove as only want decorator
                 # arguments
                 del kwargs["name_prefix"]
+            except KeyError:
+                pass
+            try:
+                del kwargs["reclassifier"]
             except KeyError:
                 pass
 
@@ -109,6 +120,12 @@ def StrategyTransformerFactory(strategy_wrapper, name_prefix=None):
                 # Modify the Player name (class variable inherited from Player)
                 name = ' '.join([name_prefix, PlayerClass.name])
 
+            original_classifier = copy.deepcopy(PlayerClass.classifier) # Copy
+            if reclassifier is not None:
+                classifier = reclassifier(original_classifier, *args, **kwargs,)
+            else:
+                classifier = original_classifier
+
             # Define the new __repr__ method to add the wrapper arguments
             # at the end of the name
             def __repr__(self):
@@ -136,6 +153,7 @@ def StrategyTransformerFactory(strategy_wrapper, name_prefix=None):
                     "strategy": strategy,
                     "__repr__": __repr__,
                     "__module__": PlayerClass.__module__,
+                    "classifier": classifier,
                 })
             return new_class
     return Decorator
@@ -229,9 +247,14 @@ def noisy_wrapper(player, opponent, action, noise=0.05):
         return flip_action(action)
     return action
 
+def noisy_reclassifier(original_classifier, noise):
+    """Function to reclassify the strategy"""
+    if noise not in [0, 1]:
+        original_classifier["stochastic"] = True
+    return original_classifier
 
 NoisyTransformer = StrategyTransformerFactory(
-    noisy_wrapper, name_prefix="Noisy")
+    noisy_wrapper, name_prefix="Noisy", reclassifier=noisy_reclassifier)
 
 
 def forgiver_wrapper(player, opponent, action, p):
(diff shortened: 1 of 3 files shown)


axelrod/strategy_transformers.py
```diff
@@ -424,15 +461,33 @@ def joss_ann_wrapper(player, opponent, proposed_action, probability):
...
 
+def jossann_reclassifier(original_classifier, probability):
+    """
+    Reclassify: note that if probabilities are (0, 1) or (1, 0) then we override
+    the original classifier.
+    """
+    if sum(probability) > 1:
+        probability = tuple([i / sum(probability) for i in probability])
+
+    if probability[0] in [0, 1]:
```
eric-s-s: will this fail on probability=(0.0, 0.5) or the very useless (0.0, 0.0)?

drvinceknight: I don't think so because of the normalisation step that happens before. I'm fixing some remaining failures and will take a look at the (0, 0) case! :+1:

drvinceknight: You were right, that did fall over on those cases :) In 149e2e6 I've added more tests and a fix.

marcharper:
Do we have a test that catches the original issue with cloning these players?

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -349,6 +420,7 @@ def test_mixed(self):
         probability = (0, 0, 1)  # Note can also pass tuple
         strategies = [axelrod.TitForTat, axelrod.Grudger, axelrod.Defector]
         MD = MixedTransformer(probability, strategies)(axelrod.Cooperator)
+        self.assertFalse(MD.classifier["stochastic"])
 
         p1 = MD()
         # Against a cooperator we see that we only cooperate
```
eric-s-s: i've been reading through this to try and wrap my head around decorators, which drive me up the wall.  it has helped but it's also hard reading.  This comment that sneaked in from a copy-paste made me laugh.

drvinceknight:
> Do we have a test that catches the original issue with cloning these players?

I've incorporated a number of tests that check that players are now correctly classified.

This was the original issue (stochastic players were not being recognised as such and they were starting off the cache).

I haven't specifically incorporated a test of the exact symptom but the new tests for the fingerprint (showing that the previous values weren't right) show that this fixes #972

axelrod/tests/unit/test_strategy_transformers.py
drvinceknight: Removing that now :)

drvinceknight: > which drive me up the wall.

Tell me about it! There be dragons... :'(

axelrod/strategy_transformers.py
```diff
@@ -37,6 +39,8 @@ def __call__(self, player, opponent, action)
         Any keyword arguments to pass to the wrapper
     name_prefix: string, "Transformed "
         A string to prepend to the strategy and class name
+    reclassifier: function,
+        A function to use to update the class's classifier.
```
meatballs: Minor quibbles:

1. Somebody new to this might reasonably ask "What class?"
2. "to use" is redundant

How about: "A function which will update the classifier of the strategy being transformed"?

axelrod/strategy_transformers.py
```diff
@@ -229,9 +242,14 @@ def noisy_wrapper(player, opponent, action, noise=0.05):
         return flip_action(action)
     return action
 
+def noisy_reclassifier(original_classifier, noise):
+    """Function to reclassify the strategy"""
+    if noise not in [0, 1]:
```
meatballs: Really nitpicky comment: I'd use a tuple here as there is no need for a list

axelrod/strategy_transformers.py
```diff
@@ -241,9 +259,15 @@ def forgiver_wrapper(player, opponent, action, p):
         return random_choice(p)
     return C
 
+def forgiver_reclassifier(original_classifier, p):
+    """Function to reclassify the strategy"""
+    if p not in [0, 1]:
```
meatballs: same nitpicky tuple vs list comment as previous

drvinceknight:
Have addressed both those @meatballs. I still often slip in the bad habit of using lists when I shouldn't... :+1:

meatballs:
If only that were my only bad habit.....

Vince Knight wrote:

axelrod/strategy_transformers.py
```diff
@@ -294,9 +318,15 @@ def final_sequence(player, opponent, action, seq):
         return seq[-index]
     return action
 
+def final_reclassifier(original_classifier, seq):
```
marcharper: We probably need to update the memory depth as well (also for initial transformer)

drvinceknight: Sure doing that now.

drvinceknight:
I've added reclassification of the memory depth too.

I think this should go in now under the bug fix rule. If there's anything remaining we can pick it up as other issues (I don't think there are but there might be other reclassifiers that need to be added but the main bug was being caused by the stochastic ones).

PR #940, opened 2017-03-28
885

YetiYetYet:
Hiiiiiiiiiiiiiiiiiiiiiii, So I did the `__eq__` method with your help and I added some test. Please tell me if it's to your convenience.
<3

PR #975, opened 2017-04-16
Implement an eq method for players

### LINKED ISSUE #885: Implement an `__eq__` method for players
A way to test equality of players. This should not only check the class, str but also that all attributes are equal.

That sort of check is currently used in the `TestPlayer` class so as well as implemented an `__eq__` metho a very minor refactor of the `TestPlayer` class can be carried out.

drvinceknight:
- Add equality method and tests to player.
- Add documentation for player equality.
- Refactored some other tests to make use of this method
- Refactor a match generator test which was failing as a result of the
eq method. Previously this was testing equality of instance.

Some other minor changes I've made as a result of this:

- Change when qlearner makes random choice. I've also removed the reset
  test. This is now being tested by the parent test class for all
  strategies.
- Fix small bug in contrite. Without this, contrite was not resetting
  correctly (the recorded history was being maintained in the class
  attribute).

This closes #885 but also closes #940 which was worked on a while ago
but which has not seen any progress.

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -149,10 +149,17 @@ class Player(object):
                 generator, original_value = itertools.tee(value)
                 other_generator, original_other_value = itertools.tee(other_value)
 
-                setattr(self, attribute,
-                        (ele for ele in original_value))
-                setattr(other, attribute,
-                        (ele for ele in original_other_value))
+
+                if isinstance(value, types.GeneratorType):
+                    setattr(self, attribute,
+                            (ele for ele in original_value))
+                    setattr(other, attribute,
+                            (ele for ele in original_other_value))
+                else:
+                    setattr(self, attribute,
+                            itertools.cycle(original_value))
+                    setattr(other, attribute,
+                            itertools.cycle(original_other_value))
 
                 if not (all(next(generator) == next(other_generator)
                         for _ in range(200))):
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -164,6 +164,12 @@ class Player(object):
                 if not (all(next(generator) == next(other_generator)
                         for _ in range(200))):
                     return False
+
+            # Code for a strange edge case where each strategy points at each
+            # other
+            elif (value is other and other_value is self):
+                    pass
+
             else:
                 if value != other_value:
                     return False
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -3,6 +3,9 @@ from functools import wraps
 import random
 import copy
 import inspect
+import types
+import numpy as np
+import itertools
 
 from axelrod.actions import Actions, flip_action, Action
 from .game import DefaultGame
@@ -121,6 +124,44 @@ class Player(object):
         self.defections = 0
         self.state_distribution = defaultdict(int)
         self.set_match_attributes()
+
+
+    def __eq__(self, other):
+        """
+        Test if two players are equal.
+        """
+        if self.__repr__() != other.__repr__():
+            return False
+        for attribute in set(list(self.__dict__.keys()) +
+                             list(other.__dict__.keys())):
+
+            value = getattr(self, attribute, None)
+            other_value = getattr(other, attribute, None)
+
+            if isinstance(value, np.ndarray):
+                if not (np.array_equal(value, other_value)):
+                    return False
+
+            elif isinstance(value, types.GeneratorType) or \
+                 isinstance(value, itertools.cycle):
+
+                # Split the original generator so it is not touched
+                generator, original_value = itertools.tee(value)
+                other_generator, original_other_value = itertools.tee(other_value)
+
+                setattr(self, attribute,
+                        (ele for ele in original_value))
+                setattr(other, attribute,
+                        (ele for ele in original_other_value))
+
+                if not (all(next(generator) == next(other_generator)
+                        for _ in range(200))):
+                    return False
+            else:
+                if value != other_value:
+                    return False
+        return True
+
 
     def receive_match_attributes(self):
         # Overwrite this function if your strategy needs
--- a/axelrod/tests/strategies/test_player.py
+++ b/axelrod/tests/strategies/test_player.py
@@ -207,6 +207,9 @@ class TestPlayerClass(unittest.TestCase):
         self.assertEqual(list(p1.generator), list(range(10)))
         self.assertEqual(list(p2.generator), list(range(1, 10)))
 
+        # Check that type is generator
+        self.assertIsInstance(p2.generator, types.GeneratorType)
+
     def test_equality_for_cycle(self):
         """Test equality works with cycle attribute and that the cycle attribute
         is not altered during checking of equality"""
@@ -231,11 +234,15 @@ class TestPlayerClass(unittest.TestCase):
         self.assertEqual(next(p1.cycle), 0)
         self.assertEqual(next(p2.cycle), 1)
 
+        # Check that type is cycle
+        self.assertIsInstance(p2.cycle, itertools.cycle)
+
     def test_equaity_on_init(self):
         """Test all instances of a strategy are equal on init"""
         for s in axelrod.strategies:
             p1, p2 = s(), s()
-            # Check twice (so testing equality doesn't change anything)
+            # Check three times (so testing equality doesn't change anything)
+            self.assertEqual(p1, p2)
             self.assertEqual(p1, p2)
             self.assertEqual(p1, p2)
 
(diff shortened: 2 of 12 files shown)


axelrod/player.py
```diff
@@ -122,6 +125,44 @@ def __init__(self):
...
+                # Split the original generator so it is not touched
+                generator, original_value = itertools.tee(value)
+                other_generator, original_other_value = itertools.tee(other_value)
+
+                setattr(self, attribute,
+                        (ele for ele in original_value))
+                setattr(other, attribute,
+                        (ele for ele in original_other_value))
+
+                if not (all(next(generator) == next(other_generator)
```
marcharper: This could cause unintended behavior if used during a match or tournament.

axelrod/tests/strategies/test_player.py
```diff
@@ -150,6 +152,78 @@ def test_clone(self):
             self.assertEqual(len(player1.history), turns)
             self.assertEqual(player1.history, player2.history)
 
+    def test_equality(self):
+        """Test the equality method for some bespoke cases"""
+        # Check repr
+        p1 = axelrod.Cooperator()
+        p2 = axelrod.Cooperator()
+        self.assertEqual(p1, p2)
+        p1.__repr__ = lambda :"John Nash"
```
marcharper: spacing

axelrod/player.py
drvinceknight: I don't see what you mean? 

I've done this in such  a way that the generator gets copied and there are tests that show that using `__eq__` does not alter the generator.

drvinceknight: lines 199 and 215:

```python
187     def test_equality_for_generator(self):                                                                                                        
188         # Check generator attribute (a special case)                                                                                              
189         p1 = axelrod.Cooperator()                                                                                                                 
190         p2 = axelrod.Cooperator()                                                                                                                 
191         # Check that the generator is still the same                                                                                              
192         p1.generator = (i for i in range(10))                                                                                                     
193         p2.generator = (i for i in range(10))                                                                                                     
194         self.assertEqual(p1, p2)                                                                                                                  
195                                                                                                                                                   
196         _ = next(p2.generator)                                                                                                                    
197         self.assertNotEqual(p1, p2)                                                                                                               
198                                                                                                                                                   
199         # Check that internal generator object has not been changed                                                                               
200         self.assertEqual(list(p1.generator), list(range(10)))                                                                                     
201         self.assertEqual(list(p2.generator), list(range(1, 10)))                                                                                  
202                                                                                                                                                   
203     def test_equality_for_cycle(self):                                                                                                            
204         # Check cycle attribute (a special case)                                                                                                  
205         p1 = axelrod.Cooperator()                                                                                                                 
206         p2 = axelrod.Cooperator()                                                                                                                 
207         # Check that the cycle is still the same                                                                                                  
208         p1.cycle = itertools.cycle(range(10))                                                                                                     
209         p2.cycle = itertools.cycle(range(10))                                                                                                     
210         self.assertEqual(p1, p2)                                                                                                                  
211                                                                                                                                                   
212         _ = next(p2.cycle)                                                                                                                        
213         self.assertNotEqual(p1, p2)                                                                                                               
214                                                                                                                                                   
215         # Check that internal generator object has not been changed                                                                               
216         self.assertEqual(next(p1.cycle), 0)                                                                                                       
217         self.assertEqual(next(p2.cycle), 1)  
```

marcharper: So when the generator is copied it's state is also reset?

drvinceknight: Yup that's right. What `itertools.tee` does is essentially fork a generator in to two versions of itself from the state it currently is in. 

That is also being tests:

1. Create two `Cooperators` and give them two cycle generators and tests that they are equal (at this stage their generator are in the same state:

    ```
    205         p1 = axelrod.Cooperator()                                                                                                                 
    206         p2 = axelrod.Cooperator()                                                                                                                 
    207         # Check that the cycle is still the same                                                                                                
    208         p1.cycle = itertools.cycle(range(10))                                                                                                    
    209         p2.cycle = itertools.cycle(range(10))                                                                                                    
    210         self.assertEqual(p1, p2)   
    ```

2. Bump `p2`s cycle and now the players are not the same:

    ```
    212         _ = next(p2.cycle)                                                                                                                       
    213         self.assertNotEqual(p1, p2)  
    ```

3. Check what the next values in the cycle are for each player, for `p1` it should be `0` because the player has not iterated through any values and for `p2` it should be `1` because it has iterated through the first value:


    ```
    215         # Check that internal generator object has not been changed                                                                              
    216         self.assertEqual(next(p1.cycle), 0)                                                                                                       
    217         self.assertEqual(next(p2.cycle), 1)  
    ```

Let me know if you think I could clarify that in the code and/or add other tests.

drvinceknight: (The walk through I've written there is for cycles but that test is more or less repeated exactly for generators too although things are slightly different in that case.)

drvinceknight: In 308741e, I've rebased on to `master` and also (hopefully) improved the readability of these two tests.

Chadys: Instead of
```python
generator, original_value = itertools.tee(value)
other_generator, original_other_value = itertools.tee(other_value)

setattr(self, attribute,
    (ele for ele in original_value))
 setattr(other, attribute,
    (ele for ele in original_other_value))
```
Couldn't you just do :
```python
generator, value = itertools.tee(value)
other_generator, other_value = itertools.tee(other_value)
```
?

drvinceknight:
No. I tried that originally but then you get in to quirky errors because of
the difference between a cycle and a generator. Here I'm turning them all
in to generators.

axelrod/player.py
drvinceknight: No. I tried that originally but then you get in to quirky errors because of
the difference between a cycle and a generator. Here I'm turning them all
in to generators.

drvinceknight: FYI `test_equality_on_init` is the test that fails in this case.

This should clarify why:

```python
>>> import itertools
>>> import types
>>> cycle = itertools.cycle('ABCD')
>>> new_cycle, backup = itertools.tee(cycle)
>>> type(cycle), type(new_cycle), type(backup)
(itertools.cycle, itertools._tee, itertools._tee)
>>> isinstance(cycle, itertools.cycle), isinstance(cycle, types.GeneratorType)
(True, False)
>>> isinstance(new_cycle, itertools.cycle), isinstance(backup, types.GeneratorType)
(False, False)
```

So the second time around the generator/cycle is not compared in the correct way. I did have a special case for a cycle but found it more elegant to just treat them all the same way. :)

drvinceknight: We could add `isinstance(value, collections.Iterable)` to the first check but that would also catch lists etc which we don't want :)

drvinceknight: > I did have a special case for a cycle but found it more elegant to just treat them all the same way. 

This might actually be a better way as it ensures we don't change anything at all (even the type of the attributes). I've also realised something else that's not right in this test. Going to tweak and push.

drvinceknight: In c5e95d6 I've added a third equality test (theoretically things could break on the second one). I've also added test that ensure the type of a generator/cycle stays the same (which implies I've changed the way I've dealt with the output of `itertools.tee`). :+1: Let me know what you think :)

marcharper: Thanks for the explanations!

marcharper:
I approve. One comment for thought -- it seems possible to get into an infinite loop if player A contains an attribute referencing player B and player B contains a reverse reference. I couldn't think of a legitimate reason to do so -- even with the meta players we don't let them be team members of other meta players -- but maybe worth a few thought for the possibility of other cases.

drvinceknight:
That's a really good point. Let me have a play with a couple of ideas I've got. Will push/comment later today.

drvinceknight:
I've covered this with: eb612f5

Have added a bunch of tests that check we get what we want :)

(I also added a couple of other specific tests for some cases I thought were worth checking :))

drvinceknight:
> Just a small typo on the word equality ;)

Thanks :+1: :)

axelrod/tests/strategies/test_player.py
```diff
@@ -246,6 +260,40 @@ def test_equaity_on_init(self):
             self.assertEqual(p1, p2)
             self.assertEqual(p1, p2)
 
+    def test_equaity_on_with_player_attributes(self):
```
marcharper: equality

marcharper:
@meatballs you're up when you get a chance

PR #977, opened 2017-04-17
Add double resurrection to strategies list.

drvinceknight:
We missed this on #876. (Arguably this is a bug)


meatballs:
Merged as a bug fix

PR #979, opened 2017-04-18
Add scipy to requirements

souravsingh:
The PR adds scipy dependency for Axelrod's Stein and Rapoport Strategy.

Would it make sense to add ``scipy`` as an optional dependency, since we require it for only one strategy (so far)?


drvinceknight:
Not sure why it failed but let's not add this as a dependency until it's needed (so until it comes with the PR for the strategy itself: there might be another way of doing). I'm closing this but it can be reopened if need be.

PR #980, opened 2017-04-19
Add strategy index test to travis.

### LINKED ISSUE #957: A script to automatically detect if all strategy modules are listed in `docs/reference/all_strategies.rst`
Currently, whenever a new strategy module (file) is added to `axelrod/strategies` it needs to be also added to `docs/reference/all_strategies.rst` to ensure that an index of all strategies is built as can be seen here: http://axelrod.readthedocs.io/en/latest/reference/all_strategies.html

This needs to be checked during the review process but it would be helpful to have a script that could be run (by travis) to indicate that all the modules have been listed. So for example it would look like follows if all were listed (no output):

```
$ python check_strategy_index.py
$ 
```

Of if the `axelrod/strategies/tit_for_tat.py` file was not listed it would return ` sys.exit(1)` (so that it's recognised as a failure):

```
$ python check_strategy_index.py
ERROR: `axelrod/strategies/tit_for_tat.py` is not included in the index.
```

This could be done with an inspection of the `axelrod/strategies` directory as well as using some regex on the `docs/reference/all_strategies.rst` file.

Note that we have explored automatically generating the `all_strategies.rst` file and came to a dead end (we can't get read the docs to do that for us for example).

drvinceknight:
Closes #957. I am expecting this to fail on travis: two modules are not currently indexed (`hmm` and `memorytwo`). Using that as a test that this script works (once travis fails, will push the fix).

### DIFF
--- a/run_strategy_indexer.py
+++ b/run_strategy_indexer.py
@@ -2,80 +2,45 @@
 A script to check that all strategy modules have been included in
 `./docs/reference/all_strategies.rst`
 """
-import glob
 import sys
-import os
+import pathlib
 
-def read_index(index_path: str) -> str:
-    """
-    Read the index of strategies
+default_index_path = pathlib.Path("./docs/reference/all_strategies.rst")
+excluded_modules = ("_strategies", "__init__", "_filters", "human")
 
-    Parameters
-    ----------
-    index_path : str
-        A file path for the index file where all strategies are auto documented
-
-    Returns
-    -------
-    strategies_reference : str
-        The string value of the contents of the index file.
-    """
-    with open(index_path, "r") as f:
-        strategies_reference = f.read()
-    return strategies_reference
-
-def get_module_name(module_path: str) -> str:
-    """
-    Take string of the form `./axelrod/strategies/titfortat.py` and returns
-    `titfortat`.
-
-    Parameters
-    ----------
-    module_path : str
-        A file path for a module file.
-
-    Returns
-    -------
-    module_name : str
-        The name of the module
-    """
-    filename = os.path.basename(module_path)
-    module_name = os.path.splitext(filename)[0]
-    return module_name
-
-def check_module(module_path: str,
-                 index_path: str="./docs/reference/all_strategies.rst",
-                 excluded: tuple=("_strategies", "__init__",
-                                  "_filters", "human")) -> bool:
+def check_module(module_path: pathlib.Path,
+                 index_path: pathlib.Path=default_index_path,
+                 excluded: tuple=excluded_modules) -> bool:
     """
     Check if a module name is written in the index of strategies.
 
     Parameters
     ----------
-    module_path : str
+    module_path :
         A file path for a module file.
-    index_path : str
+    index_path :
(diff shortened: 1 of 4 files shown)


drvinceknight:
4738a96 has failed as expected:

```
 0.04s$ python run_strategy_indexer.py
memorytwo not in index
The command "python run_strategy_indexer.py" exited with 1.
```

Just pushing the fix.

run_strategy_indexer.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    filename = os.path.basename(module_path)
+    module_name = os.path.splitext(filename)[0]
+    return module_name
+
+def check_module(module_path,
+                 index_path="./docs/reference/all_strategies.rst",
+                 excluded=("_strategies", "__init__", "_filters", "human")):
+    """
+    Check if a module is contained in the index of strategies
```
marcharper: Can we improve the docstring? Technically these aren't modules. Also let's use the standard style for parameters (just module_path is ok?) and the return value.

run_strategy_indexer.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Returns boolean: True/False if module is referenced.
+    """
+    strategies_reference = read_index(index_path)
+    module_name = get_module_name(module_path)
+    if module_name not in excluded:
```
marcharper: could use a single if (either way is ok)

run_strategy_indexer.py
```diff
@@ -0,0 +1,49 @@
...
+        if module_name not in strategies_reference:
+            print("{} not in index".format(module_name))
+            return False
+    return True
+
+
+if __name__ == "__main__":
+
+    modules = glob.glob("./axelrod/strategies/*py")
```
marcharper: "*.py" might be slightly more future proof

drvinceknight:
I think I've addressed your comments @marcharper. I've also added type hinting :)

meatballs:
I'm happy with this but I think an improvement might be to use the Pathlib module for file paths.

I don't think there's a problem here, but I've have had problems with file paths on windows before and Pathlib handles all that perfectly.

drvinceknight:
That's a good shout. I'll do that and push today :+1:

drvinceknight:
> but I've have had problems with file paths on windows before and Pathlib handles all that perfectly.

Great shout on Pathlib, it cleared up a lot of the code (no real point in 2 of the 3 functions as they're just simple method calls of the `Path` class). I'd seen it used before but never played around with it myself: it's lovely!.

Note: I removed the type specification in the docstring. This is mainly because I noticed just before pushing that it had become a lie (I changed the type hint to be a `Path` and didn't change the docstring). Seems sensible to me to remove the types from the docstrings when we have type hints in place but I'm happy to throw them back in for this function if we want them.

meatballs:
> Great shout on Pathlib, it cleared up a lot of the code (no real point in 2 of the 3 functions as they're just simple method calls of the Path class). I'd seen it used before but never played around with it myself: it's lovely!.

Yep. I'm a big fan!

PR #983, opened 2017-04-19
Add the random_ module to mypy tester.

drvinceknight:


marcharper:
I think we can safely merge without a second review (it's sort of a bug anyway).

PR #981, opened 2017-04-19
Refactor tests for handshake.

drvinceknight:
For #884

PR #989, opened 2017-04-25
884 gobymajority, gradualkiller, grudger, grumpy

eric-s-s:
#884 

There is a likely off-by-one error in ForgetfulGrudger.  The docstring was the following.

```
        """Begins by playing C, then plays D for mem_length rounds if the
        opponent ever plays D."""
```

mem_length was 10 but played D for 11 rounds.  I changed the code to follow the docstring.

### DIFF
--- a/axelrod/strategies/grudger.py
+++ b/axelrod/strategies/grudger.py
@@ -67,11 +67,11 @@ class ForgetfulGrudger(Player):
             self.grudge_memory = 0
             self.grudged = False
 
+        if not self.grudged and D in opponent.history[-1:]:
+            self.grudged = True
+
         if self.grudged:
             self.grudge_memory += 1
-            return D
-        elif D in opponent.history[-1:]:
-            self.grudged = True
             return D
         return C
 
@@ -206,9 +206,9 @@ class GrudgerAlternator(Player):
         """Begins by playing C, then plays Alternator for the remaining rounds
         if the opponent ever plays D."""
         if opponent.defections:
-            if self.history[-1] == Actions.C:
-                return Actions.D
-        return Actions.C
+            if self.history[-1] == C:
+                return D
+        return C
 
 
 class EasyGo(Player):
@@ -240,13 +240,14 @@ class EasyGo(Player):
             return C
         return D
 
+
 class GeneralSoftGrudger(Player):
     """
-    A generalization of the SoftGrudger strategy. SoftGrudger punishes by playing:
-    D, D, D, D, C, C. after a defection by the opponent.  GeneralSoftGrudger
-    only punishes after its opponent defects a specified amount of times consecutively.
-    The punishment is in the form of a series of defections followed by a 'penance' of
-    a series of consecutive cooperations.
+    A generalization of the SoftGrudger strategy. SoftGrudger punishes by
+    playing: D, D, D, D, C, C. after a defection by the opponent.
+    GeneralSoftGrudger only punishes after its opponent defects a specified
+    amount of times consecutively. The punishment is in the form of a series of
+    defections followed by a 'penance' of a series of consecutive cooperations.
 
     Names:
 
@@ -268,11 +269,11 @@ class GeneralSoftGrudger(Player):
         """
         Parameters
         ----------
-        n, int
+        n: int
             The number of defections by the opponent to trigger punishment
-        d, int
+        d: int
             The number of defections to punish the opponent
-        c, int
+        c: int
             The number of cooperations in the 'penance' stage
 
         Special Cases
--- a/axelrod/tests/strategies/test_gobymajority.py
+++ b/axelrod/tests/strategies/test_gobymajority.py
@@ -2,6 +2,7 @@
 
 import axelrod
 from .test_player import TestPlayer
+from axelrod import MockPlayer
 
 C, D = axelrod.Actions.C, axelrod.Actions.D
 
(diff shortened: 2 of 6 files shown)


axelrod/strategies/grudger.py
```diff
@@ -240,13 +240,14 @@ def strategy(opponent: Player) -> Action:
...
 
+
 class GeneralSoftGrudger(Player):
     """
-    A generalization of the SoftGrudger strategy. SoftGrudger punishes by playing:
-    D, D, D, D, C, C. after a defection by the opponent.  GeneralSoftGrudger
-    only punishes after its opponent defects a specified amount of times consecutively.
-    The punishment is in the form of a series of defections followed by a 'penance' of
-    a series of consecutive cooperations.
+    A generalization of the SoftGrudger strategy. SoftGrudger punishes by
```
eric-s-s: this is simply fitting docstring to 80 chars per line.

axelrod/strategies/grudger.py
```diff
@@ -67,12 +67,12 @@ def strategy(self, opponent: Player) -> Action:
             self.grudge_memory = 0
             self.grudged = False
 
+        if not self.grudged and D in opponent.history[-1:]:
```
eric-s-s: The code and docstring didn't match.  the docstring said to play D for 10 rounds, but the code played D for 11 rounds.  I wasn't sure which to follow, so for now, I changed the code to fit the docstring.  

for a small issue, is this a correct way to raise it?

drvinceknight: > The code and docstring didn't match. the docstring said to play D for 10 rounds, but the code played D for 11 rounds. I wasn't sure which to follow, so for now, I changed the code to fit the docstring.
> 
> for a small issue, is this a correct way to raise it?

Yup: this is fine and I think you've changed it correctly. The behaviour was not matching the description. :+1:

axelrod/tests/strategies/test_grudger.py
```diff
@@ -128,24 +154,26 @@ class TestSoftGrudger(TestPlayer):
...
-        self.responses_test([D], [C, C, D, D], [C, D, C, C])
-        self.responses_test([D], [C, C, D, D, D], [C, D, C, C, C])
-        self.responses_test([C], [C, C, D, D, D, D], [C, D, C, C, C, C])
-        self.responses_test([C], [C, C, D, D, D, D, C], [C, D, C, C, C, C, C])
-        self.responses_test([D], [C, C, D, D, D, D, C, C],
-                            [C, D, C, C, C, C, C, D])
-        self.responses_test([D], [C, C, D, D, D, D, C, C, D],
-                            [C, D, C, C, C, C, C, D, C])
+        vs_cooperator = [(C, C)] * 30
+        self.versus_test(axl.Cooperator(), expected_actions=vs_cooperator)
```
drvinceknight: To be in line with tests already written could we stick with `actions=[(C, C)] * 30` instead of `vs_cooperator`. (Similarly elsewhere).

axelrod/tests/strategies/test_grudger.py
```diff
@@ -236,22 +259,40 @@ class TestGeneralSoftGrudger(TestPlayer):
...
-        actions = [(C, D), (D, D), (D, C), (D, C), (D, D), (C, D), (C, C), (C, C)]
-        self.versus_test(axelrod.MockPlayer(actions=[D, D, C, C]), expected_actions=actions)
-
-        # Testing n=2, d=4, c=2
-        actions = [(C, D), (C, D), (D, C), (D, C), (D, D), (D, D), (C, C), (C, C)]
-        self.versus_test(axelrod.MockPlayer(actions=[D, D, C, C]), expected_actions=actions,
-                         init_kwargs={"n": 2})
-
-        # Testing n=1, d=1, c=1
-        actions = [(C, D), (D, D), (C, C), (C, C), (C, D), (D, D), (C, C), (C, C)]
```
drvinceknight: Why are you re writing these tests? They are already in the correct form.

axelrod/tests/strategies/test_gradualkiller.py
```diff
@@ -19,78 +19,59 @@ class TestGradualKiller(TestPlayer):
...
-        P2.history = [C, C, C, C, C, D, D, C, C]
-        self.assertEqual(P1.strategy(P2), 'D')
-        P1.history = [D, D, D, D, D, C, C, D, D, D]
-        P2.history = [C, C, C, C, C, D, D, C, C, D]
-        self.assertEqual(P1.strategy(P2), 'D')
+        expected, opponent = self.get_actions_for_test(C, [D, D], D)
+        self.versus_test(axl.MockPlayer(actions=opponent),
+                         expected_actions=expected)
+
+    def get_actions_for_test(self, opponent_start, action_six_and_seven,
```
drvinceknight: Unless they're truly necessary it's nicer to not make use of these helper functions and write out tests explicitly: otherwise technically we also want a test for this helper function. 

If we do actually need this helper function then it should certainly be better documented.

axelrod/tests/strategies/test_gobymajority.py
```diff
@@ -24,40 +25,55 @@ class TestHardGoByMajority(TestPlayer):
...
-        player = self.player(soft=False)
-        opponent = axelrod.Cooperator()
-        self.assertEqual('D', player.strategy(opponent))
-
-    def test_soft(self):
+    soft = True
+
+    def test_set_soft_to_false(self):
+        self.eq_play = D
+        expected, opponent_actions = self.get_infinite_memory_depth_actions()
```
drvinceknight: Not a fan of these helper functions. I have no idea what this does and subsequently am not sure what this test is doing. Can you just write out the test explicitly?

I'd prefer it if we kept the tests very clear (technically these helper functions need tests). Of course there might be a valid argument for using a helper function here but as is it just obfuscates the test here (no docstring etc). In this case for example it looks like you're just concatenating 3 lists together: so I'd just go with that.

axelrod/tests/strategies/test_gobymajority.py
```diff
@@ -24,40 +25,55 @@ class TestHardGoByMajority(TestPlayer):
...
+        self.versus_test(MockPlayer(actions=opponent_actions),
+                         expected_actions=expected, init_kwargs={'soft': False})
+        self.eq_play = C
```
drvinceknight: Not sure why this needs to be a class attribute. (I'm not actually sure what this is doing.)

drvinceknight:
If it's helpful: please do feel free to open multiple PRs for each module (it helps with reviewing especially when/if you're going to make more than just the straightforward changes converting `test_responses` to `versus_test`). For example I did that recently with #981 and #982, which I opened one after the over, even though they were standard conversions. :)

eric-s-s:
> If it's helpful: please do feel free to open multiple PRs for each module (it helps with reviewing especially when/if you're going to make more than just the straightforward changes converting test_responses to versus_test).

yes.  that seems like a good idea.  i'll open PR's for each one.  close this one?  (i'll, of course be working off the comments you made and change those things before pushing up the new PR's.  this will be in a couple of days)

eric-s-s:
bah! wrong button.  i meant to ask if i should close, not close. >(

drvinceknight:
> yes. that seems like a good idea. i'll open PR's for each one. close this one? (i'll, of course be working off the comments you made and change those things before pushing up the new PR's. this will be in a couple of days)

Sounds good to me :) :+1:

PR #993, opened 2017-04-28
884 grudger - with 1 to 3 issues

eric-s-s:
grudger has one issue and possibly three.

- ForgetfulGrudger: changed in PR
    - docstring and code did not match.  docstring said grudge for 10 rounds and code said grudge for 11    rounds
    - changed code to fit docstring
- GeneralSoftGrudger: changed in PR
    - possibly an issue IF edge cases are valid.  
    - I tested for edge cases of n=0, d=0 and c=0.  original code did not follow expected.  updated code.
- Possible issue not changed:
    - should ForgetfulGrudger and SoftGrudger inherit from GeneralSoftGrudger?  they are `GeneralSoftGrudger(n=1, d=10, c=0)` and `GeneralSoftGrudger(n=1, d=4, c=2)`


axelrod/strategies/grudger.py
```diff
@@ -240,13 +240,14 @@ def strategy(opponent: Player) -> Action:
...
 
+
 class GeneralSoftGrudger(Player):
     """
-    A generalization of the SoftGrudger strategy. SoftGrudger punishes by playing:
-    D, D, D, D, C, C. after a defection by the opponent.  GeneralSoftGrudger
-    only punishes after its opponent defects a specified amount of times consecutively.
-    The punishment is in the form of a series of defections followed by a 'penance' of
-    a series of consecutive cooperations.
+    A generalization of the SoftGrudger strategy. SoftGrudger punishes by
```
eric-s-s: changed docstring to fit line-llen = 80.

axelrod/strategies/grudger.py
```diff
@@ -63,16 +63,16 @@ def __init__(self) -> None:
     def strategy(self, opponent: Player) -> Action:
         """Begins by playing C, then plays D for mem_length rounds if the
         opponent ever plays D."""
-        if self.grudge_memory >= self.mem_length:
```
eric-s-s: strategy docstring says defect for 10 rounds.  strategy code said detect for 11 rounds.  I changed the code to fit the docstring.

axelrod/strategies/grudger.py
```diff
@@ -293,16 +294,18 @@ def strategy(self, opponent: Player) -> Action:
         The punishment is in the form of 'd' defections followed by a penance of
         'c' consecutive cooperations.
         """
```
eric-s-s: This startegy was changed to fit the edge cases where n=0 or d=0 or c=0.  see the tests here for an example [get url to line 322 of test].  If the edge cases are invalid, then this change is totally unnecessary.

axelrod/tests/strategies/test_grudger.py
```diff
@@ -236,22 +259,105 @@ class TestGeneralSoftGrudger(TestPlayer):
...
-                         init_kwargs={"n": 2})
-
-        # Testing n=1, d=1, c=1
-        actions = [(C, D), (D, D), (C, C), (C, C), (C, D), (D, D), (C, C), (C, C)]
-        self.versus_test(axelrod.MockPlayer(actions=[D, D, C, C]), expected_actions=actions,
-                         init_kwargs={"n": 1, "d": 1, "c": 1})
-
+        # the default follows SoftGrudger
+        super(TestGeneralSoftGrudger, self).test_strategy()
+
```
eric-s-s: @drvinceknight - in the previous PR you asked me why i changed from the original `test_strategy` to the following three tests.  so I thought it over and here is my reasoning.  

1 - I thought tests physically separated from each other and the default behavior tests were more clear.
2 - for `n != 1` I wanted tests that showed the edges at and below the threshold.  (and what happens with constant defections, which, for reasons unknown to me, I kept getting surprised about)

whether or not these are good reasons, I do think it useful to have this inherit from `TestSoftGrudger` which will test the default behavior and `reset`

axelrod/tests/strategies/test_grudger.py
```diff
@@ -236,22 +259,105 @@ class TestGeneralSoftGrudger(TestPlayer):
...
+                         init_kwargs=init_kwargs)
+
+        opponent_actions = [C] * 10 + [D]
+        opponent = axl.MockPlayer(actions=opponent_actions)
+        actions_start = [(C, C)] * 10 + [(C, D)]
+        subsequent = grudge_response_c + [(C, C)] * 4 + [(C, D)]
+        actions = actions_start + subsequent * 5
+        self.versus_test(opponent, expected_actions=actions,
+                         init_kwargs=init_kwargs)
+
```
eric-s-s: Are these edge cases valid or beyond the scope of what this class should be allowed to do?

axelrod/tests/strategies/test_grudger.py
```diff
@@ -236,22 +259,105 @@ class TestGeneralSoftGrudger(TestPlayer):
...
+        opponent_actions = [C] * 10 + [D]
+        opponent = axl.MockPlayer(actions=opponent_actions)
+        actions_start = [(C, C)] * 10 + [(C, D)]
+        subsequent = [(D, C)] * 4 + [(C, C)] * 6 + [(C, D)]
+        actions = actions_start + subsequent * 5
+        self.versus_test(opponent, expected_actions=actions,
+                         init_kwargs=init_kwargs)
+
```
eric-s-s: I noticed in GoByMajority that it change `self.name` to reflect the parameters but that this updates `repr` to reflect the parameters.  is one preferable to the other?

axelrod/strategies/grudger.py
drvinceknight: :+1:

axelrod/strategies/grudger.py
drvinceknight: >  If the edge cases are invalid,

I don't understand what you mean.

axelrod/tests/strategies/test_grudger.py
drvinceknight: What do you meant by "valid"?

axelrod/tests/strategies/test_grudger.py
drvinceknight: Let's not change that for now.

axelrod/tests/strategies/test_grudger.py
drvinceknight: I am a bit weary of PRs that change things that have been (very recently) added when the motivation for change seems stylistic. This takes a significant cognitive effort by the reviewers to make sure we're not breaking anything. If you recall there were a fair few times I had to explain why something was done a particular way in the fingerprint class. 

Please keep these as they are. #884 is a simple issue: refactor tests to use `versus_test`. If you catch a case that is missing by all means add it.

> whether or not these are good reasons, I do think it useful to have this inherit from TestSoftGrudger which will test the default behavior and reset

`test_reset` can actually be removed entirely by the way, this is now being tested by the parent class `TestPlayer`.

axelrod/strategies/grudger.py
eric-s-s: sorry.  that wasn't clear.  

are n=0, d=0 and c=0 legal inputs?  

if they are, the original code failed on n=0 and d=0.  This change is to fix that.  

if they are not allowed as inputs, then the original code works just fine and i can remove this change.  (but perhaps it would be useful to add to `n>=1 and d>=1` to the docstring)

axelrod/tests/strategies/test_grudger.py
eric-s-s: are n=0, d=0 and c=0 legal/allowed inputs?  or is it required that n>=1, d>=1, c>=1?

axelrod/tests/strategies/test_grudger.py
eric-s-s: i apologize.  i have great difficulty figuring these kinds of things out.  this is an entirely new set of social rules far outside of my usual experience. the very explicit "this is annoying for this reason and causes this kind of trouble" is quite helpful.  thank you for that.

eric-s-s: i'll wait on replies to the other two questions and fix this in the next PR.

axelrod/strategies/grudger.py
eric-s-s: and on reflection, those cases would never be called.  i will remove the changes.

axelrod/tests/strategies/test_grudger.py
drvinceknight: :+1: Sorry for the late reply, I've been busy on other things. Looking through it now :)

axelrod/tests/strategies/test_grudger.py
```diff
@@ -44,34 +51,36 @@ class TestForgetfulGrudger(TestPlayer):
...
-        # Forgets the grudge eventually
-        i = 10
-        self.responses_test([C], [C, C] + [D] * i + [C], [C, D] + [C] * i + [C],
-                            attrs={"grudged": False, "grudge_memory": 0,
-                                   "mem_length": 10})
+        # If opponent defects at any point then the player will respond with
+        # D ten times and then continue to check for defections.
+        opponent = axl.Cooperator()
+        actions = [(C, C)] * 20
+        self.versus_test(opponent, expected_actions=actions)
```
drvinceknight: These look great but we're losing the `attrs` test above. Could you add them in?

From: http://axelrod.readthedocs.io/en/latest/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html:

> The function versus_test also accepts a dictionary parameter of attributes to check at the end of the match. For example this test checks if the player’s internal variable opponent_class is set to "Cooperative":

Let me know if I can assist.

PR #997, opened 2017-04-30
Add Alexei strategy

souravsingh:
Strategy is adapted from http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -636,3 +636,30 @@ class SlowTitForTwoTats2(Player):
 
         # Otherwise play previous move
         return self.history[-1]
+
+class Alexei(Player):
+    """
+    Plays similar to Tit-for-Tat, but always defect on last turn.
+    
+    Names:
+
+    - From http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/    
+    """
+
+    name = 'Alexei'
+    classifier = {
+        'memory_depth': 1,
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if not self.history:
+            return C
+        if opponent.history[-1] == D:
+            return D
+        return D
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -637,7 +637,7 @@ class SlowTitForTwoTats2(Player):
         # Otherwise play previous move
         return self.history[-1]
 
-@FinalTransformer(D, name_prefix=None)
+@FinalTransformer((D, ) name_prefix=None)
 class Alexei(Player):
     """
     Plays similar to Tit-for-Tat, but always defect on last turn.
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -637,7 +637,7 @@ class SlowTitForTwoTats2(Player):
         # Otherwise play previous move
         return self.history[-1]
 
-@FinalTransformer((D,) name_prefix=None)
+@FinalTransformer((D,), name_prefix=None)
 class Alexei(Player):
     """
     Plays similar to Tit-for-Tat, but always defect on last turn.
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -1,6 +1,6 @@
 from axelrod.actions import Actions, Action
 from axelrod.player import Player
-from axelrod.strategy_transformers import TrackHistoryTransformer
+from axelrod.strategy_transformers import TrackHistoryTransformer, FinalTransformer
 
 C, D = Actions.C, Actions.D
 
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -644,7 +644,7 @@ class Alexei(Player):
     
     Names:
 
-    - From http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/    
+    - Alexei's Strategy: [LessWrong2011]_    
     """
 
     name = 'Alexei'
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -651,7 +651,7 @@ class Alexei(Player):
     classifier = {
         'memory_depth': 1,
         'stochastic': False,
-        'makes_use_of': set(),
+        'makes_use_of': {'length'},
         'long_run_time': False,
         'inspects_source': False,
         'manipulates_source': False,
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -647,7 +647,7 @@ class Alexei(Player):
     - From http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/    
     """
 
-    name = 'Alexei: D'
+    name = 'Alexei'
     classifier = {
         'memory_depth': 1,
         'stochastic': False,
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -637,6 +637,7 @@ class SlowTitForTwoTats2(Player):
         # Otherwise play previous move
         return self.history[-1]
 
+@FinalTransformer(D, name_prefix=None)
 class Alexei(Player):
     """
     Plays similar to Tit-for-Tat, but always defect on last turn.
@@ -662,4 +663,4 @@ class Alexei(Player):
             return C
         if opponent.history[-1] == D:
             return D
-        return D
+        return C
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -647,7 +647,7 @@ class Alexei(Player):
     - From http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/    
     """
 
-    name = 'Alexei'
+    name = 'Alexei: D'
     classifier = {
         'memory_depth': 1,
         'stochastic': False,
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -637,7 +637,7 @@ class SlowTitForTwoTats2(Player):
         # Otherwise play previous move
         return self.history[-1]
 
-@FinalTransformer((D, ) name_prefix=None)
+@FinalTransformer((D,) name_prefix=None)
 class Alexei(Player):
     """
     Plays similar to Tit-for-Tat, but always defect on last turn.
@@ -649,7 +649,7 @@ class Alexei(Player):
 
     name = 'Alexei'
     classifier = {
-        'memory_depth': 1,
+        'memory_depth': float('inf'),
         'stochastic': False,
         'makes_use_of': {'length'},
         'long_run_time': False,
--- a/axelrod/tests/strategies/test_titfortat.py
+++ b/axelrod/tests/strategies/test_titfortat.py
@@ -608,7 +608,7 @@ class TestAlexei(TestPlayer):
                          seed=1)
 
         opponent = axelrod.MockPlayer(actions=[C, D])
-        actions = [(C, C), (C, D), (D, C), (C, D)]
+        actions = [(C, C), (C, D), (D, C), (D, D)]
         self.versus_test(opponent, expected_actions=actions)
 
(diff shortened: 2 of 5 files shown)


drvinceknight:
Tests are failing on a syntax error.

meatballs:
This is failing at  `axelrod/tests/strategies/test_player.py", line 361`
```
AssertionError: 'D' != 'C'
```

souravsingh:
I am not sure if the strategy written is correct.

souravsingh:
Can someone check the strategy written? It plays like Tit-for-Tat with a defection at the end.

drvinceknight:
The strategy you have written does not do what you describe (there's no reference to the number of turns in it which is what is needed to be able to defect on the last move). Take a look at strategy transformers for an way to do this easily: http://axelrod.readthedocs.io/en/latest/tutorials/advanced/strategy_transformers.html

In particular use the Final Transformer: take a look at backstabber.

axelrod/strategies/titfortat.py
```diff
@@ -636,3 +636,31 @@ def strategy(self, opponent: Player) -> Action:
...
+    Names:
+
+    - From http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/    
+    """
+
+    name = 'Alexei'
+    classifier = {
+        'memory_depth': 1,
+        'stochastic': False,
+        'makes_use_of': set(),
```
Nikoleta-v3: your strategy makes use of length.

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -563,3 +563,54 @@ def test_strategy(self):
...
+
+    def test_strategy(self):
+        self.first_play_test(C)
+        self.second_play_test(rCC=C, rCD=D, rDC=C, rDD=D)
+
+        actions = [(C, C), (C, D), (D, C), (C, D), (D, C)]
+        self.versus_test(axelrod.Alternator(), expected_actions=actions)
+
+        actions = [(C, C), (C, C), (C, C), (C, C), (C, C)]
+        self.versus_test(axelrod.Cooperator(), expected_actions=actions)
```
Nikoleta-v3: You are getting a failure against Cooperator.

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -563,3 +563,54 @@ def test_strategy(self):
...
         self.versus_test(opponent, expected_actions=actions)
+
+class TestAlexei(TestPlayer):
+    """
+    Note that this test is referred to in the documentation as an example on
+    writing tests.  If you modify the tests here please also modify the
+    documentation.
+    """
+
+    name = "Alexei"
```
Nikoleta-v3: Now that your are using `@FinalTransformer` the name of your strategy is different.
`AssertionError: 'Alexei: D' != 'Alexei'`

Take a look at the backstabber tests https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/strategies/test_backstabber.py.

Nikoleta-v3:
@souravsingh thank you for the changes, though you did not get everything correct. 

1. now that you have included that your strategy is making use of length you should alter your test
2. Looking at how the `@FinalTransformer` works the name of the strategy should be `Given Name: the action you specified`. Thus 'Alexei: D'. 
3. when you play against strategies could you make sure you write down the expected actions by hand first. For example your strategy should always defect on the last round but that's not the case in your tests.

I would recommend you run your tests before pushing and if you are getting an error you can't solve feel free to throw it in here!

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -563,3 +563,54 @@ def test_strategy(self):
...
+        actions = [(C, C), (C, D), (D, D), (D, C)]
+        self.versus_test(axelrod.Random(), expected_actions=actions,
+                         seed=1)
+
+        opponent = axelrod.MockPlayer(actions=[C, D])
+        actions = [(C, C), (C, D), (D, C), (D, D)]
+        self.versus_test(opponent, expected_actions=actions)
+
+        opponent = axelrod.MockPlayer(actions=[C, C, D, D, C, D])
+        actions = [(C, C), (C, C), (C, D), (D, D), (D, C), (C, D)]
```
Nikoleta-v3: Try to keep in mind how your strategy plays. Again I would do the expected actions by hand before writing the test. Your strategy has a rule, Defect on the last round

souravsingh:
@Nikoleta-v3 All tests pass except for one which is unrelated to the strategy.

drvinceknight:
> All tests pass except for one which is unrelated to the strategy.

@souravsingh that failing test might be related to the strategy in some way (I'm not sure how) but all tests pass on the library so that failure is being cause by this addition.

axelrod/strategies/titfortat.py
```diff
@@ -636,3 +636,31 @@ def strategy(self, opponent: Player) -> Action:
...
         return self.history[-1]
+
+@FinalTransformer(D, name_prefix=None)
+class Alexei(Player):
+    """
+    Plays similar to Tit-for-Tat, but always defect on last turn.
+    
+    Names:
+
+    - From http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/
```
drvinceknight: Can you reference this as for other strategies please.

souravsingh:
Was there an update made to ``test/unit/test_tournament.py`` before version 2.11 was released? If so, I just have to do a rebase and it should be fixed.

drvinceknight:
> Was there an update made to test/unit/test_tournament.py before version 2.11 was released? If so, I just have to do a rebase and it should be fixed.

I don't believe so but rebasing anyway sounds like a good idea. Then you can run the full test locally `python -m unittest discover` from the master branch and this branch.

drvinceknight:
I've just confirmed that all tests on `master` pass (locally).

drvinceknight:
And I've just confirmed the failure locally. It's failing on the test for probalistic ending tournaments with probability 1. So it's something to do with the strategy not behaving as expected when it does not know the length of the game. Could point to something not right with the Transformer. I'm investigating.

drvinceknight:
This brings up the failure:

```
python -m unittest axelrod.tests.unit.test_tournament.TestProbEndingSpatialTournament
```

drvinceknight:
The error seems to be something to do with the `Final` transformer not working when there sequence passed has length 1.

axelrod/strategies/titfortat.py
```diff
@@ -636,3 +636,31 @@ def strategy(self, opponent: Player) -> Action:
 
         # Otherwise play previous move
         return self.history[-1]
+
+@FinalTransformer(D, name_prefix=None)
```
drvinceknight: `FinalTransformer` takes a sequence. Change this to `(D, )` (although that doesn't fix what's going wrong).

axelrod/strategies/titfortat.py
```diff
@@ -636,3 +636,31 @@ def strategy(self, opponent: Player) -> Action:
...
+    Plays similar to Tit-for-Tat, but always defect on last turn.
+    
+    Names:
+
+    - From http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/    
+    """
+
+    name = 'Alexei'
+    classifier = {
+        'memory_depth': 1,
```
drvinceknight: This strategy has memory depth: `float('inf')` as it needs to keep count of how many turns it has played to be able to know when to defect on the last round. 

Fixing this will fix the other test failure as it was being picked up as a basic strategy (which it is not).

axelrod/strategies/titfortat.py
```diff
@@ -636,3 +636,31 @@ def strategy(self, opponent: Player) -> Action:
 
         # Otherwise play previous move
         return self.history[-1]
+
+@FinalTransformer((D, ) name_prefix=None)
```
drvinceknight: `(D,)` (no space)

drvinceknight:
> The error seems to be something to do with the Final transformer not working when there sequence passed has length 1.

For the record: this was not the issue, the transformer works exactly as expected.

axelrod/strategies/titfortat.py
```diff
@@ -636,3 +636,31 @@ def strategy(self, opponent: Player) -> Action:
 
         # Otherwise play previous move
         return self.history[-1]
+
+@FinalTransformer((D,) name_prefix=None)
```
drvinceknight: You now have a syntax error, type exactly this:

```
@FinalTransformer((D,), name_prefix=None)
```

There's information on running the tests (that would have picked this up immediately) here: http://axelrod.readthedocs.io/en/stable/tutorials/contributing/running_tests.html

axelrod/strategies/titfortat.py
souravsingh: This is the only strategy implemented from the site.

drvinceknight: Yes that's fine but can you move the site to the bibliography and make the `Names` section fit the guidelines: http://axelrod.readthedocs.io/en/stable/tutorials/contributing/strategy/docstrings.html

drvinceknight: Take a look at how `Prison` is written in the bibliography. Let me know if you'd like a hand.

souravsingh: Oh, I had forgotten the docstrings conventions for Axelrod.

drvinceknight: (I think you actually wrote that file I linked to :) :rofl: :+1: )

souravsingh:
@drvinceknight I have made the fixes. Can you review and suggest any more changes?

docs/reference/bibliography.rst
```diff
@@ -20,6 +20,7 @@ documentation.
 .. [Frean1994] Frean, Marcus R. “The Prisoner's Dilemma without Synchrony.” Proceedings: Biological Sciences, vol. 257, no. 1348, 1994, pp. 75–79. www.jstor.org/stable/50253.
 .. [Hilde2013] Hilbe, C., Nowak, M.A. and Traulsen, A. (2013). Adaptive dynamics of extortion and compliance, PLoS ONE, 8(11), p. e77886. doi: 10.1371/journal.pone.0077886.
 .. [Kraines1989] Kraines, D. & Kraines, V. Theor Decis (1989) 26: 47.  doi:10.1007/BF00134056
+.. [LessWrong2011] Zoo of Strategies (2011) LessWrong. Available at: http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/
```
drvinceknight: All on one line please. I think this doesn't play properly with sphinx otherwise.

souravsingh: I think I have written in one line. Unless I am misunderstanding something.

drvinceknight: You have indeed. I'll review properly when I get a moment.

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -563,3 +563,52 @@ def test_strategy(self):
...
+        actions = [(C, C), (C, D), (D, C), (C, D), (D, C)]
+        self.versus_test(axelrod.Alternator(), expected_actions=actions)
+
+        actions = [(C, C), (C, C), (C, C), (C, C), (D, C)]
+        self.versus_test(axelrod.Cooperator(), expected_actions=actions)
+
+        actions = [(C, D), (D, D), (D, D), (D, D), (D, D)]
+        self.versus_test(axelrod.Defector(), expected_actions=actions)
+
+        actions = [(C, C), (C, D), (D, C), (C, D), (D, C)]
```
drvinceknight: Could you add one more round to this `actions`, the point of having `match_attributes={"length":-1}` is to show how the strategy acts when playing without knowledge of the when the last round is so the last actions would be:

```python
actions = [(C, C), (C, D), (D, C), (C, D), (D, C), (C, D)]
```

Then can you repeat the exact same test but without `match_attributes` which should then have:

```python
actions = [(C, C), (C, D), (D, C), (C, D), (D, C), (D D)]
```

Which shows that the "last round" instruction is working. (So just move your other alternator test here and add the extra round).

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -563,3 +563,52 @@ def test_strategy(self):
...
+        self.versus_test(axelrod.Alternator(), expected_actions=actions,
+                         match_attributes={"length": -1})
+
+        actions = [(C, D), (D, D), (D, C), (C, C), (D, D)]
```
drvinceknight: I think all the test case will be covered by the cases above, we can simplify this and remove everything below.

drvinceknight:
This looks great to me. Thanks @souravsingh :+1:

PR #999, opened 2017-05-01
Update bibliography.rst

GGOUSSEAUD:
Added source for BM player.

PR #1001, opened 2017-05-01
Bush Mosteller(BM) player added

GGOUSSEAUD:
Added Bush Mosteller test player and player


drvinceknight:
Please add the reference from #999 to the bibliography. My suggestion for closing that PR was so that all the relevant commits stay together (in this PR). :)

PR #1007, opened 2017-05-14
Stein and rapoport

MariosZoulias:
#379
Stein and Rapoport strategy tobe implemented into axelrod_first

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -79,7 +79,7 @@ from .titfortat import (
 from .verybad import VeryBad
 from .worse_and_worse import (WorseAndWorse, KnowledgeableWorseAndWorse,
                               WorseAndWorse2, WorseAndWorse3)
-
+from .axelrod_first import SteinAndRapoport
 # Note: Meta* strategies are handled in .__init__.py
 
 all_strategies = [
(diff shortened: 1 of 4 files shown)


drvinceknight:
> Accidentally i commited again the refactoring of mutual.py test

OK it would be better if we could avoid these extra commits (they're already in there so it's confusing things). Could you either:

1. Go back to the `master` branch and create a new branch from which to make this changes, then push and create a new PR (the easiest way from a `git` point of view).
2. Go back to the `master` branch, pull all the changes from upstream (`git pull upstream`) then go back to this branch and `git rebase master` and then force push: `git push -f origin SteinAndRapoport`. (This is technically the correct thing to do but might create some errors that we could help work you through).

Let me know how you'd prefer to do this.

Note that the automatic tests are currently failing, this is because you also need to add `scipy` to `requirements.txt`.

drvinceknight:
> Note that the automatic tests are currently failing, this is because you also need to add scipy to requirements.txt.

You will also need to add it to: https://github.com/Axelrod-Python/Axelrod/blob/master/rtd_environment.yml#L5 so that it would read:

```
name: py35
dependencies:
- python=3.5.2
- numpy>=1.9.2
- matplotlib>=1.4.2
- scipy>=0.19.0
```

But don't worry about that just yet, let's get the PR situation sorted first :)

MariosZoulias:
Hello 
I face problems (again ) in submiting my changes :/
I have tryied evetyhing. I dont know whats wrong.
Thank yu (if you want send me a private in gitter)

MariosZoulias:
Forget it i think i made it ;)

drvinceknight:
I'm closing this because I believe that's what you want to do. (We can always reopen if not)

PR #1008, opened 2017-05-16
Refactoring mutual.py using versus_test

MariosZoulias:
#379
Stein and Rapoport strategy to be implemented into axelrod_first
Also requirements tying to be ficed


MariosZoulias:
Sorry

PR #1009, opened 2017-05-16
all changes

MariosZoulias:
#379
Stein and Rapoport strategy to be implemented into axelrod_first
Please tell me if its everything ok with these 3 files and if so what i have to do in order to fix the test problem ;)

Thanks a lot

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -16,7 +16,6 @@ from .cycler import (
     AntiCycler, Cycler, CyclerCCD, CyclerCCCD, CyclerCCCCCD,
     CyclerDC, CyclerDDC, CyclerCCCDCD)
 from .darwin import Darwin
-from .dbs import DBS
 from .defector import Defector, TrickyDefector
 from .doubler import Doubler
 from .finite_state_machines import (
@@ -64,7 +63,7 @@ from .punisher import Punisher, InversePunisher, LevelPunisher
 from .qlearner import (
     RiskyQLearner, ArrogantQLearner, HesitantQLearner, CautiousQLearner)
 from .rand import Random
-from .resurrection import Resurrection, DoubleResurrection
+from .resurrection import Resurrection
 from .retaliate import (
     Retaliate, Retaliate2, Retaliate3, LimitedRetaliate, LimitedRetaliate2,
     LimitedRetaliate3)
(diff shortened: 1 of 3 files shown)


axelrod/strategies/_strategies.py
```diff
@@ -16,7 +16,6 @@
     AntiCycler, Cycler, CyclerCCD, CyclerCCCD, CyclerCCCCCD,
     CyclerDC, CyclerDDC, CyclerCCCDCD)
 from .darwin import Darwin
-from .dbs import DBS
```
drvinceknight: You seem to have deleted lines of code that are not part of your PR.

axelrod/strategies/_strategies.py
```diff
@@ -64,7 +63,7 @@
 from .qlearner import (
     RiskyQLearner, ArrogantQLearner, HesitantQLearner, CautiousQLearner)
 from .rand import Random
-from .resurrection import Resurrection, DoubleResurrection
```
drvinceknight: More lines have been deleted.

axelrod/strategies/_strategies.py
```diff
@@ -76,18 +75,17 @@
     TitForTat, TitFor2Tats, TwoTitsForTat, Bully, SneakyTitForTat,
     SuspiciousTitForTat, AntiTitForTat, HardTitForTat, HardTitFor2Tats,
     OmegaTFT, Gradual, ContriteTitForTat, SlowTitForTwoTats, AdaptiveTitForTat,
-    SpitefulTitForTat, SlowTitForTwoTats2, Alexei)
+    SpitefulTitForTat, SlowTitForTwoTats2)
 from .verybad import VeryBad
 from .worse_and_worse import (WorseAndWorse, KnowledgeableWorseAndWorse,
                               WorseAndWorse2, WorseAndWorse3)
-
+from .axelrod_first import SteinAndRapoport
```
drvinceknight: This can just be put in the earlier line where we have multiple strategies with `from .axelrod_first ...`

axelrod/strategies/_strategies.py
```diff
@@ -76,18 +75,17 @@
...
 # Note: Meta* strategies are handled in .__init__.py
 
 all_strategies = [
     Adaptive,
     AdaptiveTitForTat,
     Aggravater,
-    Alexei,
```
drvinceknight: More things being deleted.

axelrod/strategies/axelrod_first.py
```diff
@@ -487,3 +490,76 @@ class UnnamedStrategy(Player):
...
+                return D
+            else :
+                return C
+
+        if round_number >= 16 :
+            # >=16 before for the first 15 round he plays for sure titfortat
+            times15_in_list = (round_number // 15) * 15
+            # lets suppose we are in 28th round
+            # So we still play according to chi-squared test in the 15th round
+            all_c = 0 # Each time we consider the whole history
```
drvinceknight: You don't need this, you can just use `opponent.cooperations` and `opponent.defections`, so your test would just be `chisquare([opponent.cooperations, opponent.defections])`, you only need to do that every fifteen moves though so something like:


```python
if len(self.history) % 15 == 0:
    chisquare(...)
```

drvinceknight: Let's get the PR and the commits aligned first before we worry about this though.

PR #1023, opened 2017-05-23
Refactor tests for Prober.

drvinceknight:
Addresses #884

Also reformats strategy docstrings (which addresses #225).

### DIFF
--- a/axelrod/tests/strategies/test_prober.py
+++ b/axelrod/tests/strategies/test_prober.py
@@ -21,15 +21,20 @@ class TestCollectiveStrategy(TestPlayer):
     }
 
     def test_strategy(self):
-        # Starts by playing CD.
-        self.responses_test([C, D])
-        # Defects forever unless opponent matched first two moves
-        self.responses_test([D] * 10, [C, D], [C, C])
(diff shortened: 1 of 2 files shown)


axelrod/tests/strategies/test_prober.py
```diff
@@ -232,24 +252,33 @@ class TestNaiveProber(TestPlayer):
     }
 
     def test_strategy(self):
-        # Randomly defects and always retaliates like tit for tat.
         self.first_play_test(C)
         # Always retaliate a defection
-        self.responses_test([D], [C] * 2, [C, D])
+        actions = [(C, D)] + [(D, D)] * 10
```
marcharper: Is this one equivalent? The original test had a cooperation then a defection by the opponent. If we want to replace the test then one with alternator seems more appropriate.

drvinceknight: Against an Alternator this test becomes stochastic because the strategy randomly defects against a cooperation. The test we're replacing just checked the first defection and not a set of future rounds. I'll change this to use a `MockPlayer` that starts with `C` and then `D` for the remaining rounds.

PR #1027, opened 2017-05-28
Fake name se dynamic tit for tat

FakeNameSE:
I added a new strategy to Tit for Tat which is based on two tits for tat but outperforms most variants of the strategy by dynamically matching its probability of forgiveness to to ratio of defects to cooporations of the opponent. Unfortunately, I could not manage to write a correct test for it, so if that is an obstacle please just let me know and I will give it another shot.

Have a nice day!

### DIFF
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -94,10 +94,10 @@ class TwoTitsForTat(Player):
     
 class DynamicTitForTat(Player):
     """A player starts by cooperating and then mimics previous move by
-     opponent with a dynamic bias based off of the opponents ratio of
-      defections to cooperations towards cooporating regardless of 
-      the move."""
-
+    opponent with a dynamic bias based off of the opponents ratio of
+    defections to cooperations towards cooporating regardless of 
+    the move."""
+    
     name = 'Dynamic Tit for Tat'
     classifier = {
         'memory_depth': 2,  # Long memory, memory-2
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -91,7 +91,36 @@ class TwoTitsForTat(Player):
     @staticmethod
     def strategy(opponent: Player) -> Action:
         return D if D in opponent.history[-2:] else C
+    
+class DynamicTitForTat(Player):
+    """A player starts by cooperating and then mimics previous move by
+     opponent with a dynamic bias based off of the opponents ratio of
+      defections to cooperations towards cooporating regardless of 
+      the move."""
 
+    name = 'Dynamic Tit for Tat'
+    classifier = {
+        'memory_depth': 2,  # Long memory, memory-2
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    @staticmethod
+    def strategy(self, opponent):
+        try:
+            if 'D' in opponent.history[-2:]:
+                if random.random() < (sum([s == 'C' for s in opponent.history]) / len(opponent.history)):
+                    return 'C'
+                else:
+                    return 'D'
+            else:
+                return 'C'
+        except IndexError:
+            return 'C'    
 
 class Bully(Player):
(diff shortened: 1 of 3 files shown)


axelrod/strategies/titfortat.py
```diff
@@ -91,7 +91,36 @@ class TwoTitsForTat(Player):
...
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
 
+    @staticmethod
+    def strategy(self, opponent):
```
drvinceknight: As you have set this as a `staticmethod` it cannot have the `self`. This needs to be:

```python
def strategy(opponent)
```

(This is what is currently tripping a number of tests)

axelrod/strategies/titfortat.py
```diff
@@ -91,7 +91,36 @@ class TwoTitsForTat(Player):
...
+        try:
```
drvinceknight: Instead of using `try` `except` clause we usually just check the len of `opponent.history` so:

```python
if len(opponent.history) == 0:
    return C
...
```

However in this case, as you're using `opponent.history[-2:]`  I believe you don't need a check anyway:

```python
>>> history = []
>>> history[-2:]
[]
>>> history = ["C"]
>>> history[-2:]
["C"]
```

So you don't need this `try` `except` clause at all (I believe).

axelrod/strategies/titfortat.py
```diff
@@ -91,7 +91,36 @@ class TwoTitsForTat(Player):
...
+            if 'D' in opponent.history[-2:]:
```
drvinceknight: Throughout, instead of using `'C'` and `'D'`, use `C` and `D` which have been imported at the top of this module.

axelrod/strategies/titfortat.py
```diff
@@ -91,7 +91,36 @@ class TwoTitsForTat(Player):
...
+                if random.random() < (sum([s == 'C' for s in opponent.history]) / len(opponent.history)):
```
drvinceknight: `opponent.cooperations` should be used as a more efficient count of the cooperation in it's history.

drvinceknight: Instead of `random.random() <  ...` we have a random place function that is preferable to use, this line can then be:

```python
return random_choice(opponent.cooperations / len(opponent.history))
``` 

You'll need to add the following to the top of this module:

```python
```

axelrod/strategies/titfortat.py
```diff
@@ -91,7 +91,36 @@ class TwoTitsForTat(Player):
     @staticmethod
     def strategy(opponent: Player) -> Action:
         return D if D in opponent.history[-2:] else C
+    
+class DynamicTitForTat(Player):
+    """A player starts by cooperating and then mimics previous move by
```
drvinceknight: If you have found this strategy elsewhere we'll need to reference the source, if not we still need to say it's an original strategy. There's information about what we need in the docstring here: http://axelrod.readthedocs.io/en/stable/tutorials/contributing/strategy/docstrings.html

On that note, I'm not entirely sure this is a good name for the strategy, if it's called this somewhere else then we will stick with it though.

drvinceknight:
@FakeNameSE I see you closed this PR. If you're planning on addressing my comments please reopen it and push your new commits to the same branch. 

That way we can keep track of my comments (and others that might be made) as well as your fixes. Let us know if you need a hand with any of this.👍

FakeNameSE:
Sorry about that. I couldn't seem to edit that branch so I pulled the
request. I just put together the strategy and test and they passed the
tests, so I will put this back up later today. Also, the strategy is
original (at least in the sense that I did not see it anywhere before), so
now is your chance to give it a better name. I originally called it Joshua
in homage to wargames, but changed it to something more descriptive.

drvinceknight:
Ok, when you open the new PR include a link to this one. Leave the name for now. We can think about changing it on the new PR.

Hopefully whatever went wrong with this PR will work this time and you'll be able to push future commits to that branch as we discuss things.

FakeNameSE:
Got it. Thank you.

PR #1031, opened 2017-05-28
Refactor memory two test.

drvinceknight:
Addresses #884.

Some of the test cases were not reproducible (they relied on impossible histories) but what I've got there does test everything (I believe!) and I'm also now carefully checking the attributes.

### DIFF
--- a/axelrod/tests/strategies/test_memorytwo.py
+++ b/axelrod/tests/strategies/test_memorytwo.py
@@ -22,22 +22,27 @@ class TestMEM2(TestPlayer):
     }
 
     def test_strategy(self):
+        self.first_play_test(C)
         # Start with TFT
-        self.responses_test([C])
-        self.responses_test([C], [C], [C])
-        self.responses_test([D], [C], [D])
-        # TFT if mutual cooperation on first two rounds
-        self.responses_test([C], [C, C], [C, C])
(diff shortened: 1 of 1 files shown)


axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -22,22 +22,27 @@ class TestMEM2(TestPlayer):
...
+        actions = [(C, C), (C, C)]
+        self.versus_test(opponent=axelrod.Cooperator(),
+                         expected_actions=actions,
+                         attrs={"play_as": "TFT", "shift_counter": 1,
+                                "alld_counter": 0})
+        actions = [(C, D), (D, D)]
+        self.versus_test(opponent=axelrod.Defector(), expected_actions=actions,
+                         attrs={"play_as": "TFT", "shift_counter": 1,
+                                "alld_counter": 0})
         # TFTT if C, D and D, C
```
marcharper: Are we still testing both possible TFTT cases?

drvinceknight: I'll double check 👍🙂

drvinceknight: I've just checked by removing each case and the test fails so this test if hitting both of them. :+1:

PR #1032, opened 2017-05-29
Refactor tests for memory one

drvinceknight:
Addresses #884 

I also added some references to strategies.

### DIFF
--- a/docs/reference/bibliography.rst
+++ b/docs/reference/bibliography.rst
@@ -34,7 +34,7 @@ documentation.
   for the Iterated Prisoner's Dilemma. Proceedings of the 2015
   International Conference on Autonomous Agents and Multiagent Systems.
 .. [Nachbar1992] Nachbar J., Evolution in the finitely repeated prisoner’s dilemma, Journal of Economic Behavior & Organization, 19(3): 307-326, 1992.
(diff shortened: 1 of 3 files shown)


docs/reference/bibliography.rst
```diff
@@ -33,7 +34,8 @@ documentation.
   for the Iterated Prisoner's Dilemma. Proceedings of the 2015
   International Conference on Autonomous Agents and Multiagent Systems.
 .. [Nachbar1992] Nachbar J., Evolution in the finitely repeated prisoner’s dilemma, Journal of Economic Behavior & Organization, 19(3): 307-326, 1992.
-.. [Nowak1992] Nowak, M. a., & May, R. M. (1992). Evolutionary games and spatial chaos. Nature. http://doi.org/10.1038/359826a0
+.. [Nowak1989] Nowak, M., & Sigmund, K. (1989). Natural selection of memory-one strategies for the iterated prisoner's dilemma. Journal of Theoretical biology. http://www.sciencedirect.com/science/article/pii/0096300389900520
```
Nikoleta-v3: Thought Nowak was from the first to look at memory one strategies (I believe), the 1989  article is focused only on the reactive strategies. 

I would recommend the following article as a reference: https://www.researchgate.net/publication/225248231_Nowak_M_Sigmund_K_The_evolution_of_stochastic_strategies_in_the_Prisoner%27s_Dilemma_Acta_Appl_Math_20_247-265

It can be seen in the appendix that they give full definition of memory one strategies and the transition 
matrix in a match.

drvinceknight: Thanks @Nikoleta-v3! Fixing now.

PR #1033, opened 2017-05-29
Refactor tests for hunter.

drvinceknight:
Addresses #884

### DIFF
--- a/axelrod/tests/strategies/test_player.py
+++ b/axelrod/tests/strategies/test_player.py
@@ -337,7 +337,7 @@ class TestPlayerClass(unittest.TestCase):
 
 
 def test_responses(test_class, player1, player2, responses, history1=None,
-                   history2=None, seed=None, attrs=None):
+                   history2=None, seed=None):
     """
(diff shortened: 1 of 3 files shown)


drvinceknight:
Just noticed that this had brought coverage down, because a particular part of `responses_test` was no longer being used. I've removed that 0c34fbc

axelrod/tests/strategies/test_player.py
```diff
@@ -359,9 +359,6 @@ def test_responses(test_class, player1, player2, responses, history1=None,
     for response in responses:
         s1, s2 = simulate_play(player1, player2)
         test_class.assertEqual(s1, response)
-    if attrs:
```
marcharper: Why do we not need `attrs` anymore? Have all such cases been moved to `versus_test`?

drvinceknight: (The coverage on this PR dropped because that line wasn't being hit by anything, which is why I just removed it.)

PR #1034, opened 2017-05-29
Refactor tests for player.

drvinceknight:
This is a minor refactor of some warnings that were occurring:

- A `simulate_play` that was using a fake history;
- A warning from the generator equality

The change to the way the generator equality was being run is based on
this error:

```
.........../home/vince/src/Axelrod/axelrod/player.py:165:
PendingDeprecationWarning: generator 'Player.__eq__.<locals>.<genexpr>'
raised StopIteration
  for _ in range(200))):
```

Addresses #884

### DIFF
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -163,7 +163,7 @@ class Player(object):
 
                 for _ in range(200):
                     try:
-                        if not next(generator) == next(other_generator):
+                        if next(generator) != next(other_generator):
                             return False
                     except StopIteration:
                         break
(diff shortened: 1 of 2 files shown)


axelrod/player.py
```diff
@@ -161,9 +161,12 @@ def __eq__(self, other):
                     setattr(other, attribute,
                             itertools.cycle(original_other_value))
 
-                if not (all(next(generator) == next(other_generator)
-                        for _ in range(200))):
-                    return False
+                for _ in range(200):
+                    try:
+                        if not next(generator) == next(other_generator):
```
drvinceknight: Good call on `!=`.

I believe the error was thrown by an attribute that was not a cycle and ran out of elements. I didn't check through for which one that was though (this change seemed to make sense anyway to me).

Issue #678, opened 2016-08-02
Tournament Transformers?


marcharper:
Implement `@noise(noise)` and `@spatial(graph)` decorators for tournament classes.

drvinceknight:
Awesome idea. 👍

marcharper:
Perhaps -- I can still think of some ideas, like applying a transformer to all strategies, or further tournament variations. But we can close if you'd like and reopen later if needed.

drvinceknight:
:+1:

marcharper:
It might be worth thinking about decorators for `Match`.

PR #1042, opened 2017-06-09
Reconcile tournaments

drvinceknight:
**This would introduce backwards incompatibility at the Tournament call level. This PR would be for v3.0.0**

---

**Status quo:**

- To create a standard tournament:

   ```python
   >>> std = axl.Tournament(players=players, turns=200)
   ```

- To create a probabilistic ending tournament:

   ```python
   >>> prob_end = axl.ProbEndTournament(players=players, prob_end=.2)
   ```

- To create a standard spatial tournament:

   ```python
   >>> spatial = axl.SpatialTournament(players=players, turns=200, edges=edges)
   ```

- To create a probabilistic ending tournament:

   ```python
   >>> prob_end_spatial = axl.ProbEndSpatialTournament(players=players, prob_end=.2, edges=edges)
   ```

---
This PR changes that and removes the `ProbEndTournament`, `SpatialTournament`, `ProbEndSpatialTournament` classes so to create the above we would have:

```python
>>> std = axl.Tournament(players=players, turns=200)
>>> prob_end = axl.Tournament(players=players, prob_end=.2)
>>> spatial = axl.Tournament(players=players, turns=200, edges=edges)
>>> prob_end_spatial = axl.Tournament(players=players, prob_end=.2, edges=edges)
```

I've done this by carrying out various things:

- c695781 Pass the responsibility of sampling a match length in a prob end Match from the `MatchGenerator` to the `Match` class. The main consideration here is that I've considered `turns` and `prob_end` as stopping conditions for a match:
   - If one of them is `None` the other is used. (If `turns` is `None` and `prob_end` is not `None` then we have a probabilistic ending match)
   - If both are `None` then a default of 20 turns is used. (Currently no default length is in place, I've picked 20 out of a hat, could use 200 to be in line with tournaments?)
   - If both are not `None` then both are used (so in essence we have a finite match of max length `turns` but with a probability of finishing earlier).
- 99e8ed9 Refactor the `match_generator` module: it had 4 different generators (and 1 parent class), they have all been put in the `MatchGenerator` class which passes the match parameters to the `Match` class. The one thing this does is create an `edges` generator if no `edges` are passed to it. In this case the `edges` are the edges of the complete graph (so everyone plays everyone else).
- 0e859a7 Refactor the rest of the code: removing all the other tournament classes.

One final change is a potential fix of a bug. Currently, in a prob end tournament, the match length between any two strategies is randomly sampled and that sampled length is repeated. In cc4d537 I change this so that a new Match is created at each repetition which in turn means a new length is sampled. I've done this as a separate commit so that it's isolated but I feel that this should be the way it is (I've written a test for it too).

---

**Advantages**

- This cleans up the code quite a lot (it's basically moving a couple of things and then a bunch of deletions)
- This simplifies the creation of tournaments (users don't need to figure out which tournament class to use)
- I think this might actually remove the need for https://github.com/Axelrod-Python/Axelrod/issues/678 (tournament transformers?)
- EDIT: Also make things simpler for https://github.com/Axelrod-Python/axelrod-api? (No need to consider separate "interfaces" for different types of tournaments. Just take the input parameters and it'll know what to do.)

**Disadvantages**

- Backwards incompatibility, (basically anyone using `ProbEndSpatialTournament` etc no longer can). So this would be for `v3.0.0`.

Let me know what you think :)


meatballs:
I like this a lot! I'll have a proper look through the code....

axelrod/match.py
```diff
@@ -35,8 +40,17 @@ def __init__(self, players, turns, game=None, deterministic_cache=None,
...
         """
-        self.result = []
+
+        if prob_end is not None:
+            if turns is not None:
+                turns = min(turns, sample_length(prob_end))
+            else:
+                turns = sample_length(prob_end)
+        elif turns is None:
+            turns = 20
```
meatballs: I know it's my usual reaction to conditional statement but, especially with nesting, I'd be tempted to use a dict here. Something like:

```
turns_mapping = {
    (True, True): 20,
    (True: False): sample_length(prob_end),
    (False, True): turns,
    (False, False): min(turns, sample_length(prob_end))
}

turns = turns_mapping[(turns is None, prob_end is None)]
```

axelrod/tournament.py
```diff
@@ -11,21 +11,20 @@
...
 
 
 class Tournament(object):
 
     def __init__(self, players: List[Player],
-                 match_generator: MatchGenerator = RoundRobinMatches,
-                 name: str = 'axelrod', game: Game = None, turns: int = 200,
-                 repetitions: int = 10,
-                 noise: float = 0, with_morality: bool = True) -> None:
+                 match_generator: MatchGenerator = MatchGenerator,
```
marcharper: Is match_generator still needed as a parameter?

Would it be better for Tournament to take a dictionary of match parameters?

axelrod/tournament.py
```diff
@@ -51,13 +50,24 @@ def __init__(self, players: List[Player],
...
         self.noise = noise
         self.num_interactions = 0
         self.players = players
         self.repetitions = repetitions
-        self.match_generator = match_generator(
-            players, turns, self.game, self.repetitions, self.noise)
+        self.edges = edges
+
+        if turns is None and prob_end is None:
+            turns = 200
```
marcharper: Maybe we should have a constant DEFAULT_TURNS somewhere

axelrod/match.py
drvinceknight: I've just realised I need to move this logic anyway to the `play` method (for the cache to work as we'd want), I'll see if I see a way of getting a dict to work there (although it'd need to be different to this so that it avoids unnecessary random samples). :+1:

axelrod/tournament.py
drvinceknight: > Is match_generator still needed as a parameter?

I'd be happy to remove it. The only reason we really have it is to allow for the `making new tournaments` idea: http://axelrod.readthedocs.io/en/stable/tutorials/advanced/making_tournaments.html

But that's such a niche thing I'd be happy to take it out.

> 
> Would it be better for Tournament to take a dictionary of match parameters?

I don't think so. It's essentially a dictionary anyway (just `kwargs`) and as is is more helpful to the user (autocompletion etc...).

axelrod/tournament.py
drvinceknight: Good idea.

drvinceknight:
I've pushed a few more changes to this:

- 4ce4098 adds `axelrod.default_turns` and then uses that throughout as a default (including the Moran process - which changed some tests). I didn't go with `axelrod.DEFAULTTURNS` but only because I never really know when to use the "shouty" variables (happy to change it).
- 4ce4098 removes the `match_generator` as an argument in `Tournament` and also removes the corresponding tutorial.
- ae2c878 changes how the Match decides how long it's going to play for. This is now done in the `play` method. So every Match has a `turns` and a `prob_end` attribute which are never `None` (but for example can be `turns=float('inf')` and `prob_end=.2`) and that's just used in the `play` method. 
- As a result of that, d61c957  changes the Tournament back to using the `Match` class properly (by **not** creating new Match instances for each repetition).

The other commits are minor things (@meatballs in 1c162eb I moved over to use a dictionary for the input selection instead of nested `if` statements :+1:).

meatballs:
> I never really know when to use the "shouty" variables

When they intended to be used as a constant (which python doesn't have) rather than an ordinary variable. This would be a good example of using shoutiness correctly!

meatballs:
> I moved over to use a dictionary for the input selection instead of nested if statements

You've no idea how happy that makes me!

drvinceknight:
> This would be a good example of using shoutiness correctly!

Ok cool. I'll change it!

meatballs:
Nice piece of work!

marcharper:
LGTM but I do have one suggestion. For the tournament class `with_morality` seems out of place -- I get that it's a pass through to results set but I think it would be better as an invokable method on the results set rather than something that `Tournament` should be concerned with.

drvinceknight:
> I get that it's a pass through to results set but I think it would be better as an invokable method on the results set rather than something that Tournament should be concerned with. 

This wouldn't really work as all the calculations for the result set are done as they're read in. Invoking it on the result set would imply an entire reread of all the interactions. I agree that it's out of place. One option would be to make it no longer default and just have it as something that's always being calculated. I think having it as a optional is legacy.

drvinceknight:
> One option would be to make it no longer default and just have it as something that's always being calculated.

Another (perhaps better) option is not to make it part of the `Tournament.__init__` but an option that can be passed to `Tournament.play`.

marcharper:
I'm hitting the button, let's table the morality discussion

PR #1043, opened 2017-06-10
Reconcile Moran processes

drvinceknight:
Use spatial Moran Processes throughout. 

Also add ability to pass a prob_end to the Matches.

In terms of performance this actually seems faster:

```
%%timeit
# Branch implementation of Moran Process
for players in [[axl.Cooperator()] * 3 + [axl.Defector()] * 3,
                [axl.Random(), axl.Grudger(), axl.TitFor2Tats()],
                [s() for s in axl.demo_strategies]]:

    for seed in range(100):
        axl.seed(seed)
        mp = axl.MoranProcess(players)
        mp.play()
        winner = mp.winning_strategy_name
```

output: `1 loop, best of 3: 8.7 s per loop`

whereas current `master`:

```
%%timeit
# Master branch implementation of Moran Process
for players in [[axl.Cooperator()] * 3 + [axl.Defector()] * 3,
                [axl.Random(), axl.Grudger(), axl.TitFor2Tats()],
                [s() for s in axl.demo_strategies]]:

    for seed in range(100):
        axl.seed(seed)
        mp = axl.MoranProcess(players)
        mp.play()
        winner = mp.winning_strategy_name
```
Output: `1 loop, best of 3: 9.92 s per loop`

This complements #1042.

### DIFF
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -136,8 +136,8 @@ class MoranProcess(object):
         self.interaction_graph = interaction_graph
         self.reproduction_graph = reproduction_graph
         # Map players to graph vertices
-        self.locations = list(interaction_graph.vertices())
-        self.index = dict(zip(interaction_graph.vertices(),
+        self.locations = sorted(interaction_graph.vertices())
+        self.index = dict(zip(sorted(interaction_graph.vertices()),
                               range(len(players))))
 
     def set_players(self) -> None:
@@ -190,7 +190,7 @@ class MoranProcess(object):
             # Select locally
             # index is not None in this case
             vertex = random.choice(
-                self.reproduction_graph.out_vertices(self.locations[index]))
+                sorted(self.reproduction_graph.out_vertices(self.locations[index])))
             i = self.index[vertex]
         return i
 
@@ -290,12 +290,12 @@ class MoranProcess(object):
         if self.mode == "db":
             source = self.index[self.dead]
             self.dead = None
-            sources = self.interaction_graph.out_vertices(source)
+            sources = sorted(self.interaction_graph.out_vertices(source))
         else:
             # birth-death is global
-            sources = self.locations
+            sources = sorted(self.locations)
         for i, source in enumerate(sources):
-            for target in self.interaction_graph.out_vertices(source):
+            for target in sorted(self.interaction_graph.out_vertices(source)):
                 j = self.index[target]
                 if (self.players[i] is None) or (self.players[j] is None):
                     continue
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -4,6 +4,7 @@ import random
 import numpy as np
 
 from axelrod import DEFAULT_TURNS
+from .graph import complete_graph
 from .deterministic_cache import DeterministicCache
 from .match import Match
 from .random_ import randrange
@@ -31,9 +32,10 @@ def fitness_proportionate_selection(scores):
 
 
 class MoranProcess(object):
-    def __init__(self, players, turns=DEFAULT_TURNS, noise=0,
+    def __init__(self, players, turns=DEFAULT_TURNS, prob_end=None, noise=0,
                  deterministic_cache=None, mutation_rate=0., mode='bd',
-                 match_class=Match):
+                 match_class=Match, interaction_graph=None,
+                 reproduction_graph=None):
         """
         An agent based Moran process class. In each round, each player plays a
         Match with each other player. Players are assigned a fitness score by
@@ -51,11 +53,21 @@ class MoranProcess(object):
         population. This is not the only method yet emulates the common method
         in the literature.
 
+		It is possible to pass interaction graphs and reproduction graphs to the
+        Moran process. In this case, in each round, each player plays a
+        Match with each neighboring player according to the interaction graph.
+        Players are assigned a fitness score by their total score from all
+        matches in the round. A player is chosen to reproduce proportionally to
+        fitness, possibly mutated, and is cloned. The clone replaces a randomly
+        chosen neighboring player according to the reproduction graph.
+
         Parameters
         ----------
         players, iterable of axelrod.Player subclasses
         turns: int, 100
             The number of turns in each pairwise interaction
+        prob_end : float
+            The probability of a given turn ending a match
         noise: float, 0
             The background noise, if any. Randomly flips plays with probability
             `noise`.
@@ -68,9 +80,15 @@ class MoranProcess(object):
             Birth-Death (bd) or Death-Birth (db)
         match_class: subclass of Match
             The match type to use for scoring
+        interaction_graph: Axelrod.graph.Graph
+            The graph in which the replicators are arranged
+        reproduction_graph: Axelrod.graph.Graph
+            The reproduction graph, set equal to the interaction graph if not
+            given
         """
         self.match_class = match_class
         self.turns = turns
+        self.prob_end = prob_end
         self.noise = noise
         self.initial_players = players  # save initial population
         self.players = []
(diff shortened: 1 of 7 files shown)


docs/tutorials/further_topics/moran_processes_on_graphs.rst
```diff
@@ -4,7 +4,7 @@ Moran Process on Graphs
 =======================
 
 The library also provides a graph-based Moran process [Shakarian2013]_ with
-:code:`MoranProcessGraph`.  To use this class you must supply at least one
+:code:`MoranProcess`.  To use this class you must supply at least one
```
marcharper: class --> feature since the class can be used without a graph.

axelrod/moran.py
```diff
@@ -31,9 +32,10 @@ def fitness_proportionate_selection(scores):
 
 
 class MoranProcess(object):
-    def __init__(self, players, turns=DEFAULT_TURNS, noise=0,
+    def __init__(self, players, turns=DEFAULT_TURNS, prob_end=None, noise=0,
                  deterministic_cache=None, mutation_rate=0., mode='bd',
-                 match_class=Match):
+                 match_class=Match, interaction_graph=None,
```
marcharper: Do we need match_class any longer?

axelrod/moran.py
```diff
@@ -101,6 +119,23 @@ def __init__(self, players, turns=DEFAULT_TURNS, noise=0,
             mutation_targets[key] = [v for (k, v) in sorted(d.items()) if k != key]
         self.mutation_targets = mutation_targets
 
+        if interaction_graph is None:
+            interaction_graph = complete_graph(len(players), loops=False)
+            if reproduction_graph is None:
+                reproduction_graph = complete_graph(len(players))
+        if reproduction_graph is None:
```
marcharper: Is this the right default? I think we should consider using the same default in all cases.

axelrod/moran.py
```diff
@@ -124,15 +159,21 @@ def mutate(self, index):
         return new_player
 
     def death(self, index=None):
-        """Selects the player to be removed. Note that the in the birth-death
```
marcharper: This docstring is still relevant, no?

marcharper:
One concern is that without the non-graph implementation we don't have a canonical implementation to compare to. We should consider adding some tests that really demonstrate that it is equivalent (e.g. explicitly tests the interaction pairs for various population sizes).

axelrod/moran.py
drvinceknight: I don't think so, I'll take it out.

axelrod/moran.py
drvinceknight: This is the same default as it was implemented in `MoranProcessGraph` but I'll change it. I've got a further idea for how I'd like to refactor this as well.

axelrod/moran.py
drvinceknight: Yup, not sure why I got rid of it.

drvinceknight:
> We should consider adding some tests that really demonstrate that it is equivalent (e.g. explicitly tests the interaction pairs for various population sizes).

Sure. I'm going to add to the tests a bit anyway, I'll try to incorporate your suggestion but I might need more guidance. :+1:

drvinceknight:
I've just pushed a bunch of changes. Amongst other things I've added numpy docstrings and type hints. I've also gone through and added more unit type tests. I took a look at adding specific interaction tests but I'm not sure it's necessary:

- Each method is now tested (it wasn't before);
- There are some tests that **do** test the expected overall outcomes.

I'm not at all averse to adding more tests if you're happy to spell them out for me (or perhaps they can be left as a further PR). :+1:

axelrod/moran.py
```diff
@@ -389,14 +395,15 @@ class ApproximateMoranProcess(MoranProcess):
     Instead of playing the matches, the result is sampled
     from a dictionary of player tuples to distribution of match outcomes
     """
-    def __init__(self, players, cached_outcomes, mutation_rate=0.):
+    def __init__(self, players: List[Player], cached_outcomes: dict,
```
marcharper: We need all the parameters to pass through?

axelrod/tests/unit/test_moran.py
```diff
@@ -151,30 +261,15 @@ def test_three_players_with_mutation(self):
...
-        random.seed(10)
+        axelrod.seed(29)
         mp = MoranProcess(players)
         populations = mp.play()
         self.assertEqual(len(mp), 9)
         self.assertEqual(len(populations), 9)
         self.assertEqual(populations, mp.populations)
         self.assertEqual(mp.winning_strategy_name, str(axelrod.Defector()))
 
-    def test_standard_fixation(self):
```
marcharper: Why remove this test?

docs/tutorials/further_topics/moran_processes_on_graphs.rst
```diff
@@ -17,8 +17,9 @@ The nodes can be any hashable object (integers, strings, etc.). For example::
     >>> edges = [(0, 1), (1, 2), (2, 3), (3, 1)]
     >>> graph = Graph(edges)
 
-Graphs are undirected by default. Various intermediates such as the list of
-neighbors are cached for efficiency by the graph object.
+Graphs are undirected by default but you can pass :code:`directed=True` to
+create a directed graph.. Various intermediates such as the list of neighbors
```
marcharper: ..

axelrod/moran.py
drvinceknight: Are you asking about the `turns` `prob_end` etc? In which case: no we don't. They are not used as all match results sampled from the `cached_outcomes`.

If you're asking about the `graphs`. I didn't want to tackle that here. It would be an easy thing to add.

axelrod/tests/unit/test_moran.py
drvinceknight: Because it was for a mock match class (it was a test for the `match_class` attribute).

axelrod/moran.py
drvinceknight: > It would be an easy thing to add.

But let's do that at a later date.

drvinceknight:
Failures are py3.6 vs py3.5... Working on basically ensuring the dicts are ordered as expected.

axelrod/tests/unit/test_moran.py
marcharper: Actually it tested the constant fitness case (or else, that was the original intent).

drvinceknight: Most of the tests in here were testing multiple things (which I've tried to address). This test also tested that you could pass a different match class and has been removed because I've removed the `match_class` attribute (as you suggested earlier).

drvinceknight: If you wanted to test constant fitness we can do that by passing a cache where all strategies have the same score.

drvinceknight: Actually, I don't understand what that test was doing at all with the Mock class (apart from testing that it worked). I'm fixed the py3.6 problems and about to push with a version of this test (just using a normal match).

drvinceknight:
- 5c26eee gets tests to pass on both py3.6 and py3.5
- b3960ca adds a constant fitness test case

axelrod/tests/unit/test_moran.py
marcharper: It was returning a constant score (fitness) in a 2:1 ratio to see if the outcome was in the ballpark of what would be expected for an r=2 Moran process. This appears to be what the new test essentially does.

drvinceknight: Cool :) Let me know if you think there's anything else that we can add.

PR #1045, opened 2017-06-18
PEP8 / style fixes and some test updates

marcharper:


axelrod/strategies/axelrod_first.py
```diff
@@ -522,15 +523,15 @@ def __init__(self, alpha: float=0.05) -> None:
...
-        alpha, float
-            The significant level of pvalue from chi-squared test
-            0.05 by default according to literature
+        alpha: float
+            The significant level of p-value from chi-squared test
+            005 by default according to literature
         """
         super().__init__()
         self.alpha = alpha
-        self.opponent_is_random = None
```
drvinceknight: This is tripping the test. I suggest changing the test to `assertFalse`. (You did this on line 389 but not on line 385).

axelrod/fingerprint.py
```diff
@@ -327,14 +327,14 @@ def fingerprint(
...
-    def plot(self, col_map: str = 'seismic', interpolation: str = 'none',
+    def plot(self, cmap: str = 'seismic', interpolation: str = 'none',
              title: str = None, colorbar: bool = True,
              labels: bool = True) -> plt.Figure:
         """Plot the results of the spatial tournament.
 
         Parameters
         ----------
-        col_map : str, optional
+        cmap : str, optional
```
drvinceknight: Based on the fact that `mpl.imshow` uses `cmap` right (https://matplotlib.org/devdocs/api/_as_gen/matplotlib.axes.Axes.imshow.html)?

I'm happy with this (should have gone with this from the start really), just checking.

drvinceknight: (Theoretically this adds a backwards incompatibility, I'm happy with the change and I'd rather not bump the next release to `4.0.0` which I think in this instance is ok?)

axelrod/tests/strategies/test_player.py
```diff
@@ -124,6 +121,20 @@ def test_noisy_play(self):
         self.assertEqual(player1.history[0], D)
         self.assertEqual(player2.history[0], D)
 
+    def test_update_history(self):
```
drvinceknight: Why do we need this? (Not suggesting we don't, just curious what we were/are missing.)

marcharper: It was in a different file.

axelrod/strategies/axelrod_first.py
marcharper: By the way, this was incorrect but the `Player__eq__` method was not catching it because it assumes that attributes that it doesn't find are `None`. The reset method was resetting ` self.random_opponent` instead of `opponent_is_random`.

axelrod/tests/strategies/test_player.py
drvinceknight: Cool. :+1:

axelrod/strategies/axelrod_first.py
drvinceknight: Great catch.

drvinceknight:
Looks good to me assuming CI is happy.

drvinceknight:
This was merged with CI still failing. I'm going to push a fix straight to master.

PR #1052, opened 2017-06-27
Enum actions #816

eric-s-s:
There is still some work to do, but this should pass all tests.  (i hope. i'll have to wait for TravisCI. i'm on windows).  I tried to keep changes to an absolute minimum.  I needed to add a test in `test_resultset.py` (it has a comment why it's there) and I wanted to check inputs on Human, so I added 4 tests in `test_human.py` .

work not done yet - 

- flip_action can be replaced by the action's `.flip()` method, but basic tests first.
- it was mentioned moving `random_choice` to an Action method. I have not touched that yet.
- the alias `Action = Actions` can now be removed. I didn't want to do it yet, because it affects the import statement of almost every module and there's more than enough code to look through already.
- if this looks good, should the enum be "Action" or "Actions"?

### DIFF
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -13,9 +13,9 @@ from enum import Enum
 from typing import Iterable
 
 
-class UnknownAction(ValueError):
+class UnknownActionError(ValueError):
     def __init__(self, *args):
-        super(UnknownAction, self).__init__(*args)
+        super(UnknownActionError, self).__init__(*args)
 
 
 class Actions(Enum):
@@ -49,7 +49,7 @@ class Actions(Enum):
         elif character == 'D':
             return cls.D
         else:
-            raise UnknownAction('Character must be "C" or "D".')
+            raise UnknownActionError('Character must be "C" or "D".')
 
 # Type alias for actions.
 Action = Actions
@@ -57,7 +57,7 @@ Action = Actions
 
 def flip_action(action: Action) -> Action:
     if not isinstance(action, Action):
-        raise UnknownAction('Not an Action')
+        raise UnknownActionError('Not an Action')
     return action.flip()
 
 
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -11,6 +11,7 @@ C, D = Actions.C, Actions.D
 
 from enum import Enum
 import random
+from typing import Union
 
 
 class UnknownAction(ValueError):
@@ -36,6 +37,15 @@ class Actions(Enum):
             return Action.C
         else:
             raise UnknownAction('Cannot flip action: {!r}'.format(self))
+
+    @classmethod
+    def from_char(cls, character):
+        if character == 'C':
+            return cls.C
+        elif character == 'D':
+            return cls.D
+        else:
+            raise UnknownAction('Character must be "C" or "D".')
 
     @classmethod
     def random_choice(cls, p: float = 0.5) -> 'Action':
@@ -71,4 +81,8 @@ def str_to_actions(actions: str) -> tuple:
     except KeyError:
         raise UnknownAction(
             'The characters of "actions" str may only be "C" or "D"')
+
+
+def action_sequence_to_str(actions: Union[tuple, list]) -> str:
+    return "".join(map(repr, actions))
 
--- a/axelrod/actions.py
+++ b/axelrod/actions.py
@@ -11,7 +11,7 @@ C, D = Actions.C, Actions.D
 
 from enum import Enum
 import random
-from typing import Union
+from typing import Iterable
 
 
 class UnknownAction(ValueError):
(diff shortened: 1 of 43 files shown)


axelrod/tests/unit/test_actions.py
```diff
@@ -1,32 +1,61 @@
...
+        self.assertEqual(C.flip(), D)
+        self.assertEqual(D.flip(), C)
+
+    def test_from_char(self):
+        self.assertEqual(Actions.from_char('C'), C)
+        self.assertEqual(Actions.from_char('D'), D)
+
+    def test_from_char_error(self):
+        self.assertRaises(UnknownAction, Actions.from_char, '')
+        self.assertRaises(UnknownAction, Actions.from_char, 'c')
```
marcharper: Can we add 'd' as well?

axelrod/tournament.py
```diff
@@ -210,8 +211,8 @@ def _write_interactions_to_file(self, results):
                 row = list(index_pair)
                 row.append(str(self.players[index_pair[0]]))
                 row.append(str(self.players[index_pair[1]]))
-                history1 = "".join([i[0] for i in interaction])
-                history2 = "".join([i[1] for i in interaction])
+                history1 = action_sequence_to_str([i[0] for i in interaction])
```
marcharper: `actions_to_str` might be a better name

marcharper:
Thanks for the contribution! This is looking pretty good to me. We have a 2 review policy for merges and some of us are traveling but we should be able to get this in soon.

I left a couple of comments and there's no major concerns from me on a first pass.

axelrod/tournament.py
eric-s-s: it certainly is.

eric-s-s:
i was happy to do it.  i'll push up those changes and a couple of minor things (i forgot to remove a self.maxDiff = None, and i tested repr, but not str).

marcharper:
Looks great to me! Only one suggestion: `UnknownAction` should be `UnknownActionError` since it's an exception. (I should have realized this before).

@meatballs I approved it regardless of the suggestion above.

axelrod/actions.py
```diff
@@ -9,31 +9,65 @@
...
+    def from_char(cls, character):
+        """Converts a single character into an Action. `Action.from_char('C')`
+        returns `Action.C`. `Action.from_char('CC')` raises an error. Use
+        `str_to_actions` instead."""
+        if character == 'C':
+            return cls.C
+        elif character == 'D':
+            return cls.D
+        else:
+            raise UnknownActionError('Character must be "C" or "D".')
```
meatballs: How about using a dict here to make it a little more 'pythonic'? e.g.

```
actions = {
    'C': cls.C,
    'D': cls.D
}
try:
    return actions[character]
except KeyError:
    raise UnknownActionError('Character must be "C" or "D".')
```

meatballs:
Nice! I've made one suggestion but, other than that, it looks good to go and a nice contribution.

axelrod/actions.py
meatballs: And, doing it this way would remove the need for the UnknownActionError - it could just raise a KeyError with a message

marcharper: That's how it was before I think, I suggested this way thinking it would be more efficient.

axelrod/actions.py
```diff
@@ -9,31 +9,65 @@
...
+        return '{}'.format(self.name)
 
-class Actions(object):
-    C = 'C'  # type: Action
-    D = 'D'  # type: Action
+    def __str__(self):
+        return '{}'.format(self.name)
+
+    def flip(self):
+        """Returns the opposite Action. """
```
eric-s-s: I believe i originally wrote this as also using a dict. Is it better to use the if statements or the dict?  I changed it to follow the style of the function below.

```
    def flip(self):
        """Returns the opposite Action. """
        opposites = {Actions.C: Actions.D,
                     Actions.D: Actions.C}
        return opposites[self]
```

should i use this code instead?

axelrod/actions.py
eric-s-s: that's pretty close.  It originally caught the KeyError and raised a ValueError.  Any way you slice it, it has catch the KeyError and raise a new error to have a useful message, so it could be whatever error we please.  

So:

`UnknownActionError('Character must be "C" or "D".')` or `ValueError('Character must be "C" or "D".')` or `KeyError('The only accepted action keys are "C" and "D".')` 

fyi, i got curious and timed it:
10000 loops of if/else = 4ms
10000 loops of dict = 5.5ms 
so if/else saves about 150ns.

I don't have a strong feeling on this.  i kind of prefer ValueError or UnknownActionError to KeyError, and dictionary to if/else, but that's about the extent of it.  I'll happily go with "do it .... way, please." or I'll make the call if you wish.  what shall i do?

marcharper: AKA approx 1.5/5.5 * 100 = 27.3 % faster :)

meatballs: > i kind of prefer ValueError or UnknownActionError to KeyError, and dictionary to if/else

Me too. It's disappointing (and suprising) that using a dict causes such performance hit. Let's leave things as they are. I'll merge this now.

Issue #1047, opened 2017-06-18
Add a little Pizzaz to the Readme


marcharper:
Some suggestions:

* Image of a small tournament (see examples repo)
* Sparklines of a match
* A plot of the Moran process like the one in [this notebook](https://github.com/Axelrod-Python/Axelrod-notebooks/blob/master/Moran%20Process.ipynb)
* A cooperation heatmap from the examples repository
* A couple of fingerprints
* Remove some of the installation instructions (like mpl issues on ubuntu) -- could go into a wiki page
* Make the reference paper more prominent

Thoughts @drvinceknight @meatballs ?

meatballs:
Yep - sounds good. We should have installation and usage stuff in the docs. The readme can then be a 'sales pitch'

drvinceknight:
I also reckon the initial paragraph ("Currently ...") is a bit heavy, by the time I get to the `README` I'd like short and sharp information/sale pitches. If I want details I'll look at the documentation.

drvinceknight:
We should have a very minimum of installation stuff still:

```
pip install axelrod
```

and 

```
python setup.py develop
```

marcharper:
Might I humbly suggest something more like [this repository](https://github.com/marcharper/python-ternary)?

drvinceknight:
There is a lot in that README. Do you have separate documentation elsewhere
for that or is that your user documentation also?

It might just be the nature of tertiary but I'd be cautious about overlap
with our docunentation (our README is not our user documentation). I like
@meatballs way of phrasing it: a sales pitch.

I'd suggest the audience of the README is either developers who will want
to be pointed at our contribution documentation (without having to read a
lot). Or potential users who'll want that sales pitch and a pointer to the
documentation.

To add to the examples here's my README for Nashpy
https://github.com/drvinceknight/Nashpy

drvinceknight:
And here's another example: https://github.com/PyconUK/ConferenceScheduler

drvinceknight:
I wouldn't be surprised if we're actually all on the same page and it might be helpful for someone to just take a stab at this to give us a reference point...

drvinceknight:
For further reference (I'm not suggesting that any of these are what we want for Axelrod, just throwing things out here):

- https://github.com/numpy/numpy is extremely short and just has a few bullet points.
- https://github.com/scipy/scipy#documentation also pretty bare but points to all relevant links.

drvinceknight:
[...]

PR #1072, opened 2017-07-16
Streamline the README

drvinceknight:
Here's my streamlining of the README - discussions at #1047. I had it lying around and think it's better than the status quo.

You can view it here: https://github.com/Axelrod-Python/Axelrod/blob/readme-revamp/README.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,21 +34,32 @@ Features
 
 With Axelrod you:
 
-- have access to over 200 strategies, extendable through parametrization and
-  a collection of strategy transformers.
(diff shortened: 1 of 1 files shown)


marcharper:
A few suggestions:
* smaller tournament with a more readable image
* reduce the size of the fingerprint image
* still mention that classics like WSLS are included
* maybe link the capabilities straight to the relevant section in the docs?

drvinceknight:
All great suggestions, I agree, I'll pick these up tomorrow :+1: :)

PR #1076, opened 2017-07-20
Add keys to Fortran code in a separate file.

drvinceknight:
This uses the following to generate the `fortrain_keys.rst` file:

```
import glob
import pathlib
strategy_files = glob.glob("./src/*.f")
with open("/home/vince/src/Axelrod/docs/reference/fortan_keys.rst", "w") as f:
    for file in strategy_files:
        path = pathlib.Path(file)
        strategy_name = path.stem.upper()
        file_name = path.name
        f.write(".. _{}: https://github.com/Axelrod-Python/TourExec/blob/master/src/{}\n".format(strategy_name, file_name))
```


meatballs:
Nice! I hadn't thought of just using an external file and including that.

meatballs:
I'll put something similar together to build the table...

PR #1077, opened 2017-07-21
Fix a bug that was being skipped.

drvinceknight:
I assume this was being skipped because of output ordering. Hopefully
CI is still happy on all py versions.


meatballs:
Merging under the bugfix protocol....

PR #1081, opened 2017-07-22
Refactor Tournament docs

meatballs:
- Use a csv_table for the first tournament and the S&P tournament
- Use :class: directives instead of :code: in both tables
- Remove the text descriptions of first tournament strategies


axelrod/strategies/axelrod_first.py
```diff
@@ -346,6 +376,8 @@ class Shubik(Player):
     Plays like Tit-For-Tat with the following modification. After each
     retaliation, the number of rounds that Shubik retaliates increases by 1.
 
+    his strategy came 5th in Axelrod's original tournament.
```
drvinceknight: `This`

drvinceknight:
I clicked approve but think I picked up a typo above.

PR #1084, opened 2017-07-24
N tit(s) for M tat(s)

### LINKED ISSUE #1073: SlowTitForTwoTats duplicate of TitFor2Tats
After poking around titfortat strategies, I found out that `TitFor2Tats` and `SlowTitForTwoTats` do the exact same thing although coded differently. Shouldn't we get rid of `SlowTitForTwoTats` ? And, in that case, should we rename `SlowTitForTwoTats2` ?

> strategies.tifortat : TitFor2Tats : l.68
```python
    @staticmethod
        return D if opponent.history[-2:] == [D, D] else C
```

> strategies.tifortat : SlowTitForTwoTats : l.517
```python

        # Start with two cooperations
        if len(self.history) < 2:
            return C

        # Mimic if opponent plays the same move twice
        if opponent.history[-2] == opponent.history[-1]:
            return opponent.history[-1]

        # Otherwise cooperate
        return C
```

Chadys:
Resolve #1073 and add the NTitsForMTats asked for in  #379


drvinceknight:
@Chadys looks like a doctest is failing:

```
File "./docs/tutorials/advanced/classification_of_strategies.rst", line 59, in classification_of_strategies.rst
Failed example:
    len(strategies)
Expected:
    28
Got:
    29
```

Chadys:
Yes I saw, `NTitsForMTats` is a memory one strategy if called with its default arguments, so I updated the returned number of memory one strategies in the example.

drvinceknight:
> Yes I saw, NTitsForMTats is a memory one strategy if called with its default arguments, so I updated the returned number of memory one strategies in the example.

If I'm correct the default is TfT so I'd suggest not including this in `axelrod.strategies` (which is what I did for `ReactivePlayer` which doesn't actually have a default) just to avoid having two instances of TfT in there? (So in turn you would revert 69723ee)

Chadys:
I thought all strategies had to be in `axelrod.strategies`, I changed it !

drvinceknight:
> I thought all strategies had to be in axelrod.strategies,

Not quite :) This is similar to `ReactivePlayer` and `LookerUp` which are strategies that are intended to be used to create specific strategies so they're not added to the `axelrod.strategies` list. (And also the cheating strategies are not in there either.).

I'll review this properly tomorrow, logging off for the evening.

marcharper:
Can we add one instance to the library as a strategy, with say N=3 M=2 or similar?

drvinceknight:
Sounds good to me. Perhaps that could be the default?

axelrod/strategies/titfortat.py
```diff
@@ -750,3 +716,59 @@ def strategy(self, opponent: Player) -> Action:
...
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self, N: int=1, M: int=1) -> None:
+        """
+        Parameters
+        ----------
+        N, int
```
drvinceknight: Because of the type hints let's drop the type in the docstrings (in case at some point they are changed and we forget to change them in the docstrings).

axelrod/strategies/titfortat.py
```diff
@@ -750,3 +716,59 @@ def strategy(self, opponent: Player) -> Action:
...
+        self.M = M
+        self.classifier['memory_depth'] = max([M,N])
+        self.retaliate_count = 0
+
+    def strategy(self, opponent: Player) -> Action:
+        # if opponent defected consecutively M times, start the retaliation
+        if not self.M or opponent.history[-self.M:].count(D) == self.M:
+            self.retaliate_count = self.N
+        if self.retaliate_count:
+            self.retaliate_count-=1
```
drvinceknight: PEP8: `self.retaliate_count -= 1`

Chadys:
All done, I went with the change in default values solution !

drvinceknight:
Looks good to me :+1:

PR #1090, opened 2017-07-27
Remove strategies that were assumed implemented.

drvinceknight:
See
https://github.com/Axelrod-Python/axelrod-fortran/issues/31#issuecomment-318279265

Note I removed the "Work in progress" also because, similarly, that is
being implemented based on the text description in Axelrod's first
paper.


marcharper:
Some of these, like Joss, were sufficiently described in later sources IIRC.

drvinceknight:
> Some of these, like Joss, were sufficiently described in later sources IIRC.

Yes but we do not know that, until the actual source code in the Fortran strategies is "figured out" it makes sense to mark these as "not implemented".

drvinceknight:
So to be super clear: the name of this PR is clumsy, we're not removing the strategies from the library, just tweaking the docs to denote that we don't know that these 3 strategies from Axelrod's second tournament are implemented. (We don't really have a reason to believe that they correspond to the 3 strategies from the first tournament).

TLDR: This is just a tweak to documentation.

PR #1091, opened 2017-07-27
Michaelos strategy

dmanc:
Implemented a strategy from this [link](http://lesswrong.com/lw/7f2/prisoners_dilemma_tournament_results/) (specifically letter H) mentioned in #379. For some reason it shows commits that I pulled from upstream, are those supposed to be there?


axelrod/strategies/titfortat.py
```diff
@@ -772,3 +774,52 @@ def strategy(self, opponent: Player) -> Action:
...
+    the game.
+
+    Names:
+
+    - Michaelos: [LessWrong2011]_
+    """
+
+    name = 'Michaelos'
+    classifier = {
+        'memory_depth': 1,
```
drvinceknight: I believe that the memory depth is not 1.

If on turn `n` Michaelos defects and the opponent cooperates then 50% of the time I defect and defect throughout, However if I have already **become** a defector than I will defect 100% of the time so I not only need to know the previous turns (that I D and you C but I also need to know if I D'd because I have already "transformed" in to a defector).

I believe the memory depth is in fact `float('inf')`:

If our actions are `...(D, D), (D, C)` then I do not know if my previous plays of `D` are TfT retaliations **or** because I have previously been turned in to a defector so I do not know if I need to randomly sample as a result of the last round being `(D, C)`.

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -717,3 +717,50 @@ def test_varying_memory_depth(self):
...
+        actions = [(C, D), (D, D), (D, D), (D, D)]
+        self.versus_test(axelrod.Defector(), expected_actions=actions,
+                         attrs={"is_defector": False}, seed=2)
+
+        actions = [(C, D), (D, D), (D, D), (D, D)]
+        self.versus_test(axelrod.Defector(), expected_actions=actions,
+                         attrs={"is_defector": False},
+                         match_attributes={"length": -1}, seed=2)
+
+        # Chance of becoming a defector is 50% after (D, C) occurs.
```
drvinceknight: Could you repeat this test with a different seed to show differing behaviour after `(C, D)` (so only need 4 round please):

```
actions = [(C, C), (C, D), (C, C), (C, D)]
```

dmanc: Wait, I don't think this scenario is possible for the strategy unless I am misinterpreting something. After (C, D) Michaelos will play D since it is still playing like Tit-for-Tat.

dmanc: For the second test, does tweaking the test mean changing the expected actions?

drvinceknight: > After (C, D) Michaelos will play D since it is still playing like Tit-for-Tat.

Sorry you're correct, that was a typo from me, I simply mean to include a random seed version to illustrate that 50% of the time after `(D, C)` (not `(C, D)` - sorry) the strategy will cooperate **and ** `is_defector` will be False (so kind of like your `seed=1` test but without the second `(C, D)`.

> For the second test, does tweaking the test mean changing the expected actions?

Yes.

dmanc:
How do I fix the Travis build error for python 3.5?

drvinceknight:
I've set the build going again. It's hypothesis (a tool to find random test scenarios) falling over because it didn't find anything in time. I'll keep an eye on it, if it fails again we might need to tweak the settings but I'll assist with that.

drvinceknight:
This looks good to me, thanks @dmanc :+1:

PR #1092, opened 2017-07-27
pickling for strategies and transformed strategies

eric-s-s:
#718 

had to change `test_fingerprint.py` because the `axl.seed` behavior has changed for `DualTransformer`.  `DualTransformer` still passes all its own tests.

The equality testing to Game was added so that I could test `pickle.loads(pickle.dumps(player)) == player`. Since it creates a new Game instance, these tests evaluated to false until i added the `__eq__` method to `Game`.


drvinceknight:
Thanks for this @eric-s-s, I've only scanned for now but so that I understand the context: this is a first step towards fixing #718 by making everything (including decorated strategies) pickable?

axelrod/tests/unit/test_pickling.py
```diff
@@ -0,0 +1,349 @@
...
+    pass
+
+
+@st.InitialTransformer((C, D), name_prefix=None)
+class Initial(axl.Cooperator):
+    pass
+
+
+@st.JossAnnTransformer((0.2, 0.2), name_prefix=None)
+class JossAnn(axl.Cooperator):
```
drvinceknight: What are these for?

eric-s-s:
@drvinceknight yes.  that's correct.  Since windows multiprocess requires everything be pickled, that was the aim.

>pickable? Pickalable? Made possible to become pickles 😆 ?

yeah, i've settled for "pickle-able".  it does roll off the tongue, though.  i've had Peter Piper running through my head for a few days now.

drvinceknight:
> Since windows multiprocess requires everything be pickled, that was the aim.

OK great thanks, that makes sense (and useful to have anyway I guess). Will try to review this first thing tomorrow :+1:

axelrod/tests/unit/test_pickling.py
eric-s-s: There seem to be roughly 3 cases.

1) a decorated class declaration with no name prefix
    - based off of `Player`
    - based off of another strategy
2) a decorated class declaration with a name prefix
    - see `DoubleFlip` for an example
3) a dynamic wrapping of a class (with a name prefix)

i wanted to make sure that each transformer was tested on case-3 and either case-1 or case-2.  So these are module level class declarations for case-1, dynamically wrapped strategies for case-3, and some other module level class declarations for special cases and case-2.

and probably some overkill.  problems popped up in strange ways and i got paranoid.  pickling problems are weird and have little useful stack trace information when they throw an error.

drvinceknight: Ok great thanks :+1:

axelrod/strategy_transformers.py
```diff
@@ -149,15 +173,60 @@ def __repr__(self):
...
+    def __call__(self, strategy_wrapper, name_prefix, reclassifier,
+                 args, kwargs, instance_name_prefix):
+        decorator_class = StrategyTransformerFactory(
+            strategy_wrapper, name_prefix, reclassifier
+        )
+        kwargs['name_prefix'] = instance_name_prefix
+        return decorator_class(*args, **kwargs)
+
+
+class StrategyReBuilder(object):
```
drvinceknight: Could you write a docstring for this.

axelrod/strategy_transformers.py
```diff
@@ -149,15 +173,60 @@ def __repr__(self):
...
+    :param player_class: Any class that inherits from axelrod.Player
+    :return: bool
+    """
+    for klass in player_class.mro():
+        method = inspect.getattr_static(klass, 'strategy', default=None)
+        if method is not None:
+            return isinstance(method, staticmethod)
+
+
+class DecoratorReBuilder(object):
```
drvinceknight: Doc string please.

axelrod/strategy_transformers.py
```diff
@@ -84,23 +92,18 @@ def __call__(self, PlayerClass):
...
-                is_static = False
-
             # Define the new strategy method, wrapping the existing method
             # with `strategy_wrapper`
             def strategy(self, opponent):
-
-                if is_static:
-                    # static method
-                    proposed_action = PlayerClass.strategy(opponent)
+                if strategy_wrapper == dual_wrapper:
```
drvinceknight: You've lost me here, isn't this just setting the `proposed_action` of anything wrapped with `Dual` to be `C`?

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -15,6 +15,20 @@ class TestClass(object):
 
 class TestTransformers(unittest.TestCase):
```
drvinceknight: Could we also add specific unit tests for those two extra classes you wrote in `strategy_transformers`: `DecoratorReBuilder` and `StrategyReBuilder`?

axelrod/strategy_transformers.py
```diff
@@ -228,11 +297,21 @@ def dual_wrapper(player, opponent, proposed_action):
     -------
     action: an axelrod.Action, C or D
     """
+
     if not player.history:
-        player.original_player = player.original_class(**player.init_kwargs)
+        player.use_history = []
```
drvinceknight: Actually @eric-s-s , I'm not sure this is going to work. 

The reason we originally didn't just take the history but the player itself is that the Dual must keep track of the state of the original player, so that must include any other internal variables that would get changed by the original player.

I see two course of action:

- Keeping the `original_player` and finding a way to get it to pickle
- Add tests with more complex players and check that they both pass with or without this change (no matter what this should probably happen).
- Not only copy the player but copy all the attributes back and forth (least pleasing of all the options in my opinion)

drvinceknight: @theref could you take a look at this and confirm I've got things right? It's not as simple as just using the history, we need to keep track of the whole state, I don't believe the change written here ensures that.

No matter what we should probably write a test that would catch this.

theref: Correct, the internal state of the original player is what's important, which means all internal variables must also be retained. Another option is that it may be possible to 'rebuild' the original player from the history.

I agree that some tests definitely need to be included that ensure this.

drvinceknight:
> I agree that some tests definitely need to be included that ensure this.

Perhaps a test against a player that makes use of a lot of internal variables.

axelrod/strategy_transformers.py
drvinceknight: > I agree that some tests definitely need to be included that ensure this.

Perhaps a test against a player that makes use of a lot of internal variables.

eric-s-s: i'll comment the code where state-change is recorded, and the possible hidden issue (that was also true with the original)

here's the test i just tried. `assert_dual_wrapper_correct` is lifted from code in `test_strategy_transformers.py`.  (it was a separate method as I used it in a few other tests (which can be mostly be removed)).

```
    def test_all_strategies_with_dual(self):
        for s in axl.strategies:
            self.assert_dual_wrapper_correct(s)

    def assert_dual_wrapper_correct(self, player_class):
        p1 = player_class()
        p2 = st.DualTransformer()(player_class)()
        p3 = axl.CyclerCCD()  # Cycles 'CCD'

        axl.seed(0)
        for _ in range(10):
            p1.play(p3)

        p3.reset()

        axl.seed(0)
        for _ in range(10):
            p2.play(p3)

        self.assertEqual(p1.history, [x.flip() for x in p2.history])
```

it failed on  `fails = [axl.EvolvedANN , axl.EvolvedANN5, axl.EvolvedANNNoise05, axl.Golden, axl.e]`

after i added a function to switch player.cooperations and player.defections (and then switch them back) they all passed.  

Does this work, testing-wise?

eric-s-s:
@drvinceknight i'll get to work on all of that in the next few days.  just imagine a thumbs-up on every comment

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -15,6 +15,20 @@ class TestClass(object):
...
+            pass
+
+        class NewAlternator(axelrod.Alternator):
+            pass
+
+        self.assertTrue(is_strategy_static(NewCooperator))
+        self.assertFalse(is_strategy_static(NewAlternator))
+
     def test_all_strategies(self):
         # Attempt to transform each strategy to ensure that implementation
```
eric-s-s: you mean that i forgot to test IdentityTransformer, right?  oops.

eric-s-s: oh no!!!!  it fails!  i think i know why, and it may or may not be possible to fix.  yikes!

axelrod/strategy_transformers.py
eric-s-s: it is.  but more importantly, it's avoiding calling the wrapped player twice.  since dual wrapper requires a `proposed_action` variable, but doesn't use it, the `C` is just a stand-in value.

axelrod/strategy_transformers.py
```diff
@@ -228,11 +297,21 @@ def dual_wrapper(player, opponent, proposed_action):
...
+
+    temp = player.history[:]
+    player.history = player.use_history[:]
+
```
eric-s-s: here's where the state-change is happening.  when the player calls its `original_class.strategy` it still affects all the state in the player.  so, for instance, a `DualFSMPlayer` would still own an FSM, and update its state based on the call to `self.FSMPlayer.strategy`.

axelrod/tests/unit/test_strategy_transformers.py
eric-s-s: arrrrggghh.  brought low by the lowliest of transformers!!!

axelrod/tests/unit/test_pickling.py
```diff
@@ -0,0 +1,349 @@
...
+        self.assert_original_plays_same_as_pickled(axl.Alexei(), turns=10)
+
+    def test_nice_transformer_class(self):
+        player = axl.NMWEDeterministic()
+        self.assert_original_plays_same_as_pickled(player, turns=10)
+
+    def test_pickling_all_strategies(self):
+        for s in axl.strategies:
+            player = s()
+            player.play(axl.Cooperator())
```
eric-s-s: what about just using `assert_mutated_instance_the_same_as_pickled` from line 168?

axelrod/strategy_transformers.py
drvinceknight: > Does this work, testing-wise?

Yes I think it does. @theref can you confirm? 

I expect we would get more failures for more than 10 turns (especially for the complex FSMs) but in the interest of run time I think this is a sufficient test.

> after i added a function to switch player.cooperations and player.defections (and then switch them back) they all passed.

The trick will be to get an automatic function that doesn't just switch `player.cooperation` and `.defections` but every and any attribute that can affect a strategy (take a look at the player equality method, we essentially go in and grab relevant attributes there).

axelrod/strategy_transformers.py
drvinceknight: Sorry but I still don't see why this is necessary and/or follow. If it's just a minor efficiency boost then lets remove it in the interest of clarity.

axelrod/strategy_transformers.py
drvinceknight: Yes but that's the problem, it should have the state of the original player.

axelrod/tests/unit/test_strategy_transformers.py
drvinceknight: My comment here was just that you wrote `def test_all_strategies` and a comment underneath but that you hadn't actually written any code. If you've found a fail case though that's good, although sucky if it's not fixable. Fingers crossed...

drvinceknight: Woops my comment was actually a mistake because I misread the diff but if you've found a fail case that's good I guess although fingers crossed we can get a fix :confused: :)

eric-s-s: phew.  fixed.  just needed to sleep on it.  And it helped me make the code cleaner.  :)

axelrod/strategy_transformers.py
eric-s-s: tested with turns=100 . all tests pass.  i'll keep it at turns=100 for the next PR.  it adds about 5-10 seconds to the test.

axelrod/strategy_transformers.py
eric-s-s: Here's my thinking process:

`op` = original_player
`wp` = wrapped_player

`play` only mutates history, cooperations, defections, state_distributions.

all other player attributes are mutated in `player.strategy`

`play_attrs` = history, cooperations, defections, state_distributions

`wp` strategy:

1) switch to `op.play_attrs`
2) play `op` strategy
3) this will update `wp` state as if it was `op`
4) update history with `op` action - this step could be skipped and done on-the-fly in step 1 (see code below)
5) switch back to `wp.play_attrs`
6) flip action
7) return (assuming play was called, `play` will now update `wp.play_attrs`.  this assumption was also implicit in the original implementation.)

the resulting player is a `wp` with `wp` history but `op` state.  

code to switch play_attrs (as yet untested.  This is just to get the gist):

```
def switch_play_attrs(player):
    new_history = [action.flip() for action in player.history]
    player.history = new_history
    
    temp = player.cooperations
    player.cooperations = player.defections
    player.defections = temp
 
    new_distributions = defaultdict(int)
    for key, val in player.state_distribution.items():
        new_key = (key[0].flip(), key[1])
        new_distributions[new_key] = val
    player.state_distribution = new_distributions
```

axelrod/strategy_transformers.py
eric-s-s: (see the explanation of how the state change works first)

`wp` = wrapped_player
`op` = original_player

the calls inside `dual_wrapper` call `op.strategy` **on** `wp`.  this mutates the state of `wp` as if it was `op` .  if the `op.strategy` gets called here again, it will again mutate the state of `wp`.  this would be bad.  

ex of bad way (if uses proposed_action = PlayerClass.strategy(opponent):

player = DualCycler('CD') not yet played
expected_action = `C.flip()` = `D`
cycle now advanced to `D`
expected_player_state = next_action_to_be_returned: `D.flip()`

1) `wp.strategy` called
2) `proposed_action` calls `op.strategy`
3) returns `C` (which will be discarded)
4) mutates state of `wp` (cycle now advanced to 'D')
5) `dual_wrapper` called
6) `dual_wrapper` calls `op.strategy` 
7) returns `D`
8) mutates state of `wp`  (cycle now advanced to `C`, the wrong state)
9) return action.flip() (which is now `C`, the wrong action)

removing steps 2, 3, 4 makes `wp` return the correct action and stay in the correct state.

axelrod/strategy_transformers.py
drvinceknight: I think this looks right. 

To ensure I'm following: step 5 and 6 could be swapped? (Not suggesting they are, just checking I'm following).

drvinceknight:
and step 3 is the important one right? Step 3 is the step that's ensuring the state is updated "properly".

axelrod/strategy_transformers.py
drvinceknight: and step 3 is the important one right? Step 3 is the step that's ensuring the state is updated "properly".
--

eric-s-s: > To ensure I'm following: step 5 and 6 could be swapped? (Not suggesting they are, just checking I'm following).

yes, that is correct.

> and step 3 is the important one right? Step 3 is the step that's ensuring the state is updated "properly"

yes.  step 3 automatically happens as a result of step 2. if it didn't, this would all fall apart.

eric-s-s: i'm in the middle of refactoring, but perhaps the new `dual_wrapper` code (doc_str edited out) will make things more clear.

```
def flip_wrapper(player, opponent, proposed_action):
    flip_play_attributes(player)

    if is_strategy_static(player.original_class):
        action = player.original_class.strategy(opponent)
    else:
        action = player.original_class.strategy(player, opponent)

    flip_play_attributes(player)

    return action.flip()
```

`flip_play_attributes` flips `history`, `cooperations`, `defections` and `state_distribution`

eric-s-s:
i've hit a major snag with dual and could use some advice.  

first, the old version of dual has a bug.  The following test passes.

```
    # (on master)

    def test_bug(self):
        p = IdentityTransformer()(DualTransformer()(axelrod.Cooperator))()
        p2 = DualTransformer()(axelrod.Cooperator)()

        for _ in range(3):
            p.play(p2)

        self.assertEqual(p.history, [C, C, C])
        self.assertEqual(p2.history, [D, D, D])
```

`DualDualCooperator` returns `C`'s
`DualDualDualCooperator` returns `D`'s

but if any of the `Dual`'s are not in the outer group of decorators, it doesn't work.

`DualIdentityDualCooperator` returns `D`'s
`IdentityDualDualCooperator` returns `D`'s

I have a feeling that if i can understand what's going on there, it'll help solve the current problem.  

current problem:
`dual_wrapper` calls player.original_class, which will always be the top class in a group of decorators.

so if you have `FlippedDualCooperator` , `dual_wrapper` tries to play the strategy for `DualCooperator`, which tries to play the strategy for `DaulCooperator` ...   and then you reach maximum recursion depth.

current "solution" (shudder):
give the class a dedicated attr for dual_wrapper (currently called "for_dual"), that allows the `dual_wrapper` to call the correct strategy.  Now, `FlippedDualCooperator` flips the `DualCooperator` which uses the "for_dual" attribute to do the opposite of a `Cooperator`.  It's GREAT! PROBLEM SOLVED! except if `Dual` is used twice.  

i've tried:
- making dual_wrapper a class that owns an "original_player". (kinda-ugly, didn't work but probably the best direction)
- making the new player class own a list of classes to call for dual_wrappers (very ugly, couldn't get it to work)
- in the player_class, dynamically creating numbered attributes ("for_dual_1", "for_dual_2") for each dual_wrapper (super ugly and couldn't get it to work)

I would love some ideas.  and/or i would love to know what's going on in the original.

drvinceknight:
Am I correct in understanding that this bug "only" affects using the Dual twice?

eric-s-s:
yes, that is correct.

drvinceknight:
> yes, that is correct.

I would be quite tempted to simply include a catch if that's ever called and raise an error. From the use case point of view being able to call them twice is not really necessary (these Duals are mainly used in fingerprinting). Obviously ideally and for completeness it would be nice to have but if this makes the code too ugly/hard to maintain and/or is not possible I'd be happy to just raise an error saying "Calling the Dual recursively is not implemented."

eric-s-s:
hmmmmm.  i've found a solution that works beautifully, but is very programmatically questionable (it passes a set of ugly test)

```
            def strategy(self, opponent):
                if strategy_wrapper == dual_wrapper:
                    
                    flip_play_attributes(self)

                if is_strategy_static(PlayerClass):
                    proposed_action = PlayerClass.strategy(opponent)
                else:
                    proposed_action = PlayerClass.strategy(self, opponent)

                if strategy_wrapper == dual_wrapper:
                    
                    flip_play_attributes(self)

                # Apply the wrapper
                return strategy_wrapper(self, opponent, proposed_action,
                                        *args, **kwargs)
```

mixed with

```
def dual_wrapper(player, opponent: Player, proposed_action: Action) -> Action:
    """Wraps the players strategy function to produce the Dual.

    The Dual of a strategy will return the exact opposite set of moves to the
    original strategy when both are faced with the same history.

    A formal definition can be found in [Ashlock2010]_.
    http://doi.org/10.1109/ITW.2010.5593352

    Parameters
    ----------
    player: Player object or subclass (self)
    opponent: Player object or subclass
    proposed_action: axelrod.Action, C or D
        The proposed action by the wrapped strategy

    Returns
    -------
    action: an axelrod.Action, C or D
    """

    # flip_play_attributes(player)
    #
    # if is_strategy_static(player.for_dual):
    #     action = player.for_dual.strategy(opponent)
    # else:
    #     action = player.for_dual.strategy(player, opponent)
    #
    # flip_play_attributes(player)

    return proposed_action.flip()
```

so, the player_class is responsible for all the work of the dual_wrapper.  in this case, it passes all the tests that original dual_wrapper failed on and the tests that the new dual_wrapper failed on.

eric-s-s:
it adds a specialized code for `dual_wrapper` in the strategy (which was there anyways, just less intrusively) and removes the need for a special attribute added to the dynamically created player_class.  so it solves the problem better, makes one part neater and another part uglier.

drvinceknight:
> so, the player_class is responsible for all the work of the dual_wrapper. in this case, it passes all the tests that original dual_wrapper failed on and the tests that the new dual_wrapper failed on.

Ok I my initial thought it that this is a nice/ok solution. We should document in the dual wrapper why it has been implemented differently.

eric-s-s:
>Ok I my initial thought it that this is a nice/ok solution. We should document in the dual wrapper why it has been implemented differently.

i'll push that up soon and we can see how it looks

drvinceknight:
> i'll push that up soon and we can see how it looks

I'll review properly tomorrow :+1: Thanks :)

axelrod/strategy_transformers.py
```diff
@@ -84,24 +95,25 @@ def __call__(self, PlayerClass):
...
+                    # played the opposite actions of its current history.
+                    flip_play_attributes(self)
+
+                if is_strategy_static(PlayerClass):
                     proposed_action = PlayerClass.strategy(opponent)
                 else:
                     proposed_action = PlayerClass.strategy(self, opponent)
 
+                if strategy_wrapper == dual_wrapper:
+                    # after dual_wrapper_figures calls the strategy, it returns
```
eric-s-s: typo here - "dual_wrapper_figures" should be "dual_wrapper".  i'll wait on a full review before pushing this change.

axelrod/tests/unit/test_pickling.py
```diff
@@ -0,0 +1,335 @@
...
+    def test_pickling_all_transformers_as_instance_called_on_a_class(self):
+        for transformer in transformer_instances:
+            player = transformer(axl.Cooperator)()
+            self.assert_orignal_equals_pickled(player)
+
+    def test_created_on_the_spot_multiple_transformers(self):
+        klass = st.FlipTransformer()(axl.Cooperator)
+        klass = st.DualTransformer()(klass)
+        player = st.FinalTransformer((C, D))(klass)()
+        copy = pickle.loads(pickle.dumps(player))
```
eric-s-s: this should be `self.assert_original_equals_pickled(player)`

axelrod/strategy_transformers.py
```diff
@@ -141,6 +153,27 @@ def __repr__(self):
...
 
+            def reduce_for_decorated_class(self_):
+                class_module = import_module(self_.__module__)
+                import_name = self_.__class__.__name__
+
+                if player_can_be_pickled(self_):
+                    return self_.__class__, (), self_.__dict__
+
+                decorators = []
+                for klass in self_.__class__.mro():
```
drvinceknight: `class_` perhaps?

axelrod/strategy_transformers.py
```diff
@@ -141,6 +153,27 @@ def __repr__(self):
                     prefix = ', '
                 return name
 
+            def reduce_for_decorated_class(self_):
```
drvinceknight: Docstring please.

axelrod/strategy_transformers.py
```diff
@@ -141,6 +153,27 @@ def __repr__(self):
...
+                class_module = import_module(self_.__module__)
+                import_name = self_.__class__.__name__
+
+                if player_can_be_pickled(self_):
```
drvinceknight: Instead of this can we use a `try/except`?

axelrod/tests/unit/test_pickling.py
```diff
@@ -0,0 +1,335 @@
...
+            player = s()
+            self.assert_orignal_equals_pickled(player)
+
+    def test_pickling_all_transformers_as_instance_called_on_a_class(self):
+        for transformer in transformer_instances:
+            player = transformer(axl.Cooperator)()
+            self.assert_orignal_equals_pickled(player)
+
+    def test_created_on_the_spot_multiple_transformers(self):
+        klass = st.FlipTransformer()(axl.Cooperator)
```
drvinceknight: `class_`? (not terribly insistent on this)

axelrod/tests/unit/test_pickling.py
```diff
@@ -0,0 +1,335 @@
...
+
+class TestPickle(unittest.TestCase):
+
+    def assert_orignal_equals_pickled(self, player, turns=10):
+        self.assert_original_plays_same_as_pickled(player, turns)
+        self.assert_mutated_instance_same_as_pickled(player)
+
+    def assert_original_plays_same_as_pickled(self, player, turns=10):
+        copy = pickle.loads(pickle.dumps(player))
+        opponent_1 = axl.CyclerCCCDCD()
```
drvinceknight: Can we loop this over a collection of opponents:

```
for opponent_class in [axl.Defector(), axl.Cooperator(), axl.Random(), axl.CyclerCCCDC]:
```

axelrod/tests/unit/test_pickling.py
```diff
@@ -0,0 +1,335 @@
...
+        axl.seed(0)
+        match_2 = axl.Match((copy, opponent_2), turns=turns)
+        result_2 = match_2.play()
+
+        self.assertEqual(result_1, result_2)
+
+    def assert_mutated_instance_same_as_pickled(self, player):
+        player.reset()
+        turns = 5
+        opponent = axl.Alternator()
```
drvinceknight: Same comment about looping over a set of opponents.

axelrod/tests/unit/test_pickling.py
```diff
@@ -0,0 +1,335 @@
...
+        klass = st.FlipTransformer()(axl.Cooperator)
+        klass = st.DualTransformer()(klass)
+        player = st.FinalTransformer((C, D))(klass)()
+        copy = pickle.loads(pickle.dumps(player))
+
+        self.assertEqual(player, copy)
+        self.assert_mutated_instance_same_as_pickled(player)
+
+    def test_dual_transformer_special_case(self):
+        player = TestDualTransformerIssues()
```
drvinceknight: This isn't terribly clear?  (even the class definition before)

Could the class def be moved here and made a bit more explicit.

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -101,38 +248,60 @@ def test_dual_wsls_transformer(self):
...
+        self.assert_dual_wrapper_correct(klass)
 
-        p3.reset()
-        for _ in range(10):
-            p2.play(p3)
+        klass = JossAnnTransformer((0.5, 0.4))(axelrod.EvolvedLookerUp2_2_2)
+        self.assert_dual_wrapper_correct(klass)
 
-        self.assertEqual(p1.history, [x.flip() for x in p2.history])
+    def test_dual_transformer_special_cases_1(self):
```
drvinceknight: Am I right that these tests are the ones that are fixed by your re write of Dual?

axelrod/tests/unit/test_pickling.py
eric-s-s: i picked "klass" as it seems to be convention in **a lot** of code.  i also have no real preference.  

then i search on PEP8 and here's what i found.

>If a function argument's name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus class_ is better than clss . (Perhaps better is to avoid such clashes by using a synonym.)

sooo... `class_` seems official. i'll go with that.

axelrod/tests/unit/test_strategy_transformers.py
eric-s-s: yes, that is correct.  it addresses issues with the original code and a few different tries on the re-write

axelrod/tests/unit/test_pickling.py
eric-s-s: unfortunately, this can't be moved inside the test.  no pickling of non-module-lvl classes.  i'll work on clarity in class name, test name, and perhaps comments.

axelrod/tests/unit/test_pickling.py
eric-s-s: adding a few seconds to total test time is not an issue, correct?  since the play mutates the original player, perhaps i can just absorb this into the previous test.

axelrod/strategy_transformers.py
eric-s-s: i'm not sure how.  no error is raised in the __reduce__ function, so it wouldn't be able to catch it, right?  since this is called by the pickler, that is where you'd need the try/catch.

axelrod/strategy_transformers.py
eric-s-s: while checking PEP8 about `klass` , i noticed about the importance of sticking with `self` and `cls`.  even though it shadows a variable from outer scope, is it better to shadow the `self` variable than to use `self_` ?

here's the quote

> Function and method arguments
  Always use self for the first argument to instance methods.
  Always use cls for the first argument to class methods.

axelrod/strategy_transformers.py
eric-s-s: or `child_class` or `sub_class`  might be better?

axelrod/tests/unit/test_pickling.py
drvinceknight: Sounds good.

axelrod/tests/unit/test_pickling.py
drvinceknight: Cool.

axelrod/tests/unit/test_strategy_transformers.py
drvinceknight: Cool.

axelrod/strategy_transformers.py
drvinceknight: Ok sounds like you've already thought about it. :+1:

axelrod/strategy_transformers.py
drvinceknight: I'm not terribly sure to be honest. I'm happy as is, I'm guessing that quote does not necessarily apply to nested def and class?

axelrod/strategy_transformers.py
drvinceknight: Fine with that also.

axelrod/tests/unit/test_pickling.py
```diff
@@ -238,19 +230,37 @@ def test_pickling_all_transformers_as_instance_called_on_a_class(self):
...
+        transformer or when other transformers were between multiple
+        DualTransformers."""
+        player = InterspersedDualTransformersNamePrefixAbsent()
         self.assert_orignal_equals_pickled(player)
 
-        player = TestDualTransformerIssues2()
+        player = InterspersedDualTransformersNamePrefixPresent()
         self.assert_orignal_equals_pickled(player)
 
+        interspersed_dual_transformers = (
```
drvinceknight: This isn't terribly nice to look at. Perhaps we can make use of intermediate variables to improve readability?

axelrod/tests/unit/test_strategy_transformers.py
drvinceknight: This has been done: :+1:

axelrod/tests/unit/test_pickling.py
eric-s-s: i wasn't sure how to go about it.  i'll try that.

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -101,38 +248,67 @@ def test_dual_wsls_transformer(self):
...
+        It has also failed when DualTransformer was not the outermost
+        transformer or when other transformers were between multiple
+        DualTransformers."""
+        dual_not_first_transformer = IdentityTransformer()(DualTransformer()(axelrod.EvolvedANN))
+        self.assert_dual_wrapper_correct(dual_not_first_transformer)
+
+        multiple_dual_transformers = DualTransformer()(DualTransformer()(axelrod.WinStayLoseShift))
+        self.assert_dual_wrapper_correct(multiple_dual_transformers)
+
+    def assert_dual_wrapper_correct(self, player_class):
```
theref: Yeah I really like this test, great work!

drvinceknight: Thanks @theref :+1:

eric-s-s: i'd love to take credit, but my contribution to this particular test was pulling it out of the original testing code to make a function and adding `axl.seed(0)`. whoever wrote the original (`test_dual_somestrategy`) made a nice concise test that showed me exactly what `dual_transformer` needed to do.  if that was you, @theref , thanks!

axelrod/strategy_transformers.py
```diff
@@ -9,9 +9,12 @@
 import copy
 import inspect
 import random
+from typing import Any
```
marcharper: Please reorder the imports (alphabetical, std lib, then 3rd party, then project specific)

axelrod/strategy_transformers.py
```diff
@@ -57,6 +60,14 @@ def __init__(self, *args, **kwargs):
             else:
                 self.name_prefix = name_prefix
 
+        def __reduce__(self):
```
marcharper: Can we add a docstring here? I get that we're rebuilding the class but if I didn't know what was going on I wouldn't know until I read through the file, and this is a fairly complicated part of the library so I'd like to err on the side of more documentation.

axelrod/strategy_transformers.py
```diff
@@ -84,24 +95,25 @@ def __call__(self, PlayerClass):
...
+                    # played the opposite actions of its current history.
+                    flip_play_attributes(self)
+
+                if is_strategy_static(PlayerClass):
                     proposed_action = PlayerClass.strategy(opponent)
                 else:
                     proposed_action = PlayerClass.strategy(self, opponent)
 
+                if strategy_wrapper == dual_wrapper:
+                    # after dual_wrapper calls the strategy, it returns
```
marcharper: Capital After

axelrod/strategy_transformers.py
```diff
@@ -228,12 +322,47 @@ def dual_wrapper(player, opponent, proposed_action):
...
+
+    return proposed_action.flip()
+
+
+def flip_play_attributes(player: Player) -> None:
+    """
+    Flips all the attributes created by `player.play`:
+        - `player.history`,
+        - `player.cooperations`,
+        - `player.defectsions`,
```
marcharper: defections

marcharper:
@eric-s-s Please rebase onto master.

axelrod/strategy_transformers.py
eric-s-s: good to know!

eric-s-s:
@marcharper 

ummmmm.  i committed my changes, checkout master, pulled from upstream to master, checkout my_branch, typed `git rebase master` in my branch, and then followed all the conflict instructions (any resolve went to whatever was in master).  at first glance, all the changes i made seem to be there. but i have no idea about all the other 25-ish files changed.

marcharper:
I'm not sure what you actually did -- it looks like there are twice as many commits. When I tried to rebase again I got a ton of merge conflicts, so I suspect that you somehow rewrote the commits then pushed them to the head of the old branch, while picking up some old commits. Maybe you rebased master onto your branch?

I was able to squash all the commits, see #1106. I think that #1107 has the original commits, rebased.

eric-s-s:
yeah, i figured that something weird happened.  i did the steps i said.  i have no idea why it did that.  the examples i looked at on the github manpage looked so much nicer. :(

drvinceknight:
> should i have pushed up the rebased branch as a new PR?

Yeah I think that was the error, you should have `git push -f` git force pushed and just over written this branch or indeed just opened a second PR :)

eric-s-s:
@drvinceknight thanks.  

i reset the head, re-rebased it and am going to try that now.

marcharper:
I think this is fine to merge now and we can close the others.

drvinceknight:
I'm just going to re start the builds to triple check nothing got lost during the rebases.

drvinceknight:
Thanks @eric-s-s  :+1: Great work getting to the bottom of this. :+1:

eric-s-s:
it was my pleasure.  :)

Issue #1093, opened 2017-07-27
K86R Strategy


meatballs:
This is one of the strategies from [Axelrod's Second Tournament](http://axelrod.readthedocs.io/en/stable/reference/overview_of_strategies.html#axelrod-s-second-tournament) for which we have the [orginal Fortran code](https://github.com/Axelrod-Python/TourExec/blob/v0.3.0/src/strategies/k86r.f).

From my analysis of the fortran code, the strategy:

- Cooperates on the first two rounds
- For rounds 3 -7 inclusive, plays the same as the opponent's last move
- thereafter, applies the following logic
   - If its own previous move was C and the opponent has defected less than twice in the last 7 rounds, cooperate
   - If its own previous move was C and the opponent has defected twice or more in the last 7 rounds, defect
   - If its own previous move was D and the opponent has defected only once or not at all in the last 7 rounds, cooperate
   - If its own previous move was D and the opponent has defected more than once in the last 7 rounds, defect

The class to implement this strategy should be added to the [source code for the second tournament](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/axelrod_second.py) and its name should be `MoreGrofman`

buckbaskin:
@meatballs I spent some time today doing a first pass at implementing this strategy. I'd appreciate any feedback.

PR #1094, opened 2017-07-28
Update authors and K92R from table2 of paper

meatballs:
https://www.jstor.org/stable/173638 had some extra information compared with the existing docs.

The author in the current docs was taken from comment lines in the Fortran code. 

For [K71R](https://github.com/Axelrod-Python/TourExec/blob/v0.2.0/src/strategies/k71r.f), the code comment gives the author as 'James E Hill' whilst the paper says 'James E Hall.' I've taken the paper as the correct source.

I've also added the TitForTat class as the implementation for [K92R](https://github.com/Axelrod-Python/TourExec/blob/v0.2.0/src/strategies/K92R.f). That decision is based on two factors:

1. Looking at the code for K92R, it's clear that it is TitForTat
2. When ticking off the lists of strategies listed in our docs against those listed in the paper, there is no other strategy that K92R could be


meatballs:
I've also added a link to #1093 for K86R

docs/reference/overview_of_strategies.rst
```diff
@@ -141,29 +141,29 @@ repository.
...
-   "K80R_", "Robyn M Dawes and Mark Batel", Not Implemented
+   "K80R_", "Robyn M Dawes and Mark Batell", Not Implemented
    "K81R_", "Martyn Jones", "Not Implemented"
    "K82R_", "Robert A Leyland", "Not Implemented"
    "K83R_", "Paul E Black", "Not Implemented"
    "K84R_", "T Nicolaus Tideman and Paula Chieruzz", "Not Implemented"
-   "K85R_", "Rober B Falk and James M Langsted", "Not Implemented"
-   "K86R_", "Bernard Grofman", "Not Implemented"
+   "K85R_", "Robert B Falk and James M Langsted", "Not Implemented"
+   "K86R_", "Bernard Grofman", "`Issue 1093 <https://github.com/Axelrod-Python/Axelrod/issues/1093>`_"
```
drvinceknight: I think I'd prefer to just leave `Not Implemented`? This means that a reader has to click on it and read to find more information whereas the point of the table is just that it's not implemented yet.

docs/reference/overview_of_strategies.rst
```diff
@@ -141,29 +141,29 @@ repository.
...
-   "K85R_", "Rober B Falk and James M Langsted", "Not Implemented"
+   "K85R_", "Robert B Falk and James M Langsted", "Not Implemented"
    "K86R_", "Bernard Grofman", "Not Implemented"
    "K87R_", "E E H Schurmann",  "Not Implemented"
    "K88R_", "Scott Appold", "Not Implemented"
    "K89R_", "Gene Snodgrass", "Not Implemented"
    "K90R_", "John Maynard Smith", "Not Implemented"
    "K91R_", "Jonathan Pinkley", "Not Implemented"
-   "K92R_", "Anatol Rapoport", "Not Implemented"
+   "K92R_", "Anatol Rapoport", ":class:`TitForTat <axelrod.strategies.titfortat.TitForTat>`"
```
drvinceknight: Is there some confusion here (not on your part) with `KTITFORTATC`? I was the one who suggested Rapoport as the author for that (because he is the author of TfT) but perhaps we should change that to be `Unknown` as we don't know who wrote that file <https://github.com/Axelrod-Python/TourExec/blob/v0.2.0/src/strategies/KTitForTatC.f>?

I wonder if they were both run in `TourExec` in Axelrod's second tournament... Perhaps `KTITFORTATC` was not included in the original run but just there as an example? (Just wondering.)

drvinceknight: I've just read: https://github.com/Axelrod-Python/axelrod-fortran/issues/36 :+1:

Cool. Yeah so I'd suggest changing the author of `KTITFORTATC` to be `Unknown`

docs/reference/overview_of_strategies.rst
meatballs: I did this because otherwise we have no other reference point for the question "Which of the second tournament strategies do we still need to look at and reverse engineer"

drvinceknight: How about opening an "Reverse engineer axelrod-fortran stragegies" and keep track of that there?

So that issue would essentially be a list of strategies that get ticked etc...

With the idea that once a strategy is reverse engineered a further issue is opened (like you've done for #1093) for it to be "re-engineered" in python.

PR #1096, opened 2017-07-29
Add ability to pass match attributes in tournament

drvinceknight:
It is possible to pass a dictionary to `axelrod.Match` is
`match_attributes` which correspond to the information given to players.

This is motivated by wanting to reproduce Axelrod's second tournament in
which all matches were 1 of 5 values.

### DIFF
--- a/axelrod/tests/unit/test_match_generator.py
+++ b/axelrod/tests/unit/test_match_generator.py
@@ -127,14 +127,15 @@ class TestMatchGenerator(unittest.TestCase):
                                     game=test_game,
                                     repetitions=test_repetitions,
                                     turns=5,
-                                    match_attributes={"length": -1})
+                                    match_attributes={"length": float('inf')})
         match_params = rr.build_single_match_params()
         self.assertIsInstance(match_params, dict)
         self.assertEqual(match_params["turns"], 5)
         self.assertEqual(match_params["game"], test_game)
         self.assertEqual(match_params["prob_end"], None)
         self.assertEqual(match_params["noise"], 0)
-        self.assertEqual(match_params["match_attributes"], {"length": -1})
+        self.assertEqual(match_params["match_attributes"],
+                         {"length": float('inf')})
 
 
         # Check that can build a match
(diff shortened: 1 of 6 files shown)


axelrod/tests/unit/test_match_generator.py
```diff
@@ -122,6 +122,29 @@ def test_build_single_match_params_with_prob_end_and_turns(self):
         self.assertGreater(len(match), 0)
         self.assertLessEqual(len(match), 10)
 
+    def test_build_single_match_params_with_fixed_length_unknown(self):
+        rr = axelrod.MatchGenerator(players=self.players,
+                                    game=test_game,
+                                    repetitions=test_repetitions,
+                                    turns=5,
+                                    match_attributes={"length": -1})
```
marcharper: Is `length` -> `None` a better default than `-1`?

docs/tutorials/advanced/player_information.rst
```diff
@@ -0,0 +1,31 @@
+.. _player_information:
+
+Player information
+==================
+
+It is possible to determine what information players know about their matches.
+By default all known information is given.
+For example let us create a match with 5 turns between :code:`SteinAndRapoport`
+and :code:`Aternator`. The latter of these two always defects on the last 2
```
marcharper: Alternator

axelrod/tests/unit/test_match_generator.py
drvinceknight: For some reason I thought `-1` was the default already in place. It's not! Let me change it to `float('inf')` which is what is currently there. Good catch.

docs/tutorials/advanced/player_information.rst
drvinceknight: :+1:

axelrod/tests/unit/test_match_generator.py
marcharper: We should check the behavior of any strategies that use the match attributes...

drvinceknight: We already do this for the tests in the `matches`? (This `matches_attribute` argument is not new, it's used with `prob_end` matches, this PR is just adding the ability to get the tournament to pass it to the matches).

drvinceknight:
@marcharper with 1648ddf I've changed the `-1` to `float('inf')` which is the default we already use. For some reason I thought it was `-1`. No idea why.

marcharper:
Cool. I think it used to be -1.

drvinceknight:
> Cool. I think it used to be -1.

That must be why :)

PR #1097, opened 2017-07-29
Three new ZD strategies

marcharper:
Also:
* moved ZD strategies to their own file
* fixed a reset bug on NTitsForMTats
* various style cleanups


axelrod/strategies/zero_determinant.py
```diff
@@ -0,0 +1,322 @@
...
+        self.l = l
+        super().__init__()
+
+    def receive_match_attributes(self):
+        super().receive_match_attributes(self.phi, self.s, self.l)
+
+
+class ZDExtort3(LRPlayer):
+    """
+    An extortionate strategy from Press and Dyson's paper an extortion factor
```
drvinceknight: `with an extortion`

axelrod/tests/strategies/test_zero_determinant.py
```diff
@@ -0,0 +1,303 @@
...
+        'stochastic': True,
+        'makes_use_of': set(["game"]),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_four_vector(self):
+        expected_dictionary = {(C, C): 0.64, (C, D): 0.18, (D, C): 0.28,
```
drvinceknight: Can we write these as fractions here? (Just to avoid floating point errors)

axelrod/strategies/titfortat.py
```diff
@@ -772,3 +774,7 @@ def strategy(self, opponent: Player) -> Action:
             self.retaliate_count -= 1
             return D
         return C
+
+    def reset(self):
+        super().reset()
+        self.retaliate_count = 0
```
drvinceknight: This not tripping a test means we can probably improve the test...

marcharper: Hypothesis caught it

drvinceknight: Nice :+1:

drvinceknight: I've just written another one. Pushing it now :)

axelrod/tests/strategies/test_zero_determinant.py
marcharper: I took them from the reference like that. Can change if you want...

drvinceknight: Only to avoid the numerical manipulation carried out by the strategy to throw out numerical floating point errors at some point? (if it never fails then I'm completely fine with them being numeric, just trying to avoid a trivial fix in the future)

```
>>> import sympy as sym
>>> sym.nsimplify(.64), sym.nsimplify(0.18), sym.nsimplify(0.28)
(16/25, 9/50, 7/25)
```

drvinceknight:
Assuming appveyor is happy. :+1:

PR #1106, opened 2017-08-03
Squashed commits for refactor to make all strategies pickleable.

marcharper:


eric-s-s:
i looked through this to be sure.  it looks correct.

PR #1107, opened 2017-08-03
Eric-s-s orig

marcharper:


eric-s-s:
i looked through it twice.  it looks correct.  the latest commit with your requested changes is there, and on two look throughs i believe everything else is there.

does that travis-ci fail on a hypothesis test have anything to do with this code, or is that a separate issue?

drvinceknight:
> does that travis-ci fail on a hypothesis test have anything to do with this code, or is that a separate issue

I've reset that build. I expect it's nothing to do with your code.

Can you or @marcharper offer some guidance, what should we be doing with this PR versus #1106 and #1092?

eric-s-s:
i just tried resetting, re-rebasing and force pushing in #1092.  if that worked, i vote to merge that and close this.  if it didn't work, then i'd vote to merge this and close that.

Issue #1102, opened 2017-07-30
K76R Strategy


meatballs:
#This is one of the strategies from [Axelrod's Second Tournament](http://axelrod.readthedocs.io/en/stable/reference/overview_of_strategies.html#axelrod-s-second-tournament) for which we have the [orginal Fortran code](https://github.com/Axelrod-Python/TourExec/blob/v0.3.0/src/strategies/k76r.f).

A description of this strategy is given in Axelrod, R. (1980). More Effective Choice in the Prisoner’s Dilemma. Journal of Conflict Resolution, 24(3), 379-403, where it is known as 'TESTER':

> The rule is unusual in that it defects on the very first move in order to test the other's response. If the other player ever defects, it apologizes by cooperating and playing tit-for-tat for the rest of the game. Otherwise, it defects as much as possible subject to the constraint that the ratio of its defections to moves remains under 0.5, not counting the first defection.

From our analysis of the Fortran code, the strategy:
- Has a boolean state variable named 'PATSY' which is initialised as True.
  This variable represents whether the strategy currently 'believes' that its opponent can be exploited. i.e. it believes that its opponent is a [patsy](https://www.merriam-webster.com/dictionary/patsy).
- Always Defects on the first round. On subsequent rounds:
  - If PATSY is False, play the opponent's last move
  - If PATSY is True and the opponent's last move was D, set PATSY to False and Cooperate
  - If PATSY is True and the opponent's last move was C:
    - Cooperate unless the ratio of (total number of its own cooperations) / (total number of moves) is greater than or equal to 0.5, in which case Defect

drvinceknight:
Following a good discussion on this I agree with @meatballs. Note that the textual description mentions "not counting the first defection" to ensure that the ratio "remains" below 0.5. We are in agreement that this is equivalent to simply counting the total number of moves to ensure that the **current** ratio is below 0.5.

:+1:

dmanc:
I would like to work on this but there is a strategy in axelrod_second.py already called "Tester". It is somewhat different than the description given here so i'm not sure if I should modify it or just make a new class.

drvinceknight:
That's great @dmanc!

I suggest the class name for this strategy be `Gladstein` (the name of the author) but then include `Tester` in the docstring as one of the other names (also clarify in the docstring that this is based on reverse engineering the Fortran code from the second tournament).

dmanc:
[...]

PR #1110, opened 2017-08-04
Gladstein strategy

dmanc:
This is the implementation of the K76R Fortran strategy mentioned in #1102. The strategy was named Gladstein after the author's last name.

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -153,12 +153,12 @@ class Gladstein(Player):
     """
     Submitted to Axelrod's second tournament by David Gladstein.
 
-    This strategy is also known as Tester and is based of the reverse
+    This strategy is also known as Tester and is based on the reverse
     engineering of the Fortran strategies from Axelrod's second tournament.
 
     This strategy is a TFT variant that defects on the first round in order to
     test the opponent's response. If the opponent ever defects, the strategy
-    'apologies' by cooperating and then plays TFT for the rest of the game.
+    'apologizes' by cooperating and then plays TFT for the rest of the game.
     Otherwise, it defects as much as possible subject to the constraint that
     the ratio of its defections to moves remains under 0.5, not counting the
     first defection.
@@ -166,11 +166,12 @@ class Gladstein(Player):
     Names:
 
     - Gladstein: [Axelrod1980b]_
+    - Tester: [Axelrod1980b]_
     """
 
     name = "Gladstein"
     classifier = {
-        'memory_depth': 1,
+        'memory_depth': float('inf'),
         'stochastic': False,
         'makes_use_of': set(),
         'long_run_time': False,
@@ -183,7 +184,6 @@ class Gladstein(Player):
         super().__init__()
         # This strategy assumes the opponent is a patsy
         self.patsy = True
-        self.cooperation_ratio = 0
 
     def strategy(self, opponent: Player) -> Action:
         # Defect on the first move
@@ -196,8 +196,8 @@ class Gladstein(Player):
                 self.patsy = False
                 return C
             # Cooperate as long as the cooperation ratio is below 0.5
-            self.cooperation_ratio = self.cooperations / len(self.history)
-            if self.cooperation_ratio >= 0.5:
+            cooperation_ratio = self.cooperations / len(self.history)
+            if cooperation_ratio >= 0.5:
                 return D
             return C
         else:
@@ -207,4 +207,3 @@ class Gladstein(Player):
     def reset(self):
         super().reset()
         self.patsy = True
-        self.cooperation_ratio = 0
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -52,7 +52,7 @@ class Champion(Player):
         defection_prop = opponent.defections / len(opponent.history)
         if opponent.history[-1] == D:
             r = random.random()
-            if defection_prop > max(0.4, r):
+            if defection_prop >= max(0.4, r):
                 return D
         return C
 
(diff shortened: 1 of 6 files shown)


axelrod/strategies/axelrod_second.py
```diff
@@ -147,3 +147,64 @@ def strategy(self, opponent: Player) -> Action:
...
+class Gladstein(Player):
+    """
+    Submitted to Axelrod's second tournament by David Gladstein.
+
+    This strategy is also known as Tester and is based of the reverse
+    engineering of the Fortran strategies from Axelrod's second tournament.
+
+    This strategy is a TFT variant that defects on the first round in order to
+    test the opponent's response. If the opponent ever defects, the strategy
+    'apologies' by cooperating and then plays TFT for the rest of the game.
```
meatballs: typo - should be 'apologizes'

axelrod/strategies/axelrod_second.py
```diff
@@ -147,3 +147,64 @@ def strategy(self, opponent: Player) -> Action:
     def reset(self):
         super().reset()
         self.is_TFT = False
+
+
+class Gladstein(Player):
+    """
+    Submitted to Axelrod's second tournament by David Gladstein.
+
+    This strategy is also known as Tester and is based of the reverse
```
meatballs: typo - should be 'is based on'

axelrod/strategies/axelrod_second.py
```diff
@@ -147,3 +147,64 @@ def strategy(self, opponent: Player) -> Action:
...
+    This strategy is a TFT variant that defects on the first round in order to
+    test the opponent's response. If the opponent ever defects, the strategy
+    'apologies' by cooperating and then plays TFT for the rest of the game.
+    Otherwise, it defects as much as possible subject to the constraint that
+    the ratio of its defections to moves remains under 0.5, not counting the
+    first defection.
+
+    Names:
+
+    - Gladstein: [Axelrod1980b]_
```
meatballs: Could you add the name Tester here, please?

axelrod/strategies/axelrod_second.py
```diff
@@ -147,3 +147,64 @@ def strategy(self, opponent: Player) -> Action:
...
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self) -> None:
+        super().__init__()
+        # This strategy assumes the opponent is a patsy
+        self.patsy = True
+        self.cooperation_ratio = 0
```
meatballs: There is no need to maintain this ratio between calls to the play method. The original fortran strategy had to do that, but we pass the entire history to the player at each turn, so it can always be calculated within the play method.

axelrod/strategies/axelrod_second.py
```diff
@@ -147,3 +147,64 @@ def strategy(self, opponent: Player) -> Action:
...
+        if not self.history:
+            return D
+        # Is the opponent a patsy?
+        if self.patsy:
+            # If the opponent defects, apologize and play TFT.
+            if opponent.history[-1] == D:
+                self.patsy = False
+                return C
+            # Cooperate as long as the cooperation ratio is below 0.5
+            self.cooperation_ratio = self.cooperations / len(self.history)
```
meatballs: As above, this can be calculated directly from the history without the need for an instance variable. For example:_ratio = 

```
from collections import Counter

collection_ratio = Counter(self.history)[C] / len(self.history)
```

meatballs: Actually, I was forgetting we have that count already built into the Player class, so it becomes even simpler:

```
cooperation_ratio = self.cooperations / len(self.history)
```

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -117,3 +116,41 @@ def test_strategy(self):
...
+        actions = [(D, D), (C, D), (D, D), (D, D), (D, D)]
+        self.versus_test(axelrod.Defector(), expected_actions=actions,
+                         attrs={'patsy': False, 'cooperation_ratio': 0.0})
+
+        # Ratio is 1/3 when MockPlayer defected for the first time.
+        opponent = axelrod.MockPlayer(actions=[C, C, C, D, D])
+        actions = [(D, C), (C, C), (D, C), (C, D), (C, D)]
+        self.versus_test(opponent, expected_actions=actions,
+                         attrs={'patsy': False,
+                                'cooperation_ratio': 0.3333333333333333})
```
drvinceknight: Could we have this as `'cooperation_ratio': 1/3`.

axelrod/strategies/axelrod_second.py
```diff
@@ -147,3 +147,64 @@ def strategy(self, opponent: Player) -> Action:
...
+    first defection.
+
+    Names:
+
+    - Gladstein: [Axelrod1980b]_
+    """
+
+    name = "Gladstein"
+    classifier = {
+        'memory_depth': 1,
```
drvinceknight: This strategy has infinite memory depth `: float('inf')` as it keeps track of it's cooperations.

(You'll need to adjust the corresponding doctest.)

axelrod/strategies/axelrod_second.py
marcharper: Using `self.cooperations` is better, recomputing each time is O(n^2).

dmanc:
I added the requested changes. I had to change a test in test_meta.py for `TestMetaMajorityLongMemory` since it was giving me errors after changing the memory depth of the strategy.

meatballs:
Could you also amend the [overview of strategies doc](https://github.com/Axelrod-Python/Axelrod/blob/master/docs/reference/overview_of_strategies.rst) to show that K76R has been implemented by this new class?

drvinceknight:
Thanks for this @dmanc!

PR #1111, opened 2017-08-04
Alter inequality condition to match docstring

meatballs:
The docstring text (from the original paper) and the K61R code in axelrod-fortran both test whether the cooperation rate is less than 0.6.

We use the defection rate, so the equivalent condition should test whether it is greater than or equal to 0.6

In the current implementation, it is simply testing whether the defection rate is greater than 0.6


drvinceknight:
Looks good assuming CI :+1:

PR #1113, opened 2017-08-05
Fix a small bug in Gladstein.

### LINKED ISSUE #1112: Newly implemented Gladstein isn't correct
Woops we should have checked this on @dmanc's #1110. Using `axelrod_fortran` (https://github.com/Axelrod-Python/axelrod-fortran):

```python
>>> import axelrod as axl
>>> axl_gladstein = axl.Gladstein()
>>> gladstein = axlf.Player('k76r')
>>> opponent = axl.AntiTitForTat()
>>> axl.Match((axl_gladstein, opponent), turns=5).play()
[(D, C), (C, C), (D, D), (C, C), (C, D)]
>>> axl.Match((gladstein, opponent), turns=5).play()
[(D, C), (C, C), (C, D), (C, D), (D, D)]
```

In the 3rd round the strategy implemented by #1102 is defecting but the fortran strategy (https://github.com/Axelrod-Python/TourExec/blob/master/src/strategies/k76r.f) cooperates.

drvinceknight:
I have tested this locally with axelrod_fortran and this strategy now
agrees with the fortran version against all basic strategies.

Closes #1112

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -197,7 +197,7 @@ class Gladstein(Player):
                 return C
             # Cooperate as long as the cooperation ratio is below 0.5
             cooperation_ratio = self.cooperations / len(self.history)
-            if cooperation_ratio >= 0.5:
+            if cooperation_ratio > 0.5:
                 return D
             return C
         else:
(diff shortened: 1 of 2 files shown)


drvinceknight:
I have modified the tests that were there and also added the test for the known behaviour from axelrod fortran written up at #1112.

PR #1115, opened 2017-08-07
Add MoreGraaskamp strategy

### LINKED ISSUE #1051: Specific strategy request: Graaskamp
This strategy is one of the strategies from Axelrod's original tournaments. It is described here http://axelrod.readthedocs.io/en/stable/reference/overview_of_strategies.html#graaskamp as:

1. Play Tit For Tat for the first 50 rounds;
2. Defects on round 51;
4. A check is then made to see if the opponent is playing randomly in which case it defects for the rest of the game;
5. The strategy also checks to see if the opponent is playing Tit For Tat or another strategy from a preliminary tournament called ‘Analogy’. If so it plays Tit For Tat. If not it cooperates and randomly defects every 5 to 15 moves.

Step 4 can be implemented using a chi squared test similarly to the Stein and Rapaport strategy which is implemented here: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/axelrod_first.py#L496

Step 5 is a bit tricky as it refers to a strategy called "Analogy". I have not been able to find any information describing this strategy. I suggest that anyone who tries to work on this does their best to find that information but if there is non available let us simply ignore that part of the check **but clearly point that out in the documentation of the strategy**.

SplitInfinity:
This pull request adds an implementation of the Graaskamp strategy from
Axelrod's second tournament.

Resolves #1051

### DIFF
--- a/axelrod/strategies/graaskamp.py
+++ b/axelrod/strategies/graaskamp.py
@@ -24,7 +24,7 @@ class Mode(Enum):
 class MoreGraaskamp(Player):
     """
     This strategy can be described as follows:
-
+    
     1. Play tit for tat for the first 50 rounds.
     2. Defect on round 51.
     3. Play 5 further rounds of tit for tat.
@@ -34,7 +34,8 @@ class MoreGraaskamp(Player):
         2 - If the opponent's score is greater than 135 and moves 52 through 55 were
             [C, D, C, D] or [D, D, D, D], play tit for tat thereafter starting with
             cooperate
-        3 - If the opponent's score is less than or equal to 135, always defect hereafter
+        3 - If the opponent's score is less than or equal to 135, always defect 
+            hereafter
         4 - If the opponent's score is greater than 135 and moves 52 through 56 were
             [C, C, D, C, C], cooperate until and including move 118, and tit for tat
             thereafter
@@ -44,7 +45,7 @@ class MoreGraaskamp(Player):
     - GRASR: [Axelrod1993]_
     """
 
-    # These are various properties for the
+    # These are various properties of the strategy
     name = 'MoreGraaskamp'
     classifier = {
         'memory_depth': 5,
--- a/axelrod/strategies/graaskamp.py
+++ b/axelrod/strategies/graaskamp.py
@@ -6,20 +6,6 @@ from axelrod._strategy_utils import calculate_scores
 from enum import Enum
 
 C, D = Action.C, Action.D
-
-class Mode(Enum):
-    """
-    Enumerated type for the four "modes" for MoreGraaskamp after round 56:
-        1. Tit for tat starting with cooperate, defect randomly every 5 - 15 turns
-        2. Tit for tat for rest of the game starting with cooperate
-        3. Defect for the rest of the game
-        4. Cooperate until move 118 and then transition to tit for tat
-    """
-    Unknown = 0
-    DefectRandomly = 1
-    TitForTat = 2
-    AlwaysDefect = 3
-    CooperateThenTitForTat = 4
         
 class MoreGraaskamp(Player):
     """
@@ -29,26 +15,27 @@ class MoreGraaskamp(Player):
     2. Defect on round 51.
     3. Play 5 further rounds of tit for tat.
     4. In round 57, select between four different "modes" for the rest of the game:
-        1 - If none of the below modes are selected, cooperate and defect randomly
-            every 5 to 15 turns
-        2 - If the opponent's score is greater than 135 and moves 52 through 55 were
-            [C, D, C, D] or [D, D, D, D], play tit for tat thereafter starting with
-            cooperate
-        3 - If the opponent's score is less than or equal to 135, always defect 
-            hereafter
-        4 - If the opponent's score is greater than 135 and moves 52 through 56 were
-            [C, C, D, C, C], cooperate until and including move 118, and tit for tat
-            thereafter
+        defect_randomly - If none of the below modes is selected, cooperate and
+            defect randomly every 5 to 15 turns
+        tit_for_tat - If the opponent's score is greater than 135 and moves 52
+            through 55 were [C, D, C, D] or [D, D, D, D], play tit for tat
+            thereafter starting with cooperate
+        always_defect - If the opponent's score is less than or equal to 135,
+            always defect hereafter
+        cooperate_then_tit_for_tat - If the opponent's score is greater than 135
+            and moves 52 through 56 were [C, C, D, C, C], cooperate until and
+            including move 118, and tit for tat thereafter
 
     Names:
 
     - GRASR: [Axelrod1993]_
+    - Graaskamp: [Axelrod1980b]_
     """
 
     # These are various properties of the strategy
     name = 'MoreGraaskamp'
     classifier = {
-        'memory_depth': 5,
+        'memory_depth': float('inf'),
         'stochastic': True,
         'makes_use_of': set(),
         'long_run_time': False,
@@ -59,7 +46,12 @@ class MoreGraaskamp(Player):
 
     def __init__(self) -> None:
         super().__init__()
-        self.mode = Mode.Unknown
+        self.mode = 'unknown'
+        self.defect_round = 0
+
+    def reset(self):
+        super().reset()
+        self.mode = 'unknown'
         self.defect_round = 0
 
     # The strategy itself
@@ -69,21 +61,21 @@ class MoreGraaskamp(Player):
         # Tit for tat for first 50 rounds
         if round_number == 1:
             return C
-        elif round_number < 51:
+        if round_number < 51:
             return opponent.history[-1]
         # Defect on round 51
-        elif round_number == 51:
+        if round_number == 51:
             return D
         # Play 5 further rounds of tit for tat
-        elif round_number < 57:
+        if round_number < 57:
             return opponent.history[-1]
-        elif round_number == 57:
+        if round_number == 57:
             game = self.match_attributes['game']
 
             self_score, opponent_score = calculate_scores(self, opponent, game)
 
             if opponent_score <= 135:
-                self.mode = Mode.AlwaysDefect
+                self.mode = 'always_defect'
                 return D
             else:
                 # Opponent's moves 52 through 56
@@ -92,31 +84,29 @@ class MoreGraaskamp(Player):
                 prev_four_opponent_actions = opponent.history[-5:-1]
 
                 if last_five_opponent_actions == [C, C, D, C, C]:
-                    self.mode = Mode.CooperateThenTitForTat
+                    self.mode = 'cooperate_then_tit_for_tat'
                 elif prev_four_opponent_actions == [C, D, C, D] or prev_four_opponent_actions == [D, D, D, D]:
-                    self.mode = Mode.TitForTat
+                    self.mode = 'tit_for_tat'
                 else:
-                    self.mode = Mode.DefectRandomly
+                    self.mode = 'defect_randomly'
                     self.defect_round = round_number + randrange(5, 15) + 1
 
                 return C
 
         # Beyond round 57, obey the rules of the mode entered during round 57
-        else:
-            if self.mode == Mode.TitForTat:
-                return opponent.history[-1]
-            elif self.mode == Mode.AlwaysDefect:
-                return D
-            elif self.mode == Mode.CooperateThenTitForTat:
-                if round_number == 118:
-                    self.mode = Mode.TitForTat
+        if self.mode == 'tit_for_tat':
+            return opponent.history[-1]
+        if self.mode == 'always_defect':
+            return D
+        if self.mode == 'cooperate_then_tit_for_tat':
+            if round_number == 118:
+                self.mode = 'tit_for_tat'
 
-                return C
-            else:
-                # self.mode must be Mode.DefectRandomly
-                if round_number == self.defect_round:
-                    self.defect_round = round_number + randrange(5, 15) + 1
-                    return D
-                else:
-                    return C
-            
+            return C
+
+        # self.mode must be 'defect_randomly'
+        if round_number == self.defect_round:
+            self.defect_round = round_number + randrange(5, 15) + 1
+            return D
+
+        return C
--- a/axelrod/strategies/graaskamp.py
+++ b/axelrod/strategies/graaskamp.py
@@ -1,0 +1,121 @@
+from axelrod.action import Action, actions_to_str
+from axelrod.player import Player
+from axelrod.random_ import randrange
+from axelrod._strategy_utils import calculate_scores
+
+from enum import Enum
+
+C, D = Action.C, Action.D
+
+class Mode(Enum):
+    """
+    Enumerated type for the four "modes" for MoreGraaskamp after round 56:
+        1. Tit for tat starting with cooperate, defect randomly every 5 - 15 turns
+        2. Tit for tat for rest of the game starting with cooperate
+        3. Defect for the rest of the game
+        4. Cooperate until move 118 and then transition to tit for tat
+    """
+    Unknown = 0
+    DefectRandomly = 1
+    TitForTat = 2
+    AlwaysDefect = 3
+    CooperateThenTitForTat = 4
+        
+class MoreGraaskamp(Player):
+    """
+    This strategy can be described as follows:
+
+    1. Play tit for tat for the first 50 rounds.
+    2. Defect on round 51.
+    3. Play 5 further rounds of tit for tat.
+    4. In round 57, select between four different "modes" for the rest of the game:
+        1 - If none of the below modes are selected, cooperate and defect randomly
+            every 5 to 15 turns
+        2 - If the opponent's score is greater than 135 and moves 52 through 55 were
+            [C, D, C, D] or [D, D, D, D], play tit for tat thereafter starting with
+            cooperate
+        3 - If the opponent's score is less than or equal to 135, always defect hereafter
+        4 - If the opponent's score is greater than 135 and moves 52 through 56 were
+            [C, C, D, C, C], cooperate until and including move 118, and tit for tat
+            thereafter
+
+    Names:
+
+    - GRASR: [Axelrod1993]_
+    """
+
+    # These are various properties for the
+    name = 'MoreGraaskamp'
+    classifier = {
+        'memory_depth': 5,
+        'stochastic': True,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.mode = Mode.Unknown
+        self.defect_round = 0
+
+    # The strategy itself
+    def strategy(self, opponent: Player) -> Action:
+        round_number = len(self.history) + 1
+
+        # Tit for tat for first 50 rounds
+        if round_number == 1:
+            return C
+        elif round_number < 51:
+            return opponent.history[-1]
+        # Defect on round 51
+        elif round_number == 51:
+            return D
+        # Play 5 further rounds of tit for tat
+        elif round_number < 57:
+            return opponent.history[-1]
+        elif round_number == 57:
+            game = self.match_attributes['game']
+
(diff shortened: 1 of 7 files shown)


axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Names:
+
+    - GRASR: [Axelrod1993]_
+    """
+
+    # These are various properties of the strategy
+    name = 'MoreGraaskamp'
+    classifier = {
+        'memory_depth': 5,
```
marcharper: The `memory_depth` is `float('inf')` (because of the use of the total score)

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
...
+        self.defect_round = 0
+
+    # The strategy itself
+    def strategy(self, opponent: Player) -> Action:
+        round_number = len(self.history) + 1
+
+        # Tit for tat for first 50 rounds
+        if round_number == 1:
+            return C
+        elif round_number < 51:
```
marcharper: These `elif` can all be `if` because of the returns

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
...
+                if round_number == 118:
+                    self.mode = Mode.TitForTat
+
+                return C
+            else:
+                # self.mode must be Mode.DefectRandomly
+                if round_number == self.defect_round:
+                    self.defect_round = round_number + randrange(5, 15) + 1
+                    return D
+                else:
```
marcharper: This `else` is unnecessary

axelrod/tests/strategies/test_graaskamp.py
```diff
@@ -0,0 +1,196 @@
...
+from .test_player import TestPlayer
+
+C, D = axelrod.Action.C, axelrod.Action.D
+
+class TestMoreGraaskamp(TestPlayer):
+
+    name = 'MoreGraaskamp'
+    player = axelrod.MoreGraaskamp
+    expected_classifier = {
+        'memory_depth': 5,
```
marcharper: `float('inf')`

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
+from axelrod.action import Action, actions_to_str
+from axelrod.player import Player
+from axelrod.random_ import randrange
+from axelrod._strategy_utils import calculate_scores
+
+from enum import Enum
+
+C, D = Action.C, Action.D
+
+class Mode(Enum):
```
drvinceknight: Can we remove this class and just use a string as the corresponding attribute in the player class.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
...
+        1 - If none of the below modes are selected, cooperate and defect randomly
+            every 5 to 15 turns
+        2 - If the opponent's score is greater than 135 and moves 52 through 55 were
+            [C, D, C, D] or [D, D, D, D], play tit for tat thereafter starting with
+            cooperate
+        3 - If the opponent's score is less than or equal to 135, always defect 
+            hereafter
+        4 - If the opponent's score is greater than 135 and moves 52 through 56 were
+            [C, C, D, C, C], cooperate until and including move 118, and tit for tat
+            thereafter
```
drvinceknight: We would then need to move the mode descriptions to here.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
...
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.mode = Mode.Unknown
```
drvinceknight: `self.mode = "unknown"`

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
...
+        # Play 5 further rounds of tit for tat
+        elif round_number < 57:
+            return opponent.history[-1]
+        elif round_number == 57:
+            game = self.match_attributes['game']
+
+            self_score, opponent_score = calculate_scores(self, opponent, game)
+
+            if opponent_score <= 135:
+                self.mode = Mode.AlwaysDefect
```
drvinceknight: `self.mode = "always_defect"`

drvinceknight: Same for other modes.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
...
+            cooperate
+        3 - If the opponent's score is less than or equal to 135, always defect 
+            hereafter
+        4 - If the opponent's score is greater than 135 and moves 52 through 56 were
+            [C, C, D, C, C], cooperate until and including move 118, and tit for tat
+            thereafter
+
+    Names:
+
+    - GRASR: [Axelrod1993]_
```
drvinceknight: Can you also add:

```
- Graaskamp: [Axelrod1980b]_
```

axelrod/tests/strategies/test_graaskamp.py
```diff
@@ -0,0 +1,196 @@
+"""Tests for the Graaskamp strategies."""
+
+import axelrod
+from .test_player import TestPlayer
+
+C, D = axelrod.Action.C, axelrod.Action.D
+
+class TestMoreGraaskamp(TestPlayer):
```
drvinceknight: These tests look good but can you add an `attrs` check (for each) which verifies the strategy is in the mode it should be. An example here: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/strategies/test_oncebitten.py#L38

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,122 @@
...
+        
+class MoreGraaskamp(Player):
+    """
+    This strategy can be described as follows:
+    
+    1. Play tit for tat for the first 50 rounds.
+    2. Defect on round 51.
+    3. Play 5 further rounds of tit for tat.
+    4. In round 57, select between four different "modes" for the rest of the game:
+        1 - If none of the below modes are selected, cooperate and defect randomly
```
meatballs: typo: should be 'is selected' rather than 'are selected'

drvinceknight:
> Thanks for this, I have tested it locally and it agrees with the implemented strategy.

Actually my test didn't pick this strategy up because it's stochastic. Once you've made the requested changes we will check the behaviour as best as possible to ensure it matches up. One way to do this would be to confirm the fingerprints are the same. Here's the Fortran code's fingerprint: https://github.com/Axelrod-Python/Axelrod-fingerprint#grasr

The code to run this would be:

```
>>> import axelrod as axl
>>> fp = axl.AshlockFingerprint(axl.MoreGrasskamp)
>>> fp.fingerprint(turns=200, repetitions=100)
>>> fp.plot();
```

They should give similar images. If they don't that indicates there's something not quite right.

marcharper:
Was there something wrong with the enum?

SplitInfinity:
@marcharper 

I don't think so. @drvinceknight requested that I replace it with a simple string member that records the mode, so that's what I did.

drvinceknight:
> Was there something wrong with the enum?

I didn't see what advantage it offered to just a string for the mode (which is akin to what is done elsewhere). Less code is good code :)

Thanks for all the modifications @SplitInfinity, I'm going to run the fingerprint locally and will post it here :+1:

drvinceknight:
It looks like the logic must be a bit off. Here is the fingerprint of the fortran code:



Here is a reproduced one (over 20 repetitions):

download 2

SplitInfinity:
@drvinceknight Do the visual differences between the fingerprints give any hints as to where the differences in logic are?

drvinceknight:
> @drvinceknight Do the visual differences between the fingerprints give any hints as to where the differences in logic are?

Ish... These fingerprints aren't great if I'm honest (not their implementation but the theory behind them), they're implemented directly from the literature.

So the area where they seem to disagree is for x >.5 and x>y>.25 (ish). In that area: `grasr` seem to be getting a higher score than `MoreGrassman`. This would indicate that in that area `grasr` is defecting more than `MoreGrassman`.

The fingerprints are calculated by pitting the strategies `grasr` and `MoreGrassman` against a given strategy for every (x, y) with the following definition (http://axelrod.readthedocs.io/en/stable/tutorials/further_topics/fingerprinting.html):

- x of the time the strategy cooperates
- y of the time the strategy defects
- 1-x-y of the time the strategy plays Tit For Tat

There is an interpretation for when x + y >= 1 which (it looks like is where our problem region is):

- 1 - x of the time the strategy cooperates
- 1 - y of the time the strategy defects
- 1 - x - y of the time the strategy plays the opposite of Tit For Tat

**TLDR**

I THINK (but these fingerprints are hard to interpret) that when the opponent defects a lot (because it's mainly playing a defector (1-x is low) or the opposite of TfT) `MoreGrassman` is not defecting enough.

drvinceknight:
Another (better) clue:

```python
>>> for opponent in axl.basic_strategies:
...     turns = 500
...     matches = [axl.Match((grassr, opponent()), turns=turns),
...                axl.Match((moreG, opponent()), turns=turns)]
...     for match in matches:
...         axl.seed(0)
...         match.play()
...     print([match.final_score_per_turn()[0] for match in matches], opponent.name)
[2.492, 1.726] Alternator
[4.692, 4.692] Anti Tit For Tat
[4.684, 4.684] Bully
[3.176, 3.156] Cooperator
[2.548, 2.502] Cycler DC
[0.998, 0.998] Defector
[1.204, 1.204] Grudger
[1.15, 1.15] Suspicious Tit For Tat
[2.994, 2.918] Tit For Tat
[2.526, 1.752] Win-Shift Lose-Stay
[2.164, 1.812] Win-Stay Lose-Shift
```

Seems to suggest that there is a discrepancy against `Win-Shift Lose-Stay`.

Confirmation:

```python
>>> import numpy as np
>>> repetitions = 50
>>> match = axl.Match((grassr, axl.WinShiftLoseStay()), turns=turns)
>>> scores = []
>>> for seed in range(repetitions):
...     axl.seed(seed)
...     match.play()
...     scores.append(match.final_score_per_turn()[0])
>>> np.mean(scores)
2.5259999999999998
``` 

```python
>>> match = axl.Match((moreG, axl.WinShiftLoseStay()), turns=turns)
>>> scores = []
>>> for seed in range(repetitions):
...     axl.seed(seed)
...     match.play()
...     scores.append(match.final_score_per_turn()[0])
>>> np.mean(scores)
1.7615599999999998
```

For reference here is the source code for `WinShiftLoseStay`: http://axelrod.readthedocs.io/en/stable/_modules/axelrod/strategies/memoryone.html#WinShiftLoseStay. It's not terribly well described there but:

1. It starts by defecting;
2. If it cooperates and the opponent defects it cooperates in the next round
3. If it defects and the opponent cooperates it cooperates in the next round
4. If both cooperate it defects
5. If both defect it defects

drvinceknight:
Here's the confirmation against `Alternator`:

```python
>>> match = axl.Match((grassr, axl.Alternator()), turns=turns)
>>> scores = []
>>> for seed in range(repetitions):
...     axl.seed(seed)
...     match.play()
...     scores.append(match.final_score_per_turn()[0])
>>> np.mean(scores)
2.4920000000000004
```

```python
>>> match = axl.Match((moreG, axl.Alternator()), turns=turns)
>>> scores = []
>>> for seed in range(repetitions):
...     axl.seed(seed)
...     match.play()
...     scores.append(match.final_score_per_turn()[0])
>>> np.mean(scores)
1.7347600000000003
```

drvinceknight:
The fact that the scores seem to be relatively similar against `CyclerDC` which is essentially `Alternator` but with a different initial action (starts with `D`) is probably a big clue:

```python
>>> match = axl.Match((grassr, axl.CyclerDC()), turns=turns)
>>> scores = []
>>> for seed in range(repetitions):
...     axl.seed(seed)
...     match.play()
...     scores.append(match.final_score_per_turn()[0])
>>> np.mean(scores)
2.5445199999999999
```

```python
>>> match = axl.Match((moreG, axl.CyclerDC()), turns=turns)
>>> scores = []
>>> for seed in range(repetitions):
...     axl.seed(seed)
...     match.play()
...     scores.append(match.final_score_per_turn()[0])
>>> np.mean(scores)
2.5019999999999993
```

drvinceknight:
With 100 reps (I had it running, feel like I might as well throw it in here):

download 3

SplitInfinity:
I'm trying to run these little snippets myself so I can change the number of turns and figure out when the scores start to differ, but I'm having some trouble using the `axelrod_fortran` module. This is what happens when I try to import it in the interpreter after installing it with `pip install axelrod-fortran`:

`OSError: libstrategies.so: cannot open shared object file: No such file or directory`

Are there additional setup steps I'm not aware of?

meatballs:
@SplitInfinity We've not written much in the way of documentation for this, but you need to have compiled and installed the fortran code from https://github.com/Axelrod-Python/TourExec

The axelrod-fortran library is a Python wrapper for the compiled result.

Note - at present, this isn't available on Windows

marcharper:
@SplitInfinity I can give you some sample outputs of the Fortran strategy for testing purposes, say every opponent history of length 10 and the response? Would that help?

SplitInfinity:
@marcharper 

The first 55 or so turns of Graaskamp are essentially tit for tat so I don't expect the issue to be there. Round 57 and the ones after are more likely to reveal the issue, since the strategy enters one of the four modes at that point. To start with, could you send moves made by the Fortran strategy in rounds 1 - 100 against an Alternator? Thanks in advance!

marcharper:
```
import axelrod as axl
import axelrod_fortran as axlf
player = axlf.Player("grasr")
opponent = axl.Alternator()
m = axl.Match((player, opponent), turns=100)
m.play()
print(m.result)
```
```
[(C, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (C, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D), (D, C), (C, D)]

**Comment by marcharper** [comment]
[This gist](https://gist.github.com/marcharper/eab8ad75c11f9da5e4e32004144a1040) has matches with many more opponents.

**Comment by drvinceknight** [comment]
Keep in mind that Graaskamp is stochastic, trying to match up to exact interactions is perhaps tricky but looking at the limiting scores in my snippets (over repetitions of matches) is a good indication. Although you can't reproduce the fortran ones locally at least you know what you're aiming for with `MoreGraaskamp`.

**Comment by marcharper** [comment]
I can fix the random seed if that will help

**Comment by drvinceknight** [comment]
> I can fix the random seed if that will help

It would do no harm but still wouldn't allow for a direct comparison.

The fortran strategies don't necessarily follow the same stochastic path (a random number is sampled every turns regardless of whether or not it is used for example). So really the scores are the best indication we can get.

**Comment by drvinceknight** [comment]
> I suggest using my matches above as an indication of what MoreGraaskamp should score.

This code snippet (and similar above) does not need `axelrod_fortran`:

```python
>>> import axelrod as axl
>>> moreG = axl.MoreGraaskamp()
>>> match = axl.Match((moreG, axl.CyclerDC()), turns=turns)
>>> scores = []
>>> for seed in range(repetitions):
...     axl.seed(seed)
...     match.play()
...     scores.append(match.final_score_per_turn()[0])
>>> np.mean(scores)
2.5019999999999993
```

marcharper:
This one has been stale for a while -- is there a way we can make progress? Or merge (with a docstring that it's not exactly correct as is) and open an issue to figure out the mismatch?

drvinceknight:
> is there a way we can make progress? 

Yes: one of us can take the time to go through the Fortran and the PR and offer guidance and help.

> Or merge (with a docstring that it's not exactly correct as is) and open an issue to figure out the mismatch?

No. We should not do that.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,112 @@
...
+                # Opponent's moves 52 through 55
+                prev_four_opponent_actions = opponent.history[-5:-1]
+
+                if last_five_opponent_actions == [C, C, D, C, C]:
+                    self.mode = 'cooperate_then_tit_for_tat'
+                elif prev_four_opponent_actions == [C, D, C, D] or prev_four_opponent_actions == [D, D, D, D]:
+                    self.mode = 'tit_for_tat'
+                else:
+                    self.mode = 'defect_randomly'
+                    self.defect_round = round_number + randrange(5, 15) + 1
```
marcharper: This line may be slightly off: the Fortran code has `N = RANDO * 10.0 + 5.0`. If I compare `randrange(5, 15) + 1` with `5 + int(10 * random.random())` there seems to be an off by one shift.

Can we try that (also change the instance below), and also change to `defect_round` `rounds_until_defection` and count down by one each round rather than trying to compute what future round to defect on? I know it may be equivalent but it's closer to the original implementation, and a little easier to match them up.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,112 @@
...
+
+                return C
+
+        # Beyond round 57, obey the rules of the mode entered during round 57
+        if self.mode == 'tit_for_tat':
```
marcharper: Can we add comments for the original enum here? I.e. this mode `IGAME = 2` for example.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,112 @@
...
+            if opponent_score <= 135:
+                self.mode = 'always_defect'
+                return D
+            else:
+                # Opponent's moves 52 through 56
+                last_five_opponent_actions = opponent.history[-5:]
+                # Opponent's moves 52 through 55
+                prev_four_opponent_actions = opponent.history[-5:-1]
+
+                if last_five_opponent_actions == [C, C, D, C, C]:
```
marcharper: If the random issue doesn't fix everything my guess is that there is a transcription error in this line or the next few compared to the Fortran:

```
31    IF (NMOV(1) - 3) 12, 35, 12
35    IF (NMOV(3) - 3) 12, 16, 12
32    IF (NMOV(1) - 5) 12, 33, 12
33    IF (NMOV(3) - 5) 12, 16, 12
30    IF (NMOV(1) - 2) 12, 34, 12
34    IF (NMOV(3) - 4) 12, 40, 12
40 IF (NMOV(4) - 2) 12, 41, 12
```
In particular, it seems like there is at least one more way to get to the `round_number == 118` check (the jump to line 41 on the last line above) than is currently possible with the python implementation.

drvinceknight:
@SplitInfinity just checking if you were going to work on this anymore? No problem if you've got other things on. I was thinking of finding a moment to work on this, I'd fork from your branch so when (eventually) merged in, all your commits would be there and it would auto merge this PR (which we can leave open until then). :) :+1:

SplitInfinity:
@drvinceknight Probably not :(
Sorry guys!

drvinceknight:
> Probably not :(
> Sorry guys!

No problem at all, I'll pick it up, thanks for your efforts on this! :)

drvinceknight:
I doubt I'm going to find time to come around to this so am closing, if anyone would like to pick it up please do.

PR #1118, opened 2017-08-13
TidemanAndChieruzzi strategy

### LINKED ISSUE #1105: Specific strategy: Tideman and Chieruzzi
This is one of the remaining strategies from the first tournament and is described here: http://axelrod.readthedocs.io/en/stable/reference/overview_of_strategies.html#tideman-and-chieruzzi:

This strategy begins by playing Tit For Tat and then things get slightly complicated:

1. Every run of defections played by the opponent increases the number of defections that this strategy retaliates with by 1.
2. The opponent is given a ‘fresh start’ if:
        - it is 10 points behind this strategy
        - and it has not just started a run of defections
        - and it has been at least 20 rounds since the last ‘fresh start’
        - and there are more than 10 rounds remaining in the tournament
        - and the total number of defections differs from a 50-50 random sample by at least 3.0 standard deviations.

A ‘fresh start’ is a sequence of two cooperations followed by an assumption that the game has just started (everything is forgotten).

dmanc:
This is the implementation of Tideman and Chieruzzi's strategy from Axelrod's first tournament. 
Resolves #1105.

### DIFF
--- a/axelrod/strategies/axelrod_first.py
+++ b/axelrod/strategies/axelrod_first.py
@@ -589,3 +589,135 @@ class SteinAndRapoport(Player):
     def reset(self):
         super().reset()
         self.opponent_is_random = False
+
+
+class TidemanAndChieruzzi(Player):
+    """
+    This strategy begins by playing Tit For Tat and then things get slightly
+    complicated:
+
+    1. Every run of defections played by the opponent increases the number of
+    defections that this strategy retaliates with by 1.
+    2. The opponent is given a ‘fresh start’ if:
+        - it is 10 points behind this strategy
+        - and it has not just started a run of defections
+        - and it has been at least 20 rounds since the last ‘fresh start’
+        - and there are more than 10 rounds remaining in the tournament
+        - and the total number of defections differs from a 50-50 random sample
+        by at least 3.0 standard deviations.
+
+    A ‘fresh start’ is a sequence of two cooperations followed by an assumption
+    that the game has just started (everything is forgotten).
+
+    This strategy came 2nd in Axelrod’s original tournament.
+
+    Names:
+
+    - TidemanAndChieruzzi: [Axelrod1980]_
+    """
+
+    name = 'Tideman and Chieruzzi'
+    classifier = {
+        'memory_depth': float('inf'),
+        'stochastic': False,
+        'makes_use_of': {"game", "length"},
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.is_retaliating = False
+        self.retaliation_length = 0
+        self.retaliation_remaining = 0
+        self.current_score = 0
+        self.opponent_score = 0
+        self.last_fresh_start = 0
+        self.fresh_start = False
+
+    def _decrease_retaliation_counter(self):
+        """Lower the remaining owed retaliation count and flip to non-retaliate
+        if the count drops to zero."""
+        if self.is_retaliating:
+            self.retaliation_remaining -= 1
+            if self.retaliation_remaining == 0:
+                self.is_retaliating = False
+
+    def _fresh_start(self):
+        """Give the opponent a fresh start by forgetting the past"""
+        self.is_retaliating = False
+        self.retaliation_length = 0
+        self.retaliation_remaining = 0
+
+    def _score_last_round(self, opponent: Player):
+        """Updates the scores for each player."""
+        # Load the default game if not supplied by a tournament.
+        game = self.match_attributes["game"]
+        last_round = (self.history[-1], opponent.history[-1])
+        scores = game.score(last_round)
+        self.current_score += scores[0]
+        self.opponent_score += scores[1]
+
+    def strategy(self, opponent: Player) -> Action:
+        if not opponent.history:
+            return C
+
+        # Calculate the scores.
+        self._score_last_round(opponent)
+
+        # Check if we have recently given the strategy a fresh start.
+        if self.fresh_start:
+            self.fresh_start = False
+            return C  # Second cooperation
+
+        # Check conditions to give opponent a fresh start.
+        current_round = len(self.history) + 1
+        if self.last_fresh_start == 0:
+            valid_fresh_start = True
+        # There needs to be at least 20 rounds before the next fresh start
+        else:
+            valid_fresh_start = current_round - self.last_fresh_start >= 20
+
+        if valid_fresh_start:
+            valid_points = self.current_score - self.opponent_score >= 10
+            valid_rounds = self.match_attributes['length'] - current_round >= 10
+            opponent_is_cooperating = opponent.history[-1] == C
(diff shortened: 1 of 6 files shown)


axelrod/strategies/axelrod_first.py
```diff
@@ -589,3 +589,135 @@ def strategy(self, opponent: Player) -> Action:
...
+
+class TidemanAndChieruzzi(Player):
+    """
+    This strategy begins by playing Tit For Tat and then things get slightly
+    complicated:
+
+    1. Every run of defections played by the opponent increases the number of
+    defections that this strategy retaliates with by 1.
+    2. The opponent is given a ‘fresh start’ if:
+        - it is 10 points behind this strategy
```
drvinceknight: Need a space between this line and `2. The opponent is `

axelrod/strategies/axelrod_first.py
```diff
@@ -589,3 +589,135 @@ def strategy(self, opponent: Player) -> Action:
...
+        - and it has not just started a run of defections
+        - and it has been at least 20 rounds since the last ‘fresh start’
+        - and there are more than 10 rounds remaining in the tournament
+        - and the total number of defections differs from a 50-50 random sample
+        by at least 3.0 standard deviations.
```
drvinceknight: This needs to be at the same level of indentation as the previous line:

```
+        - and the total number of defections differs from a 50-50 random sample
           by at least 3.0 standard deviations.
```

axelrod/strategies/axelrod_first.py
```diff
@@ -589,3 +589,136 @@ def strategy(self, opponent: Player) -> Action:
...
+    complicated:
+
+    1. Every run of defections played by the opponent increases the number of
+    defections that this strategy retaliates with by 1.
+
+    2. The opponent is given a ‘fresh start’ if:
+        - it is 10 points behind this strategy
+        - and it has not just started a run of defections
+        - and it has been at least 20 rounds since the last ‘fresh start’
+        - and there are more than 10 rounds remaining in the tournament
```
drvinceknight: `in the match`

axelrod/tests/strategies/test_axelrod_first.py
```diff
@@ -437,3 +437,54 @@ def test_strategy(self):
...
+                   (D, D), (D, D), (D, C), (C, D), (D, D),
+                   (D, D), (D, D), (D, C), (D, D), (D, D)]
+
+        self.versus_test(opponent, expected_actions=actions,
+                         attrs={'current_score': 40, 'opponent_score': 35,
+                                'last_fresh_start': 14,
+                                'retaliation_length': 4,
+                                'retaliation_remaining': 1})
+
+
```
drvinceknight: To be a slight pain (this strategy really is excellently written), could we add a couple more tests please:

- Check the standard deviation condition (test similar actions where one is and one isn't within bounds)
- Check the `self.fresh_start` attribute

dmanc: In the description of the strategy
> and the total number of defections differs from a 50-50 random sample by at least 3.0 standard deviations.

I implemented this as ` if self.defections <= lower or self.defections >= upper:`
But I am now question what "total defections" means, would it be self.defections + opponent.defections?

drvinceknight: I agree that the wording is not completely clear but I am 99.9% sure that the way you have interpreted/implemented it is correct.

It makes sense to check the behaviour of the opponent.

drvinceknight:
This looks great. Thanks :+1:

One final request from me (sorry I should have realised this earlier). Can you modify https://github.com/Axelrod-Python/Axelrod/blob/master/docs/reference/overview_of_strategies.rst to show that this is now implemented please.

drvinceknight:
Thanks this looks good to me :+1:

axelrod/strategies/axelrod_first.py
```diff
@@ -589,3 +589,136 @@ def strategy(self, opponent: Player) -> Action:
...
+        current_round = len(self.history) + 1
+        if self.last_fresh_start == 0:
+            valid_fresh_start = True
+        # There needs to be at least 20 rounds before the next fresh start
+        else:
+            valid_fresh_start = current_round - self.last_fresh_start >= 20
+
+        if valid_fresh_start:
+            valid_points = self.current_score - self.opponent_score >= 10
+            valid_rounds = self.match_attributes['length'] - current_round >= 10
```
marcharper: What will happen if `length` is not supplied?

drvinceknight: You have dealt with this the correct way. When a length is not supplied the match tells the players that it's infinite (we have a few players that do this). 👍

drvinceknight: Could we add a test where we don't tell the match length to the players? 

Similar to the test for `Alexei` in this file: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/strategies/test_titfortat.py

drvinceknight: Except go for `match_attributes={"length": float('inf')}`, that's essentially equivalent to `-1` but more readable (I'll probably open a PR to that test file and change it).

dmanc: Wouldn't the behavior be different if I use -1? 
`valid_rounds = self.match_attributes['length'] - current_round >= 10` would be true if length was float('inf') and false if length was -1.

drvinceknight: Yup you're correct. 'float('inf')' is the correct choice here: in essence if match length is not known then that corresponds to an infinite match length from the point of view of the strategy.

In this case: are we within ten turns of the end? Probably not: we don't know when the end is. So it's essentially going to assume there is always more than 10 rounds remaining.

dmanc: Alright thanks, i'll make a quick change to the test.

axelrod/tests/strategies/test_axelrod_first.py
```diff
@@ -437,3 +437,78 @@ def test_strategy(self):
...
+        self.versus_test(opponent, expected_actions=actions,
+                         attrs={'current_score': 40, 'opponent_score': 35,
+                                'last_fresh_start': 14,
+                                'retaliation_length': 4,
+                                'retaliation_remaining': 1})
+
+        # Similar to the last test except the length is not given.
+        opponent = axelrod.Cycler('DDCDD')
+        self.versus_test(opponent, expected_actions=actions,
+                         match_attributes={'length': float('inf')})
```
drvinceknight: Is there a way to devise a test where the play plays differently when the match length is not given? Or even if the actions are the same, are the `attrs` different?

dmanc: Yes, I can change this test so with 15 rounds of play this strategy will give a fresh start on the last two rounds when the length is not given. Otherwise if the length was 15 it would defect on the last two rounds.

drvinceknight: Perfect!

drvinceknight:
Perfect! Thanks :+1:

Issue #1114, opened 2017-08-06
Some refinement and reduction of Hypothesis


drvinceknight:
We've been getting what feels like a few more health warning type errors with hypothesis including indications of some depreciations of features we're using (we always pip install the latest hypothesis on travis). I'll take a look at this (just opening an issue to keep track). If we get any hypothesis type failures on travis could we copy the error message in to this issue to give me some pointers :) :+1:

drvinceknight:
At: https://travis-ci.org/Axelrod-Python/Axelrod/jobs/261514887

```
ERROR: test_property_serial_play (axelrod.tests.unit.test_tournament.TestTournament)
Test serial play using hypothesis
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/Axelrod-Python/Axelrod/axelrod/tests/unit/test_tournament.py", line 221, in test_property_serial_play
    max_turns=10, min_repetitions=2,
  File "/home/travis/virtualenv/python3.6.2/lib/python3.6/site-packages/hypothesis/core.py", line 721, in wrapped_test
    random, settings, test_runner, search_strategy)
  File "/home/travis/virtualenv/python3.6.2/lib/python3.6/site-packages/hypothesis/core.py", line 398, in perform_health_checks
    HealthCheck.too_slow,
  File "/home/travis/virtualenv/python3.6.2/lib/python3.6/site-packages/hypothesis/core.py", line 272, in fail_health_check
    raise FailedHealthCheck(message)
hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 10 valid examples in 1.01 seconds (0 invalid ones and 0 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.average_size or max_leaves parameters).
See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.too_slow to the suppress_health_check settings for this test.
```

drvinceknight:
Closed by #1121

PR #1121, opened 2017-08-18
Implement a global reset and trim test suite

### LINKED ISSUE #1119: Minor refactor of __init__ and reset
Often we set and clear `Player` attributes in our `__init__` and `reset` methods, resulting in duplicate code and sometimes subtle bugs when we miss something in the `reset` method (or leave it out altogether).

I think it makes sense to have either a special method called something like `set_default_attributes` that handle all the attribute initialization and clearing that is called by the default `reset` method, or that we alter `Player.__init__` to call `reset` by default and set all the variables in the reset method. I prefer the former.

Thoughts?

drvinceknight:
This makes use of the `init_kwargs` and in fact we no longer need to
write reset methods (unless we want them for cheating strategies etc).

There are two relevant (simple) changes:

- Modify `player.reset` to call `self.__init__(**self.kwargs)`
- Tweak the match to pass the match attributes to the player just after resetting it.

The rest are me just removing the `reset` method except for the `Darwin` strategy which does things it's own way (and is classified as such).

Closes #1119

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ script:
   - cd docs; make html
   # Run the test suit with coverage
   - cd ..
-  - coverage run --source=axelrod -m unittest discover
+  - travis_wait 30 coverage run --source=axelrod -m unittest discover
   - coverage report -m
   # Run the doctests
   - python doctests.py
--- a/axelrod/match.py
+++ b/axelrod/match.py
@@ -132,6 +132,7 @@ class Match(object):
         if self._stochastic or (cache_key not in self._cache):
             for p in self.players:
                 p.reset()
+                p.set_match_attributes(**self.match_attributes)
             for _ in range(turns):
                 self.players[0].play(self.players[1], self.noise)
             result = list(
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -247,13 +247,9 @@ class Player(object):
         This method is called at the beginning of each match (between a pair
         of players) to reset a player's state to its initial starting point.
         It ensures that no 'memory' of previous matches is carried forward.
-
-        The default method resets a player's history, cooperations, defections
-        and state_distribution. Players which have further attributes need to
-        override this method and ensure those additional attributes are also
-        reset.
         """
         self.history = []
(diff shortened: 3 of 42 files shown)


drvinceknight:
This PR was having numerous CI failures due to hypothesis timeouts so I've gone through and trimmed a number of tests:

- Reduce examples of hypothesis
- Reduce some of the sizes of the larger integration tests
- Set defaults for some property based tests to only use short run time
  strategies

This now closes #1114

drvinceknight:
There are two things going on in this PR. Usually we'd want to avoid that but in this case the first thing (adding the `reset`) was proving to be a good thing to "fix" as far as the tests go. The commits are tidy so if helpful:

- a04cfb6 implements the new reset method
- efefd13 trims our tests (less examples in hypothesis, use of short run time strategies etc) and tweaks the travis config to allow for a longer run time

There are two further commits:

- 4157070 adds an error just to make sure travis reports this correctly;
- 553e131 reverts 4157070

.travis.yml
```diff
@@ -24,7 +24,7 @@ script:
   - cd docs; make html
   # Run the test suit with coverage
   - cd ..
-  - coverage run --source=axelrod -m unittest discover
+  - travis_wait 30 coverage run --source=axelrod -m unittest discover
```
marcharper: Why is this needed?

drvinceknight:
@marcharper and @meatballs you've both approved so I'm going to assume this is ok to go in?

.travis.yml
drvinceknight: If travis receives no output for a default 10 minutes then it stops and returns a failure. This changes the default to 30 mins but also changes how it runs it (https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) - that's why I tested that failures were still being reported correctly.

Even with the test trimming the 10 minutes default was still causing failures on travis (I think due to travis being a bit slow and some hypothesis tests - I ran numerous experiments and wasn't getting this locally) so this just ensures that it shouldn't happen.

PR #1122, opened 2017-08-20
Moran plot

### LINKED ISSUE #1048: Standard plots for the Moran process
We should have plots for the output, like the population distribution per round over time.

Perhaps we could consider one for fixation probabilities (time consuming for some parameters perhaps).

marcharper:
Add a plot for the Moran process. Closes #1048 .


axelrod/moran.py
```diff
@@ -385,6 +386,28 @@ def __len__(self) -> int:
         """
         return len(self.populations)
 
+    def populations_plot(self, ax=None):
```
drvinceknight: Could we just have a docstring?

```
Plot the population distribution over iterations of the Moran process.

Parameters
----------------
    ax: a matplotlib axis
       Allows to plot this on a given matplotlib axis. Default is None.

Returns
-----------
    A matplotlibt axis object
```

docs/tutorials/getting_started/moran.rst
```diff
@@ -78,14 +77,33 @@ The scores in each round::
...
     [7.0, 7.0, 7.0, 0.0]
 
+We can plot the results of a Moran process with `mp.populations_plot()`. Let's
+use a larger population to get a bit more data::
+
+   >>> import random
+   >>> import matplotlib.pyplot as plt
+   >>> axl.seed(15)  # for reproducible example
+   >>> N = 10
+   >>> players = [random.choice(axl.basic_strategies)() for i in range(N)]
```
drvinceknight: Could we hard code a list of strategies please. Otherwise, if and when `basic_strategies` changes (with new strategies being added) this would not necessarily give the same plot.

docs/tutorials/getting_started/moran.rst
```diff
@@ -78,14 +77,33 @@ The scores in each round::
...
+   >>> mp = axl.MoranProcess(players=players, turns=200)
+   >>> populations = mp.play()
+   >>> mp.winning_strategy_name
+   'Tit For Tat'
+    >>> ax = mp.populations_plot()
+   >>> plt.show()  #doctest: +SKIP
```
drvinceknight: No need to skip this.

axelrod/tests/unit/test_moran.py
```diff
@@ -330,6 +331,20 @@ def test_iter(self):
         mp = MoranProcess((p1, p2))
         self.assertEqual(mp.__iter__(), mp)
 
+    def test_population_plot(self):
+        # Test that can plot on a given matplotlib axes
+        axelrod.seed(15)
+        N = 5
+        players = [random.choice(axelrod.basic_strategies)() for _ in range(N)]
```
drvinceknight: Same comment about the `basic_strategies` would this potentially change the axes in lines 345, 346 if and when new strategies come in?

For the purpose of the test let's just use `demo_strategies` (or any other hard coded list: perhaps just a cooperator and a defector?)

axelrod/tests/unit/test_moran.py
```diff
@@ -330,6 +331,20 @@ def test_iter(self):
...
+        mp = axelrod.MoranProcess(players=players, turns=200)
```
drvinceknight: For the purpose of the test can we reduce turns? (Run time?)

axelrod/tests/unit/test_moran.py
```diff
@@ -330,6 +331,20 @@ def test_iter(self):
...
+        mp.play()
+        fig, axarr = plt.subplots(2, 2)
+        ax = axarr[1, 0]
+        mp.populations_plot(ax=ax)
```
drvinceknight: Can we also have a test where we just call `mp.populations_plot()` (to test the default runs without a problem).

drvinceknight: (This is where the coverage is missing)

docs/tutorials/getting_started/moran.rst
marcharper: Actually the test hangs and all the windows stay open otherwise, something to do with `plt` not exiting perhaps because the plot object isn't gc'd.

axelrod/tests/unit/test_moran.py
marcharper: Sure but there's a much longer test in the `moran_test.py` file.

docs/tutorials/getting_started/moran.rst
drvinceknight: gc'd (curious)?

No problem to skip.

axelrod/tests/unit/test_moran.py
drvinceknight: Feel free to reduce that one too :)

marcharper:
FYI all the tests pass locally, looks like a timeout issue.

docs/tutorials/getting_started/moran.rst
marcharper: garbage collected

drvinceknight:
> FYI all the tests pass locally, looks like a timeout issue.

Is this rebased on to master with the stuff I did on #1121 ?

axelrod/tests/unit/test_moran.py
```diff
@@ -330,6 +330,21 @@ def test_iter(self):
...
+        players = [random.choice(axelrod.demo_strategies)() for _ in range(5)]
+        mp = axelrod.MoranProcess(players=players, turns=30)
+        mp.play()
+        fig, axarr = plt.subplots(2, 2)
+        ax = axarr[1, 0]
+        mp.populations_plot(ax=ax)
+        self.assertEqual(ax.get_xlim(), (-0.8, 16.8))
+        self.assertEqual(ax.get_ylim(), (0, 5.25))
+
+        # Run with a give axis
```
drvinceknight: `# Run without a given axis` (I guess)

axelrod/tests/unit/test_moran.py
```diff
@@ -330,6 +330,21 @@ def test_iter(self):
...
+        mp.populations_plot()
```
drvinceknight: Can we test something. 

Perhaps (to avoid adding to the previous plot):

```python
plt.figure()
self.assertl(mp.populations_plot(), matplotlib.pyplot.Figure)  # I'm guessing?
```

docs/tutorials/getting_started/moran.rst
```diff
@@ -78,14 +77,33 @@ The scores in each round::
...
     [7.0, 7.0, 7.0, 0.0]
 
+We can plot the results of a Moran process with `mp.populations_plot()`. Let's
+use a larger population to get a bit more data::
+
+   >>> import random
+   >>> import matplotlib.pyplot as plt
+   >>> axl.seed(15)  # for reproducible example
+   >>> N = 10
+   >>> players = [random.choice(axl.demo_strategies)() for i in range(N)]
```
drvinceknight: Thanks for changing this, the plot file now needs to be updated to run with these strategies (eg `Suspicious Tit For Tat` is not in `demo_strategies`). Another alternative is to just hard code the names of the players in that plot file (instead of `random.choice(axl.demo_strategies)`) either way they need to match.

drvinceknight:
> Is this rebased on to master with the stuff I did on #1121 ?

It's looking like it's not, the PR run worked fine but the push didn't. If you rebase on to `master` it should sort that out.

docs/tutorials/getting_started/moran.rst
drvinceknight: I think I would prefer the hard coded approach as there's less code to (potentially) understand but your call. As long as they match up.

PR #1124, opened 2017-08-23
Add the K86R (MoreGrofman) Strategy

buckbaskin:
This strategy is implemented in response to issue #1093 

I've implemented the MoreGrofman strategy as well as added tests to verify the functionality.

I'm opening a PR for early feedback and to check test coverage. I don't know if the pull request is ready to merge (probably not yet) and I'd appreciate feedback for how to improve the pull request.


axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,62 @@ def strategy(self, opponent: Player) -> Action:
         else:
             # Play TFT
             return opponent.history[-1]
+
+
+class MoreGrofman(Player):
+    '''
```
drvinceknight: Can we change these to `"`

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,62 @@ def strategy(self, opponent: Player) -> Action:
...
+    Submitted to Axelrod's second tournament by Bernard Grofman.
+
+    This strategy has 3 phases:
+    1. First it cooperates on the first two rounds
+    2. For rounds 3-7 inclusive, it plays the same as the opponent's last move
+    3. Thereafter, it applies the following logic
+      - If its own previous move was C and the opponent has defected less than
```
drvinceknight: CI is failing because this isn't building in the documentation. Need a space:

```
    3. Thereafter, it applies the following logic

      - If its own previous move was C and the opponent has defected less than
```

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,62 @@ def strategy(self, opponent: Player) -> Action:
...
+    def strategy(self, opponent: Player) -> Action:
+        # Cooperate on the first two moves
+        if not self.history or len(self.history) in [1]:
+            return C
+        # For rounds 3-7, play the opponent's last move
+        elif 2 <= len(self.history) <= 6:
+            return opponent.history[-1]
+        # Logic for the rest of the game
+        else:
+            opponent_defections_last_7_rounds = 0
```
drvinceknight:
```
opponent_defections_last_7_rounds = opponent.history[-7:].count(D)
```

buckbaskin:
@drvinceknight I believe I've updated the code to fix the areas you highlighted in your review comments. I also added some tests that use matches run with the Fortran library to verify that the Python implementation matches how the Fortran library behaves (up to 60 rounds vs. AntiTitForTat). I've pushed the changes to see if the docs-building is fixed, and I'd appreciate any additional feedback.

drvinceknight:
> I also added some tests that use matches run with the Fortran library to verify that the Python implementation matches how the Fortran library behaves (up to 60 rounds vs. AntiTitForTat).

Thanks @buckbaskin this is looking good. Before reviewing properly just to verify that we have a good implementation. I've run some of the new fingerprints implemented on #1125:

```python
>>> import axelrod as axl
>>> import numpy as np
>>> import matplotlib.pyplot as plt

>>> moreG = axl.MoreGrofman()
>>> fortranG = axlf.Player('k86r')

>>> moreGtf = axl.TransitiveFingerprint(moreG, number_opponents=20)
>>> fortranGtf = axl.TransitiveFingerprint(fortranG, number_opponents=20)
```

Against a spectrum of random opponents:

```python
>>> axl.seed(0)
>>> _ = moreGtf.fingerprint(processes=0, repetitions=20000)  # Huge reps for stochasticity
>>> moreGtf.plot()
```

download 31


```python
>>> axl.seed(0)
>>> _ = fortranGtf.fingerprint(processes=0, repetitions=20000)
>>> fortranGtf.plot();
```

download 32

These match up well.

```python
>>> plt.imshow(np.abs(moreGtf.data - fortranGtf.data))
>>>plt.colorbar()
```
download 33

Against given deterministic strategies:

```python
>>> basic = [s() for s in axl.basic_strategies]
>>> moreGtf_against_basic = axl.TransitiveFingerprint(moreG, opponents=basic)
>>> fortranGtf_against_basic = axl.TransitiveFingerprint(fortranG, opponents=basic)
>>> _ = moreGtf_against_basic.fingerprint(processes=0)
>>> moreGtf_against_basic.plot(display_names=True);
```

download 34

```python
>>> _ = fortranGtf_against_basic.fingerprint(processes=0)
>>> fortranGtf_against_basic.plot(display_names=True);
```

download 35


Again a good match:

```python
>>> np.array_equal(moreGtf_against_basic.data, fortranGtf_against_basic.data)
True
```

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,60 @@ def strategy(self, opponent: Player) -> Action:
...
+    def strategy(self, opponent: Player) -> Action:
+        # Cooperate on the first two moves
+        if not self.history or len(self.history) in [1]:
+            return C
+        # For rounds 3-7, play the opponent's last move
+        elif 2 <= len(self.history) <= 6:
+            return opponent.history[-1]
+        # Logic for the rest of the game
+        else:
+            opponent_defections_last_7_rounds = opponent.history[-8:-1].count(D) # (-7:) fails at 13, (-7:-1), (-8:), (-8:-1), (-6:), (-6:-1)
```
drvinceknight: This doesn't seem correct: why doesn't `opponent.history[-7:]` work?

```python
>>> ten_numbers = list(range(10))
>>> len(ten_numbers[-7:])  # 7 numbers
7
>>> len(ten_numbers[-7:-1])  # Only 6 numbers
 6
```

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,60 @@ def strategy(self, opponent: Player) -> Action:
...
+            if self.history[-1] == C:
+                if opponent_defections_last_7_rounds <= 2:
```
drvinceknight: The description says:

```
If its own previous move was C and the opponent has defected twice or more in the last 7 rounds, defect
```

So I believe this should be a strict inequality.

axelrod/strategies/axelrod_second.py
buckbaskin: I originally tried `[-7:]` but that resulted in what amounted to an off by one error for certain cases. This came to light when comparing the Python implementation with the Fortran implementation. I believe that the difference in indexes comes from the difference in Fortran indexing during do-loops/arrays (1 indexed and inclusive-inclusive) vs Python indexing into lists. The `[-8:-1]` indexing appears to match the behavior of the Fortran. I will remove the comment after because I was writing down potential off by one candidates to experiment with because I didn't really understand the Fortran indexing at the time.

axelrod/strategies/axelrod_second.py
buckbaskin: Based on the behavior in testing compared with the Fortran implementation and my attempt to understand the Fortran implementation, the logic is correct and I will adjust the comment. Is that sufficient?

axelrod/strategies/axelrod_second.py
drvinceknight: Ok can you adjust the docstring description then because they're not what you are implemented:

```
If its own previous move was C and the opponent has defected less than twice in the last 7 rounds
```

I think we should say something like the mouthful that is `the 7 rounds before the last one` ... ?

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -158,3 +158,125 @@ def test_strategy(self):
...
         self.versus_test(opponent, expected_actions=actions,
                          attrs={'patsy': False})
+
+
+class TestMoreGrofman(TestPlayer):
+
+    name = "MoreGrofman"
+    player = axelrod.MoreGrofman
+    expected_classifier = {
+        'memory_depth': 7,
```
drvinceknight: This now needs to be 8.

drvinceknight:
>  I believe that the difference in indexes comes from the difference in Fortran indexing during do-loops/arrays (1 indexed and inclusive-inclusive) vs Python indexing into lists. The [-8:-1] indexing appears to match the behavior of the Fortran. 

Having looked through the code I agree. Although I think it's actually due to the fact that the Fortran strategies all assume that prior to actually starting a match the opponent has cooperated (it's this initial cooperation that I think is making the difference here).

Here is another example that shows that it's not in fact the last 7 moves but the 7 moves prior to the last one:

```
>>> import axelrod as axl
>>> player = axlf.Player("k86r")
>>> sequence = "C" * 6 + "DDD"
>>> opponent = axl.Cycler(sequence)
>>> match = axl.Match((player, opponent), turns=11)
>>> match.play()
[(C, C),
 (C, C),
 (C, C),
 (C, C),
 (C, C),
 (C, C),
 (C, D),
 (C, D),
 (C, D),
 (C, C),
 (D, C)]
```

- We see that the first 6 rounds are just pairs of cooperations.
- The 7th round: the player cooperates
- The 8th round: the player considers the first 6th moves of the opponent (**keeping in mind that the Fortran strategies use an initial dummy cooperation move of the opponent** I suspect this is the critical thing going on here) and there are no defections so it cooperates.
- The 9th round: the player considers the first 7 moves: there is a single defection so it cooperates
- The 10th round: the player considers moves from round 2 till 8: there are two defections so it cooperates
- The 11th round: the player considers moves from round 3 till 9: there are 3 defections so it defects.

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,67 @@ def strategy(self, opponent: Player) -> Action:
...
+        # Logic for the rest of the game
+        else:
+            # Note: the Fortran code behavior ignores the opponent behavior
+            #   in the last round and instead looks at the first 7 of the last
+            #   8 rounds.
+            opponent_defections_last_8_rounds = opponent.history[-8:-1].count(D)
+            if self.history[-1] == C:
+                if opponent_defections_last_8_rounds <= 2:
+                    return C
+                else:
```
drvinceknight: no need to an `else`, just replace the `else` with `return D`

Similarly below (line 264).

(Not a big deal: just a stylistic suggestion.)

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,67 @@ def strategy(self, opponent: Player) -> Action:
         else:
             # Play TFT
             return opponent.history[-1]
+
+
+class MoreGrofman(Player):
+    """
+    Submitted to Axelrod's second tournament by Bernard Grofman.
+
+    This strategy has 3 phases:
```
drvinceknight: Also need a space here.

```
This strategy has 3 phases:

1. First
```

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,66 @@ def strategy(self, opponent: Player) -> Action:
...
+        'memory_depth': 8,
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self) -> None:
```
marcharper: I don't think we need this `__init__` if it simply calls the parent class `__init__`

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,66 @@ def strategy(self, opponent: Player) -> Action:
         else:
             # Play TFT
             return opponent.history[-1]
+
+
+class MoreGrofman(Player):
+    """
+    Submitted to Axelrod's second tournament by Bernard Grofman.
```
marcharper: Can we add the Fortran rule name (K86R) to the docstring?

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,66 @@ def strategy(self, opponent: Player) -> Action:
...
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def strategy(self, opponent: Player) -> Action:
+        # Cooperate on the first two moves
+        if not self.history or len(self.history) in [1]:
```
marcharper: `if len(self.history) < 2:`

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,66 @@ def strategy(self, opponent: Player) -> Action:
...
+        # For rounds 3-7, play the opponent's last move
+        elif 2 <= len(self.history) <= 6:
+            return opponent.history[-1]
+        # Logic for the rest of the game
+        else:
+            # Note: the Fortran code behavior ignores the opponent behavior
+            #   in the last round and instead looks at the first 7 of the last
+            #   8 rounds.
+            opponent_defections_last_8_rounds = opponent.history[-8:-1].count(D)
+            if self.history[-1] == C:
```
marcharper: Can this be simplified to:
```
if self.history[-1] == C and opponent_defections_last_8_rounds <= 2:
    return C
if self.history[-1] == D and opponent_defections_last_8_rounds <= 1:
   return C
return [D
```
?

axelrod/strategies/axelrod_second.py
```diff
@@ -199,3 +199,66 @@ def strategy(self, opponent: Player) -> Action:
...
+        super().__init__()
+
+    def strategy(self, opponent: Player) -> Action:
+        # Cooperate on the first two moves
+        if not self.history or len(self.history) in [1]:
+            return C
+        # For rounds 3-7, play the opponent's last move
+        elif 2 <= len(self.history) <= 6:
+            return opponent.history[-1]
+        # Logic for the rest of the game
```
marcharper: This comment isn't necessary

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -158,3 +158,125 @@ def test_strategy(self):
...
+        opponent_actions = ([D] * 7) + [D]
+        opponent = axelrod.MockPlayer(actions=opponent_actions)
+        actions = list(zip(moregrofman_actions, opponent_actions))
+        self.versus_test(opponent, expected_actions=actions)
+
+        # Test to make sure logic matches Fortran (discrepancy found 8/23/2017)
+        opponent = axelrod.AntiTitForTat()
+        # Actions come from a match run by Axelrod Fortran using Player('k86r')
+        actions = [
+            (C, C),
```
marcharper: The spacing here is perhaps a little excessive, can we compress to just a couple of lines?

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -158,3 +158,125 @@ def test_strategy(self):
...
+            (D, C),
+            (D, C),
+            (C, C)
+            ]
+        self.versus_test(opponent, expected_actions=actions)
+
+        # Test to match the Fortran implementation for 30 rounds
+        opponent = axelrod.AntiTitForTat()
+        actions = [(C, C),
+            (C, D),
```
marcharper: same

drvinceknight:
@marcharper I believe all your request have now been addressed and this is ready to go? :)

drvinceknight:
Merging as all the stylistic requests have been addressed. Thanks @buckbaskin👍👍

PR #1127, opened 2017-08-26
Properly handle recieve_match_attributes for Stalker, simplify for LR Players, and various style fixes

### LINKED ISSUE #877: What is `player.receive_match_attributes` for?
As far as I can tell we don't use it?

```
125     def receive_match_attributes(self):                                                                                                           
126         # Overwrite this function if your strategy needs                                                                                          
127         # to make use of match_attributes such as                                                                                                 
128         # the game matrix, the number of rounds or the noise                                                                                      
129         pass 
```

We only use the `set_match_attributes` method in the `Match` class. Can this be refactor out or am I missing something?

marcharper:
Closes #877

### DIFF
--- a/axelrod/tests/strategies/test_stalker.py
+++ b/axelrod/tests/strategies/test_stalker.py
@@ -1,10 +1,10 @@
 """Tests for the Stalker strategy."""
 
 import axelrod as axl
-import random
 from .test_player import TestPlayer
 
 C, D = axl.Action.C, axl.Action.D
+
 
 class TestStalker(TestPlayer):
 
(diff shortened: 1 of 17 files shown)


axelrod/tests/strategies/test_stalker.py
```diff
@@ -62,3 +62,12 @@ def test_strategy(self):
         self.versus_test(opponent=axl.Cooperator(),
                          expected_actions=actions,
                          match_attributes={"length": 4})
+
+    def test_reset(self):
```
drvinceknight: The parent `test_reset` method also plays against `Alternator` so I don't believe this is necessary (but it does no harm). Just mentioning it.

drvinceknight:
(Appveyor is falling over because of some upstream connectivity issues - setting up the conda envs -, as long as one of the builds on it is fine I'm happy.)

axelrod/tests/strategies/test_stalker.py
marcharper: The intent is to catch a regression in the new `reset` / `__init__` methods or the `set_match_attributes` / `receive_match_attributes` chain by looking for a specific attribute that should be reset. (There may be other such tests and/or it may be in fact useless.)

drvinceknight: Cool. Thanks. Looks good to me. 👍

PR #1128, opened 2017-09-04
Write description/ex of `receive_match_attributes`

drvinceknight:
Note we only need to do this when the inputs from the match are used to
change some other variable.

Some other minor tweaks here:

- Remove unneeded data file.
  This was checked in at some point but shouldn't have been.
- Remove outdated documentation about repr.
  This is now done automatically by the parent class.
- Make some minor stylistic changes.
  - PEP8 (column width)
  - Change to use class hyperlink instead of code syntax highlighting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ script:
   - cd docs; make html
   # Run the test suit with coverage
   - cd ..
-  - travis_wait 30 coverage run --source=axelrod -m unittest discover
+  - travis_wait 60 coverage run --source=axelrod -m unittest discover
   - coverage report -m
   # Run the doctests
   - python doctests.py
--- a/docs/tutorials/contributing/strategy/writing_the_new_strategy.rst
+++ b/docs/tutorials/contributing/strategy/writing_the_new_strategy.rst
@@ -22,7 +22,8 @@ The code
 --------
 
 There are a couple of things that need to be created in a strategy.py file.  Let
-us take a look at the :code:`TitForTat` class (located in the
+us take a look at the :class:`TitForTat
+<axelrod.strategies.titfortat.TitForTat>` class (located in the
 :code:`axelrod/strategies/titfortat.py` file)::
 
     class TitForTat(Player):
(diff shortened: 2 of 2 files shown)


docs/tutorials/contributing/strategy/writing_the_new_strategy.rst
```diff
@@ -115,30 +117,19 @@ opponent.history[-1]`)::
...
-The variables :code:`C` and :code:`D` represent the cooperate and defect actions respectively.
+The variables :code:`C` and :code:`D` represent the cooperate and defect actions
+respectively.
 
-You can also modify the name of the strategy with the :code:`__repr__` method,
-which is invoked when :code:`str` is applied to a player instance. For example,
-the :code:`Random` strategy takes a parameter :code:`p` for how often it
-cooperates, and the :code:`__repr__` method adds the value of this parameter to
-the name::
+Some strategies make specific use of the variables of a match to create their
```
marcharper: Maybe a bit more detail here. After:

+Some strategies make specific use of the variables of a match to create their
+own attributes.

Then:

In principle these attributes could change throughout a match or tournament if the match properties (like the game matrix) change, so we require that this logic live in the 
the :code:`receive_match_attributes` method for correct dynamic updating. Here is how this is done for
+:class:`Stalker <axelrod.strategies.stalker.Stalker>`:

drvinceknight: Done! :+1:

.travis.yml
```diff
@@ -24,7 +24,7 @@ script:
   - cd docs; make html
   # Run the test suit with coverage
   - cd ..
-  - travis_wait 30 coverage run --source=axelrod -m unittest discover
+  - travis_wait 60 coverage run --source=axelrod -m unittest discover
```
marcharper: Maybe we should work on decreasing the time it takes all the tests to run

drvinceknight: > Maybe we should work on decreasing the time it takes all the tests to run

That was mainly the purpose of the work I did on https://github.com/Axelrod-Python/Axelrod/pull/1121.

There I reduced the test runs times on travis - where it's very long - from approx 40 mins to approx 20 mins. (Note that locally the full test suite takes about 10 mins on my small laptop - similar on appveyor.)

Just now it failed on a timeout which would have just been a random event. The 30 minute limit was arbitrary there's no reason not to increase it to 60. It hadn't timed out for a while since #1121 (before that it was essentially timing out every other run) but we'll keep an eye on it.

drvinceknight: > That was mainly the purpose of the work I did on #1121.

(Just in case you only read the title of that issue, the rest of the comments there detail the work I went through minimising test suite run time and why I did it there. :+1:)

PR #1130, opened 2017-10-09
Fix the data set dimensions for the sdvplot.

### LINKED ISSUE #1129: Plot incompatibility with mpl 2.1
Something seems to be broken with mpl 2.1, and possibly, the violon plots, working on a fix now :+1:

drvinceknight:
This closes #1129.

I could not find anything in the release notes
<http://matplotlib.org/users/whats_new.html> however I think this is due to the fact that mpl 2.1 got rid of the ability to handle the unflattened sub list (so this fix will still work with previous versions of mpl).

### DIFF
--- a/axelrod/plot.py
+++ b/axelrod/plot.py
@@ -112,8 +112,8 @@ class Plot(object):
     @property
     def _sdv_plot_dataset(self):
         ordering = self._sd_ordering
-        diffs = [[diff for repetition in diffs for diff in repetition]
-                 for diffs in self.result_set.score_diffs]
+        diffs = [[score_diff for opponent in player for score_diff in opponent]
+                 for player in self.result_set.score_diffs]
         # Reorder and grab names
         diffs = [diffs[i] for i in ordering]
         ranked_names = [str(self.players[i]) for i in ordering]
--- a/axelrod/plot.py
+++ b/axelrod/plot.py
@@ -112,7 +112,8 @@ class Plot(object):
     @property
     def _sdv_plot_dataset(self):
         ordering = self._sd_ordering
-        diffs = self.result_set.score_diffs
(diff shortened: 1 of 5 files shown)


drvinceknight:
Investigating why coverage would have dropped...

axelrod/plot.py
```diff
@@ -112,7 +112,8 @@ def _sd_ordering(self):
     @property
     def _sdv_plot_dataset(self):
         ordering = self._sd_ordering
-        diffs = self.result_set.score_diffs
+        diffs = [[diff for repetition in diffs for diff in repetition]
```
marcharper: This is confusing, can we use different iteration variables?

drvinceknight: I picked these in an attempt to not be confusing but happy to take any suggestions.

drvinceknight: 91a3ee8 renames them to what is hopefully less confusing but still happy to change further if you think it can be improved :+1: :)

marcharper: It's still a little confusing since player and opponent usually indicate instances of Player classes. But it's better than before in that I can see what it's doing now.

axelrod/tournament.py
```diff
@@ -255,8 +255,8 @@ def _run_parallel(self, processes: int=2) -> bool:
         """
         # At first sight, it might seem simpler to use the multiprocessing Pool
         # Class rather than Processes and Queues. However, this way is faster.
-        work_queue = Queue()
-        done_queue = Queue()
+        work_queue = Queue() # type: Queue
```
marcharper: nit: I think it's supposed to be two spaces before the `#`.

drvinceknight: Good catch. Have fixed.

drvinceknight:
Merging. Will do a new release :+1

PR #1131, opened 2017-10-10
Correct the implementation of Champion.

drvinceknight:
Currently champion uses a linear scale to adapt its two "phases" for any
game length. This was not the intention of the original description and
is causing errors with comparisons of the original results.

This essentially simplifies the strategy.

(I believe this was my doing)


drvinceknight:
(My editor has ripped out a bunch of unnecessary white space too. Apologies.)

PR #1132, opened 2017-10-29
Added Tricky level punisher strategy.

anurag03:
Added a new strategy called as Tricky level punisher along with the test case which is required.
This is my first PR in this repository.

### DIFF
--- a/axelrod/strategies/punisher.py
+++ b/axelrod/strategies/punisher.py
@@ -178,7 +178,7 @@ class TrickyLevelPunisher(Player):
             return D
         elif len(opponent.history) < 100:
             return C
-        elif len((len(opponent.history) - opponent.cooperations) / len(opponent.history) > 0.05:
+        elif (len(opponent.history) - opponent.cooperations) / len(opponent.history) > 0.05:
             return D
         else:
             return C
--- a/axelrod/strategies/punisher.py
+++ b/axelrod/strategies/punisher.py
@@ -147,9 +147,9 @@ class LevelPunisher(Player):
 
 class TrickyLevelPunisher(Player):
     """
-    A player starts by cooperating however, after 10 rounds
+    A player starts by cooperating however, after 10, 50 and 100 rounds
     will defect if at any point the number of defections
-    by an opponent is greater than 20%.
+    by an opponent is greater than 20%, 10% and 5% respectively.
 
     Names:
 
@@ -170,15 +170,15 @@ class TrickyLevelPunisher(Player):
     def strategy(self, opponent: Player) -> Action:
         if len(opponent.history) < 10:
             return C
-        elif (len(opponent.history) - opponent.cooperations) / len(opponent.history) > 0.2:
+        if opponent.defections / (opponent.defections + opponent.cooperations) > 0.2:
             return D
-        elif len(opponent.history) < 50:
+        if len(opponent.history) < 50:
             return C
-        elif (len(opponent.history) - opponent.cooperations) / len(opponent.history) > 0.1:
+        if opponent.defections / (opponent.defections + opponent.cooperations) > 0.1:
             return D
-        elif len(opponent.history) < 100:
+        if len(opponent.history) < 100:
             return C
-        elif (len(opponent.history) - opponent.cooperations) / len(opponent.history) > 0.05:
+        if opponent.defections / (opponent.defections + opponent.cooperations) > 0.05:
             return D
         else:
             return C
--- a/axelrod/strategies/punisher.py
+++ b/axelrod/strategies/punisher.py
@@ -153,10 +153,10 @@ class TrickyLevelPunisher(Player):
 
     Names:
 
-    - Level Punisher: [Eckhart2015]_
+    - Tricky Level Punisher: [Eckhart2015]_
     """
 
-    name = 'Level Punisher'
+    name = 'Tricky Level Punisher'
     classifier = {
         'memory_depth': float('inf'), # Long Memory
         'stochastic': False,
--- a/axelrod/strategies/punisher.py
+++ b/axelrod/strategies/punisher.py
@@ -148,7 +148,7 @@ class LevelPunisher(Player):
 class TrickyLevelPunisher(Player):
     """
     A player starts by cooperating however, after 10, 50 and 100 rounds
-    will defect if at any point the number of defections
+    will defect if at any point the percentage of defections
     by an opponent is greater than 20%, 10% and 5% respectively.
 
     Names:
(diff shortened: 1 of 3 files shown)


axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
...
+    A player starts by cooperating however, after 10 rounds
+    will defect if at any point the number of defections
+    by an opponent is greater than 20%.
+
+    Names:
+
+    - Level Punisher: [Eckhart2015]_
+    """
+
+    name = 'Level Punisher'
```
drvinceknight: This name needs to be changed (one of the test failures is because of this):

```
- Tricky Level Punisher: [Eckhart2015]_
```

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
...
+
+class TrickyLevelPunisher(Player):
+    """
+    A player starts by cooperating however, after 10 rounds
+    will defect if at any point the number of defections
+    by an opponent is greater than 20%.
+
+    Names:
+
+    - Level Punisher: [Eckhart2015]_
```
drvinceknight: `Tricky Level Punisher` (I have not checked the reference yet but I assume that it's called that there.)

drvinceknight:
That's just an initial sweep through looking at the failing tests (will aim to review properly once that's done).

drvinceknight:
Tests are failing because of the meta player whose behaviour has changed (because it makes use of this new strategy). Make this modification (to both `actions`) in `axelrod/tests/strategies/test_meta.py`:

```python
333     def test_strategy(self):                                                                                                                    
334         actions = [(C, C), (C, D), (C, C), (C, D), (D, C)]                                                                                      
335         self.versus_test(opponent=axelrod.Alternator(),                                                                                         
336                          expected_actions=actions, seed=0)                                                                                      
337                                                                                                                                                 
338         actions = [(C, C), (C, D), (C, C), (C, D), (D, C)]                                                                                        339         self.versus_test(opponent=axelrod.Alternator(),                                                                                           340                          expected_actions=actions, seed=1)  
```

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
...
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if len(opponent.history) < 10:
+            return C
+        elif (len(opponent.history) - opponent.cooperations) / len(opponent.history) > 0.2:
```
drvinceknight: these can all just be `if` statements (no need to be `elif` because of the `return`).

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
             return D
         else:
             return C
+
+
+class TrickyLevelPunisher(Player):
+    """
+    A player starts by cooperating however, after 10 rounds
```
drvinceknight: This docstring is not complete (it does not match up with what the strategy does).

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
...
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if len(opponent.history) < 10:
+            return C
+        elif (len(opponent.history) - opponent.cooperations) / len(opponent.history) > 0.2:
+            return D
+        elif len(opponent.history) < 50:
```
drvinceknight: Looking at the strategy at https://github.com/jecki/CoopSim/blob/master/Strategies.py#L1022 this does not look like it's implemented correctly. The ratio checks are done after checking the number of turns.

So I believe this should look something like:

```python
if len(opponent.history) < 50:
    if (len(opponent.history) - opponent.cooperations) / len(opponent.history) > 0.2:
        return C
    return D
```

and similarly for the rest.

Essentially the strategy always checks the cooperation difference ratio against some level and that level changes depending on the number of turns.

Also (very minor) to avoid `len` calculations you can use:

```
opponent.defections / (opponent.defections + opponent.cooperations)
```

instead of:

```
(len(opponent.history) - opponent.cooperations) / len(opponent.history)
```

(I know you've done it your way because that's how it's done in `LevelPunisher`)

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
...
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if len(opponent.history) < 10:
+            return C
+        if opponent.defections / (opponent.defections + opponent.cooperations) > 0.2:
```
drvinceknight: For the sake of readability could we write this like this::

```

if len(opponent.history) < 50:
    if opponent.defections / len(opponent.history) > 0.2:
        return D
    return C
if len(opponent.history) < 100:
    if opponent.defections / len(opponent.history) > 0.1:
        return D
    return C
if opponent.defections / len(opponent.history) > 0.1:     
    return D
return C
```

This might be equivalent to what you have but it's more in line with the original (https://github.com/jecki/CoopSim/blob/master/Strategies.py#L1022) and I feel is also more readable.

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
...
+            return D
+        if len(opponent.history) < 50:
+            return C
+        if opponent.defections / (opponent.defections + opponent.cooperations) > 0.1:
+            return D
+        if len(opponent.history) < 100:
+            return C
+        if opponent.defections / (opponent.defections + opponent.cooperations) > 0.05:
+            return D
```
drvinceknight: There is no test that covers this possibility. Please add a test to ensure 100% coverage.

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
...
+
+    def strategy(self, opponent: Player) -> Action:
+        if len(opponent.history) < 10:
+            return C
+        if opponent.defections / (opponent.defections + opponent.cooperations) > 0.2:
+            return D
+        if len(opponent.history) < 50:
+            return C
+        if opponent.defections / (opponent.defections + opponent.cooperations) > 0.1:
+            return D
```
drvinceknight: There is no test that covers this possibility. Please add a test to ensure 100% coverage.

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,42 @@ def strategy(self, opponent: Player) -> Action:
             return D
         else:
             return C
+
+
+class TrickyLevelPunisher(Player):
+    """
+    A player starts by cooperating however, after 10, 50 and 100 rounds
+    will defect if at any point the number of defections
```
drvinceknight: `the percentage of defections`

axelrod/strategies/punisher.py
```diff
@@ -143,3 +143,40 @@ def strategy(self, opponent: Player) -> Action:
...
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if len(opponent.history) < 10:
+            if opponent.defections / len(opponent.history) > 0.2:
```
drvinceknight: This gives a division error (dividing by zero on first ten). 

Change to something like:

```
    if len(opponent.history) == 0:
        return C
    if len(opponent.history) < 10:
        ....
```

drvinceknight:
Closing this as it's been implemented by #1178

PR #1133, opened 2017-10-29
[probably not ready to merge] Add experimental graph.py type annotations

eshansingh:
This is my best guess on the types. They're still returning errors, but only a few, and most everything else is OK.

### DIFF
--- a/axelrod/graph.py
+++ b/axelrod/graph.py
@@ -6,7 +6,11 @@ https://github.com/marcharper/stationary/blob/master/stationary/utils/graph.py
 """
 
 from collections import defaultdict
+from typing import List, Dict, Hashable, Tuple
+from numbers import Number
 
+Vertice = Hashable
+Edge = Tuple[Vertice, Vertice]
 
 class Graph(object):
     """Weighted and directed graph object intended for the graph associated to a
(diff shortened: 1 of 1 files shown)


drvinceknight:
Could you add this file to https://github.com/Axelrod-Python/Axelrod/blob/master/run_mypy.py, this will ensure that the continuous integration (the automatic testing) also checks the types.

axelrod/graph.py
```diff
@@ -6,7 +6,11 @@
 """
 
 from collections import defaultdict
+from typing import List, Dict, Hashable, Tuple
+from numbers import Number
 
+Vertice = Hashable
```
marcharper: Vertice -> Vertex (and below)

axelrod/graph.py
```diff
@@ -6,7 +6,11 @@
 """
 
 from collections import defaultdict
+from typing import List, Dict, Hashable, Tuple
+from numbers import Number
```
marcharper: We can assume that the weight is a float IMO rather than a Number

eshansingh: Sure, will make changes ASAP.

PR #1138, opened 2017-11-12
 Implemented K32R by Charles Kluepfel.

gaffney2010:
Strategy submitted to Axelrod's second tournament by Charles Kluepfel (K32R).

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -3,7 +3,7 @@ Additional strategies from Axelrod's second tournament.
 """
 
 import random
-import numpy
+import numpy as np
 
 from axelrod.action import Action
 from axelrod.player import Player
@@ -478,7 +478,7 @@ class Kluepfel(Player):
     - `dd_count` counts: Opponent defects as response to player defecting.
     - `cc_count` counts: Opponent cooperates as response to player cooperating.
     - `dc_count` counts: Opponent defects as response to player cooperating.
-    
+
     After 26 turns, the player then tries to detect a random player.  The
     player decides that the opponent is random if
     cd_counts >= (cd_counts+dd_counts)/2 - 0.75*sqrt(cd_counts+dd_counts) AND
@@ -496,7 +496,7 @@ class Kluepfel(Player):
       with probability 60%.
 
     Names:
-    
+
     - Kluepfel: [Axelrod1980b]_
     """
 
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -3,6 +3,7 @@ Additional strategies from Axelrod's second tournament.
 """
 
 import random
+import numpy
 
 from axelrod.action import Action
 from axelrod.player import Player
@@ -480,8 +481,8 @@ class Kluepfel(Player):
     
     After 26 turns, the player then tries to detect a random player.  The
     player decides that the opponent is random if
-    C1 >= (C1+C2)/2 - 1.5*sqrt(C1+C2)/2  AND
-    C4 >= (C3+C4)/2 - 1.5*sqrt(C3+C4)/2.
+    cd_counts >= (cd_counts+dd_counts)/2 - 0.75*sqrt(cd_counts+dd_counts) AND
+    cc_counts >= (dc_counts+cc_counts)/2 - 0.75*sqrt(dc_counts+cc_counts).
     If the player decides that they are playing against a random player, then
     they will always defect.
 
@@ -530,8 +531,8 @@ class Kluepfel(Player):
 
         # Check for randomness
         if len(self.history) > 26:
-            if self.cd_counts >= (self.cd_counts+self.dd_counts)/2 - 0.75*(self.cd_counts+self.dd_counts)**0.5 and \
-                self.cc_counts >= (self.dc_counts+self.cc_counts)/2 - 0.75*(self.dc_counts+self.cc_counts)**0.5:
+            if self.cd_counts >= (self.cd_counts+self.dd_counts)/2 - 0.75*np.sqrt(self.cd_counts+self.dd_counts) and \
+                self.cc_counts >= (self.dc_counts+self.cc_counts)/2 - 0.75*np.sqrt(self.dc_counts+self.cc_counts):
                 return D
 
         # Otherwise respond to recent history
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -474,30 +474,29 @@ class Kluepfel(Player):
     (K32R).
 
     This player keeps track of the the opponent's responses to own behavior:
-
-                      I did:
-                    Def    Coop
-    She does: Coop   C1  |  C3
-                   -------------
-              Def    C2  |  C4
+    - `cd_count` counts: Opponent cooperates as response to player defecting.
+    - `dd_count` counts: Opponent defects as response to player defecting.
+    - `cc_count` counts: Opponent cooperates as response to player cooperating.
+    - `dc_count` counts: Opponent defects as response to player cooperating.
     
     After 26 turns, the player then tries to detect a random player.  The
     player decides that the opponent is random if
     C1 >= (C1+C2)/2 - 1.5*sqrt(C1+C2)/2  AND
     C4 >= (C3+C4)/2 - 1.5*sqrt(C3+C4)/2.
-    If the player decides that she is playing against a random player, then
-    she will always defect.
+    If the player decides that they are playing against a random player, then
+    they will always defect.
 
-    Otherwise, she will use a straight-forward set of rules:
+    Otherwise respond to recent history using the following set of rules:
     - If opponent's last three choices are the same, then respond in kind.
     - If opponent's last two choices are the same, then respond in kind with
-    probability 90%.
+      probability 90%.
     - Otherwise if opponent's last action was to cooperate, then cooperate
-    with probability 70%.
+      with probability 70%.
     - Otherwise if opponent's last action was to defect, then defect
-    with probability 60%.
+      with probability 60%.
 
     Names:
+    
     - Kluepfel: [Axelrod1980b]_
     """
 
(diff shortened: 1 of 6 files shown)


axelrod/strategies/axelrod_second.py
```diff
@@ -466,3 +467,107 @@ def strategy(self, opponent: Player) -> Action:
...
+                    self.C4 += 1
+
+        # Check for randomness
+        if len(self.history) > 26:
+            if self.C1 >= (self.C1+self.C2)/2 - 1.5*np.sqrt(self.C1+self.C2)/2 and \
+                self.C4 >= (self.C3+self.C4)/2 - 1.5*np.sqrt(self.C3+self.C4)/2:
+                return D
+
+        # Otherwise, straight-forward strategy
+        r = random.random() # Will use later
```
drvinceknight: Let's move the `r = random.random()` to line 559 so this random sample only happens if it's necessary.

axelrod/strategies/axelrod_second.py
```diff
@@ -466,3 +467,107 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        # Use the defaults from Python
```
drvinceknight: Can you explain what you mean by this?

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -406,3 +406,90 @@ def test_strategy(self):
...
+        actions = [(C, C), (C, D),
+                    (D, C), (C, D), (C, C), (D, D)]
+        self.versus_test(axelrod.Alternator(), expected_actions=actions, seed=2)
+
+        actions = [(C, C), (C, D),
+                    (D, C), (C, D), (C, C), (C, D)]
+        self.versus_test(axelrod.Alternator(), expected_actions=actions, seed=3)
+
+        # Now we have to test the detect-random logic, which doesn't pick up
+        # until after 26 trips.  So we need a big sample.
```
drvinceknight: `turns`

axelrod/strategies/axelrod_second.py
```diff
@@ -466,3 +467,107 @@ def strategy(self, opponent: Player) -> Action:
...
+    Otherwise, she will use a straight-forward set of rules:
+    - If opponent's last three choices are the same, then respond in kind.
+    - If opponent's last two choices are the same, then respond in kind with
+    probability 90%.
+    - Otherwise if opponent's last action was to cooperate, then cooperate
+    with probability 70%.
+    - Otherwise if opponent's last action was to defect, then defect
+    with probability 60%.
+
+    Names:
```
drvinceknight: Blank line after `Names` please.

axelrod/strategies/axelrod_second.py
```diff
@@ -466,3 +467,107 @@ def strategy(self, opponent: Player) -> Action:
...
+class Kluepfel(Player):
+    """
+    Strategy submitted to Axelrod's second tournament by Charles Kluepfel
+    (K32R).
+
+    This player keeps track of the the opponent's responses to own behavior:
+
+                      I did:
+                    Def    Coop
+    She does: Coop   C1  |  C3
```
drvinceknight: Gender neutral please. Go with `player` for your current `I` and `opponent` for your current `she`. Although it's probably also worth changing this to a bullet point list (only because the rest syntax can be troublesome at times). I'd suggest:

```
This player keeps track of the the opponent's responses to own behavior:

- C1 counts: Opponent cooperates as response to player defecting.
- C2 counts: Opponent defects as response to player defecting.
- C3 counts: Opponent cooperates as response to player cooperating.
- C4 counts: Opponent defects as response to player cooperating.
```

Could we however change these variable names (unless there's a good reason) to be more explicit. Perhaps:

```
- `cd_count` counts: Opponent cooperates as response to player defecting.
- `dd_count` counts: Opponent defects as response to player defecting.
- `cc_count` counts: Opponent cooperates as response to player cooperating.
- `dc_count` counts: Opponent defects as response to player cooperating.
```

axelrod/strategies/axelrod_second.py
gaffney2010: Will do.  Changing test file too.

gaffney2010:
Thanks for all the feedback.  I made changes to the code, and will work on documentation for the second tournament a little later.  I'm not sure I understand the issue with test_meta.py...

drvinceknight:
> I'm not sure I understand the issue with test_meta.py...

There's a strategy in `meta.py` that makes use of a team of other strategies. It is called `MetaLongMemory` and with the addition of your strategy it's behaviour has changed so it just needs it's test to be updated. Currently at line 338 of `test_meta.py` we have:

```
        actions = [(C, C), (C, D), (D, C), (C, D), (D, C)]
        self.versus_test(opponent=axelrod.Alternator(),
                         expected_actions=actions, seed=1)
```

So in the third round the expected action is `(D, C)` however when making use of your strategy `MetaLongMemory` in fact cooperates in the third round so that line just needs to be changed to be:

```
        actions = [(C, C), (C, D), (C, C), (C, D), (D, C)]
        self.versus_test(opponent=axelrod.Alternator(),
                         expected_actions=actions, seed=1)
```

axelrod/strategies/axelrod_second.py
```diff
@@ -474,30 +474,29 @@ class Kluepfel(Player):
...
-                   -------------
-              Def    C2  |  C4
+    - `cd_count` counts: Opponent cooperates as response to player defecting.
+    - `dd_count` counts: Opponent defects as response to player defecting.
+    - `cc_count` counts: Opponent cooperates as response to player cooperating.
+    - `dc_count` counts: Opponent defects as response to player cooperating.
     
     After 26 turns, the player then tries to detect a random player.  The
     player decides that the opponent is random if
     C1 >= (C1+C2)/2 - 1.5*sqrt(C1+C2)/2  AND
```
drvinceknight: These (and anywhere else) needs to be updated to `cd_count, ...`

drvinceknight:
I have run fingerprints of this implementation and `K32r` and it looks like a good match:

```python
>>> import axelrod as axl
>>> fortran_player = axlf.Player("k32r")
>>> axl_player = axl.Kluepfel()
>>> repetitions = 10000
>>> fortran_player, axl_player, repetitions
(Player: k32r, libstrategies.so, Kluepfel, 10000)
```

```python
>>> axl_tf = axl.TransitiveFingerprint(axl_player)
>>> axl_tf.fingerprint(repetitions=repetitions, processes=0)
>>> axl_tf.plot();
```

download


```python
>>> fortran_tf = axl.TransitiveFingerprint(fortran_player)
>>> fortran_tf.fingerprint(repetitions=repetitions, processes=0)
>>> fortran_tf.plot();
```

download 1


```python
>>> opponents = [s() for s in axl.basic_strategies]
>>> axl_tf_v_basic = axl.TransitiveFingerprint(axl_player, opponents=opponents)
>>> axl_tf_v_basic.fingerprint(repetitions=repetitions, processes=0)
>>> axl_tf_v_basic.plot(display_names=True);
```

download 2


```python
>>> fortran_tf_v_basic = axl.TransitiveFingerprint(fortran_player, opponents=opponents)
>>> fortran_tf_v_basic.fingerprint(repetitions=repetitions, processes=0)
>>> fortran_tf_v_basic.plot(display_names=True);
```

download 3

Before merging having someone else double check the logic would be great but this is looking good to me :+1:

drvinceknight:
Re a7a6c1b

>  Was only using this for sqrt.  Seemed like it was causing problems, so I replaced with half powers.

Using `sqrt` makes the code ever so slightly more readable in my personal opinion. Not sure what trouble was being caused by using `np.sqrt` but perhaps use `math.sqrt` (from standard lib)? (Not a big deal).

axelrod/strategies/axelrod_second.py
```diff
@@ -466,3 +466,101 @@ def strategy(self, opponent: Player) -> Action:
...
+
+    This player keeps track of the the opponent's responses to own behavior:
+    - `cd_count` counts: Opponent cooperates as response to player defecting.
+    - `dd_count` counts: Opponent defects as response to player defecting.
+    - `cc_count` counts: Opponent cooperates as response to player cooperating.
+    - `dc_count` counts: Opponent defects as response to player cooperating.
+    
+    After 26 turns, the player then tries to detect a random player.  The
+    player decides that the opponent is random if
+    C1 >= (C1+C2)/2 - 1.5*sqrt(C1+C2)/2  AND
```
drvinceknight: These (and anywhere else) needs to be updated to cd_count, ...

gaffney2010:
> I have run fingerprints of this implementation and K32r and it looks like a good match

In the future is this something I should run when making a strategy?

gaffney2010:
Sorry.  I'm still trying to figure out this test piece.  I ran unittest discover on my end and was able to get the same error, and I can see that making the change fixes it.  But it looks like this test gets run for every infinite memory ordinary strategy.  How come changing this code doesn't break other strategies?  [I think that maybe the unittest detects changed code, but then will this break future changes?  I don't see why we should expect all infinite memory ordinary strategies to have this expected action.]

gaffney2010:
Alright.  I think I've fixed everything.  Please let me know if anything else needs to be fixed.  Thanks again for the help.

drvinceknight:
> But it looks like this test gets run for every infinite memory ordinary strategy. How come changing this code doesn't break other strategies?

That's not quite what's happening. For info: 

- the meta strategy has a team of strategies that it all gets to play (in this case it's team is all the infinite memory strategy), 
- every member of the team plays against the opponent but that's all done internally
- when it comes to the meta strategy making a decision it asks every member of the team to declare what it would do and then decides what to do based on that "vote" (in this case it just selects `C` or `D` depending on which action gets the majority).

So with the addition of your strategy to the team, the majority of the players in this third round turns out to be `C` (as opposed to `D`).

Don't worry too much if that's still unclear, essentially these meta players are black boxes and their tests are more regression tests than anything else. Sometimes when adding new strategies their behaviour (and so their tests) changes :)

axelrod/strategies/axelrod_second.py
```diff
@@ -466,3 +467,103 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        # Check for randomness
+        if len(self.history) > 26:
+            if self.cd_counts >= (self.cd_counts+self.dd_counts)/2 - 0.75*np.sqrt(self.cd_counts+self.dd_counts) and \
+                self.cc_counts >= (self.dc_counts+self.cc_counts)/2 - 0.75*np.sqrt(self.dc_counts+self.cc_counts):
+                return D
+
+        # Otherwise respond to recent history
+
+        # Use the defaults from Python
```
drvinceknight: I don't understand what this inline comment is saying. Perhaps remove?

drvinceknight:
> In the future is this something I should run when making a strategy?

If you could that would be super helpful. Depending on how much cpu time you have free feel free to run them with less repetitions.

axelrod/strategies/axelrod_second.py
gaffney2010: Yea, I'll remove it.

meatballs:
Excellent Job @gaffney2010 Thank you very much indeed!

PR #1141, opened 2017-11-18
Implemented strategy Cave (k49r from Axelrod's Second)

gaffney2010:
Fingerprints below:

basic_fortran_cave
basic_python_cave
prob_fortran_cave
prob_python_cave

### DIFF
--- a/axelrod/tests/strategies/test_axelrod_second.py
+++ b/axelrod/tests/strategies/test_axelrod_second.py
@@ -519,14 +519,14 @@ class TestCave(TestPlayer):
 
         # Highly-defective opponent
         # It will take until turn 20 to respond decide to repond D to C
-        opponent_actions = [D] * 16 + [C, C, C, C, C]
+        opponent_actions = [D] * 17 + [C, C, C, C]
         almost_defector = axelrod.MockPlayer(actions=opponent_actions)
 
         random.seed(1)
         actions = [(C, D)]
         actions += [(C, D), (D, D), (D, D), (C, D), (C, D), (C, D), (D, D),
                     (D, D), (C, D), (C, D), (D, D), (C, D), (D, D), (C, D),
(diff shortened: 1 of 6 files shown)


gaffney2010:
Noticed a slight difference in the fingerprints.  Found that the Fortran was calculating percentage wrong.  Put the same logic in my code.  Fingerprints match better now.
basic_fortran_cave
basic_python_cave
prob_fortran_cave
prob_python_cave

gaffney2010:
Fixed one of my tests for the calculation issue I fixed in last commit.

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -490,3 +490,55 @@ def test_strategy(self):
...
+        self.versus_test(axelrod.Cooperator(), expected_actions=actions)
+
+        # It will take until turn 18 to respond decide to repond D->D
+        random.seed(1)
+        actions = [(C, D)]
+        actions += [(C, D), (D, D), (D, D), (C, D), (C, D), (C, D), (D, D),
+                    (D, D), (C, D), (C, D), (D, D), (C, D), (D, D), (C, D),
+                    (C, D), (D, D), (C, D)] # Randomly choose
+        actions += [(D, D)] * 30 # Defect
+        self.versus_test(axelrod.Defector(), expected_actions=actions)
```
drvinceknight: Can we put the seed argument in to `self.versus_test` please:

```
self.versus_test(axelrod.Defector(), expected_actions=actions, seed=1)
```

Same below.

drvinceknight:
I have gone through the logic and agree this strategy is correctly implemented (as well as the great match of the fps).

drvinceknight:
This is looking good to me. Thanks again @gaffney2010!

PR #1142, opened 2017-11-21
Implemented WmAdams strategy (k44r from Axelrod's second)

gaffney2010:
Fingerprints below.
basic_fortran
basic_python
random_fortran
random_python


gaffney2010:
I accidentally edited/uploaded the version of test_axelrod_second that had the random.seed() calls.  I'm trying to fix it, but failing I'm afraid.

drvinceknight:
There's a merge conflict due to difference in `test_axelrod_second`.

Because you're working on the `master` branch, I'd suggest you `git pull upstream` (where you set upstream to be the remote corresponding to this repo) and fix thing locally.

Let me know if you're comfortable sorting this out or if I can assist (ping me on gitter).

drvinceknight:
> I accidentally edited/uploaded the version of test_axelrod_second that had the random.seed() calls. I'm trying to fix it, but failing I'm afraid.

No problem, messages crossed :) Going to gitter as it'll be easier to do this in sync.

axelrod/strategies/axelrod_second.py
```diff
@@ -771,3 +771,40 @@ def strategy(self, opponent: Player) -> Action:
...
+        'stochastic': True,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if len(self.history) <=1:
```
marcharper: Space after =

axelrod/strategies/axelrod_second.py
```diff
@@ -771,3 +771,40 @@ def strategy(self, opponent: Player) -> Action:
...
+            return C
+        did_d = np.vectorize(lambda action: float(action == D))
```
marcharper: We keep a running variable of the number of defections in `opponent.defections`. Please use that instead, and subtract one if the first move is a defection. That way is `O(rounds)`, the current implementation is `O(rounds^2)`.

axelrod/strategies/axelrod_second.py
```diff
@@ -771,3 +771,40 @@ def strategy(self, opponent: Player) -> Action:
...
+        number_defects = np.sum(did_d(opponent.history[1:]))
+
+        if number_defects == 4 or number_defects == 7 or number_defects == 9: return D
```
marcharper: nit: `if number_defects in [4, 7, 9]: return D` is cleaner

axelrod/strategies/axelrod_second.py
gaffney2010: Wondered about this.  I can fix in my earlier "Cave" implementation too.

gaffney2010:
I may not get to fixing these until after the Holiday (say Friday).  Let me know if I should close the PR and re-open then.  Thanks.

axelrod/strategies/axelrod_second.py
drvinceknight: If you could that would be great (my bad for missing that in the review :)).

drvinceknight:
> I may not get to fixing these until after the Holiday (say Friday).

No rush at all, happy thanksgiving for tomorrow :+1:

> Let me know if I should close the PR and re-open then. Thanks.

Whatever is easier for you, if you're able to fix the merge conflict (we can help) then it's fine to keep it on this PR but if you'd prefer to close and open another PR that's also cool.

gaffney2010:
I've made the changes you've requested.  Thanks

axelrod/strategies/axelrod_second.py
```diff
@@ -771,3 +771,41 @@ def strategy(self, opponent: Player) -> Action:
...
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        if len(self.history) <= 1:
+            return C
+        number_defects = opponent.defections
+        if opponent.history[0] == D:
+            number_defects -= 1
+
+        if number_defects in [4, 7, 9]: return D
```
drvinceknight: @gaffney2010 can you confirm which strategy this is? Looking at https://github.com/Axelrod-Python/TourExec/blob/master/src/strategies/k44r.f I don't immediately follow how this implementation corresponds to `k44r`?

gaffney2010: I can confirm that it's k44r.  I'll say more about the translation...

drvinceknight: I confirm I get the same fingerprints :) 

I think you've just done a good job refactoring and re writing the code to be much more readable, just got me scratching my head to follow along, so one or two pointers for my own benefit just to confirm would be awesome (even if it's not in the PR itself but just in a comment here).

drvinceknight: The ones I'm particularly stumped by are the comparisons of `number_defects` to 4, 7 and 9 and also where `number_defects - 9` comes from...

gaffney2010: Basically every time a defect comes up, it increments this MC variable.  While AM is still a whole number, then MC is either less than AM or equal to AM.  [While less than, line 16 returns C.]  While MC=AM it defects; this happens for the first time at 4 defects [line 11].  As soon as MC exceeds AM, then AM is halved, and MC is reset to zero.  [The 5th defect resets MC and sets AM to 2. So the 7th defect gets MC=AM.  Then the 8th defect resets MC and sets AM to 1.  So the 9th defect gets MC-AM]

After the 9th defect, the function changes because the AM is no longer whole, so MC never again equal AM (and line 17 is out of play).  So as soon as MC is incremented (any defect), lines 19-22 are triggered.  [Line 22 was 100% until now.]  And MC is immediately reset, so there's no sustained defecting, like there was earlier.  [For example, there could be exactly 4 defects for many turns, and the strategy would defect for all those turns.  But after the 100th defect, the strategy will defect (with some probability) ONLY once.]

drvinceknight: That's awesome and make sense. Thanks for explaining.

gaffney2010: `number_defects - 9` just makes the AM follow the right halving pattern.  AM starts at 4.0 and halves at 4th, 7th, and 9th defects, for AM=1.0.  After that EACH defect will trigger a halving.

drvinceknight:
I'm happy with all the changes and impressed with the refactor/translation of the original! Thanks @gaffney2010 :) :+1:

marcharper:
Changes LGTM, just need to resolve the merge conflict(s)

drvinceknight:
The conflicts were fixed in c20a106. Thanks for the contribution @gaffney2010 :+1:

PR #1143, opened 2017-11-27
Use opponent.defections in Cave and some minor styling changes

marcharper:
Re: discussion on #1142

PR #1144, opened 2017-11-28
 Implemented GraaskampKatzen strategy from Axelrod's Second (k60r)

gaffney2010:
I ran the FP (below), and I also did FP on two custom strategies that are the same as my two custom strategies in the test file.  These are in the second pair of images.  [The reason is that you need some contrived examples to get interesting enough behavior.]

fortran
python
custom_fortran
custom_python

I tried to make the same style changes as in PR #1143.


drvinceknight:
I've run the tests on `axelrod_fotran` with this strategy (it is deterministic) and everything passes. I've also looked through the logic and agree that the strategy is implemented correctly. Thanks @gaffney2010, great stuff!

PR #1145, opened 2017-11-28
Implemented Weiner, k41r from Axelrod's Second

gaffney2010:
Fingerprint below.

Some notes about the translation from Fortran:  The Fortran cycles through these three states, labeled by `ICASE`.  I replaced the third state with a flag, called `forgive_flag`.  The first two states, it goes back and forth between these as the opponent Defects; when the opponent eventually Cooperates, this strategy goes to state three (raises the `forgive_flag`) if and only if it's in the second state at that point (i.e. there have been an odd number of Defects since last Cooperate).  I wrote the code to replace the first two states with a parity-check on the number of Defects between Cooperates, which I call `defect_padding`.  This reworking should make it a little clearer, and draws out the infinite memory depth that wasn't clear with the original code.

basic_fortran
basic_python

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -595,7 +595,7 @@ class Borufsen(Player):
     re-evaluate the mode.  Enter defect mode if any of the following
     conditions hold:
 
-    - Detected random:  Opponent cooperated 7-18 times since last mode 
+    - Detected random:  Opponent cooperated 7-18 times since last mode
       evaluation (or start) AND less than 70% of opponent cooperation was in
       response to player's cooperation, i.e.
       cc_count / (cc_count+cd_count) < 0.7
@@ -757,11 +757,12 @@ class Cave(Player):
 
     def strategy(self, opponent: Player) -> Action:
         turn = len(self.history) + 1
-        if turn == 1: return C
+        if turn == 1:
+            return C
 
         number_defects = opponent.defections
         perc_defects = number_defects / turn
-        
+
         # If overly defect or appears random
         if turn > 39 and perc_defects > 0.39:
             return D
@@ -812,10 +813,12 @@ class WmAdams(Player):
         if opponent.history[0] == D:
             number_defects -= 1
 
-        if number_defects in [4, 7, 9]: return D
+        if number_defects in [4, 7, 9]:
+            return D
         if number_defects > 9 and opponent.history[-1] == D:
             return random_choice((0.5) ** (number_defects - 9))
         return C
+
 
 class GraaskampKatzen(Player):
     """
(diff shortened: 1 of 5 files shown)


drvinceknight:
(FYI, I have run the tests on `axelrod_fortran` with this deterministic strategy and they pass.)

axelrod/strategies/axelrod_second.py
```diff
@@ -812,11 +813,13 @@ def strategy(self, opponent: Player) -> Action:
         if opponent.history[0] == D:
             number_defects -= 1
 
-        if number_defects in [4, 7, 9]: return D
+        if number_defects in [4, 7, 9]:
```
drvinceknight: Hey @gaffney2010, it looks like you've copied this manually from https://github.com/Axelrod-Python/Axelrod/pull/1143?

No big deal at all, I suspect you've done this to try and avoid merge conflicts going forward but for info, that's not considered best practice in terms of etiquette. A better way to do this would be to "go get" the actual git commits that make this change, there are various ways to do this (for example a command called `git cherry-pick`. One of the main reasons for this is that it ends up causing problems on other people's branches (as well as potentially erasing their contributions in terms of git commit logs).

Not a big problem at all, and not suggesting you change anything (although once #1143 goes in you might end up with a merge conflict here), just letting you know :+1: :)

drvinceknight: @gaffney2010 now that #1143 has gone in there are a couple of merge conflicts. Let me know if you need a hand :+1:

gaffney2010: Okay.  I can see why it would be better to go get the changes directly from the commit.  How can you tell that I did it manually.  Did I miss something or does it show up differently?

drvinceknight: No it shows up the same but the commit that made the code on #1143 isn't in the history here so the conclusion is that you did it manually.

Not a big deal :)

drvinceknight:
I believe this is good to go now @marcharper @meatballs. :+1:

marcharper:
@gaffney2010 Please resolve the merge conflict(s).

gaffney2010:
@marcharper Sorry.  I'm not seeing how to do that.  [I think I saw something earlier that said fix merge conflicts, but that's not there now.]  ...

drvinceknight:
> [I think I saw something earlier that said fix merge conflicts, but that's not there now.] ...

I don't believe there are any merge conflicts on this anymore and it's good to go. Let's wait for @marcharper to perhaps expand on what he meant in case he's noticed something we're both missing. :+1:

marcharper:
Under "rebase and merge" (my default) github had a merge conflicts warning. However "merge pull request" worked.

PR #1146, opened 2017-12-04
Implemented Harrington (k75r from Axelrod's Second)

gaffney2010:
Below are the fingerprints for the basic strategies, the random, and against the six custom strategies that I used in the test file.

I've also attached a PDF that bridges the gap from the Fortran to what I have, along with some small amount of commentary on the code.

[HarringtonNotes.pdf](https://github.com/Axelrod-Python/Axelrod/files/1527303/HarringtonNotes.pdf)

additional_fortran
additional_python
basic_fortran
basic_python
prob_fortran
prob_python

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -979,10 +979,11 @@ class Weiner(Player):
 
 class Harrington(Player):
     """
-    Strategy submitted to Axelrod's second tournament by Paul Harringtn (K75R),
+    Strategy submitted to Axelrod's second tournament by Paul Harrington (K75R)
     and came in eighth in that tournament.
 
-    This strategy has three modes:  Normal, and Fair-weather, Defect.
+    This strategy has three modes:  Normal, Fair-weather, and Defect.  These
+    mode names were not present in Harrington's submission.
 
     In Normal and Fair-weather modes, the strategy begins by:
 
@@ -1013,13 +1014,17 @@ class Harrington(Player):
     The player mostly plays Tit-for-Tat for the first 36 moves, then defects on
     the 37th move.  If the opponent cooperates on the first 36 moves, and
     defects on the 37th move also, then enter Fair-weather mode and cooperate
-    this turn.
+    this turn.  Entering Fair-weather mode is extremely rare, since this can
+    only happen if the opponent cooperates for the first 36 then defects
+    unprovoked on the 37th.  (That is, this player's first 36 moves are also
+    Cooperations, so there's nothing really to trigger an opponent Defection.)
 
     Next in Normal Mode:
 
     1. Check for defect and parity streaks.
     2. Check if cooperations are scheduled.
     3. Otherwise,
+
        - If turn < 37, Tit-for-Tat.
        - If turn = 37, defect, mark this move as generous, and schedule two
          more cooperations**.
@@ -1224,10 +1229,6 @@ class Harrington(Player):
 
         if turn == 38 and opponent.history[-1] == D and opponent.cooperations == 36:
             self.mode = "Fair-weather"
-            # These flags would already be set from turn == 37 logic below.
-            # Just take care to not lower this turn.
-            # self.more_coop = 2
-            # self.generous_n_turns_ago = 1 # 1 turn ago since this turn is ending.
             return self.try_return(to_return=C, lower_flags=False)
 
 
@@ -1256,13 +1257,12 @@ class Harrington(Player):
 
         if turn < 37:
             return self.try_return(opponent.history[-1], inc_parity=True)
-        elif turn == 37:
+        if turn == 37:
             self.more_coop, self.generous_n_turns_ago = 2, 1
             return self.try_return(D, lower_flags=False)
+        if self.burned or random.random() > self.prob:
+            return self.try_return(opponent.history[-1], inc_parity=True)
         else:
-            if self.burned or random.random() > self.prob:
-                return self.try_return(opponent.history[-1], inc_parity=True)
-            else:
-                self.prob += 0.05
-                self.more_coop, self.generous_n_turns_ago = 2, 1
-                return self.try_return(D, lower_flags=False)
+            self.prob += 0.05
+            self.more_coop, self.generous_n_turns_ago = 2, 1
+            return self.try_return(D, lower_flags=False)
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1025,13 +1025,13 @@ class Harrington(Player):
     2. Check if cooperations are scheduled.
     3. Otherwise,
 
-       - If turn < 37, Tit-for-Tat.
-       - If turn = 37, defect, mark this move as generous, and schedule two
-         more cooperations**.
-       - If turn > 37, then if `burned` flag is raised, then Tit-for-Tat.
-         Otherwise, Tit-for-Tat with probability 1 - `prob`.  And with
-         probability `prob`, defect, schedule two cooperations, mark this move
-         as generous, and increase `prob` by 5%.
+    - If turn < 37, Tit-for-Tat.
+    - If turn = 37, defect, mark this move as generous, and schedule two
+      more cooperations**.
+    - If turn > 37, then if `burned` flag is raised, then Tit-for-Tat.
+      Otherwise, Tit-for-Tat with probability 1 - `prob`.  And with
+      probability `prob`, defect, schedule two cooperations, mark this move
+      as generous, and increase `prob` by 5%.
 
     ** Scheduling two cooperations means to set `more_coop` flag to two.  If in
     Normal mode and no streaks are detected, then the player will cooperate and
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -988,18 +988,16 @@ class Harrington(Player):
     In Normal and Fair-weather modes, the strategy begins by:
 
     - Update history
-    - Detects random if turn is multiple of 15 and >=30.
+    - Try to detect random opponent if turn is multiple of 15 and >=30.
     - Check if `burned` flag should be raised.
     - Check for Fair-weather opponent if turn is 38.
 
     Updating history means to increment the correct cell of the `move_history`.
     `move_history` is a matrix where the columns are the opponent's previous
-    move and rows are indexed by the combo of this player and the opponent's
-    moves two turns ago*.  [The upper-left cell must be all cooperations, but
-    otherwise order doesn't matter.]   * If the player is exiting Defect mode,
-    then the history to determine the row is taken from before the turn that
-    the player entered Defect mode.  (That is, the turn that started in Normal
-    mode, but ended in Defect mode.)
+    move and the rows are indexed by the combo of this player's and the
+    opponent's moves two turns ago.  [The upper-left cell must be all
+    Cooperations, but otherwise order doesn't matter.]  After we enter Defect
+    mode, `move_history` won't be used again.
 
     If the turn is a multiple of 15 and >=30, then attempt to detect random.
     If random is detected, enter Defect mode and defect immediately.  If the
@@ -1046,10 +1044,10 @@ class Harrington(Player):
     parity streak that we're pointing to.  If the parity streak that we're
     pointing to is then greater than `parity_limit` then reset the streak and
     cooperate immediately.  `parity_limit` is initially set to five, but after
-    its been hit eight times, it decreases to three.  The parity streak that
-    we're pointing to also gets incremented if in normal mode and WE defect but
-    not on turn 38, unless the result of a defect streak.  Note that the parity
-    streaks reset but the defect streak doesn't.
+    it has been hit eight times, it decreases to three.  The parity streak that
+    we're pointing to also gets incremented if in normal mode and we defect but
+    not on turn 38, unless we are defecting as the result of a defect streak.
+    Note that the parity streaks resets but the defect streak doesn't.
 
     If `more_coop` >= 1, then we cooperate and lower that flag here, in Normal
     mode after checking streaks.  Still lower this flag if cooperating as the
@@ -1086,33 +1084,46 @@ class Harrington(Player):
     def __init__(self):
         super().__init__()
         self.mode = "Normal"
-        self.recorded_defects = 0
-        self.exit_defect_meter = 0
-        self.coops_in_first_36 = None
-        self.was_defective = False
+        self.recorded_defects = 0  # Count opponent defects after turn 1
+        self.exit_defect_meter = 0  # When >= 11, then exit defect mode.
+        self.coops_in_first_36 = None  # On turn 37, count cooperations in first 36
+        self.was_defective = False  # Previously in Defect mode
 
-        self.prob = 0.25
+        self.prob = 0.25  # After turn 37, probability that we'll defect
 
         self.move_history = np.zeros([4, 2])
+        # Will cache value only for testing purposes, not used otherwise
         self.chi_squared = None
         self.history_row = 0
 
-        self.more_coop = 0
-        self.generous_n_turns_ago = 3
+        self.more_coop = 0  # This schedules cooperation for future turns
+        # Initial last_generous_n_turns_ago to 3 because this counts up and
+        # triggers a strategy change at 2.
+        self.last_generous_n_turns_ago = 3  # How many tuns ago was a "generous" move
         self.burned = False
 
         self.defect_streak = 0
-        self.parity_streak = [0, 0]
-        self.parity_bit = 0
-        self.parity_limit = 5
-        self.parity_hits = 0
+        self.parity_streak = [0, 0]  # Counters that get (almost) alternatively incremented.
+        self.parity_bit = 0  # Which parity_streak to increment
+        self.parity_limit = 5  # When a parity streak hits this limit, alter strategy.
+        self.parity_hits = 0  # Counts how many times a parity_limit was hit.
+        # After hitting parity_hits 8 times, lower parity_limit to 3.
 
     def try_return(self, to_return, lower_flags=True, inc_parity=False):
+        """
+        This will return to_return, with some end-of-turn logic.
+        """
+
         if lower_flags and to_return == C:
+            # In most cases when Cooperating, we want to reduce the number that
+            # are scheduled.
             self.more_coop -= 1
-            self.generous_n_turns_ago += 1
+            self.last_generous_n_turns_ago += 1
 
         if inc_parity and to_return == D:
+            # In some cases we increment the `parity_streak` that we're on when
+            # we return a Defection.  In detect_parity_streak, `parity_streak`
+            # counts opponent's Defections.
             self.parity_streak[self.parity_bit] += 1
 
         return to_return
@@ -1131,9 +1142,6 @@ class Harrington(Player):
         We say this is modified because it differs from a usual Chi-Squared
         test in that:
 
-        - It divides by turns minus 2 to get expected, whereas usually we'd
-          divide by matrix total.  Total equals turns minus 1, unless Defect
-          mode has been entered at any point.
         - Terms where expected counts are less than 1 get excluded.
         - There's a check at the beginning on the first cell of the matrix.
         - There's a check at the beginning for the recorded number of defects.
@@ -1146,17 +1154,21 @@ class Harrington(Player):
         if self.recorded_defects / denom < 0.25 or self.recorded_defects / denom > 0.75:
             return False
 
-        expected_matrix = np.outer(self.move_history.sum(axis=1), \
-                                    self.move_history.sum(axis=0))
+        # In each cell, we expect (for an independent distribution) the total
+        # number of recorded turns times the portion in that row times the
+        # portion in that column
+        expected_matrix = np.outer(self.move_history.sum(axis=1),
+                                   self.move_history.sum(axis=0)) / denom
 
         chi_squared = 0.0
         for i in range(4):
             for j in range(2):
-                expct = expected_matrix[i, j] / denom
-                if expct > 1.0:
-                    chi_squared += (expct - self.move_history[i, j]) ** 2 / expct
+                expect = expected_matrix[i, j]
+                if expect > 1.0:
+                    chi_squared += (expect - self.move_history[i, j]) ** 2 / expect
 
-        self.chi_squared = round(chi_squared, 3) # For testing
+        # Caching value only for testing purposes, not used otherwise
+        self.chi_squared = round(chi_squared, 3)
 
         if chi_squared > 3:
             return False
@@ -1164,7 +1176,7 @@ class Harrington(Player):
 
     def detect_streak(self, last_move):
         """
-        Return if and only if the opponent's last twenty moves are defects.
+        Return true if and only if the opponent's last twenty moves are defects.
         """
 
         if last_move == D:
@@ -1176,7 +1188,17 @@ class Harrington(Player):
         return False
 
     def detect_parity_streak(self, last_move):
-        self.parity_bit = 1 - self.parity_bit # Flip bit
+        """
+        Switch which `parity_streak` we're pointing to and incerement if the
+        opponent's last move was a Defection.  Otherwise reset the flag.  Then
+        return true if and only if the `parity_streak` is at least
+        `parity_limit`.
+
+        This is similar to detect_streak with alternating streaks, except that
+        these streaks get incremented elsewhere as well.
+        """
+
+        self.parity_bit = 1 - self.parity_bit  # Flip bit
         if last_move == D:
             self.parity_streak[self.parity_bit] += 1
         else:
@@ -1191,10 +1213,12 @@ class Harrington(Player):
             return C
 
         if self.mode == "Defect":
+            # There's a chance to exit Defect mode.
             if opponent.history[-1] == D:
                 self.exit_defect_meter += 1
             else:
                 self.exit_defect_meter -= 3
+            # If opponent has been mostly defecting.
             if self.exit_defect_meter >= 11:
                 self.mode = "Normal"
                 self.was_defective = True
@@ -1226,10 +1250,12 @@ class Harrington(Player):
         if self.history[-1] == D:
             self.history_row += 2
 
-        # If generous 2 turn ago and opponent defected last turn
-        if self.generous_n_turns_ago == 2 and opponent.history[-1] == D:
+        # If generous 2 turns ago and opponent defected last turn
+        if self.last_generous_n_turns_ago == 2 and opponent.history[-1] == D:
             self.burned = True
 
+        # Only enter Fair-weather mode if the opponent Cooperated the first 37
+        # turns then Defected on the 38th.
         if turn == 38 and opponent.history[-1] == D and opponent.cooperations == 36:
             self.mode = "Fair-weather"
             return self.try_return(to_return=C, lower_flags=False)
@@ -1237,8 +1263,8 @@ class Harrington(Player):
 
         if self.mode == "Fair-weather":
             if opponent.history[-1] == D:
-                self.mode = "Normal" # Post-Defect is not possible
-                #Continue below
+                self.mode = "Normal"  # Post-Defect is not possible
+                # Proceed with Normal mode this turn.
             else:
                 # Never defect against a fair-weather opponent
                 return self.try_return(C)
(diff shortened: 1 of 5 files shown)


axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,294 @@ def strategy(self, opponent: Player) -> Action:
...
+    the 37th move.  If the opponent cooperates on the first 36 moves, and
+    defects on the 37th move also, then enter Fair-weather mode and cooperate
+    this turn.
+
+    Next in Normal Mode:
+
+    1. Check for defect and parity streaks.
+    2. Check if cooperations are scheduled.
+    3. Otherwise,
+       - If turn < 37, Tit-for-Tat.
```
drvinceknight: Need a blank line here (this is what's failing on the CI):

```
    3. Otherwise,
        
       - If turn ...
```

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,294 @@ def strategy(self, opponent: Player) -> Action:
                 self.defect_padding = 0
 
             return self.try_return(opponent.history[-1])
+
+
+class Harrington(Player):
+    """
+    Strategy submitted to Axelrod's second tournament by Paul Harringtn (K75R),
```
drvinceknight: `Harrington`

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,294 @@ def strategy(self, opponent: Player) -> Action:
...
+    and came in eighth in that tournament.
+
+    This strategy has three modes:  Normal, and Fair-weather, Defect.
```
drvinceknight: `Normal, Fair-weather and Defect`

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,294 @@ def strategy(self, opponent: Player) -> Action:
...
+            if self.parity_hits >= 8:
+                self.parity_limit = 3
+            return self.try_return(C, inc_parity=True) # Inc parity won't get used here.
+
+        if self.more_coop >= 1:
+            return self.try_return(C, inc_parity=True)
+
+        if turn < 37:
+            return self.try_return(opponent.history[-1], inc_parity=True)
+        elif turn == 37:
```
drvinceknight: This can just be `if turn == 37`

axelrod/strategies/axelrod_second.py
drvinceknight: Looking at the source code https://github.com/Axelrod-Python/TourExec/blob/master/src/strategies/K75R.f, these names are of your choosing? Nothing wrong with that. Perhaps, we can add a sentence saying that though (just to avoid potential confusion):

```
..., and Defect. These mode names were not present in Harrington's submission.
```

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,294 @@ def strategy(self, opponent: Player) -> Action:
...
+        if self.history[-1] == D:
+            self.history_row += 2
+
+        # If generous 2 turn ago and opponent defected last turn
+        if self.generous_n_turns_ago == 2 and opponent.history[-1] == D:
+            self.burned = True
+
+        if turn == 38 and opponent.history[-1] == D and opponent.cooperations == 36:
+            self.mode = "Fair-weather"
+            # These flags would already be set from turn == 37 logic below.
```
drvinceknight: I think we can remove these two lines of inline comments `These flags ...lower this turn.`

axelrod/strategies/axelrod_second.py
drvinceknight: Could we also add a sentence along the lines of what you wrote in the pdf: that the `Fair-weather` mode occurs only in 1 very specific type of situation.

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -694,3 +694,128 @@ def test_strategy(self):
...
+        actions += [(C, D), (C, D)]
+        # Repeat
+        actions += [(D, D), (D, D), (D, D), (D, D), (C, D), (C, D)] * 2
+        actions += [(D, D)]
+        # 20 D have passed
+        actions += [(D, D)] * 100
+        # The defect streak will always be detected from here on, because it
+        # doesn't reset.  This logic comes before parity streaks or the turn-
+        # based logic.
+        self.versus_test(axelrod.Defector(), expected_actions=actions)
```
drvinceknight: Another thought, could we add a few `attrs={"mode": "Normal", "recorded_defects": ...}` to the `versus_test` call so that we specifically test some of these attribute values? Not necessarily overkill but at least the `mode` and perhaps one or two others as you see fit?

gaffney2010: That's cool.  Didn't know that was possible.

gaffney2010:
This new error is confusing to me.  Any idea?

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,294 @@ def strategy(self, opponent: Player) -> Action:
...
+        if turn == 1:
+            return C
+
+        if self.mode == "Defect":
+            if opponent.history[-1] == D:
+                self.exit_defect_meter += 1
+            else:
+                self.exit_defect_meter -= 3
+            if self.exit_defect_meter >= 11:
+                self.mode = "Normal"
```
drvinceknight: Coverage is failing because we don't have a test case that runs this code block:

```
axelrod/strategies/axelrod_second.py                       436      4    99%   1196-1199
```

(That's the output of coveralls showing that lines 1196-1199 are not hit.)

gaffney2010: Oh, interesting!  I'll work on it.

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+    Strategy submitted to Axelrod's second tournament by Paul Harrington (K75R)
+    and came in eighth in that tournament.
+
+    This strategy has three modes:  Normal, Fair-weather, and Defect.  These
+    mode names were not present in Harrington's submission.
+
+    In Normal and Fair-weather modes, the strategy begins by:
+
+    - Update history
+    - Detects random if turn is multiple of 15 and >=30.
```
marcharper: randomly

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+    38th turn Fair-weather check will set this.
+
+    If the opponent's last twenty moves were defections, then defect this turn.
+    Then check for a parity streak, by flipping the parity bit (there are two
+    streaks that get tracked which are something like odd and even turns, but
+    this flip bit logic doesn't get run every turn), then incrementing the
+    parity streak that we're pointing to.  If the parity streak that we're
+    pointing to is then greater than `parity_limit` then reset the streak and
+    cooperate immediately.  `parity_limit` is initially set to five, but after
+    its been hit eight times, it decreases to three.  The parity streak that
```
marcharper: it has

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+    we're pointing to also gets incremented if in normal mode and WE defect but
```
marcharper: we

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+    not on turn 38, unless the result of a defect streak.  Note that the parity
```
marcharper: unless we are defecting as a result of defect streak?

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+    streaks reset but the defect streak doesn't.
```
marcharper: resets

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+        'memory_depth': float('inf'),
+        'stochastic': True,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self):
```
marcharper: Some inline comments about these variables would be helpful

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+        self.burned = False
+
+        self.defect_streak = 0
+        self.parity_streak = [0, 0]
+        self.parity_bit = 0
+        self.parity_limit = 5
+        self.parity_hits = 0
+
+    def try_return(self, to_return, lower_flags=True, inc_parity=False):
+        if lower_flags and to_return == C:
```
marcharper: docstring please

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        Pearson's Chi Squared statistic = sum[ (E_i-O_i)^2 / E_i ], where O_i
+        are the observed matrix values, and E_i is calculated as number (of
+        defects) in the row times the number in the column over (total number
+        in the matrix minus 1).
+
+        We say this is modified because it differs from a usual Chi-Squared
+        test in that:
+
+        - It divides by turns minus 2 to get expected, whereas usually we'd
```
marcharper: to get expected ______?

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        """
+        denom = turn - 2
+
+        if self.move_history[0, 0] / denom >= 0.8:
+            return False
+        if self.recorded_defects / denom < 0.25 or self.recorded_defects / denom > 0.75:
+            return False
+
+        expected_matrix = np.outer(self.move_history.sum(axis=1), \
```
marcharper: no need for \, please align spacing

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+                                    self.move_history.sum(axis=0))
+
+        chi_squared = 0.0
+        for i in range(4):
```
marcharper: A comment about what's going on in this nested loop would be helpful

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+            for j in range(2):
+                expct = expected_matrix[i, j] / denom
```
marcharper: let's use expect or something other than "expct"

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+                if expct > 1.0:
+                    chi_squared += (expct - self.move_history[i, j]) ** 2 / expct
+
+        self.chi_squared = round(chi_squared, 3) # For testing
```
marcharper: Is this round really necessary?

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        if chi_squared > 3:
+            return False
+        return True
+
+    def detect_streak(self, last_move):
+        """
+        Return if and only if the opponent's last twenty moves are defects.
```
marcharper: defections

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+        if last_move == D:
+            self.defect_streak += 1
+        else:
+            self.defect_streak = 0
+        if self.defect_streak >= 20:
+            return True
+        return False
+
+    def detect_parity_streak(self, last_move):
+        self.parity_bit = 1 - self.parity_bit # Flip bit
```
marcharper: docstring please

marcharper: two spaces before # (and several more below)

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+                self.mode = "Defect"
+                return self.try_return(D, lower_flags=False) # Lower_flags not used here.
+
+        # history_row only gets updated if not in Defect mode AND not entering
+        # Defect mode.
+        self.history_row = 1 if opponent.history[-1] == D else 0
+        if self.history[-1] == D:
+            self.history_row += 2
+
+        # If generous 2 turn ago and opponent defected last turn
```
marcharper: 2 turns ago

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+        if self.generous_n_turns_ago == 2 and opponent.history[-1] == D:
```
marcharper: Does "generous_n_turns_ago" mean "last generous n turns ago" or something else? It's ambiguous IMO.

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,297 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        if turn == 38 and opponent.history[-1] == D and opponent.cooperations == 36:
+            self.mode = "Fair-weather"
+            return self.try_return(to_return=C, lower_flags=False)
+
+
+        if self.mode == "Fair-weather":
+            if opponent.history[-1] == D:
+                self.mode = "Normal" # Post-Defect is not possible
+                #Continue below
```
marcharper: Remove this comment

axelrod/strategies/axelrod_second.py
gaffney2010: Right.  I'll clarify.

axelrod/strategies/axelrod_second.py
gaffney2010: No need for...  Backslash?  Space?

marcharper: backslash `\`

axelrod/strategies/axelrod_second.py
gaffney2010: I'm not sure if this is proper, but I'm using self.chi_squared just to help testing.  I'm asserting in the test file that it equals the value I'd expect; which is why I rounded.  I can remove this if you want.

axelrod/strategies/axelrod_second.py
gaffney2010: Yes.  I can add "last_" here.

axelrod/strategies/axelrod_second.py
gaffney2010: I'm trying to express here that this is the only place in the code where, following a mode-switch, we don't immediately return a value.  Instead we actually treat this turn as a "Normal" mode code.  I'll take your advice if you think it's really not worth mentioning.  (Or if I should say more clearly.)

gaffney2010:
@marcharper I'll make these changes and look for places where an in-line could be helpful.  Thanks.

axelrod/strategies/axelrod_second.py
marcharper: It would be better to break this out into a function that can be directly tested, otherwise let's at least have a more detailed comment like "Caching value only for testing purposes, not used otherwise" (and also in the `__init__` function. I'd prefer a separate function.

axelrod/strategies/axelrod_second.py
marcharper: I don't feel strongly about it but it's clear that the code continues to run IMO

marcharper:
Thanks! Some of the Fortran strategies are hard to understand so the extra inline comments are really helpful and will future-proof the code more completely. We don't want to repeat the current situation where the original strategies have very little documentation and we can't really know if they are implemented correctly. For example, there's a very rarely triggered behavior in this strategy which suggests that something wasn't implemented as intended (but we can't really know).

axelrod/strategies/axelrod_second.py
gaffney2010: Actually this whole bullet should be removed; I was miscounting.

axelrod/strategies/axelrod_second.py
drvinceknight: Detects random opponent? (Not overly fussed)

axelrod/strategies/axelrod_second.py
drvinceknight: > I'd prefer a separate function.

I agree. :+1:

gaffney2010: > I'd prefer a separate function.

I'm not sure what you mean by this.

drvinceknight: Ok cool, so the idea would be to break `def detect_random(self, turn):` in to two separate functions.

It would look *something* like this:

```
def detect_random(self, turn):
    chi_squared = calculate_chi_squared(...)
    return chi_squared <= 3:
```

Then we can test the `calculate_chi_squared` function by itself without needed to set `self.chi_squared`:

```
       actions += [(D, C)]
       self.versus_test(axelrod.Random(0.5), expected_actions=actions, seed=10)
       # The history matrix will be [[0, 2], [5, 6], [3, 6], [4, 2]]

      self.assertEqual(round(axelrod_second.calculate_chi_squared([relevant attributes]), 3), 2.395)
```

The `calculate_chi_squared` function would take relevant attributes as inputs and could just be a function in the `axelrod_second` module (not necessarily a method).

Does that make sense?

gaffney2010: It does...  However, I'm using the chi-squared value to test both the chi-squared calculation _and_ the history matrix.  I tried passing history_matrix as an attrs to the versus_test, but I guess assert_equal has a hard time with numpy matrices...

drvinceknight: > It does... However, I'm using the chi-squared value to test both the chi-squared calculation and the history matrix. I tried passing history_matrix as an attrs to the versus_test, but I guess assert_equal has a hard time with numpy matrices...

Yup, testing numpy matrix equality isn't always what you'd hope:

```python
>>> A = np.array([[1, 6], [4, 3]])
>>> A == A
array([[ True,  True],
       [ True,  True]], dtype=bool)
>>> np.array_equal(A, A)
True
```

*but* I don't think that's important here, you can still use those inputs to the chi_squared calculation? Don't worry if you don't pass them to the `versus_test`.

(A note for later/elsewhere, it would be straight forward to add the ability to test `array_equal` for numpy arrays in `versus_test`: just check the type etc...)

gaffney2010: Now that I know it's working, it's fine, but when I was testing, I wanted to know that the history_matrix was getting filled out correctly.  [Lot's of different matrices will give chi_squared < 3.]

drvinceknight:
> @drvinceknight @meatballs I did not try to verify the logic in comparison to the Fortran source.

I believe this is the one last thing that needs to be done review wise apart from the requested changes from @marcharper :+1:

gaffney2010:
> I did not try to verify the logic in comparison to the Fortran source.

I think (hope) my attached PDF in original PR can be helpful here.

drvinceknight:
> I think (hope) my attached PDF in original PR can be helpful here.

yup it really did, to be clear I've gone through it myself and am happy, coupled with the agreement in the fingerprints I'm certain it's implemented correctly. Just in the case of this strategy which is pretty complex (kudos to what must have been a big translation effort @gaffney2010!) I think have a second set of eyes confirm would be good :)

gaffney2010:
In this first commit just now, I modified the comments per Marc.  In the second commit ("Cleaned up record history logic"), I very slightly modified the record history logic.  [Because in the last version I was updating move_history after the player enters and leaves Defect mode, the same way that the Fortran strategy does (which gets complicated after leaving Defect mode).  But I realized that since you can only enter Defect mode once, this matrix isn't being used again, so I just made it stop recording at that point, and was able to clean stuff up.]  Because it was a logic change.  I re-ran all the tests/FPs.  It still looks good, and I can upload the FPs if needed.

drvinceknight:
> It still looks good, and I can upload the FPs if needed.

Sounds good, if you could re upload just so we have record here that'd be great :+1:

gaffney2010:
FPs
additional_fortran
additional_python
basic_fortran
basic_python
prob_fortran
prob_python

axelrod/strategies/axelrod_second.py
```diff
@@ -975,3 +975,329 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        chi_squared = 0.0
+        for i in range(4):
+            for j in range(2):
+                expect = expected_matrix[i, j]
+                if expect > 1.0:
+                    chi_squared += (expect - self.move_history[i, j]) ** 2 / expect
+
+        # Caching value only for testing purposes, not used otherwise
+        self.chi_squared = round(chi_squared, 3)
```
drvinceknight: So it doesn't look like you've gone with the suggestion of making the `chi_squared` calculation it's own function which could be tested?

You could essentially keep this exactly as it is:

```
+ def calculate_chi_squared(self):
+        expected_matrix = np.outer(self.move_history.sum(axis=1),
 +                                   self.move_history.sum(axis=0)) / denom
 +
 +        chi_squared = 0.0
 +        for i in range(4):
 +            for j in range(2):
 +                expect = expected_matrix[i, j]
 +                if expect > 1.0:
 +                    chi_squared += (expect - self.move_history[i, j]) ** 2 / expect
 +
 +        # Caching value only for testing purposes, not used otherwise
 +        return chi_squared
```

Then in `detect_random` you just use `self.calculate_chi_squared` but then, for the test, you can directly run a match for your strategy and call the `self.calculate_chi_squared` method to ensure it gives the correct value (with no need for the `self.chi_squared`.

My other suggestion was to have this as a completely static method/function but it's also fine to leave it as a method of the class.

gaffney2010: I didn't make the change, because it sounded like it was optional AND because I'm still not seeing how, under this alternative approach, I would test that `move_history` is getting filled out right.  If I understand your comment, it was that, this doesn't matter because I can test the Chi-Squared logic on my matrix by manually entering the matrix:

self.assertEqual(round(axelrod_second.calculate_chi_squared(np.matrix([[0, 2], [5, 6], [3, 6], [4, 2]])), 3), 2.395)

While I agree that this would successfully test the chi-squared logic.  How do I test that `move_history` correctly incremented to [[0, 2], [5, 6], [3, 6], [4, 2]] ?

drvinceknight: I'm sure I see the connection between testing the `move_history` and `calculate_chi_squared`. As far as I can tell you're not testing that in the tests anyway:

```
+        actions += [(D, C)]
+        self.versus_test(axelrod.Random(0.5), expected_actions=actions, seed=10, attrs={"chi_squared": 2.395})
+        # The history matrix will be [[0, 2], [5, 6], [3, 6], [4, 2]]
```

You can keep this as a sub method of the class if you prefer in which case nothing here changes except you no longer need to artificially set `self.calculate_chi_squared` as an attribute (which is rounded so not actually an accurate name).

So that test (which if I understand correctly is the only purpose of this rounded chi_squared attribute which we'd rather not need) would be modified to be an actual match. Something like (I haven't run this):

```
axl.seed(10)
player = self.player()
match = axl.Match((player, axl.Random()), turns=len(expected_actions))
actions = match.play()
self.assertEqual(actions, expected_actions)  # Just to be consistant with the correct test.
self.assertEqual(round(player.calculate_chi_squared()), 2.395)
```

gaffney2010: > I'm not sure I see the connection between testing the move_history and calculate_chi_squared.

calculate_chi_squared uses move_history.  You can get 2.935 precisely if move_history is filled out (correctly) as [[0, 2], [5, 6], [3, 6], [4, 2]]

gaffney2010: I think I understand better now...

gaffney2010: > self.assertEqual(round(player.calculate_chi_squared()), 2.395)

I wasn't making the connection that the player would still be in the _end-of-turn-30_ state after you ran the axl.Match(), and I was trying to figure out how I could check everything in the Match() function, because I was imagining that the player would disappear afterwards.  [I guess the way I wrote it, it would disappear.]

I'll change it like you have it.  Thanks.

drvinceknight: > calculate_chi_squared uses move_history. You can get 2.935 precisely if move_history is filled out (correctly) as [[0, 2], [5, 6], [3, 6], [4, 2]]

Cool, so we can either:

1. create `compute_chi_squared` as it's own function (outside the class) that in the test gets passed `np.array([[0, 2], [5, 6], [3, 6], [4, 2]])` as well as `recorded_defects` (which I believe it also uses). So `compute_chi_squared` would be a function that takes two arguments which is in turn called by `detect_random`. Something like:
   
   ```python
   def detect_random(self):
        chi_squared = compute_chi_squared(self.move_history, self.recorded_defects)
        if chi_squared > 3 ...
   ```
   The test would be:

    ```python
     self.assertEqual(round(axelrod_second.calculate_chi_squared(np.matrix([[0, 2], [5, 6], [3, 6], [4, 2]])),  ?), 2.395)      # Replace ? with the correct recorded_defects :)
     ```
   
2. Create a method which means we then need to test using a match.

I think option 1 is probably the cleanest way of doing this but 2 would do the trick also :)

drvinceknight: > I think I understand better now...

Cool :+1:

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -694,3 +695,162 @@ def test_strategy(self):
...
+                   (D, C), (C, D), (D, C), (C, C), (C, D), (D, D), (D, C),
+                   (C, D), (D, D), (D, C), (C, C), (C, D), (D, C), (C, D),
+                   (D, D), (D, C), (C, D), (D, D), (D, D), (C, D), (D, C),
+                   (C, C)]
+        actions += [(D, C)]
+        # Enter defect mode.
+        actions += [(D, D)] * 14
+        # Mutual defect for a while, then exit Defect mode with two coops
+        actions += [(C, D)] * 2
+        self.versus_test(Rand_Then_Def, expected_actions=actions, seed=10, \
```
marcharper: Don't need the slash here

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -694,3 +695,162 @@ def test_strategy(self):
...
+        'memory_depth': float('inf'),
+        'stochastic': True,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_strategy(self):
```
marcharper: Thanks for moving the chi-squared attribute. This test function is pretty long now, you might consider breaking it up into several test functions. (optional)

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -694,3 +695,162 @@ def test_strategy(self):
...
+        self.versus_test(Defect37_big, expected_actions=actions, seed=1, attrs={"mode": "Normal"})
+
+        # Opponent defects on 1st turn
+        opponent_actions = [D] + [C] * 46
+        Defect1 = axelrod.MockPlayer(actions=opponent_actions)
+        # Tit-for-Tat on the first, but no streaks, no Fair-weather flag.
+        actions = [(C, D), (D, C)] + [(C, C)] * 34 + [(D, C)]
+        # Two cooperations scheduled after the 37-turn defection
+        actions += [(C, C)] * 2
+        # TFT twice, then low-random # yields a DCC combo.
```
marcharper: Did you want two lines for this comment?

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -694,3 +695,162 @@ def test_strategy(self):
...
+        actions += [(C, C)] * 2
+        actions += [(D, C), (C, C), (C, C)]
+        # Don't draw next random # until now.  Again DCC.
```
marcharper: same

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -694,3 +695,162 @@ def test_strategy(self):
...
+        actions = [(C, D), (D, C), (C, D), (D, C), (C, D), (C, C)] * 6
+        # Defect on 37th move
+        actions += [(D, D)]
+        actions += [(C, C)]
+        # This triggers the burned flag.  We should just Tit-for-Tat from here.
+        actions += [(C, D)]
+        actions += [(D, C), (C, D), (D, C), (C, D), (C, C)]
+        # This is the seventh time we've hit the limit.  So do it once more.
+        actions += [(C, D), (D, C), (C, D), (D, C), (C, D), (C, C)]
+        # Now hit thi limit sooner
```
marcharper: thi --> this?

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -694,3 +695,162 @@ def test_strategy(self):
...
+                   (C, C)]
+        # Enter defect mode.
+        expected_actions += [(D, C)]
+        random.seed(10)
+        player = self.player()
+        match = axelrod.Match((player, axelrod.Random()), turns=len(expected_actions))
+        # The history matrix will be [[0, 2], [5, 6], [3, 6], [4, 2]]
+        actions = match.play()
+        self.assertEqual(actions, expected_actions)  # Just to be consistant with the current test.
+        self.assertEqual(round(player.calculate_chi_squared(len(expected_actions)),3), 2.395)
```
marcharper: space after comma (before 3); drop the round or use `assertAlmostEqual` (which allows you to specify a number of places for equality):
```
self.assertAlmostEqual(a, b, places=4)
``` 

[Link to the docs](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertAlmostEqual) for `assertAlmostEqual`.

axelrod/tests/strategies/test_axelrod_second.py
marcharper: You could also target specific functions with their own tests, e.g. for `detect_streak` and friends.

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -694,3 +695,162 @@ def test_strategy(self):
...
+                   (D, D), (D, C), (C, D), (D, D), (D, D), (C, D), (D, C),
+                   (C, C)]
+        # Enter defect mode.
+        expected_actions += [(D, C)]
+        random.seed(10)
+        player = self.player()
+        match = axelrod.Match((player, axelrod.Random()), turns=len(expected_actions))
+        # The history matrix will be [[0, 2], [5, 6], [3, 6], [4, 2]]
+        actions = match.play()
+        self.assertEqual(actions, expected_actions)  # Just to be consistant with the current test.
```
drvinceknight: Can we remove `# Just to be consistant with current test` (that was just meant as a note to you in my comment but here it could be confusing).

drvinceknight:
Thanks for doing everything, this is looking great now, let's just remove that one inline comment which could be confusing later down the line :) Then this is good to go! :+1:

drvinceknight:
Awesome: thank you for making all the requested tweaks. :+1:

PR #1147, opened 2017-12-10
Implemented MoreTidemanAndChieruzzi, K84R from Axelrod's second

gaffney2010:
FPs below.

This was similar in some ways to TidemanAndChieruzzi from the first tournament, so I tried to make my function implementation similar to the that one.

The biggest change in my refactoring was to change the way JDR was used, by calculating this as a triangle number:  https://github.com/Axelrod-Python/TourExec/blob/v0.3.0/src/strategies/k84r.f#L18

basic_fortran
basic_python
random_python
random_fortran

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1312,7 +1312,7 @@ class MoreTidemanAndChieruzzi(Player):
     Strategy submitted to Axelrod's second tournament by T. Nicolaus Tideman
     and Paula Chieruzzi (K84R) and came in ninth in that tournament.
 
-    This strategy Cooperates if this player's socre exceeds the opponent's
+    This strategy Cooperates if this player's score exceeds the opponent's
     score by at least `score_to_beat`.  `score_to_beat` starts at zero and
     increases by `score_to_beat_inc` every time the opponent's last two moves
     are a Cooperation and Defection in that order.  `score_to_beat_inc` itself
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1305,3 +1305,114 @@ class Harrington(Player):
             self.prob += 0.05
             self.more_coop, self.last_generous_n_turns_ago = 2, 1
             return self.try_return(D, lower_flags=False)
+
+
+class MoreTidemanAndChieruzzi(Player):
+    """
+    Strategy submitted to Axelrod's second tournament by T. Nicolaus Tideman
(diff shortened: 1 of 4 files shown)


axelrod/strategies/axelrod_second.py
```diff
@@ -1305,3 +1305,114 @@ def strategy(self, opponent: Player) -> Action:
...
             self.more_coop, self.last_generous_n_turns_ago = 2, 1
             return self.try_return(D, lower_flags=False)
+
+
+class MoreTidemanAndChieruzzi(Player):
+    """
+    Strategy submitted to Axelrod's second tournament by T. Nicolaus Tideman
+    and Paula Chieruzzi (K84R) and came in ninth in that tournament.
+
+    This strategy Cooperates if this player's socre exceeds the opponent's
```
drvinceknight: `score`

gaffney2010:
@drvinceknight Sure.

https://github.com/Axelrod-Python/TourExec/blob/v0.2.0/src/strategies/k84r.f#L18 says:

IF (IS - JS - DS - 5 * JDR * (JDR - 1) / 2 .GE. 0)

I moved the term 5 * JDR * (JDR - 1) / 2 to the right-hand side of the inequality and called it `score_to_beat`, and rather than calculating it the same way, I calculated `score_to_beat` by initializing to 0 and increasing `score_to_beat_inc` every time that the Fortran code would have incremented JDR.  Immediately after, I increase `score_to_beat_inc` (initially 0) by 5.  I think it must be this kind of increasing scheme that motivated the formula 5 * JDR * (JDR - 1) / 2 in the first place.

drvinceknight:
AH yes! I see it now. Makes a lot of sense. :+1:

axelrod/strategies/axelrod_second.py
```diff
@@ -1305,3 +1305,114 @@ def strategy(self, opponent: Player) -> Action:
...
+        # Calculate the scores.
+        self._score_last_round(opponent)
+
+        # Check if we have recently given the strategy a fresh start.
+        if self.fresh_start:
+            self._fresh_start()
+            self.last_fresh_start = current_round
+            self.fresh_start = False
+            return C  # Second cooperation
+
```
marcharper: Can you explain the name `opponent_CDd`? I looks like it's just checking that `opponent.history[-2:] == [C, D]` right? Just curious what the final `d` indicates.

gaffney2010: That's correct.  The second d is because "CDd" is the past tense of "CD."

PR #1152, opened 2017-12-13
fixes implementation error in TidemanAndChieruzzi

### LINKED ISSUE #1149: Probable (certain?) implementation error in `axelrod_first.TidemanAndChieruzzi`
@gaffney2010 has recently implemented `MoreTidemanAndChieruzzi` in #1147, there's a check there that look very similar to the `TidemanAndChieruzzi` strategy in `axelrod_first`.

This specific line, https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/axelrod_first.py#L676:

```
if self.defections <= lower or self.defections >= upper:
```

Aims to check "and the total number of defections differs from a 50-50 random sample
by at least 3.0 standard deviations." (from the paper description).

@gaffney2010 has pointed out that the current implementation in `axelrod_first` doesn't look right:

1. It calculates the expected mean based on `opponent` and compare to `self`.
2. It differs to the implementation in the Fortran code (the strategies are different but this particular check is likely to be the same).

Thus, because of the ambiguity in the paper it makes sense to change the line to be:

```
if opponent.defections <= lower or opponent.defections >= upper:
```

This fix requires that simple change and adjustment of any relevant tests.

Nikoleta-v3:
This closes #1149.

Fixed implementation (by altering line 676) as suggested on the issue.

Tweaked tests to make sure they pass.

### DIFF
--- a/axelrod/strategies/axelrod_first.py
+++ b/axelrod/strategies/axelrod_first.py
@@ -673,7 +673,7 @@ class TidemanAndChieruzzi(Player):
                 std_deviation = (N ** (1 / 2)) / 2
                 lower = N / 2 - 3 * std_deviation
                 upper = N / 2 + 3 * std_deviation
-                if self.defections <= lower or self.defections >= upper:
+                if opponent.defections <= lower or opponent.defections >= upper:
                     # Opponent deserves a fresh start
                     self.last_fresh_start = current_round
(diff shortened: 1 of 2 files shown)


drvinceknight:
Thanks @Nikoleta-v3 :+1:

PR #1153, opened 2017-12-14
Implemented Leyvraz, K86R from Axelrod's second

gaffney2010:
Fingerprints below.

Pretty direct translation.  Their funny if statements reduced a lot.

      IF (J1 * J .EQ. 1) GOTO 540

J1 and J

      IF (J2 * 2 + J1 + J * 2 + J .EQ. 1) GOTO 550
      
J2 cannot be 1, because then the whole left side would exceed 1.  Ditto J.  So this becomes J1=1.
      
      IF (J2 * 2 + J1 * 2 + J .EQ. 1) GOTO 560
      
Similarly J1 and J2 must equal zero.  So this becomes J=1.
prob_python
prob_fortran
basic_python
basic_fortran

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1505,4 +1505,6 @@ class Leyvraz(Player):
             if len(opponent.history) >= go_back:
                 recent_history[-go_back] = opponent.history[-go_back]
 
-        return random_choice(self.prob_coop[tuple(recent_history)])
+        return random_choice(self.prob_coop[(recent_history[-3],
+                                             recent_history[-2],
+                                             recent_history[-1])])
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1466,9 +1466,9 @@ class Leyvraz(Player):
 
     1. If opponent Defected last two turns, then Defect with prob 75%.
     2. If opponent Defected three turns ago, then Cooperate.
-    3a. If opponent Defected two turns ago, then Defect.
-    3b. If opponent Defected last turn, then Defect with prob 50%.
-    4. Otherwise (all Coopertaions), then Cooperate.
+    3. If opponent Defected two turns ago, then Defect.
+    4. If opponent Defected last turn, then Defect with prob 50%.
+    5. Otherwise (all Coopertaions), then Cooperate.
 
     Names:
 
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1486,18 +1486,23 @@ class Leyvraz(Player):
         'manipulates_state': False
     }
 
+    def __init__(self) -> None:
+        super().__init__()
+        self.prob_coop = {
+            (C, C, C): 1.0,
+            (C, C, D): 0.5,
+            (C, D, C): 0.0,
+            (C, D, D): 0.25,
+            (D, C, C): 1.0,
+            (D, C, D): 1.0,
+            (D, D, C): 1.0,
+            (D, D, D): 0.5
+        }
+
     def strategy(self, opponent: Player) -> Action:
         recent_history = [C, C, C]  # Default to C.
         for go_back in range(1, 4):
             if len(opponent.history) >= go_back:
                 recent_history[-go_back] = opponent.history[-go_back]
 
-        if recent_history[-1] == D and recent_history[-2] == D:
-            return random_choice(0.25)
-        if recent_history[-3] == D:
-            return C
-        if recent_history[-2] == D:
-            return D
-        if recent_history[-1] == D:
-            return random_choice(0.5)
-        return C  # recent_history == [C, C, C]
+        return random_choice(self.prob_coop[tuple(recent_history)])
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1490,13 +1490,13 @@ class Leyvraz(Player):
         super().__init__()
         self.prob_coop = {
             (C, C, C): 1.0,
-            (C, C, D): 0.5,
-            (C, D, C): 0.0,
-            (C, D, D): 0.25,
-            (D, C, C): 1.0,
-            (D, C, D): 1.0,
-            (D, D, C): 1.0,
-            (D, D, D): 0.5
+            (C, C, D): 0.5,  # Rule 4
+            (C, D, C): 0.0,  # Rule 3
+            (C, D, D): 0.25,  # Rule 1
+            (D, C, C): 1.0,  # Rule 2
+            (D, C, D): 1.0,  # Rule 2
+            (D, D, C): 1.0,  # Rule 2
+            (D, D, D): 0.25  # Rule 1
         }
 
     def strategy(self, opponent: Player) -> Action:
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1300,11 +1300,11 @@ class Harrington(Player):
         if self.burned or random.random() > self.prob:
             # Tit-for-Tat with probability 1-`prob`
             return self.try_return(opponent.history[-1], inc_parity=True)
-        else:
-            # Otherwise Defect, Cooperate, Cooperate, and increase `prob`
-            self.prob += 0.05
-            self.more_coop, self.last_generous_n_turns_ago = 2, 1
-            return self.try_return(D, lower_flags=False)
+        
+        # Otherwise Defect, Cooperate, Cooperate, and increase `prob`
+        self.prob += 0.05
+        self.more_coop, self.last_generous_n_turns_ago = 2, 1
+        return self.try_return(D, lower_flags=False)
 
 
 class MoreTidemanAndChieruzzi(Player):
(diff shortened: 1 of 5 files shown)


axelrod/strategies/axelrod_second.py
```diff
@@ -1454,3 +1454,50 @@ def strategy(self, opponent: Player) -> Action:
...
+    (K68R) and came in eleventh in that tournament.
+
+    The strategy uses the opponent's last three moves to decide on an action
+    based on the following ordered rules.
+
+    1. If opponent Defected last two turns, then Defect with prob 75%.
+    2. If opponent Defected three turns ago, then Cooperate.
+    3. If opponent Defected two turns ago, then Defect.
+    4. If opponent Defected last turn, then Defect with prob 50%.
+    5. Otherwise (all Coopertaions), then Cooperate.
```
meatballs: typo: Cooperations

axelrod/strategies/axelrod_second.py
```diff
@@ -1300,11 +1300,11 @@ def strategy(self, opponent: Player) -> Action:
...
-        else:
-            # Otherwise Defect, Cooperate, Cooperate, and increase `prob`
-            self.prob += 0.05
-            self.more_coop, self.last_generous_n_turns_ago = 2, 1
-            return self.try_return(D, lower_flags=False)
+        
+        # Otherwise Defect, Cooperate, Cooperate, and increase `prob`
+        self.prob += 0.05
+        self.more_coop, self.last_generous_n_turns_ago = 2, 1
+        return self.try_return(D, lower_flags=False)
```
meatballs: What happened here? This doesn't look like it should have changed.

axelrod/strategies/axelrod_second.py
```diff
@@ -1454,3 +1454,50 @@ def strategy(self, opponent: Player) -> Action:
...
+
+        if recent_history[-1] == D and recent_history[-2] == D:
+            return random_choice(0.25)
+        if recent_history[-3] == D:
+            return C
+        if recent_history[-2] == D:
+            return D
+        if recent_history[-1] == D:
+            return random_choice(0.5)
+        return C  # recent_history == [C, C, C]
```
meatballs: Just a suggestion to consider rather than a request, but might this be easier to read with a dict mapping recent_history to action? Something like:

```
history_to_action = {
    [C, C, C]: C,
    [C, C, D]: random_choice(0.5)
etc....
}

return  history_to_action[recent_history]
```

drvinceknight: I think this actually is not ideal in this case (unless I'm misunderstanding):

```
[C, C, D]: random_choice(0.5)
```

Implies `history_to_action` would need to be redefined at every call of `strategy` to ensure the random sample is different which in turn means we'd be using the random module even when not needed (which bumps the seed and makes the strategy slower).

drvinceknight: One way around it (if you did want to use a `dict`) would be to use the dict to map to probabilities:

{
    [C, C, C]: 1,
    [C, C, D]: 0.1
}

and then use `random_choice(history_to_action[recent_history])` (the `random_choice` is smart so if passed 0 or 1 doesn't sample).

axelrod/strategies/axelrod_second.py
gaffney2010: Sorry, maybe I shouldn't have slipped this in here.  I got rid of a useless else.

axelrod/strategies/axelrod_second.py
gaffney2010: I think @drvinceknight is right here, but I'll keep in mind.

drvinceknight:
Strategy logic looks good and a correct implementation to me. Will leave the stylistic issue of the `dict` up to @gaffney2010 and @meatballs :+1:

axelrod/strategies/axelrod_second.py
gaffney2010: I went ahead and implemented this.  I agree that it's a little more readable.

axelrod/strategies/axelrod_second.py
```diff
@@ -1454,3 +1454,55 @@ def strategy(self, opponent: Player) -> Action:
...
+            (C, C, C): 1.0,
+            (C, C, D): 0.5,
+            (C, D, C): 0.0,
+            (C, D, D): 0.25,
+            (D, C, C): 1.0,
+            (D, C, D): 1.0,
+            (D, D, C): 1.0,
+            (D, D, D): 0.5
+        }
+
```
meatballs: nice!

axelrod/strategies/axelrod_second.py
```diff
@@ -1454,3 +1454,55 @@ def strategy(self, opponent: Player) -> Action:
...
+    Strategy submitted to Axelrod's second tournament by Fransois Leyvraz
+    (K68R) and came in eleventh in that tournament.
+
+    The strategy uses the opponent's last three moves to decide on an action
+    based on the following ordered rules.
+
+    1. If opponent Defected last two turns, then Defect with prob 75%.
+    2. If opponent Defected three turns ago, then Cooperate.
+    3. If opponent Defected two turns ago, then Defect.
+    4. If opponent Defected last turn, then Defect with prob 50%.
```
drvinceknight: I think the docstrings don't match the description:

```
(D, D, D): 0.5
```

That doesn't correspond to ` If opponent Defected last turn, then Defect with prob 50%.` as indicated for example by `(D, C, D): 1.0`?

drvinceknight: Apologies: I see you said "ordered" rules.

gaffney2010: You're right.  I was hoping the tests would catch any issues.  I'll go through this more carefully and re-run the fingerprints.

drvinceknight: No sorry apologies again, that's still not exact then, according to the ordered description: `(D, D, D)` should defect with prob 75%.

Now that we have this `dict`, we could almost just describe the strategy by saying:

```
Depending on the 3 most recent actions by the opponent the strategy cooperates with probability:

- If (C, C, C): probability 1;
- If (C, C, D): probability 0.5;
...
```

and essentially write out the 8 cases that correspond to the `dict` (which indeed really helps with readability here :+1:).

gaffney2010: I think (D, D, D) should be 0.25, by the first rule.  (These rules get applied in order.)  These others seem okay; (D, C, D) is 1.0 by rule 2.

drvinceknight: I think you are correct:

```python
>>> import axelrod as axl
>>> players = (axl.Defector(), axlf.Player("k68r"))
>>> match = axl.Match(players, turns=1000)
>>> results = match.play()
>>> match.cooperation()
(0,261)
```

so `k68r` is cooperating 25% of the time against `D, D, D`.

drvinceknight: Whereas your current branch:

```
>>> players = (axl.Defector(), axl.Leyvraz())
>>> match = axl.Match(players, turns=1000)
>>> axl.seed(0)
>>> results = match.play()
>>> match.cooperation()
(0, 508)
```

gaffney2010:
Re-updated with fixed probability.  I added some comments and re-ran FPs for more confidence here.

prob_fortran
basic_fortran
prob_python
basic_python

axelrod/strategies/axelrod_second.py
drvinceknight: A direct map of the previous implementation:

```
-        if recent_history[-1] == D and recent_history[-2] == D:
 -            return random_choice(0.25)
 -        if recent_history[-3] == D:
 -            return C
 -        if recent_history[-2] == D:
 -            return D
 -        if recent_history[-1] == D:
 -            return random_choice(0.5)
 -        return C  # recent_history == [C, C, C]
```

gives:

```
{
   (C, D, D): 0.25,
   (D, D, D): 0.25,
   (D, C, C): 1,
   (D, C, D): 1,
   (D, D, C): 1,
   (C, D, C): 0,
   (C, C, D): 0.5,
   (C, C, C): 1
}
```

which I believe gives the same as you have, apart from, as you say `(D, D, D)`.

Would be good to confirm the fps just in case :+1:

drvinceknight:
FPs look good.

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -930,3 +930,33 @@ def test_strategy(self):
...
+    expected_classifier = {
+        'memory_depth': 3,
+        'stochastic': True,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
```
drvinceknight: Could we add a `test_init_of_dict(self)` just to add a second anchor for the internal dictionary.

```
def test_init_of_dict(self):
    player = self.player()
    expected_prob_coop = {...}
    self.assertEqual(player.prob_coop, expected_proob_coop)
```

drvinceknight:
For completeness have rerun at dacc9c8:

```python
>>> players = (axl.Defector(), axl.Leyvraz())
>>> match = axl.Match(players, turns=1000)
>>> axl.seed(0)
>>> results = match.play()
>>> match.cooperation()
(0, 248)
```

:+1:

axelrod/strategies/axelrod_second.py
```diff
@@ -1454,3 +1454,55 @@ def strategy(self, opponent: Player) -> Action:
...
+            (D, D, D): 0.25  # Rule 1
+        }
+
+    def strategy(self, opponent: Player) -> Action:
+        recent_history = [C, C, C]  # Default to C.
+        for go_back in range(1, 4):
+            if len(opponent.history) >= go_back:
+                recent_history[-go_back] = opponent.history[-go_back]
+
+        return random_choice(self.prob_coop[tuple(recent_history)])
```
drvinceknight: I've hunted down why the type testing is failing on travis. It doesn't know what size `tuple` `tuple(recent_history)` returns so you need to change this to be:

```  
1508         return random_choice(self.prob_coop[(recent_history[-3],                │                                                              
+ 1509                                              recent_history[-2],                │                                                              
+ 1510                                              recent_history[-1])])   

```

So specifically say this is a tuple of size 3, which is a bit of a mouthful.

drvinceknight:
Thanks for making all the changes :+1:

PR #1154, opened 2017-12-18
Implemented White strategy, k72r from Axelrod's Second

gaffney2010:
Fingerprints below.

Notes on refactor from https://github.com/Axelrod-Python/TourExec/blob/v0.3.0/src/strategies/k72r.f :
N, M, and JCOUNT are all integers, so everything gets rounded.  So the probability statement (line 12) actually only ever gets run with 0% or 100% (evidenced by the basic-strategy fingerprints).  And actually unless the if-statement in line 11 gets run, then the probability statement always Cooperates, since JCOUNT can be at most M-1.

random_fortran
basic_python
basic_fortran
random_python

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1512,15 +1512,13 @@ class Leyvraz(Player):
 
 class White(Player):
     """
-    Strategy submitted to Axelrod's second tournament by Edward C Whtie (K72R)
+    Strategy submitted to Axelrod's second tournament by Edward C White (K72R)
     and came in thirteenth in that tournament.
 
-    If the opponent Cooperated last turn or in the first ten turns, then
-    Cooperate.
-
-    Otherwise Defect if and only if:
-
-    floor(log(turn)) * opponent Defections >= turn
+    * If the opponent Cooperated last turn or in the first ten turns, then
+      Cooperate.
+    * Otherwise Defect if and only if:
+        floor(log(turn)) * opponent Defections >= turn
 
     Names:
 
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1137,7 +1137,7 @@ class Harrington(Player):
         In this function, the statistic is non-standard in that it excludes
         summands where E_i <= 1.
         """
-        
+
         denom = turn - 2
 
         expected_matrix = np.outer(self.move_history.sum(axis=1),
@@ -1165,7 +1165,7 @@ class Harrington(Player):
         self.history, and returns True (is random) if and only if the statistic
         is less than or equal to 3.
         """
-        
+
         denom = turn - 2
 
         if self.move_history[0, 0] / denom >= 0.8:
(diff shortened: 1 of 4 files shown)


axelrod/strategies/axelrod_second.py
```diff
@@ -1508,3 +1508,42 @@ def strategy(self, opponent: Player) -> Action:
         return random_choice(self.prob_coop[(recent_history[-3],
                                              recent_history[-2],
                                              recent_history[-1])])
+
+
+class White(Player):
+    """
+    Strategy submitted to Axelrod's second tournament by Edward C Whtie (K72R)
+    and came in twelfth in that tournament.
```
drvinceknight: Looking at https://github.com/Axelrod-Python/axelrod-fortran/blob/master/src/axelrod_fortran/strategies.py#L214 it looks like it ranked 13th?

gaffney2010: It should say thirteenth, and actually the last few would be off.  I'll fix them.

gaffney2010: I got thrown off at Kluepfel, which had already been implemented since I've started doing these in order.

gaffney2010: > changes the behaviour of the fortran function

Changes how?  https://github.com/Axelrod-Python/TourExec/blob/v0.3.0/src/strategies/k72r.f#L12 must be rounding down the right-hand side, otherwise the 100% cooperation in the first ten moves (in the fingerprints) wouldn't make sense.

drvinceknight: If you pass a 0 (which is what we do when marking a function as `stochastic: False`) then the function seems to always defect. 

It's not an issue, just noting it. I think that essentially that `stochastic: ` key is whether or not the original Fortran code used the random number *anywhere*, even if as you've noticed here, it ended up not being relevant.

drvinceknight:
Thanks! (my bad for missing those rankings in the docstrings previously. :+1:)

drvinceknight:
Nudge @marcharper and @meatballs (in case you've missed this): I have checked the logic and this looks good to go :+1:

axelrod/strategies/axelrod_second.py
```diff
@@ -1508,3 +1508,42 @@ def strategy(self, opponent: Player) -> Action:
         return random_choice(self.prob_coop[(recent_history[-3],
                                              recent_history[-2],
                                              recent_history[-1])])
+
+
+class White(Player):
+    """
+    Strategy submitted to Axelrod's second tournament by Edward C Whtie (K72R)
```
marcharper: White -> White

axelrod/strategies/axelrod_second.py
```diff
@@ -1508,3 +1508,42 @@ def strategy(self, opponent: Player) -> Action:
...
+    and came in thirteenth in that tournament.
+
+    If the opponent Cooperated last turn or in the first ten turns, then
```
marcharper: *  If the opponent cooperated last turn or in the first ten turns, then
   cooperate.
*  Otherwise defect if and only if:
      floor(log(turn)) * opponent.defections >= turn

marcharper:
Looks fine to me, please fix the typo in the name at least.

gaffney2010:
Made those changes.

drvinceknight:
Great. Thanks :+1:

Issue #597, opened 2016-05-17
Algorithm for memory depth for finite state machine players


marcharper:
We need one!

drvinceknight:
👍

drvinceknight:
See for context: https://github.com/Axelrod-Python/Axelrod/pull/591#issuecomment-219623591

drvinceknight:
At the moment a generic FSM strategy has default memory 1, I think the number of states in the underlying FSM is a better estimate (until we obtain a better algorithm).

gaffney2010:
I've been giving this some thought, and I think I have an approach to analyze a FSM for memory depth.  [Would work for other games too.]  I describe my idea below.  If it seems good, I could code it up.


My idea:

The memory is the length of memory needed to determine which state we're in.  The algorithm I'm thinking about will take each state and walk backward along the edges that are coming into that state.  The edges each of one of four types (C in reponse to C, C resp to D, D resp to C, and D resp to D).  For a given edge type, if there is one or fewer states with that edge type incoming, then we could determine the state if we're given that history.  If there are two or more, then one edge worth of history is not enough, and we similarly walk backwards from each of the state we ended on.  For example if states 1 and 4 both have incoming C/C edges coming from states 5 and 3 (resp), then we can look at all the incoming edges on states 5 and 3.  If 5 only has incoming C/C and C/D edges and 3 only has incoming D/C and D/D edges, then we only two edge's worth of history would be enough to distinguish.  If on the other hand, they both had an incoming D/C edge, then we'd have to continue to trace those edges back.  [I call this breaking a tie.]  Otherwise we would have to repeat until we've broken all the ties.  When this terminates, we'd know how much history would be needed to be guaranteed to be determine the state.

We'd start this algortihm by listing all incoming edges for each state.  Then the time it'd take for for each step back would be O(E).  If we went back 4E steps (maybe too many), then we'd know that infinite memory is needed, because by then we'd certainly have entered a looping pattern.

[There's an off-by-one issue because knowing that an edge is of type C/C requires 2-turn history because it requires knowing that the opponent cooperated two turns ago.  When our algorithm ends (i.e. the last ties is broken), we need to check if we could have broken the tie with only the knowledge of what the player did in response.  That is, is knowing that the edge is of type */C enough inforamtion?]

marcharper:
[...]

PR #1155, opened 2018-01-03
Implemented Black, K83R from Axelrod's Second

gaffney2010:
Fingerprints Below

basic_fortran
basic_python
random_fortran
random_python

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -10,7 +10,7 @@ from .axelrod_first import (
 from .axelrod_second import (
     Champion, Eatherley, Tester, Gladstein, Tranquilizer, MoreGrofman,
     Kluepfel, Borufsen, Cave, WmAdams, GraaskampKatzen, Weiner, Harrington,
-    MoreTidemanAndChieruzzi, Getzler, Leyvraz, White)
+    MoreTidemanAndChieruzzi, Getzler, Leyvraz, White, Black)
 from .backstabber import BackStabber, DoubleCrosser
 from .better_and_better import BetterAndBetter
 from .bush_mosteller import BushMosteller
@@ -107,8 +107,9 @@ all_strategies = [
     Appeaser,
     ArrogantQLearner,
     AverageCopier,
-    BetterAndBetter,
     BackStabber,
+    BetterAndBetter,
+    Black,
     Borufsen,
     Bully,
     BushMosteller,
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1554,9 +1554,9 @@ class Black(Player):
 
     The strategy Cooperates for the first five turns.  Then it calculates the
     number of opponent defects in the last five moves and Cooperates with
-    probability 'prob_coop'[`number_defects`], where:
+    probability `prob_coop`[`number_defects`], where:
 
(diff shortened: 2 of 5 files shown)


marcharper:
Are we sure the memory depth is 5? It's a weird case since you need to know the round number + the last five rounds...

drvinceknight:
> Are we sure the memory depth is 5? It's a weird case since you need to know the round number + the last five rounds.

This is similar to memory one strategies that all have a memory depth of 1 but that act differently on their first move to any other right?

gaffney2010:
You don't need to know the full round number.  If you gave me just the last five moves as a list, I could check if that list is length 5 or not.

drvinceknight:
> You don't need to know the full round number. If you gave me just the last five moves as a list, I could check if that list is length 5 or not.

Even better point :+1:

marcharper:
I guess that's what I'm asking -- on the first move there's no history so it's sort of a special case and players must have a separate rule. Here on rounds say 4, 5, 6 you have to know the round number and then decide from one of two behaviors. So we need to know if there are at least 5 rounds of history already, so we need the size of the history and therefore all of the history. Or is the rule that we can ask "Are there at least X rounds of history?" and then take the minimum needed value of X as the memory depth?

marcharper:
Also think about the case where there's X rounds of initial play and then a behavior that depends on the last Y rounds, where X != Y.

marcharper:
Also, if the implementation was setting a bit if_round_greater_than_five then we usually call that kind of situation infinite memory...

drvinceknight:
> Here on rounds say 4, 5, 6 you have to know the round number and then decide from one of two behaviors. So we need to know if there are at least 5 rounds of history already, so we need the size of the history and therefore all of the history. Or is the rule that we can ask "Are there at least X rounds of history?"

It's a good question. I think that my personal preference would be to define memory as "if you give `history[-memory:]` or `history` you get the same behaviour". I believe this isn't the first strategy that poses this question so I'd suggest we merge as is (based on the mem 1 behaviour/convention it feels right to me that this has mem depth 5). And perhaps aim to implement a test for memory on all players?

gaffney2010:
In this sense, Tit-For-Tat has to know the round number, since there's a special rule for when turn=1, so you have to know the turn number.  But I think it's clear that TFT is memory=1; by analogy, this is memory=5.  And I think it's intuitive; you're allowed to be forgetful.

drvinceknight:
> "if you give history[-memory:] or history you get the same behaviour"

In this particular case if we're on the third turn, `opponent.history=[C, C, C]` then `opponent.history[-5:]` would also be `[C, C, C]`, so:

```
+        if len(opponent.history) < 5:
+            return C
```

would return `C`.

> And perhaps aim to implement a test for memory on all players?

This would potentially assist with #597 (not in finding an algorithm but at least verifying upper bounds on memory lengths).

drvinceknight:
> In this particular case if we're on the third turn with opponent.history=[C, C, C] then opponent.history[-5:] would also be [C, C, C], so:

If we're on the 150th turn then `len(opponent.history[-5:])` is `5` so the rest of the logic kicks in.

drvinceknight:
I've opened https://github.com/Axelrod-Python/Axelrod/issues/1156 as a suggestion for a test (to perhaps move the discussion about the way to go over there).

marcharper:
Ok, let's discuss elsewhere. I'll merge as is.

PR #1157, opened 2018-01-03
Add test for memory depth

### LINKED ISSUE #1156: Add test for memory depth to TestPlayer
All players have a `.classifier["memory_depth"]` attribute that defines their memory depth. 

It would be good to test that this given number is indeed a lower bound on the memory depth of players. I suggest that one way to do this would be to add a method to `TestPlayer` that takes some selection of strategies, and creates a match (seeded it stochastic) and then plays the same match but truncate the opponent and player's memories to `opponent.memory[-memory_depth:]` at each step and verifies that the result is the same.

Something like the following (the `assert` would need to be changed to the corresponding unittest call etc...):

```
def test_memory(player, opponent, seed=None, turns=10):
    
    if seed is not None:
        axl.seed(seed)
    match = axl.Match((player, opponent), turns=turns)
    expected_results = match.play()
    
    player.reset()
    opponent.reset()
    memory = player.classifier["memory_depth"]
    
    results = []
    for _ in range(turns):
        player.history = player.history[-memory:]
        opponent.history = opponent.history[-memory:]
        player.play(opponent)
        results.append(player.history[-1])
    assert results == [interactions[0] for interactions in expected_results], (results, opponent)
```

The opponent could be picked using hypothesis perhaps (but for each strategy this should only pick a small number of opponents?).

Some examples:

```
>>> test_memory(player=axl.TitForTat(), opponent=axl.Alternator())
>>> test_memory(player=axl.TitFor2Tats(), opponent=axl.Alternator())
>>> test_memory(player=axl.Defector(), opponent=axl.Alternator())
```
Note that this would not verify that we have the lowest memory depth: if we declare that the memory depth of a strategy is 5 it might pass this test even though it in fact only needs 2 turns of memory.

I would be impressed if after implementing this test all our strategies pass. I wouldn't be surprised if we have one or two that don't have the correct memory depth. This test would help find them :)

(From discussion on #1155)

drvinceknight:
This writes a test that verifies that the memory depth given in `player.classifier` is an upper bound.

It's done using a hypothesis test that only checks a single set of 5 opponents each time so it's not overly expensive (still benefiting from hypothesis's failure database).

This is the commit that writes the test: e9da4f9

Note that it picked up a few mis classifications, that I've fixed here (I'm actually surprised it wasn't more):

- c612c1e: `DynamicTwoTitsForTat` (strategy makes use of the cooperation count, so it's infinite memory)
- f6479f7: `RandomTitForTat`: classified correctly, I've just tweaked the implementation.
- ef130c2: `SteinAndRapoport`: defects on last two rounds: so it's infinite memory.
- 483d96e: all the cheating strategies had different (weird/negative) memory depths, I've set them as infinite.

This closes #1156

### DIFF
--- a/axelrod/strategies/geller.py
+++ b/axelrod/strategies/geller.py
@@ -40,7 +40,7 @@ class Geller(Player):
 
     name = 'Geller'
     classifier = {
-        'memory_depth': -1,
+        'memory_depth': float("inf"),
         'stochastic': True,
         'makes_use_of': set(),
         'long_run_time': False,
@@ -74,7 +74,7 @@ class GellerCooperator(Geller):
     """
     name = 'Geller Cooperator'
     classifier = {
-        'memory_depth': -1,
+        'memory_depth': float("inf"),
         'stochastic': False,
         'makes_use_of': set(),
         'long_run_time': False,
@@ -101,7 +101,7 @@ class GellerDefector(Geller):
     """
     name = 'Geller Defector'
     classifier = {
-        'memory_depth': -1,
+        'memory_depth': float("inf"),
         'stochastic': False,
         'makes_use_of': set(),
         'long_run_time': False,
--- a/axelrod/strategies/mindreader.py
+++ b/axelrod/strategies/mindreader.py
@@ -22,7 +22,7 @@ class MindReader(Player):
 
     name = 'Mind Reader'
     classifier = {
-        'memory_depth': -10,
+        'memory_depth': float("inf"),
         'stochastic': False,
         'makes_use_of': set(),
         'long_run_time': False,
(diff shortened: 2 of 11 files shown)


axelrod/strategies/geller.py
```diff
@@ -40,7 +40,7 @@ class Geller(Player):
 
     name = 'Geller'
     classifier = {
-        'memory_depth': -1,
```
marcharper: FYI: This one I think we did because "it looks into the future" (not suggesting we change it back necessarily).

axelrod/strategies/mindreader.py
```diff
@@ -22,7 +22,7 @@ class MindReader(Player):
 
     name = 'Mind Reader'
     classifier = {
-        'memory_depth': -10,
```
marcharper: Same

drvinceknight: Yeah I thought it was probably due to the "looking in to the future" but taking the notion of infinity to mean "backwards and forwards" I think we're good :) (I agree that we don't care too much :)).

drvinceknight:
Marking this as good to go: whenever you get a chance @meatballs :+1:

PR #1158, opened 2018-01-07
Add lower bound for python in setup.

drvinceknight:
This ensures that if you pip install axelrod with an unsupported verion
of python that you get an error message:

```
(test-axelrod-on-3.4) ~: pip install ~/src/Axelrod
Processing ./src/Axelrod
Axelrod requires Python '>=3.5' but the running Python is 3.4.5
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ a peer reviewed paper introducing the library (22 authors).
 Installation
 ------------
 
-The library requires Python 3.4 or greater. It will not run on Python 2.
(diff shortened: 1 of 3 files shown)


RobPDev:
In the installation section of README.rst it says py3.4 is required- it should probably be changed to say 3.5

drvinceknight:
> In the installation section of README.rst it says py3.4 is required- it should probably be changed to say 3.5

Good call: have tweaked! :+1:

PR #1165, opened 2018-01-29
Re write the result set

drvinceknight:
**Note this PR is to a `possible-v4` branch which is a branch of `master`**

(@marcharper @meatballs Genuinely sorry for the diff on `result_set.py` it's a mess as this kind of guts that file... If we like the idea of what I've done here I'm happy to fix that as best I can.)

At present the result set plays the matches in parallel followed by a
(sometimes computationally expensive) single process of reading in and
analysing the interactions.

TLDR: This changes how the internal result set calculations are being
done. They are more efficiently calculated.

This happens by doing the following:

1. The various match by match calculations are done by the tournament
   (the winner of a match, the cooperation count, the score etc...). Mainly thanks to [`Tournament._calculate_results`](https://github.com/Axelrod-Python/Axelrod/pull/1165/files#diff-dad184e860a161b8b929a7ebbfd21c9cR412) 
2. These calculations are written to file (at present we just write the
  actual actions of an interaction) thanks to this line of [`Tournament._write_interaction_to_file`](https://github.com/Axelrod-Python/Axelrod/pull/1165/files#diff-dad184e860a161b8b929a7ebbfd21c9cR244)
3. The form of this file has also changed: for every match there are 2
   rows. One row corresponds to each player. This might seem more costly storage
   wise but is done to enable faster analysis (more on that later).
4. The analysis is now done entirely using `dask`: "Dask is a flexible
   parallel computing library for analytic computing."
   (https://dask.pydata.org/). This ensures all calculations are done on
   disk (so no huge memory problems) but also that they can be done **in
   parallel**. This is all done using the nice Pandas-like API that dask
   has so essentially all calculations for the result set are just done
   by a few `groupby` statements. The tasks to be run are built with [`ResultSet._build_tasks`](https://github.com/Axelrod-Python/Axelrod/pull/1165/files#diff-ba927cc36dbf4d89036c3602ad830638R535) and then compute (in parallel) with [`ResultSet._compute_tasks`](https://github.com/Axelrod-Python/Axelrod/pull/1165/files#diff-ba927cc36dbf4d89036c3602ad830638R525).
5. There is *some* work being done outside of dask but that's just
   reshaping data. `dask` outputs `pandas.Series` and to be consistent
   with our current setup these are changes to be lists of list etc... This is all done in [`ResultSet._reshape_out`](https://github.com/Axelrod-Python/Axelrod/pull/1165/files#diff-ba927cc36dbf4d89036c3602ad830638R80)

**Some user facing changes** (which is why I suggest this would be for a
`v4.0.0` release):

- The `result_set.interactions` is no longer possible. This is a choice
  and not forced by the redesign: I don't think this is ever necessary or
  helpful. The data file can always be read in.
- The ability to `tournament.play(in_memory=True)` has been removed.
  Again, not entirely a forced change (although it would be a tiny bit
  of work to allow this). Given the recent work to make everything work
  on Windows I don't think this is necessary and has allowed for a big
  deletion of code (which is a good thing re maintenance).
- The interactions data file is now in a different format, this is
  forced by the design choice.
- I have made a slight modification to `result_set.cooperation`.
  Currently this sets all self interactions to be 0 but I think that's not
  the right way to display it (note that the cooperation rates were all
  being done correctly).

**As well as ensuring the work done in series is reduced and the parallel
workers also calculate the scores (which I think is more efficient)**
this also seems to be faster:

On this branch:

```python
import axelrod as axl
players = [s() for s in axl.strategies if s.classifier["memory_depth"] == 1]
tournament = axl.Tournament(players, turns=200, repetitions=100)
results = tournament.play(processes=4)
```

Took: 1min 49s

```python
import axelrod as axl
players = [s() for s in axl.short_run_time_strategies]
tournament = axl.Tournament(players, turns=200, repetitions=20)
results = tournament.play(processes=4)
```

Took: 21min 2s

On current master;

```python
import axelrod as axl
players = [s() for s in axl.strategies if s.classifier["memory_depth"] == 1]
tournament = axl.Tournament(players, turns=200, repetitions=100)
results = tournament.play(processes=4)
```

Took: 2min 36s

```python
import axelrod as axl
players = [s() for s in axl.short_run_time_strategies]
tournament = axl.Tournament(players, turns=200, repetitions=20)
results = tournament.play(processes=4)
```

Took: 28min 8s

**One final aspect to consider** (I think) is that adding `dask` as a
dependency open up the potential to use it's `delayed` decorator to do
all our parallel processing. This would have the benefit of being able
to use a distributed scheduler that `dask` has. (We'd have to
investigate if this actually works based on our parallelisation but at
least the possibility is there).


axelrod/interaction_utils.py
```diff
@@ -239,36 +240,27 @@ def compute_sparklines(interactions, c_symbol='█', d_symbol=' '):
         sparkline(histories[1], c_symbol, d_symbol))
 
 
-def read_interactions_from_file(filename, progress_bar=True,
-                                num_interactions=False):
+def read_interactions_from_file(filename,
```
marcharper: one line?

axelrod/interaction_utils.py
```diff
@@ -239,36 +240,27 @@ def compute_sparklines(interactions, c_symbol='█', d_symbol=' '):
...
-
-            if progress_bar:
-                progress_bar.update()
-
-    if progress_bar:
-        progress_bar.close()
+    for _, d in tqdm.tqdm(groupby):
+        key = tuple(d[["Player index", "Opponent index"]].iloc[0])
+        value = list(map(str_to_actions, zip(*d["Actions"])))
+        try:
```
marcharper: Could use a `defaultdict(list)` here.

drvinceknight: Good call, much nicer :+1:

axelrod/interaction_utils.py
drvinceknight: > Will review more -- github isn't loading the large diffs and I don't want to lose the comments I've already made.

Yeah, no rush with this at all. Small piece meal chunks is probably the way to go. Sorry the diff is so horrible.

> 
> Do we want to add an end user function to assist in reading the interactions file?

We've got `iu.read_interactions_from_file` which I've documented in 
843b008, I think that could be user facing (and in fact we could have another tutorial going over each function in `iu` ?) but let me know if you were thinking of something different.

axelrod/result_set.py
```diff
@@ -1,9 +1,14 @@
 from collections import namedtuple, Counter
+from multiprocessing import cpu_count
 import csv
+import itertools
 
-from numpy import mean, nanmedian, std
+from numpy import mean, nanmedian, std, array, nan_to_num
```
marcharper: minor: I'd prefer to `import numpy as np` and then use `np.mean` and such because names like mean, std, array are common and there could be collisions.

axelrod/result_set.py
```diff
@@ -28,314 +33,297 @@ def wrapper(*args):
...
-            repetitions : int
-                The number of repetitions
+        Reshape the various pandas series objects to be of the required form and
+        set the corresponding attributes.
         """
-        repetitions = max(self.repetitions_d.values())
 
-        del self.repetitions_d  # Manual garbage collection
-        return repetitions
+        (mean_per_reps_player_opponent_df,
```
marcharper: Is there a reason not to pass these in explicitly as parameters? Or to use `*args` instead?

axelrod/result_set.py
```diff
@@ -28,314 +33,297 @@ def wrapper(*args):
...
+                                                      player_index,
+                                                      opponent_index,
+                                                      ), 0))
+                matrix.append(row)
+            score_diffs.append(matrix)
+        return score_diffs
 
-            [pi1, pi2, pi3, ..., pim]
+    @update_progress_bar
+    def _build_match_lengths(self, length_series):
```
marcharper: Seems like these functions could be abstracted, they are all very similar.

axelrod/result_set.py
```diff
@@ -28,314 +33,297 @@ def wrapper(*args):
...
+    @update_progress_bar
+    def _build_cooperation(self, cooperation_series):
+        cooperation_dict = cooperation_series.to_dict()
+        cooperation = []
+        for player_index in range(self.num_players):
+            row = []
+            for opponent_index in range(self.num_players):
+                count = cooperation_dict.get((player_index, opponent_index),0)
+                if player_index == opponent_index:
+					# Address double count
```
marcharper: Spacing

axelrod/result_set.py
```diff
@@ -28,314 +33,297 @@ def wrapper(*args):
...
+        for player_index in range(self.num_players):
+            row = []
+            for opponent_index in range(self.num_players):
+                if player_index == opponent_index:
+                    # The reduce operation implies a double count of self
+                    # interactions.
+                    row.append(0)
                 else:
-                    payoff_stddevs[player][opponent] = 0
+                    row.append(good_partner_dict.get((player_index,
```
marcharper: minor: In a case like this I think wrapping the entire tuple is more readable:
```
 row.append(good_partner_dict.get(
     (player_index, opponent_index), 0))
```

axelrod/result_set.py
```diff
@@ -384,385 +392,191 @@ def _build_normalised_state_to_action_distribution(self):
...
-            winner_index = index_pair[match_winner_index]
-            self.wins[winner_index][repetition] += 1
-
-    def _update_scores(self, repetition, p1, p2, interaction):
-        """
-        During a read of the data, update the scores attribute
+    @update_progress_bar
+    def _build_initial_cooperation_count(self, initial_cooperation_count_series):
+        initial_cooperation_count_dict = initial_cooperation_count_series.to_dict()
+        initial_cooperation_count = [
```
marcharper: Spacing on this one seems weird.

axelrod/result_set.py
```diff
@@ -384,385 +392,191 @@ def _build_normalised_state_to_action_distribution(self):
...
+        sum_per_player_opponent_task = df.groupby(groups)[columns].sum()
+
+        ignore_self_interactions_task = df["Player index"] != df["Opponent index"]
+        adf = df[ignore_self_interactions_task]
+
+        groups = ["Player index", "Repetition"]
+        columns = ["Win", "Score"]
+        sum_per_player_repetition_task = adf.groupby(groups)[columns].sum()
+
+        normalised_scores_task = adf.groupby(["Player index",
```
marcharper: maybe break up into two lines or use different spacing

axelrod/tournament.py
```diff
@@ -111,83 +113,53 @@ def play(self, build_results: bool = True, filename: str = None,
...
-            self._run_parallel(processes=processes)
+            self._run_parallel(build_results=build_results, processes=processes)
 
         result_set = None
         if build_results:
-            result_set = self._build_result_set(
-                keep_interactions=keep_interactions, in_memory=in_memory
-            )
-
+            result_set = ResultSet(
```
marcharper: Spacing

axelrod/tournament.py
```diff
@@ -212,38 +216,59 @@ def _get_progress_bar(self):
...
+                        row.append(turns)
+                        row.append(score_per_turns[index])
+                        row.append(score_diffs_per_turns[index])
+                        row.append(int(winner_index is index))
+                        row.append(initial_cooperation[index])
+                        row.append(cooperations[index])
+
+                        states = [(C, C), (C, D), (D, C), (D, D)]
+                        for state in states:
+                            if index == 1:
```
marcharper: pull to the outer context; consider using `reverse()` (modifies in place)

axelrod/result_set.py
drvinceknight: Yeah I agree, I'll change this.

axelrod/result_set.py
drvinceknight: I thought that and started taking a look at it but they've all got tiny differences that made it a bit painful. I'll take another look :) :+1:

axelrod/result_set.py
drvinceknight: Yeah that's way better :+1:

meatballs:
All looks good to me. Nice piece of work!

drvinceknight:
> All looks good to me. Nice piece of work!

Great, thanks for taking the time @meatballs :+1:

Once we're happy with this (@marcharper please don't hesitate to point out anything missing), we can merge and then open a PR from the `possible-v4` in to `master`. Not sure why I felt the need for an intermediate branch...

PR #1166, opened 2018-02-05
Rewrite the result set

drvinceknight:
This just pushes #1165 to master.


marcharper:
Do we need to review again? The tests all pass. I don't see any obvious shenanigans (new files and the like).

drvinceknight:
> Do we need to review again? The tests all pass.

No there's nothing new here at all, apologies for opening the previous PR to this dummy branch in the first place: it's created a tad more work.

Will let @meatballs press the button just in case he had any thoughts.

Issue #1169, opened 2018-03-23
`MemoryOnePlayer` default player


Nikoleta-v3:
**Question:**

In the generic `MemoryOnePlayer` the following is stated:  https://github.com/Axelrod-Python/Axelrod/blob/47abcc5f33da0cb91d0ce64e0616162f758e8e7f/axelrod/strategies/memoryone.py#L17
Does that mean that the class was suppose to initialise a `four_vector` of (1, 0, 0, 1) if `four_vector` was not given?  

If yes, then this is not implemented. It's an easy fix  **BUT** I am not sure I am interpreting the sentence correctly.

drvinceknight:
I believe you're interpreting that sentence correctly and you've identified a small bug.

The init won't fail if a `four_vector` is not passed but it will not create a valid player: https://github.com/Axelrod-Python/Axelrod/blob/47abcc5f33da0cb91d0ce64e0616162f758e8e7f/axelrod/strategies/memoryone.py#L63

```python
>>> import axelrod as axl
>>> player = axl.MemoryOnePlayer()
>>> player
Generic Memory One Player: C
>>> opponent = axl.TitForTat()
>>> match = axl.Match((player, opponent), turns=10)
>>> match.play()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
```

I suggest we either:

1. Add the intended default behaviour (including a test showing that something like the above code snippet won't fall over).
2. Remove the `default` for `four_vector` (so that `player = axl.MemoryOnePlayer()` would fail and then there's no need for the `if four_vector is not None:`).

I don't feel too strongly either way. Perhaps 2. is cleaner?

Nikoleta-v3:
I agree with 2. 

Should we raise a specific error if the vector is not given like `Transition probabilities not given` ?

drvinceknight:
> Should we raise a specific error if the vector is not given like

We haven't for other classes, so I think just leaving the standard Python error (which would kick in if there are no defaults) would be fine.

marcharper:
Good catch.

Since it's an abstract base player it's not in the standard list of players it's fine to initialize a default to throw an error. I'd suggest adding a warning to the user in the default case.

Nikoleta-v3:
I had a look at the `memoryone.py` file. 

If we choose to give an error when `four_vector` is not initialised (by just removing the `if` statement as suggested by Vince), several players that inherit from `MemoryOnePlayer` are going to break.
[...]

PR #1170, opened 2018-03-30
Fixes initialisation bug for memory one

Nikoleta-v3:
This closes issue #1169.

The issue has been that if `four_vector` was not set the player would not
break. This is fixed by having a default `four_vector` which corresponds
to the strategy Win-Stay Lose-Shift. Note that when this happens a
warning message is raised to let the user know (using the library
warnings).

A test has been altered to test that the default values are passed correctly.
Test: `test_default_if_four_vector_not_set`.

An issue has been that the players that inherited from `MemoryOnePlayer`
were breaking. This was because several classes were initialising with
an empty vector and were setting the vector are retrieving information
on the game.

This was addressed using the function `get_initial_four_vector`, which
will only set a default and raise an error for the general case and
will ignore the fact that `four_vector` is empty for the classes which
have been breaking.

### DIFF
--- a/axelrod/tests/strategies/test_memoryone.py
+++ b/axelrod/tests/strategies/test_memoryone.py
@@ -1,6 +1,7 @@
 """Tests for the Memoryone strategies."""
 
 import unittest
+import warnings
 
 import axelrod
 from axelrod import Game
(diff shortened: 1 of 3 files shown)


axelrod/tests/strategies/test_memoryone.py
```diff
@@ -222,11 +222,10 @@ def test_four_vector(self):
...
 class TestMemoryOnePlayer(unittest.TestCase):
 
-    def test_exception_if_four_vector_not_set(self):
+    def test_default_if_four_vector_not_set(self):
         player = MemoryOnePlayer()
-        opponent = axelrod.Player()
-        with self.assertRaises(ValueError):
-            player.strategy(opponent)
+        self.assertEqual(player._four_vector, {(C, C): 1.0, (C, D): 0.0,
+                                               (D, C): 0.0, (D, D): 1.0})
```
drvinceknight: Could we add a test catching the warning?

Nikoleta-v3: Sure will look into it now!

marcharper:
Thanks!

Issue #8, opened 2015-02-19
Give players different classes


JasYoung314:
This will then determine which tournament they should be run in (Cheating class, All class etc)

PR #1171, opened 2018-04-05
add implementation of a generic memory two + 2 strategies

### LINKED ISSUE #1063: Add specific memory two strategies
* A generic class for memory 2 strategies (or Gambler can be used)
* ~~The specific strategy on page 21 of this [paper](http://math.uchicago.edu/~may/REU2014/REUPapers/Li,Siwei.pdf)~~
* AON_2 and Delayed AON_1 from [Memory-n strategies of direct reciprocity](http://www.pnas.org/content/114/18/4715) [See table 1, bottom of page 4].

Nikoleta-v3:
Closes #1063

A generic class for memory two players, similar to the `MemoryOnePlayer` class.
If transition rates are not given the the player is set to default which is Cooperator.

I have added the equivalent tests and a reference from issue #1063.

Moreover, this implements two strategies of interest from #1063:

- `AON2` and `Delayed AON1`

Each strategy has 16 states. I have added tests for most of the states.


Nikoleta-v3:
**Note** that in the paper  [Memory-n strategies of direct reciprocity](http://www.pnas.org/content/114/18/4715) there are:

- four possible transition vectors which are equivalent to AON_2. I chosed the one with the highest cooperation rate against itself. I did generate the fingerprints of all four vectors.

**AON_2**
aon2

**second vector**
second

**third vector**
third

**fourth vector**
fourth


- three possible transition vectors which are equivalent to Delayed AON_1. I chosed the one with the highest cooperation rate against itself again, and the fingerprints:

**Delayed AON_1**
aon_1

**second vector**
second_aon1

**third vector**
third_aon1

The fingerprints appear to be similar.

Nikoleta-v3:
Finally the tests are going to fail because of:

```
axelrod/strategies/memorytwo.py:98: error: Invalid index type "Tuple[Tuple[Any, ...], Tuple[Any, ...]]" for "Dict[Tuple[Tuple[Any], Tuple[]], float]"; expected type "Tuple[Tuple[Any], Tuple[]]"
```

which is because of the type hints. I am not really sure why it's breaking :sweat_smile:  I would appreciate some help.

axelrod/strategies/_strategies.py
```diff
@@ -57,7 +57,7 @@
     MemoryOnePlayer, ALLCorALLD, FirmButFair, GTFT, SoftJoss,
     StochasticCooperator, StochasticWSLS, WinStayLoseShift, WinShiftLoseStay,
     ReactivePlayer)
-from .memorytwo import MEM2
+from .memorytwo import MemoryTwoPlayer, AON2, DelayedAON1, MEM2
```
drvinceknight: `AON2` and `DelayedAON1` need to be added to the `all_players` list.

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,181 @@
...
+        'memory_depth': 2,  # Memory-one Four-Vector
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self, initial: Action = C) -> None:
```
drvinceknight: As these strategies are defined in the literature, do we need `initial` to be an option?

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,181 @@
...
+
+class DelayedAON1(MemoryTwoPlayer):
+    """
+    Delayed AON1 a memory two strategy also introduced in [Hilbe2017]_ and belongs to the
+    AONk family. Note that AON1 is equivalent to Win Stay Lose Swift.
+
+    In essence is a strategy that starts of by cooperating and will cooperate
+    again only after the states (CC, CC), (CD, CD), (CD, DD), (DD, CD),
+    (DC, DC) and (DD, DD).
+
```
drvinceknight: Could we specifically write the vector here and also write the other vectors reported as equivalent (but different) in `[Hilbe2017]` (mentioning the reported self cooperation rate?).

Same for `AON2`.

drvinceknight:
> I would appreciate some help.

@Nikoleta-v3 I've opened https://github.com/Nikoleta-v3/Axelrod/pull/8 that fixes the type hint issue (for me).

I've essentially been less specific in some places with the hints and been explicit in the `Dict`, that inline comment is actually a variable type hint (see http://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html) which is there specifically annotating we've got a dictionary mapping tuples to floats.

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,181 @@
...
+                                                       float, float, float, float,
+                                                       float, float, float, float,
+                                                       float, float, float, float]):
+        if not all(0 <= p <= 1 for p in sixteen_vector):
+            raise ValueError("An element in the probability vector, {}, is not "
+                             "between 0 and 1.".format(str(sixteen_vector)))
+
+        states = [(hist[:2], hist[2:]) for hist in list(itertools.product((C, D), repeat=4))]
+
+        self._sixteen_vector = dict(zip(states, map(float, sixteen_vector)))
```
marcharper: I think the automatically derived type here is not correct. Maybe if you wrap `map(...)` with `tuple(...)` it will work since it won't be a list of tuples until actually used in a list/tuple context, which the type checker may not be able to determine.

Alternatively you could explicitly set the type (perhaps better to do in the `__init__` function probably).

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,205 @@
...
+
+    2.Error correcting. A strategy is error correcting after at most k rounds if,
+    after any history, it generally takes a group of players at most k + 1 rounds
+    to re establish mutual cooperation.
+
+    3. Retaliating. A strategy is retaliating for at least k rounds if, after
+    rounds in which the focal player cooperated while at least one coplayer defected,
+    the strategy responds by defecting the following k rounds.
+
+    In [Hilbe2017]_ the following vectors are reported as equivalent to AON2
```
drvinceknight: (Very minor request:)

`reported as "equivalent" ... cooperation rate (note that these a not the same)`

drvinceknight: (Same for the other one).

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -1,11 +1,142 @@
...
+        self.assertEqual(self.p1.name,
+                         "Generic Memory Two Player: (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)")
+        self.assertEqual(self.p2.name,
+                         "Generic Memory Two Player: (1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0)")
+
+    def test_stochastic_classification(self):
+        self.assertFalse(self.p1.classifier['stochastic'])
+        self.assertFalse(self.p2.classifier['stochastic'])
+
+class TestMemoryTwoPlayer(unittest.TestCase):
```
drvinceknight: @Nikoleta-v3 could we add some tests of the player with stochastic `sixteen_vector` (so using `TestPlayer.versus_test` with `init_kwargs`). Seeded tests, including same history/different seed to show that we indeed have stochastic behaviour?

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -1,11 +1,142 @@
...
+    p2 = MemoryTwoPlayer(sixteen_vector=(1, 0, 1, 0, 1, 0, 1, 0,
+                                         1, 0, 1, 0, 1, 0, 1, 0))
+
+    def test_name(self):
+        self.assertEqual(self.p1.name,
+                         "Generic Memory Two Player: (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)")
+        self.assertEqual(self.p2.name,
+                         "Generic Memory Two Player: (1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0)")
+
+    def test_stochastic_classification(self):
```
drvinceknight: Can rename to `test_deterministic_classification` and then also include a `test_stochastic_classification` where we use `sixteen_vector` where the values aren't all in `[0, 1]`.

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+
+    def set_sixteen_vector(self, sixteen_vector: Tuple):
+        if not all(0 <= p <= 1 for p in sixteen_vector):
+            raise ValueError("An element in the probability vector, {}, is not "
+                             "between 0 and 1.".format(str(sixteen_vector)))
+
+        states = [(hist[:2], hist[2:])
+                  for hist in list(itertools.product((C, D), repeat=4))]
+
+        self._sixteen_vector = dict(zip(states, sixteen_vector)) # type: Dict[tuple, float]
```
drvinceknight: @marcharper this no longer maps `float` on to `states`, it's a change I suggested to @Nikoleta-v3 when trying to get the type hints to work https://github.com/Axelrod-Python/Axelrod/issues/8. 

I couldn't see why we needed to `map(float` but we do do it for `memory_one` so just highlighting in case this is something I've forgotten/missed (if indeed it's not needed, let's get rid of it from `memory_one` too).

axelrod/strategies/memorytwo.py
```diff
@@ -118,8 +118,22 @@ class AON2(MemoryTwoPlayer):
...
+    1. [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1], self cooperation
+    rate: 0.952
+    2. [1, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1], self cooperation
+    rate: 0.951
+    3. [1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1], self cooperation
+    rate:  0.951
+    4. [1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 1], self cooperation
+    rate: 0.952
+
+    AON2 is implemented using vector 1 due it's self cooperation rate.
```
marcharper: nit: it's --> its

axelrod/strategies/memorytwo.py
```diff
@@ -148,6 +162,18 @@ class DelayedAON1(MemoryTwoPlayer):
     Delayed AON1 a memory two strategy also introduced in [Hilbe2017]_ and belongs to the
     AONk family. Note that AON1 is equivalent to Win Stay Lose Swift.
```
marcharper: Swift -> Shift

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
 from .defector import Defector
 
+
 C, D = Action.C, Action.D
 
 
+class MemoryTwoPlayer(Player):
+    """
+    Uses a sixteen-vector for strategies based on the last two rounds of play,
+    1.  P(C|CC, CC)
```
marcharper: Maybe better as (or something similar to): 
```Based on the 16 conditional probabilities P(X | I,J,K,L) where X, I, J, K, L in [C, D] and I, J are the players last two moves and K, L are the opponents last two moves.```

File: axelrod/strategies/memorytwo.py [hunk 12]
```diff
@@ -1,11 +1,203 @@
...
+        'memory_depth': 2,  # Memory-two Sixteen-Vector
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self, sixteen_vector: Tuple[float, float, float, float,
```
marcharper: Could use `Tuple[float, ...]` for brevity but it wouldn't enforce length 16

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+        initial: C or D
+            The initial 2 moves
+        """
+        super().__init__()
+        self._initial = initial
+        self.set_initial_sixteen_vector(sixteen_vector)
+
+    def set_initial_sixteen_vector(self, sixteen_vector):
+        if sixteen_vector is None:
+            sixteen_vector = (1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1)
```
marcharper: `tuple([1] * 16)`

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+    AON2 a memory two strategy introduced in [Hilbe2017]_. It belongs to the
+    AONk (all-or-none) family of strategies. These strategies were designed to
+    satisfy the three following properties:
+
+    1. Mutually Cooperative. A strategy is mutually cooperative if there are
+    histories for which the strategy prescribes to cooperate, and if it continues
+    to cooperate after rounds with mutual cooperation (provided the last k actions
+    of the focal player were actually consistent).
+
+    2.Error correcting. A strategy is error correcting after at most k rounds if,
```
marcharper: missing space in `2.Error`

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+    after any history, it generally takes a group of players at most k + 1 rounds
+    to re establish mutual cooperation.
```
marcharper: re-establish

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+
+    3. Retaliating. A strategy is retaliating for at least k rounds if, after
+    rounds in which the focal player cooperated while at least one coplayer defected,
```
marcharper: coplayer = opponent ? (if so, I like it)

I think this is supposed to say that the opponent defected while the player was cooperating, as written it makes it seem like there are multiple coplayers.

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+    the strategy responds by defecting the following k rounds.
+
+    In [Hilbe2017]_ the following vectors are reported as "equivalent" to AON2
+    with their respective self cooperation rate (note that these a not the same):
```
marcharper: a -> are
self cooperation -> self-cooperation

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+    2. [1, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1], self cooperation
+    rate: 0.951
+    3. [1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1], self cooperation
+    rate:  0.951
+    4. [1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 1], self cooperation
+    rate: 0.952
+
+    AON2 is implemented using vector 1 due it's self cooperation rate.
+
+    In essence is a strategy that starts of by cooperating and will cooperate again
```
marcharper: it is
off

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+
+    def __init__(self) -> None:
+        sixteen_vector = (1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1)
+        super().__init__(sixteen_vector)
+
+
+class DelayedAON1(MemoryTwoPlayer):
+    """
+    Delayed AON1 a memory two strategy also introduced in [Hilbe2017]_ and belongs to the
+    AONk family. Note that AON1 is equivalent to Win Stay Lose Swift.
```
marcharper: Swift -> Shift

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+    only after the states (CC, CC), (CD, CD), (DC, DC), (DD, DD).
+
+    Names:
+
+    - AON2: [Hilbe2017]_
+    """
+
+    name = 'AON2'
+    classifier = {
+        'memory_depth': 2,  # Memory-one Four-Vector
```
marcharper: 16 vector

axelrod/strategies/memorytwo.py
```diff
@@ -1,11 +1,203 @@
...
+    (DC, DC) and (DD, DD).
+
+    Names:
+
+    - Delayed AON1: [Hilbe2017]_
+    """
+
+    name = 'Delayed AON1'
+    classifier = {
+        'memory_depth': 2,  # Memory-one Four-Vector
```
marcharper: same

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -1,11 +1,193 @@
...
+
+    def test_exception_if_four_vector_not_set(self):
+        with warnings.catch_warnings(record=True) as warning:
+            warnings.simplefilter("always")
+            player = MemoryTwoPlayer()
+
+            self.assertEqual(len(warning), 1)
+            self.assertEqual(warning[-1].category, UserWarning)
+            self.assertEqual(str(warning[-1].message),
+                              "Memory two player is set to default, Cooperator.")
```
marcharper: remove extra space

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -1,11 +1,193 @@
...
+        'memory_depth': 2,
+        'stochastic': False,
+        'makes_use_of': set(),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_class_classification(self):
```
marcharper: Does this test have to be explicit? Or is it covered by the generic player tests?

marcharper:
Nice work so far :) Thanks for adding these strategies!

axelrod/strategies/memorytwo.py
drvinceknight: I think I'd prefer them explicitly listed, the ordered isn't clear otherwise?

axelrod/strategies/memorytwo.py
drvinceknight: I agree, I don't think the comment is useful here. :+1:

axelrod/strategies/memorytwo.py
drvinceknight: Unless your suggestion is to use your sentence to clarify what the list is and also list them which is a great call :+1:

Nikoleta-v3:
Thank you for the feedback @marcharper and thank you both for the review. I believe I addressed everything.

drvinceknight:
Thanks @Nikoleta-v3, you've addressed all my requests. Will leave it to @marcharper for final ok :+1:

PR #1173, opened 2018-04-18
remove float from memory one

Nikoleta-v3:
While working on the memory two strategies we observed that when creating the transition rates dict of the strategy there as a float argument. 

The argument seems to have been left there by previous versions, thus I am removing it.


drvinceknight:
:+1:

PR #1177, opened 2018-05-11
Fix type hints for mypy 2.1

### LINKED ISSUE #1176: mypy update has broken type hints
```
Axelrod(master) ✗: python run_mypy.py
axelrod/fingerprint.py:308: error: Value of type variable "AnyStr" of "mkstemp" cannot be "Optional[Any]"
axelrod/fingerprint.py:327: error: Argument 1 to "remove" has incompatible type "Optional[Any]"; expected "Union[bytes, str]"
axelrod/fingerprint.py:446: error: Value of type variable "AnyStr" of "mkstemp" cannot be "Optional[str]"
axelrod/fingerprint.py:459: error: Argument 1 to "remove" has incompatible type "Optional[str]"; expected "Union[bytes, str]"
axelrod/moran.py:105: error: Incompatible types in assignment (expression has type "None", variable has type "str")
axelrod/moran.py:196: error: Invalid index type "Optional[int]" for "List[Any]"; expected type "int"
axelrod/moran.py:213: error: Argument 1 to "pop" of "list" has incompatible type "Optional[int]"; expected "int"
axelrod/moran.py:216: error: Unsupported operand types for >= ("int" and "Optional[int]")
axelrod/moran.py:358: error: Incompatible types in assignment (expression has type "None", variable has type "str")
axelrod/tournament.py:86: error: Incompatible types in assignment (expression has type "None", variable has type "str")
axelrod/tournament.py:87: error: Incompatible types in assignment (expression has type "None", variable has type "int")
axelrod/strategies/apavlov.py:32: error: Incompatible types in assignment (expression has type "None", variable has type "str")
axelrod/strategies/apavlov.py:96: error: Incompatible types in assignment (expression has type "None", variable has type "str")
axelrod/strategies/darwin.py:51: error: Incompatible types in assignment (expression has type "None", variable has type "Dict[Any, Any]")
axelrod/strategies/hunter.py:123: error: Incompatible types in assignment (expression has type "None", variable has type "Tuple[Any]")
axelrod/strategies/memoryone.py:196: error: Argument 1 to "round" has incompatible type "Optional[float]"; expected "float"
```

Working on this now (as it'll cause CI to break on incoming PRs).

drvinceknight:
Closes #1176

Various things I've needed to do:

- In a number of places include that the type is `Optional[]`.
- In a number of places include an `assert` so that `mypy` knows that
  the type is no longer None.
- I have made a minor refactor to the Moran process, I could have used
  asserts but this actually pointed out a spot where I believe the tests
  were incorrect: we were testing a specific situation where
  `MoranProcess.birth` or `MoranProcess.death` should have been called with
  default `index=None` but was actually being tested with `index=0`.
- **I have ignored** a type hint check in `fingerprint.py`: I was
  completely unable to get this one happy

To ensure that any new PRs don't break CI let's merge this under bug fix. :+1: :)

### DIFF
--- a/axelrod/fingerprint.py
+++ b/axelrod/fingerprint.py
@@ -268,7 +268,7 @@ class AshlockFingerprint(object):
 
     def fingerprint(
         self, turns: int = 50, repetitions: int = 10, step: float = 0.01,
-        processes: int=None, filename: str = None,
+        processes: int = None, filename: str = None,
         progress_bar: bool = True
 ) -> dict:
         """Build and play the spatial tournament.
@@ -305,7 +305,7 @@ class AshlockFingerprint(object):
 
         temp_file_descriptor = None
         if filename is None:
-            temp_file_descriptor, filename = mkstemp()
+            temp_file_descriptor, filename = mkstemp()  # type: ignore
 
         edges, tourn_players = self.construct_tournament_elements(
             step, progress_bar=progress_bar)
@@ -323,6 +323,7 @@ class AshlockFingerprint(object):
             filename, progress_bar=progress_bar)
 
         if temp_file_descriptor is not None:
+            assert filename is not None
             os.close(temp_file_descriptor)
             os.remove(filename)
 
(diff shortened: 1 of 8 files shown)


drvinceknight:
Merging.

PR #1178, opened 2018-05-13
Add Tricky Level Punisher

Nikoleta-v3:
I picked up #1132 which implements a strategy called Tricky Level Punisher. 

The `pr` was failing due a division by zero error. Moreover, I had to tweak the tests of `MetaMajorityLongMemory` because now `TrickyLevelPunisher` is in the strategy's team.

```
>> import axelrod as axl

>> player = axl.MetaMajorityLongMemory()
>> tricky = axl.TrickyLevelPunisher()
>> tricky in player.team
True
```

Note that this `pr` will fail due to #1176, I will rebase once #1177 is merged.

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -65,7 +65,7 @@ from .negation import Negation
 from .oncebitten import OnceBitten, FoolMeOnce, ForgetfulFoolMeOnce, FoolMeForever
 from .prober import (CollectiveStrategy, Prober, Prober2, Prober3, Prober4,
                      HardProber, NaiveProber, RemorsefulProber)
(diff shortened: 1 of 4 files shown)


drvinceknight:
> Rather than update the meta player tests each time, which amounts to a "change detector test" at the moment and can be a pain to update, we might consider simply having a test for the number of team members.

Yeah great call @marcharper, perhaps not just the number but a test of the makeup of the tea (specific members?) also good to have a test that matches complete too. (@Nikoleta-v3 if you could handle this that'd be great :+1:)

#1177 is merged so you should be ok to rebase.

drvinceknight:
Thanks @Nikoleta-v3 merging this in :+1:

Issue #1183, opened 2018-07-06
Library is not compatible with `prompt_toolkit>=2.0.0`wait for update to released ipython to fix.


drvinceknight:
With `prompt_toolkit` version 2.0.3 we get an import error:

```
======================================================================
ERROR: axelrod (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: axelrod
Traceback (most recent call last):
  File "/home/vince/anaconda3/lib/python3.6/unittest/loader.py", line 153, in loadTestsFromName
    module = __import__(module_name)
  File "/home/vince/src/Axelrod/axelrod/__init__.py", line 17, in <module>
    from .strategies import *
  File "/home/vince/src/Axelrod/axelrod/strategies/__init__.py", line 2, in <module>
    from ._strategies import *
  File "/home/vince/src/Axelrod/axelrod/strategies/_strategies.py", line 47, in <module>
    from .human import Human
  File "/home/vince/src/Axelrod/axelrod/strategies/human.py", line 3, in <module>
    from prompt_toolkit.token import Token
ImportError: cannot import name 'Token'

```

Note that in #1182 I've pinned the version of `prompt_toolkit` in `requirements.txt`. Once this is fixed we should unpin.

drvinceknight:
(FYI: Looks like `prompt_toolkit` version 2.0.3 breaks ipython so probably no rush to "fix" this.)

drvinceknight:
Following a discussion with @danilobellini on #1223 it looks like the latest conda release of jupyter/ipython now uses `prompt_toolkit>=2.0.0`.

```
$  conda create -q -n tmp-env-test-ipython python=3
$ source activate tmp-env-test-ipython
$ conda install jupyter
$ pip show ipython
Name: ipython
Version: 7.1.1
Summary: IPython: Productive Interactive Computing
Home-page: https://ipython.org
Author: The IPython Development Team
Author-email: ipython-dev@python.org
License: BSD
Location: /home/vince/anaconda3/envs/tmp-env-test-ipython/lib/python3.7/site-packages
Requires: prompt-toolkit, setuptools, traitlets, jedi, pickleshare, backcall, pexpect, decorator, pygments
Required-by: jupyter-console, ipywidgets, ipykernel
```

Could be worth reapplying #1212 but we should make sure it works and doesn't break common tools.

danilobellini:
#1212 didn't work here with prompt_toolkit v2.

```
[...]

PR #1182, opened 2018-07-06
WIP: Fix compute features and retrain ann

drvinceknight:
This closes #1172

@marcharper note that I trained ANNNoise with a hidden layer of size 10 due to the
docstring. This was in fact an incorrect (the layer had size 5) docstring but retraining took
quite a while so I went with changing the layers.

I've also archived the training data (link now in docstrings of strategies and also here: doi.org/10.5281/zenodo.1306598) so please check if I've gotten the moving of that to the right place in the repo right :)

I've run a quick tournament (still takes a little while) and things don't seem to have changed much:

```
>>> import axelrod as axl
>>> axl.seed(0)
>>> players = [s() for s in axl.strategies if not s.classifier["long_run_time"]]
>>> tournament = axl.Tournament(players=players, repetitions=20)
>>> results = tournament.play(processes=0)
>>> results.ranked_names[:10]
['Evolved HMM 5',
 'EvolvedLookerUp2_2_2',
 'Evolved ANN 5',
 'Evolved FSM 16',
 'PSO Gambler 2_2_2',
 'Evolved ANN',
 'Evolved FSM 16 Noise 05',
 'Evolved FSM 4',
 'PSO Gambler 1_1_1',
 'Evolved ANN 10 Noise 05']

```

PR #1184, opened 2018-07-06
Pin prompt toolkit because of backwards comp.

drvinceknight:
This is related to #1183 . Currently a clean install of Axelrod will fail because of backwards incompatible change to `prompt_toolkit` version 2+. This pins `prompt_toolkit`.

(I suggest we merge this but leave #1182 open as a place holder, there's no rush with unpinning `prompt_toolkit` as ipython pins this too I believe: https://github.com/ipython/ipython/blob/master/setup.py#L193.)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 numpy>=1.9.2
 matplotlib>=1.4.2
 tqdm>=3.4.0
-prompt-toolkit==1.0.15
+prompt-toolkit>=1.0.7,<2.0.0
 scipy>=0.19.0
 hypothesis==3.2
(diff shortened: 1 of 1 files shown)


drvinceknight:
FYI: 5f048a9 losens the pin but it's falling over on Appveyor I'll figure it out...

drvinceknight:
@meatballs this is probably good to go now, the diff is probably more complicated than it needs to be as I've taken the opportunity to alphabetically sort `requirements.txt`. The only real change is:

```
-prompt-toolkit>=1.0.7
+prompt-toolkit>=1.0.7,<2.0.0
```

Issue #1187, opened 2018-07-18
Inconsistant CI failures


drvinceknight:
I'm not entirely sure what's the cause here but there seem to be inconsistant failures of some tests (a restart usually gets them to pass), I have not been able to reproduce this locally.

Here is the traceback from a failed run on appveyor:

```
python -m unittest discover
.....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................x.................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................C:\projects\axelrod\axelrod\strategies\memoryone.py:74: UserWarning: Memory one player is set to default (1, 0, 0, 1).
  warnings.warn("Memory one player is set to default (1, 0, 0, 1).")
..C:\projects\axelrod\axelrod\strategies\memoryone.py:74: UserWarning: Memory one player is set to default (1, 0, 0, 1).
  warnings.warn("Memory one player is set to default (1, 0, 0, 1).")
...............................................................................C:\projects\axelrod\axelrod\strategies\memorytwo.py:75: UserWarning: Memory two player is set to default, Cooperator.
  warnings.warn("Memory two player is set to default, Cooperator.")
..........C:\projects\axelrod\axelrod\strategies\memorytwo.py:75: UserWarning: Memory two player is set to default, Cooperator.
  warnings.warn("Memory two player is set to default, Cooperator.")
Playing matches: 100%|##########| 10/10 [00:00<00:00, 53.33it/s]
Analysing: 100%|##########| 25/25 [00:00<00:00, 177.78it/s]
Playing matches: 100%|##########| 10/10 [00:00<00:00, 53.33it/s]
Analysing: 100%|##########| 25/25 [00:00<00:00, 200.01it/s]
100%|##########| 9/9 [00:00<00:00, 575.86it/s]
100%|##########| 50/50 [00:00<00:00, 800.08it/s]
100%|##########| 18/18 [00:00<00:00, 1152.33it/s]
100%|##########| 18/18 [00:00<00:00, 1151.86it/s]
100%|##########| 18/18 [00:00<00:00, 1152.05it/s]
100%|##########| 18/18 [00:00<00:00, 1151.91it/s]
100%|##########| 18/18 [00:00<00:00, 1151.86it/s]
100%|##########| 18/18 [00:00<00:00, 575.94it/s]
[...]

PR #1193, opened 2018-08-15
Spruce up actions module

langner:
* Improves docstrings and move to numpydoc format
* Reviewed unit tests and add a few edge cases


drvinceknight:
The CI failure here is due to #1187 which is fixed in #1191 which should be going in soon. :)

langner:
Note: this was for #347

PR #1194, opened 2018-08-16
Add a matplotlibrc

drvinceknight:
EDIT: ~~I have also added some comments to ensure we do not reorder these
imports.~~

### DIFF
--- a/axelrod/tests/unit/test_plot.py
+++ b/axelrod/tests/unit/test_plot.py
@@ -1,19 +1,10 @@
 import tempfile
 import unittest
 
-import matplotlib  # isort: skip
-matplotlib.use("agg")  # This line need to be before `import axelrod` and
-                       # `import matplotlib.pyplot` for CI to work consistently
-                       # See
-                       # https://github.com/Axelrod-Python/Axelrod/issues/1187
-
 import matplotlib.pyplot as plt
 from numpy import mean
 
 import axelrod
-
-
-
 
 class TestPlot(unittest.TestCase):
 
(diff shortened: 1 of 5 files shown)


marcharper:
How does mpl know to look for the rc file there?

drvinceknight:
> How does mpl know to look for the rc file there?

It automatically picks it up from the dir the tests are run from (https://matplotlib.org/users/customizing.html#the-matplotlibrc-file), so this does not change overall behavior of the library but just ensures the correct setting for CI.

marcharper:
Great, thanks for the explanation.

axelrod/tests/unit/test_plot.py
```diff
@@ -1,14 +1,12 @@
...
 from numpy import mean
-import tempfile
 
-import matplotlib
-matplotlib.use("agg")
-
-import matplotlib.pyplot as plt
+import axelrod
 
+import matplotlib
```
marcharper: should this one occur before `import axelrod`?

marcharper:
If it works then it's fine with me. Might want to put in a comment somewhere that explains what's happening so someone doesn't move things around later, if that still matters with the rc file.

axelrod/tests/unit/test_plot.py
drvinceknight: Import orders don't actually matter anymore, the `matplotlibrc` ooverrides everything but I got myself confused there and will reorder to fit PEP8! (and a bonus rerun of CI to make sure it's still working :crossed_fingers:)

drvinceknight:
> Might want to put in a comment somewhere that explains what's happening so someone doesn't move things around later, if that still matters with the rc file.

The import orders no longer matter (the rc file sets it all) but I'll add a comment to the top of the rc file so we now not to delete it :)

PR #1195, opened 2018-08-16
Spruce up deterministic_cache

langner:
Next bit for #347.

This includes several things:
* Improved (hopefully) docstrings and a module docstring with a little example
* Moved cache key helper functions to module level - these were static methods so it felt natural to move them out of the class... but I'm not 100% there was a reason for doing this. If so, I can move back, although `@staticmethod` IMO is best used sparingly.
* Refactored unit tests to test public API instead of the internal helper funcs. Again, assumed this is a good thing, but not 100% there wasn't some decision to test internal functions throughout the library.

### DIFF
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -1,3 +1,17 @@
+"""Tools for caching the results of deterministic matches.
+
+The cache, in most cases, can simply be treated as a dictionary:
+
+cache = DeterministicCache()
+cache[key1] = result1
+cache[key2] = result2
+...
+if some_key in cache:
+    do_something(cache[some_key])
+else:
+    ...
+"""
+
 from collections import UserDict
 import pickle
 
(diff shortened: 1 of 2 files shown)


drvinceknight:
> The failure on appveyor is (again) because of #1187: once #1194 is merged that should hopefully (almost certainly!) stop happening :)

I've set the build going again.

axelrod/deterministic_cache.py
```diff
@@ -1,3 +1,17 @@
+"""Tools for caching the results of deterministic matches.
+
+The cache, in most cases, can simply be treated as a dictionary:
+
+cache = DeterministicCache()
+cache[key1] = result1
+cache[key2] = result2
+...
+if some_key in cache:
+    do_something(cache[some_key]
```
marcharper: )

axelrod/deterministic_cache.py
```diff
@@ -10,6 +24,63 @@
...
+        A 3-tuple: (player instance, player instance, match length)
+    """
+    return key[0].name, key[1].name, key[2]
+
+
+def _is_valid_key(key: CachePlayerKey) -> bool:
+    """Validate a deterministic cache player key.
+
+    The key should always be a 3-tuple, with a pair of axelrod.Player
+    instnaces and one integer. Both players should be deterministic.
```
marcharper: nit: instances

axelrod/tests/unit/test_deterministic_cache.py
```diff
@@ -27,6 +28,9 @@ def tearDownClass(cls):
         os.remove(cls.test_save_file)
         os.remove(cls.test_load_file)
 
+    def setUp(self):
+        self.cache = DeterministicCache()
```
marcharper: Does this create a shared cache across all tests? That could cause some confusion.

axelrod/deterministic_cache.py
langner: Done.

axelrod/deterministic_cache.py
langner: Done.

axelrod/tests/unit/test_deterministic_cache.py
langner: That's a good point. But since `setUp` is called before every test method, `self.cache` is assigned a new, empty cache for each one. Was that your concern?

marcharper: Yep, thanks for the explanation.

PR #1197, opened 2018-08-20
Spruce up deterministic_cache, ecosystem and eigen

langner:
Next batch for #347 

* Minor improvements to docstrings and unit tests for ecosystem
* Simplified unit tests for deterministic_cache a little bit more
* Minor edits to eigen module, and made several function internal

PR #1198, opened 2018-08-20
Pass fitness function moran process

Nikoleta-v3:
Using a fitness function that scales the utility for the Moran process is commonly used in the literature. 

This `PR` implements that feature and now the Moran process has a new argument, the `fitness_function` which is set as `None` as a default. 

I added a test which tests that cooperators can take over a population of defectors when the selection is weak. Also added an example in the documentations.

Please note that after chatting with @drvinceknight,  I run `black` and `isort` on the files 
(commit 4919dd2). This might make the diffs more complicated than needed, please let me know if you want me to undo that :+1: .

### DIFF
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -13,9 +13,10 @@ from .match import Match
 from .random_ import randrange
 
 from typing import List, Tuple, Set, Optional
+from types import FunctionType
 
 
-def fitness_proportionate_selection(scores: List) -> int:
+def fitness_proportionate_selection(scores: List, fitness_function: FunctionType = None) -> int:
     """Randomly selects an individual proportionally to score.
 
     Parameters
@@ -27,7 +28,10 @@ def fitness_proportionate_selection(scores: List) -> int:
     An index of the above list selected at random proportionally to the list
     element divided by the total.
     """
-    csums = np.cumsum(scores)
+    if fitness_function is None:
+        csums = np.cumsum(scores)
+    else:
+        csums = np.cumsum([fitness_function(s) for s in scores])
     total = csums[-1]
     r = random.random() * total
 
@@ -44,7 +48,8 @@ class MoranProcess(object):
                  deterministic_cache: DeterministicCache = None,
                  mutation_rate: float = 0., mode: str = 'bd',
                  interaction_graph: Graph = None,
-                 reproduction_graph: Graph = None) -> None:
+                 reproduction_graph: Graph = None,
+                 fitness_function: FunctionType = None) -> None:
         """
         An agent based Moran process class. In each round, each player plays a
         Match with each other player. Players are assigned a fitness score by
@@ -138,6 +143,7 @@ class MoranProcess(object):
         assert list(v1) == list(v2)
         self.interaction_graph = interaction_graph
         self.reproduction_graph = reproduction_graph
+        self.fitness_function = fitness_function
         # Map players to graph vertices
         self.locations = sorted(interaction_graph.vertices())
         self.index = dict(zip(sorted(interaction_graph.vertices()),
@@ -212,11 +218,11 @@ class MoranProcess(object):
             # possible choices
             scores.pop(index)
             # Make sure to get the correct index post-pop
-            j = fitness_proportionate_selection(scores)
+            j = fitness_proportionate_selection(scores, fitness_function=self.fitness_function)
             if j >= index:
                 j += 1
         else:
-            j = fitness_proportionate_selection(scores)
+            j = fitness_proportionate_selection(scores, fitness_function=self.fitness_function)
         return j
 
     def fixation_check(self) -> bool:
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -16,24 +16,24 @@ from .random_ import randrange
 
 
 def fitness_proportionate_selection(
-    scores: List, fitness_function: Callable = None
+    scores: List, fitness_transformation: Callable = None
 ) -> int:
     """Randomly selects an individual proportionally to score.
 
     Parameters
     ----------
     scores: Any sequence of real numbers
-    fitness_function: A function mapping a score to a (non-negative) float
+    fitness_transformation: A function mapping a score to a (non-negative) float
 
     Returns
     -------
     An index of the above list selected at random proportionally to the list
     element divided by the total.
     """
-    if fitness_function is None:
+    if fitness_transformation is None:
         csums = np.cumsum(scores)
     else:
-        csums = np.cumsum([fitness_function(s) for s in scores])
+        csums = np.cumsum([fitness_transformation(s) for s in scores])
     total = csums[-1]
     r = random.random() * total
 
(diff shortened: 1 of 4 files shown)


axelrod/moran.py
```diff
@@ -1,21 +1,24 @@
...
-from typing import List, Tuple, Set, Optional
 
-
-def fitness_proportionate_selection(scores: List) -> int:
+def fitness_proportionate_selection(
+    scores: List, fitness_function: FunctionType = None
+) -> int:
     """Randomly selects an individual proportionally to score.
 
     Parameters
```
drvinceknight: Need to add `fitness_function` to the docstring (similarly for `MoranProcess.__init__`):

I suggest:

```
fitness_function: A function mapping a score to a float.
```

drvinceknight:
The failure is the same as on #1197:

```
FAIL: test_returns_foil_inspection_strategy_of_opponent (axelrod.tests.strategies.test_geller.TestGeller)
  File "C:\projects\axelrod\axelrod\tests\strategies\test_geller.py", line 65, in test_returns_foil_inspection_strategy_of_opponent
    self.versus_test(axelrod.MindReader(), expected_actions=[(D, D), (D, C), (D, D)], seed=seed)
  File "C:\projects\axelrod\axelrod\tests\strategies\test_player.py", line 521, in versus_test
    self.assertEqual(match.play(), expected_actions)
AssertionError: Lists differ: [(D, D), (D, D), (D, D)] != [(D, D), (D, C), (D, D)]

```

axelrod/moran.py
Nikoleta-v3: ah missed that. Fixing it

Nikoleta-v3:
Regarding the failure if it due to f474f2e, should I wait for @langner to undo the change?

marcharper:
This is a nice addition. Have you thought about making the transformation a property of the Moran process instead of an argument to fitness_proportionate_selection?

Generally in the literature we have a payoff that is associated to a game and a total payoff vector `u` associated to what we call a match. Then we typically apply a transformation such as `f_i = exp(\beta u_i)` or `f_i = 1 - w + w * u_i`. Fitness proportionate selection acts on the vector `f`. In other words, it's not clear to me that `fitness_proportion_selection` should apply the transformation, or if it should be functional composition: `fitness_proportion_selection(transformation(payoffs))`. Numerically the result is the same, but there's an API difference. It feels to me like it makes more sense to do the composition than it does to use a kwarg for fitness_proportionate_selection.

As a point of comparison, if we were to add these transformations to the tournament class, where would they go? It feels more like a property of the tournament than the game itself -- you can imagine running the same tournament with different strengths of selection.

In the same vein, we should think carefully about the names of these arguments. Applying `exp(\beta ...)` is commonly called "Fermi selection", and I think it's more appropriate to call these functions `fitness_transformations` or something similar rather than `fitness_functions`, which to me is the full composition `transformation(payoffs)` (i.e., what fitness_proportionate_selection acts on).

Regardless, glad to have this in the library. I've used these a lot in my other work / libraries (see e.g. [here](https://github.com/marcharper/stationary/blob/master/stationary/processes/incentives.py#L134)).

langner:
> Regarding the failure if it is due to f474f2e, should I wait for @langner to undo the change?

Hmm... I'm not sure it's a matter of undoing that change, since I made it in an effort to squash exactly this error (or it's reverse). It could be that I introduced some indeterminacy in #1196, but I haven't investigated yet.

drvinceknight:
> Generally in the literature we have a payoff that is associated to a game and a total payoff vector u associated to what we call a match. Then we typically apply a transformation such as f_i = exp(\beta u_i) or f_i = 1 - w + w * u_i. Fitness proportionate selection acts on the vector f. In other words, it's not clear to me that fitness_proportion_selection should apply the transformation, or if it should be functional composition: fitness_proportion_selection(transformation(payoffs)). Numerically the result is the same, but there's an API difference. It feels to me like it makes more sense to do the composition than it does to use a kwarg for fitness_proportionate_selection.

You're the expert here @marcharper and your suggestion makes sense to me.

We could even do this without add anything at all and just passing a transformed game I guess?

```python
>>> R_f, P_f, S_f, T_f = some_transformation(R, P, S, T)
>>> game = axl.Game(R_f, P_f, S_f, T_f)
>>> mp = axl.MoranProcess(..., game=game...)
```
 
(Similarly a user could wrap this for tournaments etc...)

> 
> In the same vein, we should think carefully about the names of these arguments. Applying exp(\beta ...) is commonly called "Fermi selection", and I think it's more appropriate to call these functions fitness_transformations or something similar rather than fitness_functions, which to me is the full composition transformation(payoffs) (i.e., what fitness_proportionate_selection acts on).

Yeah :+1: 

> Regardless, glad to have this in the library. I've used these a lot in my other work / libraries (see e.g. here).

If it's "simple" functional composition, does it need to be in the library at all? Perhaps just some documentation about applying a transformation to the game values? (Perhaps at the bottom of `moran.rst`?).

What do you reckon @marcharper?

drvinceknight:
> Hmm... I'm not sure it's a matter of undoing that change, since I made it in an effort to squash exactly this error (or it's reverse)

Yeah I noticed that @Karlos78, strange :) I've opened https://github.com/Axelrod-Python/Axelrod/issues/1200 and have a possible idea of a fix. @Nikoleta-v3 I suggest you wait to see first what Marc says about the overall idea and second for this to be fixed :+1:

drvinceknight:
> Regarding the failure if it is due to f474f2e, should I wait for @langner to undo the change?

If you rebase on to master @Nikoleta-v3 this error is now fixed :+1:

Nikoleta-v3:
```
$ python run_mypy.py
axelrod/moran.py:239: error: Argument "fitness_function" to "fitness_proportionate_selection" has incompatible type "Optional[MethodType]"; expected "Optional[FunctionType]"
axelrod/moran.py:245: error: Argument "fitness_function" to "fitness_proportionate_selection" has incompatible type "Optional[MethodType]"; expected "Optional[FunctionType]"
```

Nikoleta-v3:
Fixed the type hint error :+1:

axelrod/moran.py
```diff
@@ -212,11 +235,15 @@ def birth(self, index: int = None) -> int:
...
-            j = fitness_proportionate_selection(scores)
+            j = fitness_proportionate_selection(
+                scores, fitness_function=self.fitness_function
+            )
             if j >= index:
                 j += 1
         else:
-            j = fitness_proportionate_selection(scores)
+            j = fitness_proportionate_selection(
+                scores, fitness_function=self.fitness_function
```
drvinceknight: So this would be `fitness_transformation=self.fitness_transformation`.

axelrod/moran.py
```diff
@@ -1,33 +1,39 @@
...
-def fitness_proportionate_selection(scores: List) -> int:
+def fitness_proportionate_selection(
+    scores: List, fitness_function: Callable = None
+) -> int:
     """Randomly selects an individual proportionally to score.
 
     Parameters
     ----------
     scores: Any sequence of real numbers
+    fitness_function: A function mapping a score to a (non-negative) float
```
drvinceknight: Following @marcharper's suggestion can we change this to be `fitness_transformation` as the variable name? Similarly in `MoranProcess.__init__` and the docs :+1: :)

drvinceknight:
Thanks @Nikoleta-v3 :+1:

PR #1205, opened 2018-09-01
Spruce up fingerprint (mostly AshlockFingerprint)

langner:
* Prefix internal functions and methods with underscore
* Test the external API only
* Be more locally explicit in unit tests (no need for


langner:
Note: this was also for #347

Issue #1217, opened 2018-10-29
Library no longer compatible with ipython


drvinceknight:
After #1212 I now get:

```
Axelrod(master): ipython
Traceback (most recent call last):
  File "/home/vince/anaconda3/bin/ipython", line 7, in <module>
    from IPython import start_ipython
  File "/home/vince/anaconda3/lib/python3.6/site-packages/IPython/__init__.py", line 55, in <module>
    from .terminal.embed import embed
  File "/home/vince/anaconda3/lib/python3.6/site-packages/IPython/terminal/embed.py", line 16, in <module>
    from IPython.terminal.interactiveshell import TerminalInteractiveShell
  File "/home/vince/anaconda3/lib/python3.6/site-packages/IPython/terminal/interactiveshell.py", line 22, in <module>
    from prompt_toolkit.shortcuts import create_prompt_application, create_eventloop, create_prompt_layout, create_output
ImportError: cannot import name 'create_prompt_application'
```

This is because:

```
jupyter-console 5.2.0 has requirement prompt_toolkit<2.0.0,>=1.0.0, but you'll have prompt-toolkit 2.0.6 which is incompatible.
ipython 6.4.0 has requirement prompt-toolkit<2.0.0,>=1.0.15, but you'll have prompt-toolkit 2.0.6 which is incompatible.
```

Should we revert #1212 and wait for ipython to be compatible with a `prompt-toolkit>=2.0.0` or is there a better solution?

drvinceknight:
On the `revert-prompt-toolkit` branch I've pushed 35bac5a which reverts the `prompt_toolkit` part of #1212. With that and after `pip install -r requirements.txt`: `ipython` now works. 

If we want that as a fix I'll open a PR and then we should open #1183 again and perhaps change the title to be clear that it's worth waiting for ipython to update their `prompt_toolkit` dependency.

marcharper:
^ sounds good to me.

PR #1223, opened 2018-11-20
Make Human work w/ both prompt_toolkit v1 and v2

danilobellini:
As of today, some systems work with prompt_toolkit v1 by default, but
Arch Linux has migrated to v2, and it's the last version on PyPI

To maintain the Arch Linux package of this library, it was required to
revert the 8dba4c0 commit, yet that didn't suffice to make the Human
class work in that system

This commit changes the Human strategy to work on environments with
either prompt_toolkit versions, the main differences are:

- Human._history_toolbar no longer return a list with a tuple, but
  just the raw content (making its tests more straightforward)
- self.status_messages["toolbar"] is a function as expected by the
  prompt_toolkit version available

Although the "pygments.toolbar" string is there when using the
prompt_toolkit v2, the "pygments" library isn't a dependency, since
using the Human strategy works fine in a virtual environment without
pygments

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ hypothesis==3.2
 matplotlib>=2.0.0,<3.0.0
 numpy>=1.9.2
 pandas>=0.18.1
-prompt-toolkit>=1.0.7,<2.0.0
+prompt-toolkit>=1.0.7
 scipy>=0.19.0
 toolz>=0.8.0
 tqdm>=3.4.0
--- a/axelrod/strategies/human.py
+++ b/axelrod/strategies/human.py
@@ -3,13 +3,27 @@ from os import linesep
 from axelrod.action import Action
 from axelrod.player import Player
 from prompt_toolkit import prompt
-from prompt_toolkit.styles import style_from_dict
-from prompt_toolkit.token import Token
 from prompt_toolkit.validation import ValidationError, Validator
+
+try:  # prompt_toolkit v1
+    from prompt_toolkit.styles import style_from_dict
+    from prompt_toolkit.token import Token
+
+    token_toolbar = Token.Toolbar
+    bottom_toolbar_name = "get_bottom_toolbar_tokens"
+    PROMPT2 = False
+
+except ImportError:  # prompt_toolkit v2
+    from prompt_toolkit.styles import Style
+
+    style_from_dict = Style.from_dict
+    token_toolbar = "pygments.toolbar"
+    bottom_toolbar_name = "bottom_toolbar"
+    PROMPT2 = True
 
 C, D = Action.C, Action.D
 
-toolbar_style = style_from_dict({Token.Toolbar: "#ffffff bg:#333333"})
+toolbar_style = style_from_dict({token_toolbar: "#ffffff bg:#333333"})
 
 
 class ActionValidator(Validator):
@@ -65,7 +79,7 @@ class Human(Player):
         self.symbols = {C: c_symbol, D: d_symbol}
         self.opponent_history = []
 
-    def _history_toolbar(self, cli):
+    def _history_toolbar(self):
         """
         A prompt-toolkit function to define the bottom toolbar.
         Described at http://python-prompt-toolkit.readthedocs.io/en/latest/pages/building_prompts.html#adding-a-bottom-toolbar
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -4,7 +4,7 @@ hypothesis==3.2
 matplotlib>=2.0.0,<3.0.0
 numpy>=1.9.2
 pandas>=0.18.1
-prompt-toolkit>=1.0.7,<2.0.0
+prompt-toolkit>=1.0.7
```
drvinceknight: This breaks the system install of ipython which is not yet compatible with `prompt-toolkit>2.0.0`. https://github.com/Axelrod-Python/Axelrod/issues/1183

Once ipython is using the latest `prompt-toolkit` we will migrate to just use `2.0.0+`.

drvinceknight: See also: https://github.com/Axelrod-Python/Axelrod/issues/1217

danilobellini:
I'm maintaining [the package for this library in the AUR](https://aur.archlinux.org/packages/python-axelrod). For the v4.4.0-2 packaged in AUR, I've got [this patch](https://github.com/Axelrod-Python/Axelrod/commit/8dba4c0.patch) and applied it using this:

```bash
patch -Rp1 < "$srcdir/8dba4c0.patch"
```

As it didn't suffice, I've written this code I'm pulling right now.

drvinceknight:
Closing this, please reopen if further discussion is required. :) :+1:

requirements.txt
danilobellini: IPython already requires v2+: https://github.com/ipython/ipython/blob/7.1.1/setup.py#L192

danilobellini: Actually, the oldest IPython that required prompt_toolkit v1 was the 6.5.0, and the marked line is compatible with it (actually, it's compatible with both versions). https://github.com/ipython/ipython/blob/6.5.0/setup.py#L193

drvinceknight: Perhaps my diagnosis of the source of the issue of #1217 and the linked discussions on #1183 isn't correct. Nonetheless the fix in #1212 is preferable as it'll move us to just using `prompt_toolkit` v1 **but** any fix needs to not break the common data science tools like jupyter and ipython and (as of 3 minutes ago): updating to the conda distribution of Jupyter currently keeps ipython at `6.5`.

danilobellini:
Something seem wrong in Travis CI:

```
fatal: Couldn't find remote ref refs/pull/1223/merge
```

drvinceknight:
Could that be because I closed the PR? Let me reopen to see if that keeps travis happy.

drvinceknight:
I've just installed jupyter in to a clean environment and it now brings `ipython 7.1.1`.

danilobellini:
> I've just installed jupyter in to a clean environment and it now brings `ipython 7.1.1`.

Actually, I couldn't install both jupyter and this axelrod library together in a virtual environment, and that was the main reason I've written the PKGBUILD for a system install in Arch Linux.

After Arch updated to prompt_toolkit to v2 in the community repo, I had no alternative but to patch some fix.

drvinceknight:
Sounds very similar to the opposite problem I described in #1217.

I'm just jotting down some details in #https://github.com/Axelrod-Python/Axelrod/issues/1183, could I suggest we continue the conversation there? I think that if Jupyter/ipython is all on prompt_toolkit v2 then the preffered solution would be to just move to that.

drvinceknight:
https://github.com/Axelrod-Python/Axelrod/issues/1183#issuecomment-440257512

danilobellini:
This is more of a transition PR. I can change this PR to be v2-only, but if the library were mine, I'd prefer to keep both versions during a transition period (e.g. dropping the v1 support only in Axelrod v5.1, or when Ubuntu update to v2), unless it's not possible to keep both.

I'll replace the `prompt_toolkit v1` comment by the `pragma: no cover` to avoid that coverage decrease.

drvinceknight:
> keep both versions during a transition period (e.g. dropping the v1 support only in Axelrod v5.1, or when Ubuntu update to v2), unless it's not possible to keep both.

Yup, I agree, the v2/v1 stuff here looks good to me. @danilobellini thank you for taking the time to discuss this. Once merged we'll open an issue to document the removal of v1 imports when we're ready to do so.

@danilobellini we have a two core developer review rule so @marcharper and/or @meatballs will probably be along when they get a moment :)

PR #1225, opened 2018-12-14
Fix citations link on readme

marcharper:


drvinceknight:
I believe the failure is something to do with the windows platform (64 vs 32 bit?), I've made some attempts on #1226 (a PR to your branch) but without success.

drvinceknight:
> I believe the failure is something to do with the windows platform (64 vs 32 bit?), I've made some attempts on #1226 (a PR to your branch) but without success.

Scrap that, looks like #1226 fixes things, I tried one last thing which was to not use conda envs anymore and rely on the wheels (our use of conda was due to things not working nicely which is probably fixed now). I haven't seen the full test suite pass (need to take the pup out for a walk) but the tests are all running and no more 32/64 error :+1:

#1226 also has the advantage of really clearing up `appveyor.yml`.

I suggest you merge #1226 (squash the history?) in to this branch which should (famous last words) fix all test failures. :+1:

drvinceknight:
Once tests failures are fixed this is obviously good to go from my pov :+1:

PR #1228, opened 2018-12-23
 Added Rowsam (k58r) strategy from Axelrod's second.

gaffney2010:
Adapted from https://github.com/Axelrod-Python/TourExec/blob/v0.3.0/src/strategies/k58r.f

In adapting:
- I called KAM distrust_points.
- I refactored NPHA in "Coop Def cycle" modes.
- I changed the K to M comparators in lines 12-15 to a comparison on K/M, which I call points_per_turn.


gaffney2010:
Yes, I can run them.  I'm on command line only until Wednesday; will run it then.

gaffney2010:
axl_fort
axl_py

Bit of a delay on these; I had trouble getting everything setup.

marcharper:
Looks good to me. Is it possible to generate fingerprints against more strategies? I think previously we've tossed in `Random(p)` for e.g.  `p in np.arange(0, 1, 0.05)` as well.

gaffney2010:
Here they are.

rand_axl_fort
rand_axl_py

drvinceknight:
Apologies for the delay in getting to this, just got back online :)

Looks great and thanks @gaffney2010 :+1:

PR #1229, opened 2018-12-25
Speed up caching logic

gaffney2010:
The caching is inefficient when prob_end is used.  The current paradigm caches by players and number of turns in match, for non-stochastic.  But this ignores the fact that you can read a shorter match from the result of a longer match.  For example if we have a 20-turn match cached, and we wish to run a 19-turn match, we COULD just read the first 19 moves from the 20-turn match, but the current paradigm would completely re-run because no 19-turn match has been run yet.

I changed the code to
1. Cache on player-pairs only.
2. Only re-run a match if the cache is shorter than the desired match.  In that case overwrite the cached value.
3. Sliced the cached value after reading, so as to get the first n moves, as desired.

I ran the code below and found an 80% speed-up.  I don't think this adds complexity, and keying on player-pairs only is simpler and more natural.

Experimental code:
```
import axelrod as axl
import time

MATCH_SIMS = 100

player1, player2 = axl.Cooperator(), axl.DBS()
players = (player1, player2)
match = axl.Match(players, prob_end=0.01)

start_time = time.time()
prev_time = time.time()
for _ in range(MATCH_SIMS):
  match.play()
  print("Running simulation {} of {}.  Played {} rounds.  It took {} secs.".format(
      _, MATCH_SIMS, len(match.result), time.time() - prev_time))
  prev_time = time.time()
end_time = time.time()
print("================")
print("Total time for {} simulations: {} secs.".format(MATCH_SIMS, end_time - start_time))
```

### DIFF
--- a/axelrod/tests/unit/test_match.py
+++ b/axelrod/tests/unit/test_match.py
@@ -89,12 +89,8 @@ class TestMatch(unittest.TestCase):
             self.assertEqual(len(match.play()), expected_length)
             self.assertEqual(match.noise, 0)
             self.assertEqual(match.game.RPST(), (3, 1, 0, 5))
-        self.assertEqual(len(match._cache), 3)
-
-        for expected_length in expected_lengths:
-            self.assertEqual(
-                match._cache[(p1, p2, expected_length)], [(C, C)] * expected_length
-            )
+        self.assertEqual(len(match._cache), 1)
+        self.assertEqual(match._cache[(p1, p2)], [(C, C)] * 5)
 
     @given(turns=integers(min_value=1, max_value=200), game=games())
     @example(turns=5, game=axelrod.DefaultGame)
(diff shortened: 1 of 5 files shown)


axelrod/tests/unit/test_match.py
```diff
@@ -167,14 +163,29 @@ def test_play(self):
...
         # a deliberately incorrect result so we can tell it came from the cache
-        expected_result = [(C, C), (D, D), (D, C)]
-        cache[(axelrod.Cooperator(), axelrod.Defector(), 3)] = expected_result
+        expected_result = [(C, C), (D, D), (D, C), (C, C), (C, D)]
+        cache[(axelrod.Cooperator(), axelrod.Defector())] = expected_result
         match = axelrod.Match(players, 3, deterministic_cache=cache)
-        self.assertEqual(match.play(), expected_result)
+        self.assertEqual(match.play(), expected_result[:3])
+
+    def test_long_cache(self):
```
marcharper: Can you add a short doc string explaining the test?

gaffney2010: Done.

gaffney2010:
> Looks really nice to me. Can you add a test that ensures that when a longer run happens that the cache is overwritten with the longer run, similar to the newly added test?

Added.

drvinceknight:
Nice! Thanks @gaffney2010 :+1:

PR #1231, opened 2018-12-31
Added some of the strategies listed in Ashlock2009.

gaffney2010:
Added after reading Ashlock2009, and to address the bug:
https://github.com/Axelrod-Python/Axelrod/issues/379


gaffney2010:
In addition to these, there's a strategy called TFT-PSY, which is described as:
"This strategy plays like TFT until its opponent defects; then it plays like PSYCHO [AnitTitForTat] until its opponent cooperates."

That sounds to me like it will always cooperate, and I couldn't find a detailed description in any of the Ashlock papers.  Similarly PSY-TFT seems to be a defector.  But I don't think it makes sense to add these as names to those strategies.

axelrod/tests/strategies/test_finite_state_machines.py
```diff
@@ -512,6 +512,72 @@ def test_strategy(self):
...
+    transitions = (
+            (1, C, 2, D),
+            (1, D, 1, D),
+            (2, C, 1, D),
+            (2, D, 1, C)
+        )
+    """
+
+    def test_strategy(self):
+        # Never leaves state 1 if C
```
marcharper: if C --> if D

axelrod/tests/strategies/test_finite_state_machines.py
```diff
@@ -512,6 +512,72 @@ def test_strategy(self):
...
+        state_and_actions = [(1, C)] * 10
+        self.transitions_test(state_and_actions)
+        # Visits state 2, but then comes back
```
marcharper: Can you indicate in the comments that this is testing the "except for a D following a C" part? (and below)

axelrod/tests/strategies/test_finite_state_machines.py
gaffney2010: Done.

axelrod/tests/strategies/test_finite_state_machines.py
gaffney2010: I added a comment.

drvinceknight:
:+1: Thanks @gaffney2010 :)

PR #1232, opened 2019-01-03
Clean up docstrings, to match parameter-listing to actual parameters.

gaffney2010:
As I was reading, I saw some docstrings that didn't match the parameter, or in wrong order.  So I fixed most of them to match.

### DIFF
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -92,6 +92,8 @@ class MoranProcess(object):
         noise:
             The background noise, if any. Randomly flips plays with probability
             `noise`.
+        game:
+            A game.  Used to score matches.
         deterministic_cache:
             A optional prebuilt deterministic cache
         mutation_rate:
--- a/axelrod/tournament.py
+++ b/axelrod/tournament.py
@@ -295,7 +295,8 @@ class Tournament(object):
 
         Parameters
         ----------
-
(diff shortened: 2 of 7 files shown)


axelrod/moran.py
```diff
@@ -92,6 +92,8 @@ def __init__(
         noise:
             The background noise, if any. Randomly flips plays with probability
             `noise`.
+        game:
```
meatballs: could we make this
`game: axelrod.Game`
`    The game object used to score the match`
to match other docstrings

axelrod/tournament.py
```diff
@@ -295,7 +295,8 @@ def _run_parallel(self, processes: int = 2, build_results: bool = True) -> bool:
 
         Parameters
         ----------
-
+        build_results : bool
+            whether or not to build a results st
```
meatballs: typo: results set

(repeated further below)

axelrod/moran.py
gaffney2010: done

axelrod/tournament.py
gaffney2010: done

meatballs:
Thank you @gaffney2010 Nice bit of tidying up!

PR #1233, opened 2019-01-06
Algorithm for Memory Depth in FSM

gaffney2010:
As discussed here:
https://github.com/Axelrod-Python/Axelrod/issues/597


axelrod/strategies/finite_state_machines.py
```diff
@@ -1,9 +1,347 @@
...
+        state = trans[0]
+        last_opponent_action = trans[1]
+        next_state = trans[2]
+        next_action = trans[3]
+
+        back_transitions[next_state].append(BackTrans(state,
+                                                      next_action,
+                                                      last_opponent_action))
+
+    class ActionChain(object):
```
marcharper: Maybe move these classes out of the function, it doesn't seem like they use any local data in their definitions.

marcharper:
The travis error is as follows (docstring format):

```
Warning, treated as error:
/home/travis/build/Axelrod-Python/Axelrod/axelrod/strategies/finite_state_machines.py:docstring of axelrod.strategies.finite_state_machines.get_memory_from_transitions:36:Unexpected indentation.
make: *** [html] Error 2
The command "cd docs; make clean; make html" exited with 2.
```

The tests are also outputing all the branches as the algorithm proceeds, I'd make that optional (with a verbose kwarg).

marcharper:
Can you apply the algorithm to the existing FSM players in the library? The current memory depth values were best guesses at the time the strategies were added.

marcharper:
We could also consider adding a method to `FSMPlayer`'s `__init__` to compute the memory_depth if it is unspecified. This will come in handy for e.g. the dojo so we can evolve players of depth <= N.

axelrod/strategies/finite_state_machines.py
gaffney2010: You're right that they could be moved outside of the function without problem.  My reasoning for leaving them in the function is because they're not used anywhere else, and my feeling is that it's similar to a local variable, which could be made global, but encapsulation is better.  I know this is a contentious point though.

gaffney2010:
> Can you apply the algorithm to the existing FSM players in the library?

I'll do that.

> in this case something like TestFSMPlayer where it applies your algorithm to verify, at test time, that the player is of the expected depth.

Currently the runtime for large FSMs is too long to be tolerable.  [Like many times longer than the rest of the tests put together.]  I'll see if there are any easy speed-ups.

drvinceknight:
> Currently the runtime for large FSMs is too long to be tolerable. [Like many times longer than the rest of the tests put together.] I'll see if there are any easy speed-ups.

A suggestion: one option could be to have a separate repo where we run the algorithm on the fsms perhaps?

gaffney2010:
These 16-state strategies take a very long time.  I guess the number of branches is growing exponentially.  I'm going have to get this thing to run faster.  I think there's a memoization here that makes sense.

axelrod/strategies/finite_state_machines.py
meatballs: I'd prefer to see these classes outside the function and with unit tests for their behaviour.

gaffney2010:
I may end up changing stuff for the caching logic.  Closing for now.

gaffney2010:
In thinking about how to memoize, I ended up changing significantly.  It brought the runtime down from days to milliseconds.

The new approach is a little abstract, and it may be harder to see that it works, but it does match the other implementation on every FSM strategy, and it passes all of the tests.

gaffney2010:
> in this case something like TestFSMPlayer where it applies your algorithm to verify, at test time, that the player is of the expected depth.

> We could also consider adding a method to FSMPlayer's __init__ to compute the memory_depth if it is unspecified. This will come in handy for e.g. the dojo so we can evolve players of depth <= N.

I can probably do these now that it runs fast.

marcharper:
The error is on the typing syntax, which isn't supported on Python 3.5. However we generally only support the last two trailing versions of Python, so it's time we updated the library to 3.6 and 3.7, which would eliminate the error.

@drvinceknight ok with you to drop 3.5 and add 3.7?

docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -69,7 +69,7 @@ range of memory_depth values, we can use the 'min_memory_depth' and
     ... }
     >>> strategies = axl.filtered_strategies(filterset)
     >>> len(strategies)
-    61
```
marcharper: I presume this means that we misclassified a number of the FSMs as finite memory?

gaffney2010: That's right.  Almost all FSMs were marked with memory length equal to the number of states, but actually most of them turned out to be infinite memory, since all that's needed is for the same cycle to show up in two different places.

marcharper:
> > in this case something like TestFSMPlayer where it applies your algorithm to verify, at test time, that the player is of the expected depth.
> > We could also consider adding a method to FSMPlayer's init to compute the memory_depth if it is unspecified. This will come in handy for e.g. the dojo so we can evolve players of depth <= N.

> I can probably do these now that it runs fast.

We can do this in a follow-up PR

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,220 @@
...
+(D, 0, D),
+(C, 1, C),
+(C, 1, D)
+"""
+Memit = namedtuple("Memit", ["in_act", "state", "out_act"])
+
+def memits_match(x, y):
+    """In action and out actions are the same."""
+    return x.in_act == y.in_act and x.out_act == y.out_act
+def memit_sort(x, y):
```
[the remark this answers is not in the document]
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,220 @@
...
+        return (x, y)
+    else:
+        return (y, x)
+
+
+def get_accessible_transitions(transitions: dict, initial_state: int) -> dict:
+  """Gets all transitions from the list that can be reached from the
+  initial_state.
+  """
+  edge_dict: DefaultDict[int, list] = defaultdict(list)
```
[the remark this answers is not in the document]
gaffney2010: ack

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,220 @@
...
+  visited = dict()
+  for k, v in transitions.items():
```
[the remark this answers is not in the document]
gaffney2010: I changed k, v on the edge dictionaries.

I changed my "for k, v in transitions.items()" loops (this highlighted loop and others) to "for trans in transition_iterator(transitions)" as I had done in other spots.  This cleans it up.

However, I think it make sense to leave the k, v in the implementation of transition_iterator, because these don't have perfect meanings, and I'm kinda just moving around.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,220 @@
...
+                ending_node = Memit(in_action, trans.state,
+                                    trans.last_opponent_action)
+                memit_edges[starting_node].add(ending_node)
+
+    all_memits = memit_edges.keys()
+
+    pair_nodes = set()
+    pair_edges: DefaultDict[tuple, set] = defaultdict(set)
+    # Loop through all pairs of memits.
+    for x, y in [(x, y) for x in all_memits for y in all_memits]:
```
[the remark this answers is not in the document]
gaffney2010: Ack.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,220 @@
...
+  initial_state.
+  """
+  edge_dict: DefaultDict[int, list] = defaultdict(list)
+  visited = dict()
+  for k, v in transitions.items():
+      visited[k[0]] = False
+      edge_dict[k[0]].append(v[0])
+  accessible_edges = [initial_state]
+
+  edge_queue = [initial_state]
```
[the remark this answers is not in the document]
gaffney2010: Done.

axelrod/tests/strategies/test_finite_state_machines.py
```diff
@@ -2,13 +2,276 @@
...
+        transitions = (
+            (0, C, 0, C),
+            (0, D, 0, C)
+        )
+
+        trans_dict = {(current_state, input_action): (next_state, output_action)
+                      for current_state, input_action, next_state, output_action in transitions}
+        self.assertEqual(get_memory_from_transitions(trans_dict), 0)
+
+    def test_tit_for_tat(self):
```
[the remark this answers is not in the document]
gaffney2010: Done.

gaffney2010:
> We can do this in a follow-up PR

Sounds good to me.

marcharper:
Here's the build error:

```
$ python run_mypy.py
axelrod/strategies/finite_state_machines.py:67: error: Name 'List' is not defined
axelrod/strategies/finite_state_machines.py:182: error: Name 'List' is not defined
```

gaffney2010:
Fix that error; had forgotten an import.

axelrod/strategies/finite_state_machines.py
marcharper: +gaffney2010 can we change this line to the 3.5 compatible style of type hint comment? I think you just do it like so:

```
edge_dict = defaultdict(list)  # type: DefaultDict[int, list]
```

gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
 from axelrod.action import Action
 from axelrod.player import Player
+from typing import DefaultDict, Iterator, Dict, Tuple, Set, List
+from collections import defaultdict, namedtuple
```
marcharper: nit: collections comes before typing (alphabetical)

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
 
 C, D = Action.C, Action.D
+ALL_ACTIONS = [C, D]
```
marcharper: Let's make this `(C, D)` (immutable)

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
+
+"""
+Memit = unit of memory.
+
+This represents the amount of memory that we gain with each new piece of
+history.  It includes a state, our_response that we make on our way into that
+state (in_act), and the opponent's action that makes us move out of that state
+(out_act).
+
+For example for this FSM:
```
marcharper: For example, this finite state machine

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
+(1, D, 0, D)
+
+Has the memits:
+(C, 0, C),
+(C, 0, D),
+(D, 0, C),
+(D, 0, D),
+(C, 1, C),
+(C, 1, D)
+"""
```
marcharper: newline

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
+    else:
+        return (y, x)
+
+
+Transition = namedtuple("Transition", ["state", "last_opponent_action",
+                                       "next_state", "next_action"])
+TransitionDict = Dict[Tuple[int, Action], Tuple[int, Action]]
+
+def transition_iterator(transitions: TransitionDict) -> Iterator[Transition]:
+    """Changes the transition dictionary into a iterator on namedtuples, because
```
marcharper: `"""Changes the transition dictionary into a iterator on namedtuples."""`

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
+    for trans in transition_iterator(transitions):
+        visited[trans.state] = False
+        edge_dict[trans.state].append(trans.next_state)
+    # Keep track of states that can be accessed.
+    accessible_states = [initial_state]
+
+    state_queue = [initial_state]
+    visited[initial_state] = True
+    # While there are states in the queue, visit all its children, adding each
+    # to the accesible_states.  [A basic BFS.]
```
marcharper: BFS -> Breadth-First Search  (we try to be newbie friendly whenever possible)

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
+    Assume that transitions are a dict with entries like
+    (state, last_opponent_action): (next_state, next_action)
+
+    We first break down the transitions into memits (see above).  We also create
+    a graph of memits, where the successor to a given memit are all possible
+    memits that could occur in the memory immediately before the given memit.
+
+    Then we pair up memits with different states, but same in and out actions.
+    These represent points in time that we can't determine which state we're in.
+    We also create a graph of memit-pairs, where memit-pair, Y, succedes a
```
marcharper: succeeds

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
+    memit-pair, X, if the two memits in X are succeded by the two memits in Y.
```
marcharper: succeeded

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
+    for trans in transition_iterator(transitions):
+        incoming_action_by_state[trans.next_state].add(trans.next_action)
+
+    # Keys are starting memit, and values are all possible terminal memit.
+    # Will walk backwards through the graph.
+    memit_edges = defaultdict(set)  # type: DefaultDict[Memit, Set[Memit]]
+    for trans in transition_iterator(transitions):
+        # Since all actions are out-paths for each state, add all of these.
+        # That is to say that your opponent could do anything
+        for out_action in ALL_ACTIONS:
```
marcharper: Since this is then only place where `ALL_ACTIONS` is used I think it's better as a default kwarg to the function rather than a global.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,7 +1,238 @@
...
+    incoming_action_by_state = defaultdict(set)  # type: DefaultDict[int, Set[Action]]
+    for trans in transition_iterator(transitions):
+        incoming_action_by_state[trans.next_state].add(trans.next_action)
+
+    # Keys are starting memit, and values are all possible terminal memit.
+    # Will walk backwards through the graph.
+    memit_edges = defaultdict(set)  # type: DefaultDict[Memit, Set[Memit]]
+    for trans in transition_iterator(transitions):
+        # Since all actions are out-paths for each state, add all of these.
+        # That is to say that your opponent could do anything
```
marcharper: your --> the

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
gaffney2010: Done.

drvinceknight:
> Just some minor suggestions from me, approving and tagging @drvinceknight for a second review.

Taking a look now :+1: :)

axelrod/tests/strategies/test_finite_state_machines.py
```diff
@@ -2,13 +2,332 @@
...
+            (0, D, 2, D),
+            (1, C, 3, D),
+            (1, D, 0, C),
+            (2, C, 2, D),
+            (2, D, 1, C),
+            (3, C, 3, D),
+            (3, D, 1, D),
+        )
+
+        trans_dict = {(current_state, input_action): (next_state, output_action)
```
drvinceknight: Is it worth defining `def transitions_to_dict(transitions)`?

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,9 +1,239 @@
...
+(C, 1, D)
+"""
+
+Memit = namedtuple("Memit", ["in_act", "state", "out_act"])
+
+def memits_match(x: Memit, y: Memit):
+    """In action and out actions are the same."""
+    return x.in_act == y.in_act and x.out_act == y.out_act
+
+def memit_sort(x: Memit, y: Memit):
```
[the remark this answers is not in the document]
gaffney2010: There's no reason.  I made the two memit functions a while after I made the namedtuple.  I think it makes sense to implement as a class instead, so I'll change it.

gaffney2010:
> it took me a little while to get my head around UsuallyCooperates having infinite memory which had me doubting my own sanity for a bit

You were right to be suspicious.  It looks like I implemented UsuallyCooperates wrong, because a C following a D starting when the player is in state 2 won't yield a defect as promised.  I'll fix this in a different PR.

axelrod/tests/strategies/test_finite_state_machines.py
gaffney2010: Sure.  Done.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,9 +1,240 @@
...
+            starting_node = Memit(trans.next_action, trans.next_state,
+                                  out_action)
+            # All incoming paths to current state
+            for in_action in incoming_action_by_state[trans.state]:
+                # Less recent in action history
+                ending_node = Memit(in_action, trans.state,
+                                    trans.last_opponent_action)
+                memit_edges[starting_node].add(ending_node)
+
+    all_memits = [x for x in memit_edges.keys()]
```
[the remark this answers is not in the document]
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,9 +1,240 @@
...
+    for trans in transition_iterator(transitions):
+        if trans.state in accessible_states:
+            accessible_transitions[(
+                trans.state, trans.last_opponent_action)] = (trans.next_state,
+                                                             trans.next_action)
+
+    return accessible_transitions
+
+
+MemitPair = Tuple[Memit, Memit]
```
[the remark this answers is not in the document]
gaffney2010: Done.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1,9 +1,239 @@
...
+    """Returns a tuple of x in y, sorted so that (x, y) are viewed as the
+    same as (y, x).
+    """
+    if repr(x) <= repr(y):
+        return (x, y)
+    else:
+        return (y, x)
+
+
+Transition = namedtuple("Transition", ["state", "last_opponent_action",
```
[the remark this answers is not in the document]
gaffney2010: Done.

drvinceknight:
For a nice overview of FSMs (with the memory length use case)  perhaps this section of the docs would be more appropriate: https://axelrod.readthedocs.io/en/stable/tutorials/further_topics/index.html

?

gaffney2010:
I decided to leave it on the further topics page, and focus on the way you execute the code.  I think the research topics page would be better if I focused on open questions or some of the research around this, but this is sorta how dojo documentation approaches it already.

I put this under a broader "meta-strategies" heading, because I thought we could eventually add write-up for things like HMMPlayer, SequencePlayer, LookerUp, and Gambler.  I wrote the section that it would be general enough to apply to games other than IPD, in anticipation of 5.0.

docs/tutorials/further_topics/index.rst
```diff
@@ -17,3 +17,4 @@ Contents:
    morality_metrics.rst
    ecological_variant.rst
    fingerprinting.rst
+   meta_strategies.rst
```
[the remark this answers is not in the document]
gaffney2010: I think I'm confusing myself; I think the paths don't match with titles of the pages.

I'll leave it here.

axelrod/compute_finite_state_machine_memory.py
```diff
@@ -0,0 +1,265 @@
...
+        )
+
+    def __lt__(self, other_memit) -> bool:
+        return repr(self) < repr(other_memit)
+
+
+MemitPair = Tuple[Memit, Memit]
+
+
+def OrderedMemitTuple(x: Memit, y: Memit) -> tuple:
```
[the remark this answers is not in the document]
drvinceknight: :+1:

marcharper:
Here's the build error (it's in the docs):

```
Warning, treated as error:
/home/travis/build/Axelrod-Python/Axelrod/docs/tutorials/further_topics/meta_strategies.rst:20:citation not found: Harper2017
make: *** [html] Error 2
The command "cd docs; make clean; make html" exited with 2.
```

drvinceknight:
It's the bib item missing: https://github.com/Axelrod-Python/Axelrod/pull/1233#discussion_r250517679 :+1:

gaffney2010:
Can't figure out why I can't commit that file.  Ended up editing it on Github's GUI.

gaffney2010:
I was accidentally making changes to the wrong branch.  Got it cleaned up now.

drvinceknight:
I'll aim to put a new release together tomorrow :+1:

langner:
Hey guys, I just had the pleasure of reading through this thread and wanted to send a shout out. It's a very cool change. And it's encouraging to see the project being so vibrant with so much collaboration :) kudos!

Issue #1109, opened 2017-08-04
Specific strategy: Graaskamp


drvinceknight:
This strategy is one of the strategies from Axelrod's first tournament. It is described here http://axelrod.readthedocs.io/en/stable/reference/overview_of_strategies.html#graaskamp as:

1. Play Tit For Tat for the first 50 rounds;
2. Defects on round 51;
3. Plays 5 further rounds of Tit For Tat;
4. A check is then made to see if the opponent is playing randomly in which case it defects for the rest of the game;
5. The strategy also checks to see if the opponent is playing Tit For Tat or another strategy from a preliminary tournament called ‘Analogy’. If so it plays Tit For Tat. If not it cooperates and randomly defects every 5 to 15 moves.

Step 4 can be implemented using a chi squared test similarly to the Stein and Rapaport strategy which is implemented here: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/axelrod_first.py#L496

Step 5 is a bit tricky as it refers to a strategy called "Analogy". I have not been able to find any information describing this strategy. I suggest that anyone who tries to work on this does their best to find that information but if there is non available let us simply ignore that part of the check but clearly point that out in the documentation of the strategy.

(Other strategies that we'd like to implement are available here: #379)

Note that this issue was originally at #1051 however that now corresponds to an implementation of a different strategy written by Graaskamp (for Axelrod's second tournament).

souravsingh:
@drvinceknight I am interested in working on the issue.

drvinceknight:
That would be great @souravsingh :+1:

souravsingh:
@drvinceknight I have a question- Is this strategy related to the PR at #1115

drvinceknight:
> @drvinceknight I have a question- Is this strategy related to the PR at #1115

No.

That's related to:

> Note that this issue was originally at #1051 however that now corresponds to an implementation of a different strategy written by Graaskamp (for Axelrod's second tournament).

souravsingh:
The description of the strategy looks very similar to what is written here- https://github.com/Axelrod-Python/Axelrod/issues/1051

drvinceknight:
#1051 started out as this strategy but has since changed to be about a completely different strategy. That is why this issue has been opened.

> Note that this issue was originally at #1051 however that now corresponds to an implementation of a different strategy written by Graaskamp (for Axelrod's second tournament).

drvinceknight:
[...]

PR #1234, opened 2019-01-10
New strategy: Graaskamp

kjurgielajtis:
Hi, I tried my best to contribute to issue #1109 :)


axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
+import random
```
drvinceknight: Let's put this in the `axelrod_first.py` module: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/axelrod_first.py

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+from axelrod.action import Action
+from axelrod.player import Player
+from scipy.stats import chisquare
+
+C, D = Action.C, Action.D
+
+
+class Graaskamp(Player):
+    """
```
drvinceknight: Could you take a look at this documentation about the documentation: https://axelrod.readthedocs.io/en/stable/tutorials/contributing/strategy/docstrings.html

We're missing a reference (see the other strategies in https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/axelrod_first.py

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+    Play Tit For Tat for the first 50 rounds;
+	Defects on round 51;
+	Plays 5 further rounds of Tit For Tat;
+	A check is then made to see if the opponent is playing randomly
+	in which case it defects for the rest of the game;
+	The strategy also checks to see if the opponent is playing Tit For Tat.
```
drvinceknight: This is a bit incomplete from https://github.com/Axelrod-Python/Axelrod/issues/1109 see:

> Step 5 is a bit tricky as it refers to a strategy called "Analogy". I have not been able to find any information describing this strategy. I suggest that anyone who tries to work on this does their best to find that information but if there is non available let us simply ignore that part of the check but clearly point that out in the documentation of the strategy.

Have you found any information about `analogy`? If not we should very clearly point this out in the documentation.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+	If so it plays Tit For Tat. If not it cooperates and randomly defects every 5 to 15 moves.
+    """
+
+    name = "Graaskamp"
+    classifier = {
+        "memory_depth": 57,  # long_memory
```
drvinceknight: No need for the `# long_memory`

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+        "stochastic": False,
```
drvinceknight: I think the strategy is described as being stochastic:

> The strategy also checks to see if the opponent is playing Tit For Tat or another strategy from a preliminary tournament called ‘Analogy’. If so it plays Tit For Tat. If not it cooperates and randomly defects every 5 to 15 moves.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+        "makes_use_of": set(),
+        "long_run_time": True,
```
drvinceknight: Are you sure that this has a long run time? I suspect it doesn't.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+        self.opponent_is_random = False
+        self.opponent_is_TfT = False
+
+    def strategy(self, opponent: Player) -> Action:
+        """This is the actual strategy"""
+        # First move
+        # Cooperate in first round as in Tit for Tat strategy
+        if not self.history:
+            return C
+        # Copy opponent's last move as in Tit for Tat strategy
```
drvinceknight: I don't fully understand the comment here, can you clarify it please?

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+        if len(self.history) <= 50 or 52 <= len(self.history) <= 56 or len(self.history) >= 57:
```
drvinceknight: The logic here doesn't quite look right to me, should this be:

```
52 <= len(self.history) <= 56 or len(self.history) >= 57
```

is equivalent to:

```
52 <= len(self.history)
```

I think this should just be:

```
52 <= len(self.history) <= 56
```

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+            if opponent.history[-1] == D:
+                return D
+            return C
+
+        # Defect on round 51
+        if len(self.history) == 51:
+            return D
+
+        if len(self.history) >= 57:
```
drvinceknight: This won't kick in because of the `52 <= len(self.history) <= 56 or len(self.history) >= 57` check previously.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+            self.opponent_is_random = p_value >= self.alpha
+
+            if self.opponent_is_random:
+                return D
+            elif all(opponent.history[-i] == self.history[-i - 1] for i in range(1, 57)):
+                # Check if opponent plays Tit for Tat
+                    if opponent.history[-1] == D:
+                        return D
+                    return C
+            else:
```
drvinceknight: No need for the `else` block.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+                # Generating integer from 0 to 14 for defecting every 5 to 15 moves
+                numlist = []
```
drvinceknight: I would interpret: 

>  If not it cooperates and randomly defects every 5 to 15 moves.

to be something like:

```
if len(self.history) == self.next_random_defection_turn:
    self.next_defection = random.randint(5, 15) + len(self.history) # resample the next defection turn
    return D
return C
```

(We'd need to set `self.next_random_defection_turn` in the `init`

axelrod/tests/strategies/test_graaskamp.py
```diff
@@ -0,0 +1,35 @@
+import axelrod
+from .test_player import TestPlayer
+
+C, D = axelrod.Action.C, axelrod.Action.D
+
+
+class TestGraaskamp(TestPlayer):
+    """
```
drvinceknight: No need for the empty docstring here.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+            return D
+
+        if len(self.history) >= 57:
+            # Check if opponent plays randomly, if so, defect
+            p_value = chisquare([opponent.cooperations, opponent.defections]).pvalue
+            self.opponent_is_random = p_value >= self.alpha
+
+            if self.opponent_is_random:
+                return D
+            elif all(opponent.history[-i] == self.history[-i - 1] for i in range(1, 57)):
```
drvinceknight: this can just be `if`

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+            return C
+
+        # Defect on round 51
+        if len(self.history) == 51:
+            return D
+
+        if len(self.history) >= 57:
+            # Check if opponent plays randomly, if so, defect
+            p_value = chisquare([opponent.cooperations, opponent.defections]).pvalue
+            self.opponent_is_random = p_value >= self.alpha
```
drvinceknight: This should be:

```
self.opponent_is_random = (p_value >= self.alpha) or self.opponent_is_random
```

because:

> A check is then made to see if the opponent is playing randomly in which case it defects for the rest of the game;

implies that as soon as an opponent is detected as being random then the player will defect "forever". So if an opponent all of a sudden started playing "non randomly" the player would not change back.

axelrod/strategies/graaskamp.py
drvinceknight: We should also check the entire history and not just for `range(1, 57)`, I interpret the description as saying that once past 57 we continously check if it's playing TfT.

axelrod/strategies/graaskamp.py
```diff
@@ -0,0 +1,86 @@
...
+        Parameters
+        ----------
+        alpha: float
+            The significant level of p-value from chi-squared test with
+            alpha == 0.05 by default.
+        """
+        super().__init__()
+        self.alpha = alpha
+        self.opponent_is_random = False
+        self.opponent_is_TfT = False
```
drvinceknight: I don't believe you use `opponent_is_TfT` (and I don't think you need it).

kjurgielajtis:
Thank You for reply :) Now I see that I've made many mistakes, will correct them as soon as possible.

axelrod/strategies/graaskamp.py
kjurgielajtis: Yes, totally, didn't see it before, sorry

axelrod/strategies/axelrod_first.py
```diff
@@ -206,6 +206,91 @@ def strategy(self, opponent: Player) -> Action:
         return random_choice(p)
 
 
+class Graaskamp(Player):
+    """
+    This strategy follows the following rules:
```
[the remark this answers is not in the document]
kjurgielajtis: Should I replace "This strategy follows the following rules:" with this sentence?

axelrod/strategies/axelrod_first.py
```diff
@@ -206,6 +206,91 @@ def strategy(self, opponent: Player) -> Action:
...
+        self.opponent_is_random = False
+
+    def strategy(self, opponent: Player) -> Action:
+        """This is the actual strategy"""
+        # First move
+        # Cooperate in first round as in Tit for Tat strategy
+        if not self.history:
+            return C
+        # Copy opponent's last move as in Tit for Tat strategy
+        if 52 <= len(self.history) <= 56:
```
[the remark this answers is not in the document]
kjurgielajtis: That looks a lot better, totally agree :)

axelrod/strategies/axelrod_first.py
```diff
@@ -206,6 +206,91 @@ def strategy(self, opponent: Player) -> Action:
...
+            return D
+
+        if len(self.history) >= 57:
+            # Check if opponent plays randomly, if so, defect
+            p_value = chisquare([opponent.cooperations, opponent.defections]).pvalue
+            self.opponent_is_random = (p_value >= self.alpha) or self.opponent_is_random
+
+            if self.opponent_is_random:
+                return D
+            elif all(opponent.history[-i] == self.history[-i - 1] for i in range(1, 57)):
```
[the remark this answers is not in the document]
kjurgielajtis: Right but I'm not sure if it would work with memory_depth = 1, maybe I don't get it

axelrod/strategies/axelrod_first.py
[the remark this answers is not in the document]
drvinceknight: Yes please.

axelrod/strategies/axelrod_first.py
```diff
@@ -206,6 +206,83 @@ def strategy(self, opponent: Player) -> Action:
...
+        self.next_random_defection_turn = None
+
+    def strategy(self, opponent: Player) -> Action:
+        """This is the actual strategy"""
+        # First move
+        # Cooperate in first round as in Tit for Tat strategy
+        if not self.history:
+            return C
+        # Copy opponent's last move as in Tit for Tat strategy
+        if 52 <= len(self.history) <= 56 or len(self.history) == 51:
```
drvinceknight: As is this implies that the strategy will cooperate during the first 50 turns but it should not, it should play tit for tat, it should be implemented as:

```
if len(self.history) <= 56:
    if opponent.history[-1] == D or len(self.history) == 51:
        return D
    return C
```

drvinceknight: > Play Tit For Tat for the first 50 rounds;

axelrod/strategies/axelrod_first.py
```diff
@@ -206,6 +206,83 @@ def strategy(self, opponent: Player) -> Action:
...
+            return C
+
+        if len(self.history) >= 57:
+            # Check if opponent plays randomly, if so, defect
+            p_value = chisquare([opponent.cooperations, opponent.defections]).pvalue
+            self.opponent_is_random = (p_value >= self.alpha) or self.opponent_is_random
+
+            if self.opponent_is_random:
+                return D
+            elif all(opponent.history[i] == self.history[i - 1] for i in range(1, len(self.history + 1))):
```
drvinceknight: My mistake, should be:

```
range(1, len(self.history))
```

axelrod/strategies/axelrod_first.py
```diff
@@ -206,6 +206,83 @@ def strategy(self, opponent: Player) -> Action:
...
+    This strategy came 9th in Axelrod’s original tournament.
+
+    Names:
+
+    - Graaskamp: [Axelrod1980]_
+    """
+
+    name = "Graaskamp"
+    classifier = {
+        "memory_depth": 1,
```
drvinceknight: This should be `float("inf")` (as it constantly checks if the opponent is TfT)

axelrod/tests/strategies/test_axelrod_first.py
```diff
@@ -138,6 +138,31 @@ def test_strategy(self):
         self.versus_test(axelrod.Defector(), expected_actions=actions)
 
 
+class TestGraaskamp(TestPlayer):
+
+    name = "Graaskamp"
+    player = axelrod.Graaskamp
+    expected_classifier = {
+        'memory_depth': 57,
```
drvinceknight: `float("inf")`

axelrod/tests/strategies/test_axelrod_first.py
```diff
@@ -138,6 +138,31 @@ def test_strategy(self):
         self.versus_test(axelrod.Defector(), expected_actions=actions)
 
 
+class TestGraaskamp(TestPlayer):
+
+    name = "Graaskamp"
```
drvinceknight: `Graaskamp: 0.05` (the parameters get automatically added).

axelrod/strategies/axelrod_first.py
```diff
@@ -206,6 +206,83 @@ def strategy(self, opponent: Player) -> Action:
...
+            elif all(opponent.history[i] == self.history[i - 1] for i in range(1, len(self.history + 1))):
+                # Check if opponent plays Tit for Tat
+                    if opponent.history[-1] == D:
+                        return D
+                    return C
+
+            if len(self.history) == self.next_random_defection_turn:
+                self.next_random_defection_turn = random.randint(5, 15) + len(self.history)  # resample the next defection turn
+                return D
+            return C
```
drvinceknight: This last `return C` needs to be unindented: it will catch all behaviours that do not match the above statements.

axelrod/tests/strategies/test_axelrod_first.py
```diff
@@ -138,6 +138,31 @@ def test_strategy(self):
...
+        'stochastic': True,
+        'makes_use_of': set(),
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def test_strategy(self):
+        # Play against opponents
+        actions = [(C, C), (C, D), (D, C), (C, D), (D, C)]
```
drvinceknight: These tests are a good start, once you've made the changes I've requested I think we'll need to change one or two and possibly include a few more to ensure we're checking all the described behaviours :) :+1:

axelrod/strategies/axelrod_first.py
kjurgielajtis: Right, I've implemented it wrongly, already fixed :+1:

axelrod/tests/strategies/test_axelrod_first.py
kjurgielajtis: I'm not sure what should I do here. Should I add parameters manually somehow for testing? Thanks :)

kjurgielajtis:
That's almost everything :) I'm not quite sure how to implement tests for opponents playing randomly because of almost 60 rounds of playing randomly and then taking action. How should it look like?

axelrod/tests/strategies/test_axelrod_first.py
kjurgielajtis: I've added some more tests :) I think I've found a bug. After round 57 when playing with Grudger, Graaskamp only cooperates. I'm trying to fix this.

axelrod/tests/strategies/test_axelrod_first.py
marcharper: Yes, you can change the test name to "Graaskamp: 0.05"

drvinceknight:
@kjurgielajtis I've opened https://github.com/kjurgielajtis/Axelrod/pull/1 which includes a number of little tweaks. I wanted to be sure that any suggestions I made no longer included off by one errors etc so I checked them locally. You can either merge that PR (and the changes will automatically update) or just take a look and include the changes manually.

You'll notice there was an off by 1 error in the initial TfT play and that I've also written a number of tests and included a couple of comments to show how they match each case of the description. (I believe that catches them all but I might have missed something that the next reviewer might notice.)

drvinceknight:
Closed by #1244

PR #1235, opened 2019-01-12
Bump python versions up to 3.6 and 3.7

marcharper:


.travis.yml
```diff
@@ -1,8 +1,9 @@
 
+dist: xenial
 language: python
 python:
-  - 3.5
```
drvinceknight: Unless there's anything backwards incompatible with 3.5 that we'd be looking to include I'd prefer to keep it (and of course add 3.7).

(For example, on a purely selfish note, for my teaching I'd need to get anaconda reinstalled on all our lab machines if we dropped 3.5.)

setup.py
```diff
@@ -23,15 +23,15 @@
     packages=["axelrod", "axelrod.strategies", "axelrod.data"],
     url="http://axelrod.readthedocs.org/",
     license="The MIT License (MIT)",
-    description="Reproduce the Axelrod iterated prisoners dilemma tournament",
+    description="Axelrod game theory library for the iterated prisoners dilemma",
```
drvinceknight: :+1:

README.rst
```diff
@@ -74,7 +74,8 @@ a peer reviewed paper introducing the library (22 authors).
 Installation
 ------------
 
-The library requires Python 3.5 or greater.
+The library requires Python 3.6 or greater. We typically support the two latest
```
drvinceknight: I know we said that we'd support the two latest in the past but perhaps we keep previous versions until there's a particular feature we want to take advantage of? (Perhaps there already is and I'm forgetting.)

drvinceknight: I've just seen https://github.com/Axelrod-Python/Axelrod/pull/1233#issuecomment-453702661 which seems like a good reason to me. Would it be ok to just bump here to 3.6 and not include the `We typically support the two latest...`?

.travis.yml
drvinceknight: Ignore this I've just seen https://github.com/Axelrod-Python/Axelrod/pull/1233#issuecomment-453702661

marcharper: I'm fine with keeping 3.5 if you need it. We just have to use the comment form of type hint, which we've already done in some cases.

drvinceknight: If it's easy enough it could be nicer to keep things a bit more backwards compatible but I'm happy with whatever you prefer.

marcharper:
I'm trying 3.5 again. I had to bump up to xenial for Travis to get 3.7, we'll see if it gets 3.5 still.

drvinceknight:
Looks like travis isn't immediately happy, if it's a choice between 3.5 and 3.7 I'd suggest moving forward with 3.7 :)

marcharper:
Well I haven't figured out the necessary voodoo to make this work with both 3.5 and 3.7 at the same time, so I'll close for now until we decide to drop 3.5.

Issue #1214, opened 2018-10-29
Hypothesis deprecation warnings


marcharper:
```
  @settings(max_examples=5, max_iterations=20)
/home/user/repos/axelrod/Axelrod/axelrod/tests/unit/test_tournament.py:1012: HypothesisDeprecationWarning: 
The max_iterations setting has been disabled, as internal heuristics are more
useful for this purpose than a user setting.  It no longer has any effect.
```

and many other lines. Run `./test` to see them.

marcharper:
That's fine, just documenting it here. If you use a conda installation without adhering to our requirements you get these warnings.

PR #1238, opened 2019-02-05
Hypothesis warnings and speed up some tests

### LINKED ISSUE #1209: Tests fail for matplotlib 3.0.0
This has temporarily been solved by #1208, we should at some point update and lower bound mpl in the requirements.

marcharper:
The first commit removes the hypothesis decorator kwarg `max_iterations` that give deprecation warnings (and are reported as no-ops). I changed hypothesis to be pinned at >= 3.2 instead of ==.  [Edit: see below, also limited to <=3.68] See #1214 .

The second commit reduces the run time of some of the longer tests.  These changes cut the local total test run time down by 65% (excluding doctests, which take about 60 seconds to run). For example some tests now avoid the Meta strategies (like the filter set tests). Others now only test a random sample of strategies. (We may need to add some random seeding.) There are still tests that run larger tournaments so I think the risk for this change is low.

Also, I marked one test as skippable if the matplotlib version is 3+, and removed the restriction in requirements.txt. Alternatively we could require matplotlib 3+ and update the test. It's just the colorbar position, so I think we can live without the test for a short while. See #1209 .

Closes #1209

### DIFF
--- a/axelrod/tests/integration/test_filtering.py
+++ b/axelrod/tests/integration/test_filtering.py
@@ -1,12 +1,9 @@
-import random
 import unittest
 
-from axelrod import short_run_time_strategies, filtered_strategies, seed
-
+from axelrod import filtered_strategies, seed
+from axelrod.tests.property import strategy_lists
 from hypothesis import example, given, settings
 from hypothesis.strategies import integers
-
-strategies = random.sample(short_run_time_strategies, 20)
 
 
 class TestFiltersAgainstComprehensions(unittest.TestCase):
@@ -15,7 +12,8 @@ class TestFiltersAgainstComprehensions(unittest.TestCase):
     match the results from using a list comprehension.
     """
 
-    def test_boolean_filtering(self):
+    @given(strategies=strategy_lists(min_size=20, max_size=20))
+    def test_boolean_filtering(self, strategies):
 
         classifiers = [
             "stochastic",
@@ -35,15 +33,18 @@ class TestFiltersAgainstComprehensions(unittest.TestCase):
         min_memory_depth=integers(min_value=1, max_value=10),
         max_memory_depth=integers(min_value=1, max_value=10),
         memory_depth=integers(min_value=1, max_value=10),
+        strategies=strategy_lists(min_size=20, max_size=20),
     )
     @example(
         min_memory_depth=float("inf"),
         max_memory_depth=float("inf"),
         memory_depth=float("inf"),
+        strategies=strategy_lists(min_size=20, max_size=20),
     )
     @settings(max_examples=5)
     def test_memory_depth_filtering(
-        self, min_memory_depth, max_memory_depth, memory_depth
+        self, min_memory_depth, max_memory_depth, memory_depth,
+        strategies
     ):
 
         min_comprehension = set(
@@ -81,9 +82,11 @@ class TestFiltersAgainstComprehensions(unittest.TestCase):
         filtered = set(filtered_strategies(filterset, strategies=strategies))
         self.assertEqual(comprehension, filtered)
 
-    @given(seed_=integers(min_value=0, max_value=4294967295))
+    @given(seed_=integers(min_value=0, max_value=4294967295),
+           strategies=strategy_lists(min_size=20, max_size=20),
+           )
     @settings(max_examples=5)
-    def test_makes_use_of_filtering(self, seed_):
+    def test_makes_use_of_filtering(self, seed_, strategies):
         """
         Test equivalent filtering using two approaches.
 
--- a/axelrod/tests/integration/test_filtering.py
+++ b/axelrod/tests/integration/test_filtering.py
@@ -38,7 +38,7 @@ class TestFiltersAgainstComprehensions(unittest.TestCase):
         max_memory_depth=float("inf"),
         memory_depth=float("inf"),
     )
-    @settings(max_examples=5, max_iterations=20)
+    @settings(max_examples=5)
     def test_memory_depth_filtering(
         self, min_memory_depth, max_memory_depth, memory_depth
     ):
@@ -77,7 +77,7 @@ class TestFiltersAgainstComprehensions(unittest.TestCase):
         self.assertEqual(comprehension, filtered)
 
     @given(seed_=integers(min_value=0, max_value=4294967295))
-    @settings(max_examples=5, max_iterations=20)
+    @settings(max_examples=5)
     def test_makes_use_of_filtering(self, seed_):
         """
         Test equivalent filtering using two approaches.
--- a/axelrod/tests/integration/test_filtering.py
+++ b/axelrod/tests/integration/test_filtering.py
@@ -1,9 +1,12 @@
+import random
 import unittest
 
-from axelrod import all_strategies, filtered_strategies, seed
+from axelrod import short_run_time_strategies, filtered_strategies, seed
 
 from hypothesis import example, given, settings
 from hypothesis.strategies import integers
+
+strategies = random.sample(short_run_time_strategies, 20)
 
 
 class TestFiltersAgainstComprehensions(unittest.TestCase):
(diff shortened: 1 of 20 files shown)


marcharper:
OK so we still need to limit the version of hypothesis somewhat, else it reports slow tests as errors.

axelrod/tests/strategies/test_meta.py
```diff
@@ -59,6 +60,36 @@ def test_repr(self):
...
+
+        turns = 10
+        for op in [
+            axelrod.Cooperator(),
+            axelrod.Defector(),
+            axelrod.TitForTat(),
+        ]:
+            player1.reset()
+            player2.reset()
+            seed = random.randint(0, 10 ** 6)
```
drvinceknight: This could prove tricky to pin down in case of a failure (which is unlikely). Could you perhaps wrap this as a hypothesis test (with the seed being randomly picked by hypothesis)?

axelrod/tests/integration/test_tournament.py
```diff
@@ -30,8 +31,10 @@ def setUpClass(cls):
         cls.expected_outcome.sort()
 
     def test_full_tournament(self):
```
drvinceknight: How much does this test affect things? If possible I'd like to keep a full tournament test **but** if there's a big speed up could we use the [random tournament property](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/property.py#L69) (which implements this as a hypothesis "strategy"), this would ensure we don't have issues with hunting down seeds and would also shrink failures etc...

The code would look something like:

```
from axelrod.tests.property import tournaments

@given(tournaments(
    strategies=short_run_time_strategies,
    min_size=1,
    max_size=30,
    min_turns=20,
    max_turns=200,
    min_noise=0,
    max_noise=1,
    min_repetitions=1,
    max_repetitions=20,
)
@settings(max_examples=5)
def test_big_tournaments(tournament):
        self.assertIsNone(
            tournament.play(progress_bar=False, filename=filename, build_results=False)
)
```

axelrod/tests/strategies/test_meta.py
marcharper: For this I copied the test from [TestPlayer](https://github.com/Axelrod-Python/Axelrod/blob/583cc4df87d6ea5f0eb0a08000e4b8b26bae56d3/axelrod/tests/strategies/test_player.py#L447), lowered the number of rounds, and took out the stochastic opponent. I changed the seeding on that test as well.

axelrod/tests/integration/test_tournament.py
marcharper: It helps a lot, many by excluding the meta strategies and reducing the total players. I'm fine with the hypothesis way of doing it

drvinceknight: Cool, let's go with removing the full test then :)

marcharper:
On further investigation, it appears that the colorbar location is more dynamic in mpl 3+. Calling tight layout, for example, causes it to move. I removed that portion of the test, which removes the need to skip the test and to limit mpl to < 3.

marcharper:
This PR has revealed a tangential issue (causing a test to intermittently fail) that was somewhat tricky to track down. The strategy [ShortMem](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/shortmem.py) is listed as `memory_depth=10` and is failing [this memory depth test](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/strategies/test_player.py#L647). Removing the `max_iterations` hypothesis variable somehow led to discovery of a counterexample seed.

I fixed it and some other tests. Even though the strategy only uses the last 10 rounds, it's behavior changes at round 10, therefore it has to know the exact number of rounds played, i.e. the full history.

marcharper:
I also re-pinned hypothesis. I was getting some odd coverage issues we using a newer version. Essentially coverage thought that some of the custom hypothesis decorators were not covered even though they have direct tests.

drvinceknight:
> I also re-pinned hypothesis. I was getting some odd coverage issues we using a newer version. Essentially coverage thought that some of the custom hypothesis decorators were not covered even though they have direct tests.

I noticed that: how weird... 

I'll have a proper look through everything later today @marcharper, nice work :+1:

axelrod/tests/integration/test_filtering.py
```diff
@@ -1,10 +1,13 @@
+import random
 import unittest
 
-from axelrod import all_strategies, filtered_strategies, seed
+from axelrod import short_run_time_strategies, filtered_strategies, seed
 
 from hypothesis import example, given, settings
 from hypothesis.strategies import integers
 
+strategies = random.sample(short_run_time_strategies, 20)
```
drvinceknight: We have the [`property.strategy_lists`](https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/tests/property.py#L12) that we can use instead of this.

So we'd have:

```
from axelrod.tests.property import strategy_lists
```

and then use:

```
@given(strategies=strategy_lists(min_size=20, max_size=20))  # or other parameters?
```

axelrod/tests/integration/test_tournament.py
```diff
@@ -29,12 +32,19 @@ def setUpClass(cls):
...
-        strategies = [strategy() for strategy in axelrod.strategies]
-        tournament = axelrod.Tournament(
-            name="test", players=strategies, game=self.game, turns=2, repetitions=1
-        )
+    @given(tournaments(
+        strategies=axelrod.short_run_time_strategies,
+        min_size=10,
+        max_size=30,
+        min_turns=2,
+        max_turns=40,
```
drvinceknight: There are some strategies that do weird things around 200 turns (I can't recall perfectly but I think that something like that might have even been the original motivation of this test). As we're limiting the number of strategies to 30 I don't think performance would be that bad if we let `max_turns=210` which just gives us a slightly more robust check.

axelrod/tests/integration/test_filtering.py
```diff
@@ -35,15 +33,18 @@ def test_boolean_filtering(self):
         min_memory_depth=integers(min_value=1, max_value=10),
         max_memory_depth=integers(min_value=1, max_value=10),
         memory_depth=integers(min_value=1, max_value=10),
+        strategies=strategy_lists(min_size=20, max_size=20),
     )
     @example(
         min_memory_depth=float("inf"),
         max_memory_depth=float("inf"),
         memory_depth=float("inf"),
+        strategies=strategy_lists(min_size=20, max_size=20),
```
drvinceknight: I believe this should be an actual list of strategies here for the `@example`. To do this properly with hypothesis I believe we should remove the example case here, and write a second test OR we could go with just `strategies=short_run_time_strategies` (the correct import will be needed at the top). I've run this locally and it seems fine without a large speedup.

.travis.yml
```diff
@@ -24,7 +24,7 @@ script:
   - cd docs; make clean; make html
   # Run the test suit with coverage
   - cd ..
-  - travis_wait 60 coverage run --source=axelrod -m unittest discover
+  - travis_wait 60 coverage run --source=axelrod -m unittest discover -v
```
drvinceknight: Double checking that the verbose tag isn't just something you wanted for debugging @marcharper? I don't necessarily mind either way (it does make scrolling through the travis log a big long but no big deal).

marcharper: I'll take it out

axelrod/tests/integration/test_filtering.py
marcharper: It's strange because these tests are running fine for me locally but not on Travis, and we didn't explicitly list strategies in this example previously.

.travis.yml
marcharper: I'm fine with adding it back in, I did add it just for debugging Travis

axelrod/tests/integration/test_filtering.py
drvinceknight: Strange that it runs locally... Perhaps a slightly different version of hypothesis has implemented the ability to use strategies inside examples... Would be neat if it did...

Previously I think we were just using `all_strategies` from the global space.

drvinceknight:
Very nice work @marcharper, this is a welcome speedup/refactor of the tests :+1:

PR #1239, opened 2019-02-10
Remove redundant member functions of the Action class

marcharper:
On profiling a tournament I noticed that the removed functions were among the most often called. It turns out that they are unnecessary because Enums provide their own `__hash__` and `__eq__`. This PR speeds up actual tournaments by 20-25% in my (limited) testing.


meatballs:
Nicely spotted!

drvinceknight:
Nice! I swear I still don't really understand what an Enum is but hey it sure does the job! :)

PR #1241, opened 2019-02-22
History class

marcharper:
Adds a History class, refactors Player a bit, and simplifies a number of tests and constructions. Note there were some subtle errors in the Meta players.

Increased runtime for a noisy tournament is ~25% but there are some optimizations possible:
* refactoring strategies to avoid the unnecessary property wrappers on the Player class
* Using a single joint history (adding the opponent's cooperations and defections) per Match
* Adding a version of the History class that doesn't track the state distributions and such when not needed (could detect with classifiers)
* Ultimately I'd like to change Player to only take a history object rather than the full opponent object

Also, I believe we could further refactor the dual transformer to be more efficient now but I didn't try that yet.

Updates:

* Added a `LimitedHistory` class that only retains `N` rounds of history, which improves the memory_depth test
* Removed nearly all instances of direct history assignment on the `Player` class, which should always be discouraged IMO
* Made various related minor updates
* I changed the meta classes so that each player on the meta team tracks its own history. Previously they were not updated with the history and so may not have properly set their internal variables properly as play proceeds. This PR has caused or exposed the issue.

I have fixed the remaining issues:

* the generators were not equal because the meta class was updating team histories on the next round, so the generators were off by one. I introduced an `update_history` method to the `Player` and `MetaPlayer` classes to fix, this method allows "post processing" of the moves by the players to fully end the round.
* transformed `SequencePlayer`s could not be pickled (there was no test). They can now and there are various new tests. I ended up adding a `__getstate__` method to `Player` (overridden in `SequencePlayer` and needed an abstract version in `Player`).

Tournaments take about 25% longer overall. In addition to the tasks in the first comment, we can get some or all of that back by optimizing in a subsequent PR or on the 5.0.0 branch:

* removing unnecessary function calls: inlining simultaneously_play, for example, and potentially optimizing how the history class is used in various places
* revisiting how the tournament class marshalls players

Increased runtime for the full standard tournament is 15% (30 hours versus 26 on my desktop). I think we'll easily get that back since removing function call overhead in #1239 yielded an improvement of ~25%.


marcharper:
The pickle issue has something to do with the MetaPlayer having a sequence player in the team (ThueMorse or ThueMorseInverse). When unpickled the subplayer isn't equal to itself, but its own test passes...

drvinceknight:
Sounds interesting @marcharper, I'll try and get to this next week. :+1:

axelrod/history.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class History(object):
+    """
+    History class to track the history of play and metadata including
+    the number of cooperations and defections, and if available, the
+    opponents plays and the state distribution of the history of play.
+    """
+
+    def __init__(self, plays=None, coplays=None):
```
drvinceknight: Could we document this, (it wasn't immediately obvious to me what a `coplay` was).

axelrod/history.py
```diff
@@ -0,0 +1,88 @@
...
+        return self._state_distribution
+
+    def __eq__(self, other):
+        if isinstance(other, list):
+            return self._plays == other
+        elif isinstance(other, History):
+            return self._plays == other._plays and self._coplays == other._coplays
+        raise TypeError("Cannot compare types.")
+
+    def __getitem__(self, key):
```
drvinceknight: Just a comment: This could be useful if we want to incorporate misperception (as opposed to noise).  :+1:

axelrod/player.py
```diff
@@ -245,8 +213,34 @@ def reset(self):
...
+        if isinstance(obj, list):
+            self._history = History(plays=obj)
+        elif isinstance(obj, History):
+            self._history = obj
+        else:
+            raise TypeError("Cannot assign history with type {}".format(
+                type(obj)
+            ))
+
+    # Properties maintained for legacy API, can refactor to self.history.X
```
drvinceknight: :+1:

axelrod/random_.py
```diff
@@ -31,6 +31,12 @@ def random_choice(p: float = 0.5) -> Action:
     return Action.D
 
 
+def random_flip(action, threshold):
+    if random.random() < threshold:
```
drvinceknight: Could we do something similar to `random_choice` so that if `threshold in [0, 1]` we don't sample a random number.

axelrod/tests/strategies/test_backstabber.py
```diff
@@ -172,101 +167,3 @@ def test_alt_strategy_stops_after_round_180(self):
             expected_actions=expected_actions,
             match_attributes={"length": 200},
         )
-
-
-class TestModuleMethods(unittest.TestCase):
```
drvinceknight: Why can we remove all of these?

axelrod/history.py
```diff
@@ -0,0 +1,88 @@
...
+        self._actions[play] += 1
+        if coplay:
+            self._coplays.append(coplay)
+            self._state_distribution[(play, coplay)] += 1
+
+    def copy(self):
+        return History(plays=self._plays, coplays=self._coplays)
+
+    def dual(self):
+        """Creates a dual history for use with DualTransformer."""
```
drvinceknight: I think we need to be careful with the dual here, it's not the same as flipping all the moves:

```
  310     The Dual of a strategy will return the exact opposite set of moves to the                                                                   
  311     original strategy when both are faced with the same history.                                                                                
  312                                                                                                                                                 
  313     A formal definition can be found in [Ashlock2010]_.                                                                                         
  314     http://doi.org/10.1109/ITW.2010.5593352    
```

I see that this test is not modified
https://github.com/Axelrod-Python/Axelrod/pull/1241/files#diff-1a0ceac0c6b9bdd7b042f6230ee3fb82R210 so I'm pretty confident it's all fine. I'm just not sure the name of the method is the best here, perhaps `flipped_plays`? The actual `dual` happens in the transformer.

marcharper: Sure, I wasn't sure what the proper name for this was.

axelrod/history.py
marcharper: Nice idea!

axelrod/random_.py
marcharper: Sure

axelrod/strategies/meta.py
```diff
@@ -151,21 +153,24 @@ def __init__(self, team=None):
...
         # next move.
         game = self.match_attributes["game"]
-        if len(self.history):
-            for i, player in enumerate(self.team):
-                last_round = (player.history[-1], opponent.history[-1])
-                s = game.scores[last_round][0]
-                self.scores[i] += s
+        scores = []
+        for player in self.team:
+            player.history[-1]
```
marcharper: need to remove this line

axelrod/tests/strategies/test_backstabber.py
marcharper: Good question. Some are just testing Player class properties that already have their own tests. But it looks like I was a bit overzealous.

axelrod/history.py
marcharper: Sure

axelrod/random_.py
marcharper: Actually we can just reuse `random_choice`

axelrod/tests/strategies/test_backstabber.py
marcharper: Actually I think all of these tests are redundant. If we want to keep any I'd prefer we re-write them to use an actual match rather than simply overwriting the history, which is a practice that we shouldn't encourage.

marcharper:
Updates:

* Added a `LimitedHistory` class that only retains `N` rounds of history, which improves the memory_depth test
* Removed nearly all instances of direct history assignment on the `Player` class, which should always be discouraged IMO
* Made various related minor updates

I suspect the pickle errors have something to do with the decorated meta classes and how they are rebuilt. I changed the meta classes so that each player on the meta team tracks its own history. Previously they were not updated with the history and so may not have properly set their internal variables properly as play proceeds. This has caused or exposed the issue.

axelrod/tests/strategies/test_backstabber.py
drvinceknight: Yeah I agree about not encouraging using history overwriting, could we either rewrite as part of this PR or leave them in and open an issue to remove them?

marcharper: I'll check but I think these tests are just targeting the subfuctions of the strategy and are all covered but the "standard" tests above in the file.

axelrod/history.py
drvinceknight: Yup :+1:

marcharper:
I'm still not sure why the meta players with sequence players on their teams don't pass the pickle test. (It's sufficient to use just ThueMorse as the team.) When you examine the players on team you see this:

```
{'init_kwargs': OrderedDict(), '_history': [C, D, D, C, D, C, C, D, D], 'classifier': {'memory_depth': inf, 'stochastic': False, 'makes_use_of': set(), 'long_run_time': False, 'inspects_source': False, 'manipulates_source': False, 'manipulates_state': False}, 'match_attributes': {'length': -1, 'game': Axelrod game: (R,P,S,T) = (3, 1, 0, 5), 'noise': 0}, 'sequence_generator': <generator object Player.__eq__.<locals>.<genexpr> at 0x7fc6aef72c50>}

{'init_kwargs': OrderedDict(), '_history': [C, D, D, C, D, C, C, D, D], 'classifier': {'memory_depth': inf, 'stochastic': False, 'makes_use_of': set(), 'long_run_time': False, 'inspects_source': False, 'manipulates_source': False, 'manipulates_state': False}, 'match_attributes': {'length': -1, 'game': Axelrod game: (R,P,S,T) = (3, 1, 0, 5), 'noise': 0}, 'sequence_generator': <generator object Player.__eq__.<locals>.<genexpr> at 0x7fc6aef72b48>}```

Note that the players seem to get correlated by the Player.__eq__ function: the generates have the bolded part <generator object *Player.__eq__.<locals>*.<genexpr> at 0x7fc6aef72c50>.

Oddly enough when I add similar tests to the `TestPlayer` I don't get the same result.

marcharper:
Also note that I made a significant change to the metaplayers in that they now pass along the team member's individual play choices rather than all sharing a single history. I think this makes more sense but it will impact how they play.

drvinceknight:
> Also note that I made a significant change to the metaplayers in that they now pass along the team member's individual play choices rather than all sharing a single history. I think this makes more sense but it will impact how they play.

Yeah I'm fine with this change, they'll still always be looking at the correct history for the opponent. :+1:

drvinceknight:
> Note that the sequence_generator attribute seem to correlate the instances when passed through the `Player.__eq__` function .
> 
> Oddly enough when I add similar tests to the `TestPlayer` I don't get the same failure.

I've started to have a look (really sorry it's taking me so long to try and be helpful). The `__eq__` method in the player class needs to be a bit smart with the equality of a sequence, see https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/player.py#L145 where it essentially needs to split the generator in to two copies of itself. Use one of the copies to check equality and then set the attribute with the other copy.

I'm not entirely sure why this would fall over when checking pickling equality but perhaps that's something worth considering? (If you haven't already).

marcharper:
I believe I have fixed the remaining issues:
* the generators were not equal because the meta class was updating team histories on the next round, so the generators were off by one. I introduced an `update_history` method to the `Player` and `MetaPlayer` classes to fix, this method allows "post processing" of the moves by the players to fully end the round.
* transformed `SequencePlayer`s could not be pickled. They can now and there are various new tests. I ended up adding a `__getstate__` method to `Player` (overridden in SequencePlayer and needs an abstract version).

Tournaments take about 25% longer overall. We can get some of that back by optimizing in a subsequent PR or on the 5.0.0 branch by:
* removing unnecessary function calls: inlining `simultaneously_play`, for example, and potentially optimizing how the history class is used in various places
* revisiting how the tournament class marshalls players -- there's no need to rebuild them now (if more efficient)

drvinceknight:
Nice work!

> Tournaments take about 25% longer overall. In addition to the tasks in the first comment, we can get some or all of that back by optimizing in a subsequent PR or on the 5.0.0 branch:

Is there any merit in me running a full tournament to see if the 25% increase scales well?

marcharper:
> Is there any merit in me running a full tournament to see if the 25% increase scales well?

Yes, and we should take a look at the impact on the Meta strategies

drvinceknight:
> Yes, and we should take a look at the impact on the Meta strategies

Ok cool, I'll aim to get some things running for that.

drvinceknight:
I've got some code running. Will report back once it's done :+1:

marcharper:
Increased runtime for the full standard tournament is 15% (30 hours versus 26 on my desktop). I think we'll easily get that since removing function call overhead in #1239 yielded an improvement of ~25%.

marcharper:
@drvinceknight @meatballs I think this is good to go in, PTAL when you get a chance.

drvinceknight:
I agree it's pretty much there (sorry for being so slow to get back to you, I've got a tonne of marking at the moment), I've got some long (probably too long) tournaments running for timing experiments but we can merge and if they come back atrocious (which I doubt they will) we can always figure that out later.

I've already looked through and everything looks good to me, but a close look from @meatballs would be great :+1:

marcharper:
@meatballs PTAL

PR #1242, opened 2019-02-26
Fix memory-one error in compute FSM memory.

gaffney2010:
In using, I uncovered a bug.  Basically memory-zero and memory-one cases need to be handled specially.  I was checking this case with a condition on the number of memit-pairs, but in fact a memory-one strategy could have a long chain memit-pairs that all suggest the same set of reactions.

So I fixed this, and added a test that would have caught this error.


gaffney2010:
> Any chance you could rebase on to master?

I'm trying.  Afraid I'm only making it worse.

PR #1244, opened 2019-04-09
Add Graaskamp (final missing strategy from Axelrod's first tournament)

### LINKED ISSUE #1109: Specific strategy: Graaskamp

drvinceknight:
Note that technically because we do not know what the "Analogy" strategy is, this is a best possible approximation of the Graaskamp strategy. 

This PR includes commits from https://github.com/Axelrod-Python/Axelrod/pull/1234 (merging this PR will automatically merge that one).

Closes #1109


docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -47,7 +47,7 @@ strategies::
     ... }
     >>> strategies = axl.filtered_strategies(filterset)
     >>> len(strategies)
-    84
+    85
```
meatballs: Travis seems to be suggesting this needs to be bumped to 86

drvinceknight: And travis is correct! Ta.

PR #1246, opened 2019-07-12
Added a CodeOfConduct.md file

rjsu26:
Please change the email address at the line 58

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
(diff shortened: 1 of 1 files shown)


marcharper:
Hi, there is already a code of conduct in the documentation [here](https://axelrod.readthedocs.io/en/stable/community/coc.html). The source file is in the docs directory [here](https://github.com/Axelrod-Python/Axelrod/blob/master/docs/community/coc.rst).

Thanks for finding the broken links in the other issues, and if you find any more, let us know!

rjsu26:
It was very stupid of me to not check in the source code for a code of conduct. But this confusion arouse because of my observation in the insight page, community section, where the Code Of Conduct was shown to be missing. SS below. We can mend that so it shows that there is already an existing one.

image

marcharper:
We'll look into it.

PR #1255, opened 2019-09-03
Update setup.py to include python 3.7 instead of 3.5

marcharper:


drvinceknight:
I suspect we've already discussed this and I'm just forgetting but is there a reason why we cannot add 3.7 and keep 3.5?

marcharper:
No reason, just matching with Travis. We "officially" support the two latest released of Python but I don't see any problem leaving 3.5 in.

meatballs:
I can't remember precisely what they were, but there were some things we wanted to do that were only supported in 3.6+ and we decided to leave them until 3.7 came along and we could drop 3.5

drvinceknight:
Let's go ahead and merge, if something comes up that needs us to bring 3.5 back in then we can always do that.

Issue #1240, opened 2019-02-10
5.0.0 discussion -- Abstract base classes for players


marcharper:
I've been working on a number of separate branches and problems and would like some feedback on a proposal before I sink a lot of time into it.

For the namespace reorganization for Axelrod (#1175) we need an abstract player class. A lot of the core functionality of the class can move out (cloning for example); the IPD class just needs to additionally track cooperations, defections, and the state distribution (or perhaps on a History class). To use the dojo on the ultimatum game we need resetting, cloning, and such as well.

For the new style of mutation for the dojo and Moran class (see the [mutation branch](https://github.com/Axelrod-Python/Axelrod/commit/ed5ba125a6895c992bc2d1189eedb3c22de43ed7)), I think it makes sense to introduce an intermediate ABC that defines mutation and crossover methods. IIRC we needed the params classes in the dojo because Players couldn't be picked, however I think that's no longer the case, so all that functionality can (and I think should) move into the main library.

So I'd like to do the following:
* introduce an ABC chain of Player --> MutablePlayer --> IPDPlayer (and soon UltimatumPlayer)
* enhance the Moran process to handle the new style of mutation and move the functionality of the Params classes in the dojo to the main library
* Release new versions of the main library and the dojo simultaneously
* Start on 5.0.0 by reorganizing the namespace and then merge in the Ultimatum player code

Thoughts?

drvinceknight:
This all sounds very good.

>     * introduce an ABC chain of Player --> MutablePlayer --> IPDPlayer (and soon UltimatumPlayer)

Could the `MutablePlayer` class not just be the base class? I suspect you've thought about this and there's a good reason, it's not immediately apparent to me.

>     * Release new versions of the main library and the dojo simultaneously

For the axelrod library this would be on the `master` branch right?

>  however I think that's no longer the case, so all that functionality can (and I think should) move into the main library.

Is there merit in perhaps bringing the dojo in to the axelrod library itself? I'm generally more of a fan of keeping them separated *but* given the proposed changes it feels like they'll be much closer linked.

```
>>> axelrod.dojo.train # Or whatever?
```

>     * Start on 5.0.0 by reorganizing the namespace and then merge in the Ultimatum player code

This would all be on the `5.0.0` branch right?

marcharper:
Mutable could be the base class I suppose. This would be on the master branch.

We may want to fold in the dojo once the rest of this is done. I hadn't thought about that yet.

[...]

PR #1256, opened 2019-09-03
Integrate dojo params into Axelrod

marcharper:
Re: #1240, this PR merges the Parameter classes for the training algorithms as evolvable players in Axelrod by introducing a subclass of each of the archetypes. The `init_kwargs` attribute now does most of the heavy lifting and there are many new tests to ensure identical play.

The dojo will also be simultaneously updated.

There are a few uncovered lines and mypy tests to be fixed but I think it's ready for someone else to look at.


drvinceknight:
I'll aim to take a close look through on Thursday :+1:

axelrod/graph.py
```diff
@@ -120,25 +122,42 @@ def cycle(length, directed=False):
...
-    graph = Graph(edges=edges, directed=False)
+    graph = Graph(directed=directed, edges=edges)
+    if loops:
+        graph.add_loops()
+    return graph
+
 
+def attached_complete_graphs(length, loops=True, directed=False):
+    edges = []
+    # Two complete graphs
```
drvinceknight: I'm not entirely sure I follow what this method is for.

axelrod/moran.py
```diff
@@ -108,6 +110,8 @@ def __init__(
             given
         fitness_transformation:
             A function mapping a score to a (non-negative) float
```
drvinceknight: Is `mutation_method` in the doctstring?

axelrod/moran.py
```diff
@@ -56,6 +56,8 @@ def __init__(
         interaction_graph: Graph = None,
         reproduction_graph: Graph = None,
         fitness_transformation: Callable = None,
+        mutation_method="transition",
```
drvinceknight: how do we feel about having booleans instead of passing a string argument?

```
atomic=False, transition=True
```

axelrod/random_.py
```diff
@@ -63,10 +71,16 @@ def randrange(a: int, b: int) -> int:
     return a + int(r)
 
 
-def seed(seed_):
-    """Sets a seed"""
-    random.seed(seed_)
-    numpy.random.seed(seed_)
+def random_vector(size):
+    """Create a random vector of values in [0, 1] that sums to 1."""
```
drvinceknight: Is something like:

```
vector = np.random.random(size)
return vector / np.sum(vector)
```

perhaps more efficient because of numpy magic (of course it could be the opposite) and more readable?

drvinceknight:
This looks like it'll be a great addition/move. I think we might need to document some of this somewhere, perhaps in the docstrings of the particular sublayers or perhaps better to have a section in the documentation showing how one of them can be used and how it behaves (and pointing at all the others)?

axelrod/graph.py
marcharper: It's for my speciation project, just another topology. Can move it out if you want but it would be good to have a few more. I have a grid function that I want to add as well.

axelrod/moran.py
marcharper: added

axelrod/moran.py
marcharper: You can't have both at the same time so I'd prefer not to use individual booleans. I considered using an enum instead.

axelrod/random_.py
marcharper: Yeah that's fine. I don't think I originally wrote this function, it looks odd to me as well.

marcharper:
For sure, we'll want to add some more documentation. If we're settled on the API and other changes that I've made then I'll add some docs but I didn't want to yet in case there were some major objections.

We may also want to move the rest of the dojo over -- it's basically just the two algorithms and the invocation scripts now. I'd prefer to do that in another PR though.

drvinceknight:
> For sure, we'll want to add some more documentation. If we're settled on the API and other changes that I've made then I'll add some docs but I didn't want to yet in case there were some major objections.

Yeah sounds good. I think I'm settled on the API, very nice work. Having some example use cases documented will just confirm that for me :+1:

> 
> We may also want to move the rest of the dojo over -- it's basically just the two algorithms and the invocation scripts now. I'd prefer to do that in another PR though. That should be easy, this PR is the deeper integration.

I'd prefer to leave the invocation scripts out of the library but I agree with the algorithms :+1:

docs/tutorials/getting_started/moran.rst
```diff
@@ -132,3 +132,18 @@ Other types of implemented Moran processes:
 
 - :ref:`moran-process-on-graphs`
 - :ref:`approximate-moran-process`
+
+Atomic Mutation for Evolvable Players
```
drvinceknight: I wonder if this should perhaps not be in the `getting_started` section? Perhaps just include the code as an example for what you've written in the `further_topics` section? (If you feel strongly about having it here that's fine.)

marcharper: I don't feel strongly about it but there's also not a lot to write about other than the Moran process until we move the dojo algorithms over, in which case the ideal documentation structure should be more clear.

drvinceknight: I thought it about it some more and I think if you're ok with it I'd like us to move it to `further_topics`. I think in the `getting_started` section it might be a bit confusing for newcomers to game theory.

marcharper: Done

axelrod/graph.py
drvinceknight: Could you add a docstring just highlighting the purpose.

drvinceknight:
Thanks @marcharper, this looks good to me. :+1:

marcharper:
@meatballs ready for you when you get a chance

PR #1260, opened 2019-09-26
Fix bug in Kluepfel strategy

gaffney2010:
Happened to catch this bug in an old strategy I submitted.

It didn't match the logic here.
https://github.com/Axelrod-Python/TourExec/blob/master/src/strategies/k32r.f

Shouldn't come up often.  It's difficult to construct an example that would have caught this.

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -536,11 +536,11 @@ class Kluepfel(Player):
                     self.dd_counts += 1
             else:
                 if opponent.history[-1] == C:
-                    self.dc_counts += 1
-                else:
                     self.cc_counts += 1
+                else:
+                    self.dc_counts += 1
 
-        # Check for randomness
+        # Check for "randomness"
         if len(self.history) > 26:
             if self.cd_counts >= (self.cd_counts + self.dd_counts) / 2 - 0.75 * np.sqrt(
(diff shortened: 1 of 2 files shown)


drvinceknight:
Thank you @gaffney2010 

> It's difficult to construct an example that would have caught this.

Does that mean asking for a specific test is not possible? Ideally would be nice to add that.

gaffney2010:
> Does that mean asking for a specific test is not possible? Ideally would be nice to add that.

That's right.

drvinceknight:
Cool no problem.

It's hard to review without re translating the source code, I'm sure you're correct but could you perhaps point at lines in the original code that explain what's going on?

gaffney2010:
There were two errors.  (Actually they cancelled each other out on one of the tests.)

Firstly, dc_counts is how many times an opponent responded to my C with a D.  So it should be down the branch that self.history[-2]==C and opponent.history[-1]==D.  I had mixed up an if/else.

Secondly, the two inequalities that "Check for randomness" come from lines 39-40 on the Fortran; the LHS on line 40 is C4, which is a variable that means "# OF HIS DEF. AFTER MY COOPERATION," and I was accidentally using cc_counts.

drvinceknight:
Ok, thanks for the fix and walking me through it :+1:

drvinceknight:
(great catch!)

PR #1267, opened 2019-10-17
Change EvolvablePlayer serialization to write strings instead of bytes

marcharper:
for easier inclusion in CSV files (as used in axelrod-dojo).


marcharper:
Done

PR #1268, opened 2019-10-23
Fix rendering of code snippet in docs.

drvinceknight:
This ensures the `doctest: +SKIP` does not appear.


marcharper:
merging as bug fix

Issue #813, opened 2017-01-11
More example and visualization notebooks


marcharper:
We could use some more examples and demos in our [notebooks](https://github.com/Axelrod-Python/Axelrod-notebooks) repository.

nadia39:
I would like to assist in this issue, could you give me more information on what you need?

drvinceknight:
> I would like to assist in this issue, could you give me more information on what you need?

Hi @nadia39 great to have you assist! 

If you take a look at this repository: https://github.com/Axelrod-Python/Axelrod-notebooks we have a number of Jupyter notebook examples showing a variety of things that can be done with the library. We're always looking for more examples which is what this issue is asking for.

Perhaps take a look through the documentation http://axelrod.readthedocs.io/en/latest/ and see if you think there's anything specific worth adding as a Jupyter notebook?

SirRichter:
Is there a strategy where the player assumes that their opponent's probability of defecting is a Bernoulli variable, where the parameter is based on a Bayesian analysis of the opponent's previous moves?

marcharper:
@SirRichter Not one that explicitly uses a Bayesian prior AFAIK. There are some strategies that attempt to determine the opponent's probability of cooperating in various ways, such as the running counts of the opponent's plays, which would correspond to some uninformative prior (Beta(0, 0) I guess) and then taking the posterior maximum.

Feel free to implement such a strategy (ideally using scipy) and open a PR on the main repository.

danilobellini:
I've created one in Brazilian Portuguese: https://github.com/danilobellini/notebooks/blob/master/2018-11-10_Axelrod/2018-11-10_Axelrod.ipynb

Shall I make a PR based on it? The stuff from there I may add in a PR are:

- The `Detective` and `Simpleton` strategies (which I wrote in Python based on the Nicky Case example)
- The notebook itself (which is mainly a "Getting started, step by step", but in Brazilian Portuguese)

PR #1269, opened 2019-10-30
Add Detective strategy from Nicky Case game

danilobellini:
It's an implementation of the "Detective" prober strategy as described in "The Evolution of Trust" game created by Nicky Case in July 2017

  https://ncase.me/trust/

In the game, it's implemented in:

  https://github.com/ncase/trust/blob/8ff91b81358d02847f81eaa46a797e806e749178/js/sims/PD.js#L217

This strategy was referenced in #813, and it's the only one in the Nicky Case game that wasn't already implemented in this library.


drvinceknight:
Thank you :+1:

PR #1274, opened 2019-11-13
Removes Fool Me Forever

Nikoleta-v3:
Hello everyone,

While working on one of my projects I realised that the strategy `Fool Me Forever`, which is an original strategy implemented within the library, is actually the same as the strategy `EasyGo` from
[Prison1998 and Li2011](https://axelrod.readthedocs.io/en/stable/reference/bibliography.html#li2011).

Source code for `Fool me Forever` (link to source code [here](https://axelrod.readthedocs.io/en/stable/_modules/axelrod/strategies/oncebitten.html#FoolMeForever)):
```python
    @staticmethod
        if opponent.defections > 0:
            return C
        return D
```
Source code for `EasyGo` (link to source code [here](https://axelrod.readthedocs.io/en/stable/_modules/axelrod/strategies/grudger.html#EasyGo)):
```python
    @staticmethod
        """Begins by playing D, then plays C for the remaining rounds if the
        opponent ever plays D."""
        if opponent.defections:
            return C
        return D
```

The code snippets, the Ashlock fingerprints for both strategies and their performance in the Axelrod-Python standard tournament (result csv: https://github.com/Axelrod-Python/tournament/blob/gh-pages/assets/std_summary.csv) are the same. 

fool_me_forever

easygo

This `pr` removes Fool me Forever (because it's an original strategy), the imports and tests associated with it. Let me know if it's okay and if there is something I forgot to do 👍

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -160,7 +160,7 @@ from .mindcontrol import MindBender, MindController, MindWarper
 from .mindreader import MindReader, MirrorMindReader, ProtectedMindReader
 from .mutual import Desperate, Hopeless, Willing
 from .negation import Negation
-from .oncebitten import FoolMeForever, FoolMeOnce, ForgetfulFoolMeOnce, OnceBitten
+from .oncebitten import FoolMeOnce, ForgetfulFoolMeOnce, OnceBitten
 from .prober import (
     CollectiveStrategy,
     Detective,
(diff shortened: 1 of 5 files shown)


Nikoleta-v3:
I saw the failure 👍 fixing it now

marcharper:
Nice catch :)

marcharper:
Since we've mentioned FoolMeForever in our publications we should mark it as an alternative name for EasyGo

PR #1275, opened 2019-12-01
Re factor/implement first tournament strategies

### LINKED ISSUE #1273: Possible Implementation errors for first tournament strategies
@id428 has written a detailed pdf (attached to this issue)
[First Tournament Strategies.pdf](https://github.com/Axelrod-Python/Axelrod/files/3841326/First.Tournament.Strategies.pdf)
 with their cross examination of the strategies which is attached to the issue. Here is a summary they posted on gitter:

> In summary, these are the errors we found:
> - Tideman & Chieruzzi: The strategy reportedly defected in the last two rounds. This is not implemented. (There are also two minor additional points mentioned in the PDF)
> - Nydegger: The value 49 is missing in A
> - Grofmann: We dont know where the behaviour for the first seven rounds is described. Maybe something was implemented that was not existend in the first place.
> - Shubik: This strategy should cooperate at least twice after each run of retaliations. The implementation prevents that. (At least that's how we understand Axelrods description, see PDF)
> - Graaskamp: Reportedly checks for its own twin but does not in the implementation
> - Downing: We don't know where the final decision rule in the implementation comes from.

drvinceknight:
#1273 noted a number of potential implementation errors in the first tournament strategies.

This is a first draft of addressing these so we can discuss particular implementations. 

I believe that a number of errors were a result of confusion between first tournament and second tournament code so I've renamed all the strategies: `FirstBy<author>` and `SecondBy<author>`. (I'm open to other suggestions).

In some places I've added a number of things to the docstrings to make explicit the assumptions made when descriptions are not clear.
 
Closes #1273


drvinceknight:
I've hopefully kept my commits quite modular and in some cases added text describing my approach to the commit message. For example, Downing required me to sift through Downing's 1975 paper: eccd7e2.

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,202 @@ def strategy(self, opponent: Player) -> Action:
...
+    Thus at every turn, the strategy will calculate those two values and
+    cooperate if E_C > E_D and will defect if E_C < E_D.
+
+    In the case of E_C = E_D, the player will alternate from their previous
+    move. This is based on specific sentence from Axelrod's original paper:
+
+    > "Under certain circumstances, DOWNING will even determine that the best
+    > strategy is to alternate cooperation and defection."
+
+    One final important point is the early game behaviour of the strategy. It
```
marcharper: Is this the "bug" that is later referred to? If so can we explicitly mention that here in the docstring?

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,202 @@ def strategy(self, opponent: Player) -> Action:
...
+    --
+    This strategy came 10th in Axelrod's original tournament.
 
     Names:
 
     - Revised Downing: [Axelrod1980]_
     """
 
-    name = "Revised Downing"
+    name = "First tournament by Downing"
```
marcharper: How do you feel about "First by Downing"? Just considering how the longer names will appear in plots.

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,202 @@ def strategy(self, opponent: Player) -> Action:
...
+    def __init__(self) -> None:
         super().__init__()
-        self.revised = revised
-        self.good = 1.0
-        self.bad = 0.0
-        self.nice1 = 0
-        self.nice2 = 0
-        self.total_C = 0  # note the same as self.cooperations
-        self.total_D = 0  # note the same as self.defections
+        self.number_opponent_cooperations_in_response_to_C = 0
```
marcharper: I think we track this in the state distributions in the history class, but perhaps you want to be more explicit here

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,202 @@ def strategy(self, opponent: Player) -> Action:
...
-        elif (c >= 0 and c < alt) or (alt >= 0):
-            move = self.history[-1].flip()
-        else:
-            move = D
-        return move
+        if round_number == 1:
+            return D
+        if round_number == 2:
+            if opponent.history[-1] == C:
+                self.number_opponent_cooperations_in_response_to_C += 1
```
marcharper: Seems like this is a response to a D since this strategy always defects on round 1.

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,202 @@ def strategy(self, opponent: Player) -> Action:
...
+            return D
+
+
+        if self.history[-2] == C and opponent.history[-1] == C:
+            self.number_opponent_cooperations_in_response_to_C += 1
+        if self.history[-2] == D and opponent.history[-1] == C:
+            self.number_opponent_cooperations_in_response_to_D += 1
+
+        alpha = (self.number_opponent_cooperations_in_response_to_C /
+                 (self.cooperations + 1))  # Adding 1 to count for opening move
```
marcharper: I'm confused again here -- isn't the first move always a defection? Or is this a bug in the original implementation that we're preserving?

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,202 @@ def strategy(self, opponent: Player) -> Action:
...
+        beta = (self.number_opponent_cooperations_in_response_to_D /
```
marcharper: Maybe add a comment staying that we're never dividing by zero here

axelrod/strategies/axelrod_first.py
```diff
@@ -206,25 +324,36 @@ def strategy(self, opponent: Player) -> Action:
         return random_choice(p)
 
 
-class Graaskamp(Player):
+class FirstByGraaskamp(Player):
     """
+    Submitted to Axelrod's first tournament by James Graaskamp..
```
marcharper: remove one .

axelrod/strategies/axelrod_first.py
```diff
@@ -277,8 +406,8 @@ def strategy(self, opponent: Player) -> Action:
         if all(
             opponent.history[i] == self.history[i - 1]
             for i in range(1, len(self.history))
-        ):
-            # Check if opponent plays Tit for Tat
+        ) or opponent.history == self.history:
```
marcharper: Note that this is unlikely to be the original rule since strategies in the Fortran code would have to track their history themselves, and it's unlikely that the full history array was retained. I would guess that the original tracks cooperation and defection totals, the move on round 50, and the subsequent 5 moves, maybe with a bool "is_clone" or similar.

Of course we can't really know what it originally did so feel free to leave it as is.

axelrod/strategies/axelrod_first.py
```diff
@@ -446,21 +587,44 @@ def strategy(self, opponent: Player) -> Action:
...
+    > each departure from mutual cooperation. This rule is described with its
+    > strategic implications in Shubik (1970). Further treatment of its is given
+    > in Taylor (1976).
+
+    There is some room for interpretation as to how the strategy reacts to a
+    defection on the turn where it starts to cooperate once more. In Shubik
+    (1970) the strategy is described as:
 
-    This strategy came 5th in Axelrod's original tournament.
+    >  "I will play my move 1 to begin with and will continue to do so, so long
```
marcharper: - extra space

axelrod/strategies/axelrod_first.py
```diff
@@ -523,9 +695,9 @@ class Tullock(Player):
     - Tullock: [Axelrod1980]_
     """
 
-    name = "Tullock"
+    name = "First tournament by Tullock"
     classifier = {
-        "memory_depth": 11,  # long memory, modified by init
+        "memory_depth": 11,
```
marcharper: I think this one might be infinite actually since it needs to know if it's within the first 11 rounds or not

axelrod/strategies/axelrod_first.py
```diff
@@ -538,49 +710,48 @@ def __init__(self, rounds_to_cooperate: int = 11) -> None:
...
         super().__init__()
         self._rounds_to_cooperate = rounds_to_cooperate
         self.memory_depth = rounds_to_cooperate
 
     def strategy(self, opponent: Player) -> Action:
-        rounds = self._rounds_to_cooperate
-        if len(self.history) < rounds:
+        if len(self.history) < self._rounds_to_cooperate:
             return C
+        rounds = self._rounds_to_cooperate - 1
```
marcharper: Maybe we should hard-code `_rounds_to_cooperate` since setting it to 1 would cause a division by zero error

axelrod/strategies/calculator.py
```diff
@@ -2,14 +2,14 @@
 from axelrod.action import Action
 from axelrod.player import Player
 
-from .axelrod_first import Joss
```
marcharper: Since Joss is a somewhat more well-known strategy than the others in the first tournament, maybe we should use `from .axelrod_first import FirstByJoss as Joss` instead?

axelrod/tests/strategies/test_axelrod_first.py
```diff
@@ -43,55 +43,48 @@ def test_strategy(self):
...
-    def test_not_revised(self):
-        # Test not revised
-        player = self.player(revised=False)
-        opponent = axelrod.Cooperator()
-        match = axelrod.Match((player, opponent), turns=2)
-        self.assertEqual(match.play(), [(D, C), (D, C)])
-
 
-class TestFeld(TestPlayer):
+class TestFristByFeld(TestPlayer):
```
marcharper: `TestFristByFeld` -> `TestFirstByFeld`

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,202 @@ def strategy(self, opponent: Player) -> Action:
         opponent ever plays D."""
         if len(self.history) < self._rounds_to_cooperate:
             return C
-        if opponent.defections:
+        if opponent.defections > 0:
             return D
         return C
 
+class FirstByDowning(Player):
```
marcharper: Since RevisedDowning is in the second tournament, can we also preserve that implementation and/or compare to the fortran implementation?

drvinceknight: Sorry I meant to write that in my PR: we need to consider what we do with `RevisedDowning`. As it's in the second tournament is it worth waiting until we translate https://github.com/Axelrod-Python/TourExec/blob/v0.2.0/src/strategies/K59R.f or implement `RevisedDowning` as a modification of the strategy in this PR?

axelrod/strategies/calculator.py
drvinceknight: Sure good call!

axelrod/strategies/axelrod_first.py
drvinceknight: Yeah I'm happy to do that. Looking back over `axelrod_first.py` I think we went a bit overboard on parametrising everything (if I remember correctly this was something I was intent on - my bad).

axelrod/strategies/axelrod_first.py
drvinceknight: Yeah good call!

axelrod/strategies/axelrod_first.py
drvinceknight: Very good spot, I'll add something about this to the docstring though.

axelrod/strategies/axelrod_first.py
drvinceknight: Yup completely fine by me.

axelrod/strategies/axelrod_first.py
drvinceknight: So it wasn't actually a "bug", Axelrod discusses it at length in the paper so I'm relatively sure (I wouldn't bet my house on it) that the strategy was implemented as intended. However it seems that it was accepted afterwards that the intent was a mistake. I'll add that in future tournaments this strategy was implemented with a modified initial behaviour.

axelrod/strategies/axelrod_first.py
drvinceknight: I'd completely forgotten about our new shiny history class! 

In this instance though I don't think we do track this:

```python
>>> import axelrod as axl

>>> players = (axl.CyclerCCD(), axl.CyclerDC())
>>> match = axl.Match(players)
>>> interactions = match.play()
>>> players[0].history.state_distribution
Counter({(C, D): 67, (C, C): 67, (D, D): 33, (D, C): 33})
```

Unless I'm mistaken that's returning the strategy distributions however I there need to count the response to `C`. 

I've looked through the other properties and don't see what we need in there but that could be something nice to add?

axelrod/strategies/axelrod_first.py
drvinceknight: Ah no that's not what's happening here but I understand the confusion so I'm happy to improve how this is written.

What is happening:

I assume that the player is using these first two turns to build up an early estimate of alpha  (P(C|C) - probability the opponent cooperates in turn k + 1 if the player cooperated in turn k) and beta (P(C|D)  - probability the opponent cooperates in turn k + 1 if the player defected in turn k).

I use the opening player of the opponent (k = 1) to estimate alpha - ie I assume the opponent plays their opening round as if the player had cooperated in round k = 0 (a round that does not exist).

I use the second play of the opponent (k=2) to estimate beta as the player defects in round k=1.

The player does also defect in round k=2.

Thus, `if round_number == 2:` and `if opponent.history[-1] == C:` then the opponent cooperated as their opening move and so `self.number_opponent_cooperations_in_response_to_C += 1`

axelrod/strategies/axelrod_first.py
drvinceknight: I understand the confusion, this can probably be improved.

The adjustment here is to account for the "phantom" non existent round 0 cooperation by the player that is used to estimate alpha.

So for example, if the plays are:

[(D, C), (D, C)]

Then the opponent's first cooperation counts as a cooperation in response to the non existent cooperation of round 0. The total number of cooperations in response to a cooperation is 1. We need to take in to account that extra phantom cooperation to estimate the probability alpha=P(C|C) as 1 / 1 = 1.

This is all a best guess of course.

axelrod/strategies/axelrod_first.py
```diff
@@ -1,5 +1,16 @@
 """
-Additional strategies from Axelrod's first tournament.
+Strategies from Axelrod's first tournament. All strategies in this module are
+prefixed by `FirstBy` to indicate that they were submitted in Axelrod's First
```
meatballs: nitpick: submitted to

axelrod/strategies/axelrod_first.py
```diff
@@ -1,5 +1,16 @@
...
+tournament by the given author.
+
+Note that in these strategies are implemented from the descriptions presented
```
meatballs: typo: "Note that in these..." probably should be "Note that these..."

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,203 @@ def strategy(self, opponent: Player) -> Action:
...
+    playing a reactive strategy defined by these two conditional probabilities.
+
+    The strategy aims to maximise the long run utility against such a strategy
+    and the mechanism for this is described in Appendix A (more on this later).
+
+    One final point from the main text is, on page 372:
+
+    > "For the various lag-one matching strategies of O, the maximizing
+    strategies of S will be 100% C, or 100% D, or for some strategies all S
+    strategies will be functionaly equivalent."
```
meatballs: possible typo: "functionaly" (If this is how it's actually spelt in the paper, we should add (sic) here)

axelrod/strategies/axelrod_first.py
```diff
@@ -56,97 +69,203 @@ def strategy(self, opponent: Player) -> Action:
...
+
+    > "This rule starts with tit for tat and gradually lowers its probability of
+    > cooperation following the other's cooperation to .5 by the two hundredth
+    > move. It always defects after a defection by the other."
+
     This strategy plays Tit For Tat, always defecting if the opponent defects but
     cooperating when the opponent cooperates with a gradually decreasing probability
-    until it is only .5.
+    until it is only .5. Note that the description does not clearly indicate how
+    the cooperation probability should drop, this implements a linear decreasing
```
meatballs: nitpick: comma should be replaced with a period and new sentence.

axelrod/strategies/axelrod_first.py
```diff
@@ -206,25 +325,39 @@ def strategy(self, opponent: Player) -> Action:
...
+    > defections.:
+
+    This is implemented as:
 
     1. Plays Tit For Tat for the first 50 rounds;
     2. Defects on round 51;
     3. Plays 5 further rounds of Tit For Tat;
     4. A check is then made to see if the opponent is playing randomly in which
-       case it defects for the rest of the game;
+       case it defects for the rest of the game, this is implemented with a chi
```
meatballs: nitpick: new sentence for "This is implemented with..."

axelrod/strategies/axelrod_first.py
```diff
@@ -206,25 +325,39 @@ def strategy(self, opponent: Player) -> Action:
...
+       a clone of itself. If
        so it plays Tit For Tat. If not it cooperates and randomly defects every 5
        to 15 moves.
 
-
     Note that there is no information about 'Analogy' available thus Step 5 is
-    not implemented fully.
+    a "best possible" interpretation of the description in the paper.
+    Furthermore the test for the clone is implemented as checking that both
+    players have played the same moves for the entire game, this is unlikely to
```
meatballs: nitpick: New sentence for "This is unlikely..."

drvinceknight:
> What do you think about having a classifier entry for "in Axelrod's first tournament" and "in Axelrod's second tournament"? Then we can also easily add lists of first and second tournament strategies along side `short_run_term_strategies` for ease of use (and possibly a nice tutorial example, maybe even an advanced tutorial comparing to the Fortran implementations with fingerprints). We'd also need to add these classifiers to TFT.

Yeah I really like this idea. Perhaps a classifier is not worth doing as it's not something dynamic (we will never implement another strategy from the first tournament) - perhaps a hard coded list is the way to go?

Also - unrelated - I think we should just go ahead remove the cheating strategies (I know I've been on the other side of this for a long time), if only to clear up the classifiers. But that's a discussion for another time...

drvinceknight:
Thanks @meatballs I think I've addressed all those.

axelrod/strategies/axelrod_first.py
marcharper: The current `RevisedDowning` looks really similar to the Fortran code. If they are basically the same I'm in favor of leaving `RevisedDowning` (eliminating the `revised` bool) as the second tournament implementation. Maybe comparing the fingerprints will tell us if it's essentially correct or not?

marcharper:
> Yeah I really like this idea. Perhaps a classifier is not worth doing as it's not something dynamic (we will never implement another strategy from the first tournament) - perhaps a hard coded list is the way to go?

Hard-coded lists seem fine in this case.

> Also - unrelated - I think we should just go ahead remove the cheating strategies (I know I've been on the other side of this for a long time), if only to clear up the classifiers. But that's a discussion for another time...

Sure, let's (eventually) at least silo them sufficiently so that no one inadvertently uses them.

axelrod/strategies/axelrod_first.py
marcharper: Gotcha, let's leave it as is for now and we can consider adding something to the history class later.

axelrod/strategies/axelrod_first.py
marcharper: I see, thanks for the explanation. Let's comment this well in the code.

axelrod/strategies/axelrod_first.py
marcharper: I see, thanks!

axelrod/strategies/axelrod_first.py
drvinceknight: Yeah good call.

I'm struggling to get `axelrod_fortran` to work on my current machine (I blame an OS update), could you or @meatballs **if you get time** paste fingerprints for "k59r".

Something like:

```
import axelrod as axl
import axelrod_fortran as axlf

downing = axlf.Player("k59r")
ashlock_fp = axl.AshlockFingerprint(strategy=downing)
data = ashlock_fp.fingerprint()  # This will take a little while
p = ashlock_fp.plot()
p.savefig("k59r_ashlock_fingerprint.png")

transitive_fp = axl.TransitiveFingerprint(strategy=downing)
data = transitive_fp.fingerprint() 
p = transitive_fp.plot()
p.savefig("k59r_transitive_fingerprint.png")
```

Here are the equivalent for `RevisedDowning`:

Ashlock:

downing_ashlock_fingerprint



Transitive:

downing_transitive_fingerprint

marcharper: They are really similar but not 100% identical

[transative](https://imgur.com/88mYf5b)

[ashlock](https://imgur.com/nxBTUTq)

drvinceknight: Gosh they're incredibly similar though. I'm happy and looking at the history of `RevisedDowning` you implemented it from the Fortran code and I suspect you did it right.

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -2034,7 +2034,7 @@ def test_strategy(self):
 
 class TestSeconodByDowning(TestPlayer):
```
marcharper: Seconod -> Second

Are we sure we don't want to call this one RevisedDowning ? (Here is where the classifier would separate the naming concerns from the inclusion in the first or second tournament...)

axelrod/strategies/axelrod_first.py
```diff
@@ -73,107 +73,181 @@ def strategy(self, opponent: Player) -> Action:
...
+
+    Thus, the player opens with a defection in the first two rounds. Note that
+    from the Axelrod publications alone there is nothing to indicate defections
+    on the first two rounds, although a defection in the opening round is clear.
+    However there is a presentation available at
+    http://www.sci.brooklyn.cuny.edu/~sklar/teaching/f05/alife/notes/azhar-ipd-Oct19th.pdf
+    That clearly states that Downing defected in the first two rounds, thus this
+    is assumed to be the behaviour.
+
+    Note that response to the first round allows us to estimate
```
Nikoleta-v3: I now understand these two sentences but I had to look at the code. I am happy to suggest a change but I have one question before I do.

> we will use the opening play of the player to estimate alpha = P(C_o | C_s)

Why we don't assume that `P(C_o | D_s) = P(C_o | C_s) = 1/2` as stated above?

axelrod/strategies/axelrod_first.py
```diff
@@ -189,10 +189,23 @@ class FirstByDowning(Player):
...
+    "before the match starts".
+
+    So for example, if the plays are:
+
+    [(D, C), (D, C)]
+
+    Then the opponent's first cooperation counts as a cooperation in response to
+    the non existent cooperation of round 0. The total number of cooperations in
+    response to a cooperation is 1. We need to take in to account that extra
+    phantom cooperation to estimate the probability alpha=P(C|C) as 1 / 1 = 1.
```
Nikoleta-v3: I have one question, why we don't start with `alpha=P(C|C) = 0.5` as stated above?

axelrod/strategies/axelrod_first.py
marcharper: Let's keep it and then tweak it in a follow-up PR

axelrod/strategies/axelrod_first.py
drvinceknight: Because that doesn't necessarily always imply 2 defections as an opening which is one of the "stronger" points made in the various literature. (But I'm guessing here.)

axelrod/tests/strategies/test_axelrod_second.py
drvinceknight: Yup, that'll be similar to TitForTat, Grudger etc... I'll change it.

axelrod/strategies/axelrod_first.py
drvinceknight: Fine by me. :+1:

axelrod/strategies/axelrod_first.py
drvinceknight: I'll investigate more and get back to you (and add to the docstring as well) :+1:

axelrod/tests/strategies/test_axelrod_second.py
drvinceknight: a6471e8 renames this and moves it to its own file so that the description at the top of `axelrod_second.py` is still accurate.

axelrod/strategies/axelrod_first.py
drvinceknight: 6c46483 adds to the docstring on this topic and also no the other points you raised. Let me know what you think.

drvinceknight:
b9b00a2 adds a tutorial that just reproduces the first tournament (or fails to). I thought I might as well do this after your suggestion @marcharper, let me know what you think.

FYI, I'm currently running some code to iterate through random seeds to see if we can get the same results as Axelrod reported (no luck so far but it did give me the couple of examples I use in the tutorial). Here is the code I'm using to do that:

```
import axelrod as axl
import pandas as pd
import csv

def get_players():
    first_tournament_participants_ordered_by_reported_rank = [s() for s in axl.axelrod_first_strategies]
    return first_tournament_participants_ordered_by_reported_rank

def obtain_ranked_names(players, seed=0):
    axl.seed(seed)
    tournament = axl.Tournament(players=players, 
                                turns=200, 
                                repetitions=5)  # Axelrod's original tournament ran with 5 repetitions
    results = tournament.play(progress_bar=False)
    return results.ranked_names

def count_matches(ranked_names, ranked_players):
    first_tournament_ranked_names = [str(p) for p in ranked_players]
    return sum(reported == reproduced for reported, reproduced in zip(first_tournament_ranked_names, ranked_names))

def write_data(seed, number, winner, tit_for_tat_rank, filename, mode="a"):
    with open(filename, mode) as f:
        writer = csv.writer(f)
        writer.writerow([seed, number, tit_for_tat_rank, winner])
        
def check_seed(seed, players, filename):
    ranked_names = obtain_ranked_names(players=players, seed=seed)
    number_of_matches = count_matches(ranked_names=ranked_names, ranked_players=players)
    
    write_data(
        seed=seed, 
        number=number_of_matches, 
        tit_for_tat_rank=ranked_names.index("Tit For Tat"),
        winner=ranked_names[0], 
        filename=filename)
    
    return number_of_matches

players = get_players()
number_of_players = len(players)
filename = "seed_search.csv"

try:
    seed = pd.read_csv("seed_search.csv")["seed"].max() + 1
except FileNotFoundError:
    seed = 0
    write_data(seed="seed", 
               number="number", 
               winner="winner", 
               tit_for_tat_rank="tit_for_tat_rank", 
               filename=filename, 
               mode="w",
              )

while check_seed(
    seed=seed, 
    players=players, 
    filename=filename,
) != number_of_players:
    seed += 1
```

axelrod/strategies/axelrod_first.py
```diff
@@ -125,13 +125,16 @@ class FirstByDowning(Player):
...
         EV_TOT = #CC(EV_CC) + #CD(EV_CD) + #DC(EV_DC) + #DD(EV_DD)
 
-    I.E. The player aims to maximise the expected value of being in each state
-    weighted by the number of times we expect to be in that state.
+    This differs from the more modern literature where #CC, #CD, #DC and #DD
+    would imply that counts of both players playing C and C, or the first
+    playing C and the second D etc...
+    In this case the author uses an argument based on the sequence of plays by
+    the player (S) so #CC denotes the number of times the player plays C twice
+    in a row. This is then used to
```
Nikoleta-v3: `to...`

axelrod/strategies/axelrod_first.py
Nikoleta-v3: The explanation is really good now! One minor comment would be to change `P(C | C)` to `P(C_o | C_s)` to be consistent.

drvinceknight: Yup good call.

drvinceknight:
@Nikoleta-v3 could you confirm you're happy with this now when you get a moment?

Nikoleta-v3:
Looks good to me @drvinceknight!

drvinceknight:
@marcharper @meatballs when either of your have time, I believe this is good to go now.

marcharper:
LGTM. @meatballs want to take a final look since there were some post-approval changes?

PR #1282, opened 2020-02-04
Adaptive Zero Determinant Strategy

dashiellfryer:
My recent graduate Emmanuel Estrada and I made a class that adapts its four vector in response to payoff. The player maintains zero determinant bounds each round. In our tests this player eventually outperforms all the standard zero determinant strategies. We tested by using 2000 rounds instead of the standard 200. We look forward to any comments.


axelrod/strategies/adaptivezerodet.py
```diff
@@ -0,0 +1,120 @@
...
+
+	def set_four_vector(self, four_vector: Tuple[float, float, float, float]):
+		# This checks the four vector is usable and allows previous matches' output to be input for next four vector
+		if not all(0 <= p <= 1 for p in four_vector):
+			raise ValueError("An element in the probability vector, {}, is not between 0 and 1.".format(str(four_vector)))
+		self._four_vector = dict(zip([(C, C), (C, D), (D, C), (D, D)], map(float, four_vector)))
+		self.classifier['stochastic'] = any(0 < x < 1 for x in set(four_vector))
+
+	def score_last_round(self, opponent: Player):
+		# This gives the strategy the game attributes and allows the strategy to score itself properly
```
drvinceknight: Could we change this to a docstring, ie use the triple:

```
"""
```

Similarly for our other ones.

axelrod/strategies/adaptivezerodet.py
```diff
@@ -0,0 +1,120 @@
...
+	classifier = {
+        'memory_depth': float('inf'),  # Long memory
+        'stochastic': True,
+        'makes_use_of': set(["game"]),
+        'long_run_time': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+	def __init__(self, phi: float = 0.125, s: float = 0.5, l: float = 3, four_vector: Tuple[float, float, float, float] = None, initial: Action = C) -> None:
```
drvinceknight: Would you be able to line break these so they don't go over 80 characters please?

axelrod/strategies/adaptivezerodet.py
```diff
@@ -0,0 +1,120 @@
+from axelrod.action import Action
```
drvinceknight: Could you reorder these to be:

```
from typing import List, Tuple

```

axelrod/strategies/adaptivezerodet.py
```diff
@@ -0,0 +1,120 @@
...
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+	def __init__(self, phi: float = 0.125, s: float = 0.5, l: float = 3, four_vector: Tuple[float, float, float, float] = None, initial: Action = C) -> None:
+		# This Keeps track of the parameter values (phi,s,l) as well as the four vector which makes final decisions.
+		self.scores = {C: 0, D: 0}
+		self.phi = phi
+		self.s = s
+		self.l = l
```
meatballs: Could you use some more meaningful variable names instead of 's', 'l' and 'd', please?

axelrod/strategies/adaptivezerodet.py
```diff
@@ -0,0 +1,120 @@
...
+			p4 = phi * (1 - s) * (l - P)
+			four_vector = [p1, p2, p3, p4]
+			# Four vector is calculated with new parameters
+			self.set_four_vector(four_vector)
+			if not hasattr(self, "_four_vector"):
+				raise ValueError("_four_vector not yet set")
+			if len(opponent.history) == 0:
+				return self._initial
+			p = self._four_vector[(self.history[-1], opponent.history[-1])]
+			return random_choice(p)
```
meatballs: The code in the two branches of this if/else is very similar. Could it perhaps be refactored into a function controlled by a parameter or two?

docs/reference/all_strategies.rst
```diff
@@ -8,6 +8,8 @@ Here are the docstrings of all the strategies in the library.
 
 .. automodule:: axelrod.strategies.adaptive
    :members:
+.. automodule:: axelrod.strategies.adaptivezerodet
+   :members:
```
drvinceknight: Currently the tests are failing because of the documentation:

```
Warning, treated as error:
406autodoc: failed to import module 'adaptivezerodet' from module 'axelrod.strategies'; the following exception was raised:
407No module named 'axelrod.strategies.adaptivezerodet'
408Makefile:53: recipe for target 'html' failed
409make: *** [html] Error 2
410The command "cd docs; make clean; make html" exited with 2.
```

I believe we can just delete these two lines:

```suggestion

```

axelrod/strategies/adaptivezerodet.py
marcharper: refactored

axelrod/strategies/adaptivezerodet.py
marcharper: unnecessary now

docs/reference/all_strategies.rst
marcharper: Done

axelrod/tests/strategies/test_zero_determinant.py
```diff
@@ -317,3 +317,46 @@ def test_strategy(self):
...
+    expected_classifier = {
+        "memory_depth": float('inf'),
+        "stochastic": True,
+        "makes_use_of": set(['game']),
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
```
drvinceknight: Could we add few tests for the two specific methods: `_adjust_parameters` and `score_last_round` (just running the methods and checking that the attributes have changed accordingly?

axelrod/tests/strategies/test_zero_determinant.py
```diff
@@ -317,3 +317,46 @@ def test_strategy(self):
...
+    def test_strategy(self):
+        R, P, S, T = axelrod.Game().RPST()
+
+        player = self.player(l=R)
+        axelrod.seed(0)
+        match = axelrod.Match((player, axelrod.Alternator()), turns=200)
+        match.play()
```
drvinceknight: This tests need to be tweaked so that we can check expected behaviours. There's information on writing tests here: https://axelrod.readthedocs.io/en/stable/tutorials/contributing/strategy/writing_test_for_the_new_strategy.html

For example this specific paragraph would be something like (although I am not sure what the expected behaviour is):

```suggestion
        expected_actions = [(C,D), (D, C)] * 100 
        self.versus_test(opponent=axelrod.Alternator),
                 expected_actions=expected_actions, seed=0, init_kwargs={"l": R})
```

Note that we could also check that the other attributes have changed accordingly using the `attrs` keyword.

Let me know if I can assist with that (happy to PR the tests to your branch).

axelrod/strategies/zero_determinant.py
```diff
@@ -225,3 +227,77 @@ class ZDSet2(LRPlayer):
 
     def __init__(self, phi: float = 1 / 4, s: float = 0.0, l: float = 2) -> None:
         super().__init__(phi, s, l)
+
+
+class AdaptiveZeroDet(LRPlayer):
+    """A strategy that uses a zero determinant structure that updates
+    its parameters after each round of play.
```
drvinceknight: It would be helpful to have some more information here, perhaps an explanation of the update procedure? It's ok to go quite in depth and include some mathematics which we can then use to check that the `_adjust_parameters` method is working as expected.

axelrod/tests/strategies/test_zero_determinant.py
marcharper: I agree but I didn't want to search for seeds just to have to do so again once #1288 goes in. The current tests make sure that at least each branch is run (so that the limits are bounced into).

drvinceknight: I understand. Are you suggesting this are place holders and we merge #1288 and then fix these?

marcharper: Yes, after we merge this I'll update all the tests that require seeds in #1288 (there are many and most will have to change).

drvinceknight: Sounds good. Happy to take a share of that work if helpful.

marcharper:
Hi all, quick update on this one. There are really four possibly strategies here depending on how one wants to update the underlying parameters each round. It's also not clear how much to update each parameter, so I'd like to run some evolutionary algorithms to try to fit these hyperparameters.

PR #1284, opened 2020-02-11
m

ucd2sjsu:


drvinceknight:
This currently looks like a lot of arbitrary deletions and additions with no description of the contribution. I did notice you have somewhere included a `TitFor3Tats` class. The library already has an `NTitForMTats` that you can use that is here: https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/titfortat.py#L699

There are contribution guidelines available here: https://axelrod.readthedocs.io/en/stable/tutorials/contributing/index.html

I'm going to close this PR as it's not in a state to be reviewed.

PR #1288, opened 2020-03-02
Implement reproducible seeding

### LINKED ISSUE #1181: Possibly move play method away from Player
It might be more natural to put `play` onto the game class.

### LINKED ISSUE #1214: Hypothesis deprecation warnings

### LINKED ISSUE #1277: Setting a seed is ineffective when parallelising
So I just stumbled upon the following. If you run this code snippet twice:

```python
import axelrod as axl
import numpy as np
axl.seed(1)
players = [s() for s in axl.demo_strategies]
tournament = axl.Tournament(
     players=players,
     turns=200,
     repetitions=5
)
results = tournament.play(processes=0)
list(map(np.mean, results.normalised_scores))
```

You will not get the same results. 

`axl.demo_strategies` include `axl.Random` so this is due to the seed not being effective.

If you run the same code but in series multiple times:

```python
import axelrod as axl
import numpy as np
axl.seed(1)
players = [s() for s in axl.demo_strategies]
tournament = axl.Tournament(
     players=players,
     turns=200,
     repetitions=5
)
results = tournament.play()
list(map(np.mean, results.normalised_scores))
```

then you do get the same results.

This is because the `axl.seed(1)` becomes redundant as jobs don't all necessarily finish in the same order and so are not necessarily run in the same order thus offsetting the random sequence.

**Possible fix?** (Not sure)

I can think of one idea towards a fix which involves sampling random integers in the "parent" tournament process and passing those to matches which would each set their own seed.

Modifying the match generator to do something like:

```python
    def build_match_chunks(self):
        """
        A generator that returns player index pairs and match parameters for a
        round robin tournament.

        Yields
        -------
        tuples
            ((player1 index, player2 index), match object)
        """
        if self.edges is None:
            edges = complete_graph(self.players)
        else:
            edges = self.edges

        for index_pair in edges:
            match_params = self.build_single_match_params()
            yield (index_pair, match_params, self.repetitions, random.integer()). # Adding the random integer here

```

and modifying the `Match` to take a seed (and set the seed):

```python
class Match(object):
    """The Match class conducts matches between two players."""

    def __init__(
        self,
        players,
        turns=None,
        prob_end=None,
        game=None,
        deterministic_cache=None,
        noise=0,
        match_attributes=None,
        reset=True,
        seed=seed,
    ):

```

Good points:

- [If this works] This would ensure that there are no differences in results from using multi processes or series.
- If a seed is set (like in the code above) then the random integers generated by the "parent" process would be the same;
- If a seed is not set then the "parent" process would pass "random" seeds to the matches so not reproducible.

Problem/question:

I think that the only way to do this efficiently is to sample the "parent" random numbers in the generator but if I'm not mistaken that will still be affected by the offset seedings (when a match sets a seed that will offset for the parent):

- Can we "copy" the random module so we can have two random sequences on the go? (One for the parent process and the other for the matches?). (We could potentially implement our our Mersenne twister but that sound idiotic)
- Perhaps using `numpy.random.randint()` could be used to sample all random seeds needed "efficiently". This has the downside to needing to know how many matches we want. In practice this isn't a problem, in theory we've tried to keep things so that the generators did not need this information...

**Any other/better ideas?**

Whatever the fix we should include something like the original code snippet as a test. If we can't fix it we should at least document this as a downside to parallelisation...

### LINKED ISSUE #1283: Pandas 1.0.0 breaks the result set.
When running the tests for the result set:

```
python -m unittest 
```

with pandas `0.25.3` everything passes but with `1.0.0` we get errors:

```
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/dask/local.py", line 222, in execute_task
    result = _execute_task(task, data)
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/dask/core.py", line 119, in _execute_task
    return func(*args2)
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/dask/utils.py", line 29, in apply
    return func(*args, **kwargs)
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/dask/dataframe/groupby.py", line 279, in _groupby_aggregate
    return aggfunc(df.groupby(level=levels, sort=False, **dropna), **kwargs)
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/pandas/core/frame.py", line 5807, in groupby
    observed=observed,
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/pandas/core/groupby/groupby.py", line 409, in __init__
    mutated=self.mutated,
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/pandas/core/groupby/grouper.py", line 625, in get_grouper
    if not isinstance(gpr, Grouping)
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/pandas/core/groupby/grouper.py", line 287, in __init__
    ) = index._get_grouper_for_level(self.grouper, level)
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/pandas/core/indexes/multi.py", line 1267, in _get_grouper_for_level
    grouper = level_index.take(codes, fill_value=True)
  File "/Users/vince/anaconda3/envs/axelrod/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 761, in take
    f"Unable to fill values because {cls_name} cannot contain NA"
ValueError: Unable to fill values because Int64Index cannot contain NA
```

This has happened on #1282.

marcharper:
This PR
* adds an instanceable random number generator object `RandomGenerator`, and a bulk generating version `BulkRandomGenerator`
* makes Player, `EvolvablePlayer`, `Match`, `Tournament`, `MoranProcess`, and `Fingerprint` objects seed reproducible; updates every strategy to use a local RNG
* eliminates global seeding throughout the library, but does preserve a module-level seeding operation (mostly for a few tests, maybe can be eliminated)
* moves a few functions around (`play` from `Player` to `Match`, for example), see #1181
* re-writes many tests and adds several others; upgraded our use of hypothesis to 4.0+ while doing so (closes #1214)
* fixes #1277

The solution to #1277 is quite a bit more complex than discussed on the issue, so let me explain it all now so that the code changes will make sense.

The initial issue was the lack of exact reproducibility of multi-threaded tournaments.  Here the crux of the issue is that once we start using threads, a global (seeded) random number generator (RNG) isn't necessarily going to generate values in the same order for all the matches across threads because threads may get pre-empted and so on. The solution discussed at #1277 is to generate seeds in the tournament class from a single initial seed and pass those into the threads so that they can be used to make the output of Matches reproducible.

There are a lot of ins and outs at this point because tournaments and matches can be noisy, metaplayers can contain subplayers, and in principle we might eventually have a Match that is itself multi-threaded, so how we make Matches seed-reproducible isn't so obvious. Since a Match can be noisy it needs access to a seeded RNG, as do the players, which currently call global RNG. So each stochastic player has to be re-worked to use a local or referenced RNG, which they could share with the Match or they could have their own.

If Players have their own RNG they also need to be seeded. That brings up the issue of how to handle resetting and cloning in the case of a seeded Player RNG, whether to initialize `Player`s with a non-seeded RNG, and whether matches, when re-run, should reset all the seeds or not. However, given a seed, a `Match` can generate new seeds or RNGs for the Players and so maintain reproducibility, even if we don't reset any of the seeds in subsequent matches. That reduces the need to create a lot of RNG objects (and we need only generate them in the case that a Match is stochastic).

Finally note that whatever we do has to also work for axelrod-fortran. Those strategies require random values to be passed into the strategy function in each round, so those players also need to use a deterministic RNG.

Ultimately I decided on the following principles:
* Since a player must have an opponent, it is the `Match` that is the logical smallest context of reproducibility; all others derive reproducibility from it. `Players` are seedable, but since a `Match` also needs a seed for noise, the best practice is for matches to subsequently seed players in most cases.
* Players should probably not share RNGs with opponents or clones because it could cause subtle errors
* Cloning a player shouldn't reset its seed or RNG. Rather the clone should have no seed, and if reproducibility via a seed is desired, it's up to the Match or user to supply it
* Repeating an existing Match object need not necessary produce identical results to the last run. Like a player, its seed could be reset by the user if desired, otherwise the old seed can be used to generate a new seed. However creating a new match with identical parameters and the same seed should produce exactly the same sequence of match results.
* Note however that the way we run Tournaments by generating all the matches in `__init__` means that rerunning a tournament will produce the same results. If we don't want that to be the case then we need to generate the matches on each invokation of play(). This could be changed...
* Note further that the order of matches in a tournament still isn't identical, so neither are the result sets because of the reordering, but the list of ranked strategies is the same.
* I decided it could be confusing for the user to be able to supply seeds to Players and Matches because then there has to be options for the Match to not overwrite seeds in that case, and I prefer one and only one obvious way to seed. In this case my preference is that a single seed gets reproducibility for a Tournament, Match, or Moran Process, and I couldn't think of a good use case where one would need to control all the seeds.
* A Player doesn't need a seed until it's time to play. There's not a compelling reason to generate a seed on initialization, for example, but the set_seed function can be used for testing purposes. The exception is the Evolvable players, which currently randomize on initialization (and can also now be seeded)
* The players' seeds don't belong in `match_attributes` because they are not the same for all players like the other match attributes

Then the following implementation details follow:
* Since players, matches, tests, etc need access to local-ish RNG, it was necessary to wrap the existing contents of random_.py into an object that can be seeded and instanced, removing the global RNG
* In the case where a `Match` is not supplied a seed, a random seed is used, mimicking the current behavior (with no guarantee of reproducibility)
* Matches then generate seeds to pass to the players who can then generate their own RNG
* It no longer makes sense for a Player to have a play function since a stochastic player or noisy `Match` must have a local RNG for reproducibility. The `Player.play` function was applying noise so I moved it to `Match`, and updated all uses of `Player.play`
* `Player.__eq__` ignores the RNG / seed . This mimicks the current behavior where the global seed had to be set for a player and a clone to behave identically. Now a local seed has to be set for both (e.g. via the Match seed) for reproducibility to be valid.
* Seeds are set _on_ players rather than in the `__init__` kwargs, so the seed is not naturally reused on cloning or resetting. I toyed with the alternative -- it's possible to generate seeds on the fly with `__new__` but it's hacky and leads to a lot of seeds that get overwritten or are unecessarily generated (e.g. for deterministic players), makes a bunch of tests harder to reason about, and so on.
 
Note that the order of players in a `Match` can affect the outcome (since the seeds generated by the `Match` would be reversed if the order was reversed). The other parameters like noise of course also must be the same.

Consequences
* There is now more locally stable testing with seeds, but new seeds have to be chosen for many tests, which are still broken. I'll fix them once we're happy with the rest of the changes
* In practice this actually simplifies of the use of RNG throughout the library
* Now we rely only on numpy seeding rather than numpy and stdlib
* We'll need to document this all clearly

Some things still to do:
* make axelrod-fortran players seedable (can reuse the new RNG objects)
* make evolvable players and evolutionary algorithms properly reproducible (also uses threading)

------------------------

Fixes #1277
Fixes #1283 
Fixes #1181 
Fixes #1214


drvinceknight:
Wow, thanks a lot for working on this @marcharper it looks like it was a chunk of work to get to the bottom of.

I've read your explanation/solution proposal and it all looks excellent to me. I don't think this only fixes #1277 but it also makes the whole seeding much nicer from a user point of view as well . Let me know if you'd like to share out some of the remaining work (or if I can help in any other way).

marcharper:
I sat down thinking it wouldn't be that difficult and then it keep growing. I can handle the rest of the changes -- there are a few thorny cases to handle like making some of the transformers play nicely with deterministic (non-seeded players) but probably easier for me to tackle now that I'm pretty far in.

Do you have any opinions on the names of the new classes or variables? I used `self._random` for the `Player` local RNGs, but maybe just `self._rng` is better.

Perhaps you can take a look at the tests and look for more hypothesis opportunities or entirely new tests (maybe after this PR is finished).

The axelrod-fortran update should be easy, just need to reuse the new RNG classes. Folding in the evolutionary algorithms should happen soon after and then I think we're in good shape.

drvinceknight:
> Do you have any opinions on the names of the new classes or variables? I used `self._random` for the `Player` local RNGs, but maybe just `self._rng` is better.

I am happy with the names you got so far. Forgive me in advance if I make suggestions for changing one or two when I review more carefully but I don't think that'll happen and I doubt I'll feel strongly. In this specific case I think the only preference I'd have is `self._random_number_generator` but I don't feel strongly at all. Your call.

> Perhaps you can take a look at the tests and look for more hypothesis opportunities or entirely new tests (maybe after this PR is finished).

Yeah good idea. I'll pick that up afterwards.

gaffney2010:
> Note however that the way we run Tournaments by generating all the matches in init means that rerunning a tournament will produce the same results. If we don't want that to be the case then we need to generate the matches on each invokation of play().

Is this so?  The way I'm reading it, both _run_serial and _run_parallel start by building chunks using self.match_generator, which should have the seed in the state you left it at the previous run.  Am I misunderstanding?

marcharper:
I think you're right, we'd need to have match generator reuse its seed if we want the second run to match the prior run. But running two tournaments in row (e.g. calling play() twice) with different Tournament instances (and the same seed) should generate the same pair of results since the match generator is exhausted by Tournament before batching out matches.

drvinceknight:
Making a note here to make sure we don't forget to modify the documentation if https://github.com/Axelrod-Python/Axelrod/pull/1329 goes in before. 

I'd recommend we still give hypothesis an upper bound as in the past they've developed things so quickly it's often broken things but at least the new cron CI builds would catch that.

marcharper:
Update: rebased and many tests fixed. Files still containing failing tests:
```
test_moran
test_strategy_transformers
test_axelrod_second
test_calculator
test_cycler
test_gambler
test_hmm
test_memoryone
test_meta
test_mindcontrol
test_punisher
test_qlearner
test_selfsteem
test_stalker
test_titfortat
test_worse_and_worse
test_zero_determinant
```

drvinceknight:
This is a big piece of work :+1: Let me know if I can assist in any way.

peteroupc:
Let me give you another idea.

I am aware that the problem of thread schedulers and reproducible "randomness" has been tackled by using a different design from traditional PRNGs.  Notably, so-called "splittable" PRNGs have been developed for use in "isolated parallel computations that may generate subtasks", such as event processing that could be done by individual threads.  See also https://github.com/freeorion/freeorion/issues/2856#issuecomment-616026536, the JAX [PRNG Design Notes](https://github.com/google/jax/blob/master/design_notes/prng.md), and Java's [SplittableRandom](https://docs.oracle.com/javase/9/docs/api/java/util/SplittableRandom.html) documentation.

marcharper:
Hi @peteroupc thanks for the links. I'm a fan of Jax :)

IIUC what we're doing here is quite similar in that each task ends up with it's own PRNG (in fact each match and player, though they are all deterministically linked in a tree depending on the splitting path), so in this case I'm not sure that we gain much from using splittable PRNGs. Since we're going for 100% reproducibility, we'll still have the same issue in that we would need to split out all the PRNGs initially just as we're generating seeds here because we can't guarantee that the order of dispatch will be the same in the distributed case. Am I missing another advantage of using splittable PRNGs that's relevant?

drvinceknight:
> marcharper added the ready-for-review  label yesterday

Huge piece of work @marcharper 👏 👏. I've set aside tomorrow (Tuesday) to review :+1:

.github/workflows/config.yml
```diff
@@ -37,7 +36,6 @@ jobs:
     - name: Run tests
       run: |
         python -m pip install coverage
```
drvinceknight, line 38: We have `hypothesis` in `requirements.txt` (and specifically excluded from `setup.py`: https://github.com/Axelrod-Python/Axelrod/blob/master/setup.py#L4). Should we do the same with `coverage`, `mypy`, `isort`? (Just do be consistent.)

(Happy for this to be done in a later PR - just a thought.)

axelrod/__init__.py
```diff
@@ -3,10 +3,18 @@
...
 from axelrod.version import __version__
+from axelrod.action import Action
+from axelrod.random_ import Pdf, RandomGenerator, BulkRandomGenerator
+
+# Initialize module level Random
+# This is initially seeded by the clock / OS entropy pool
+# It is not used if user specifies seeds everywhere and should only be
+# used internally by the library and in certain tests that need to set
+# its seed.
+_module_random = RandomGenerator()
```
drvinceknight, line 14: What's the rational of calling this `_module_random` ?  Would `_random_generator` or `_runtime_rng` be a bit more descriptive? (Working through `test_filtering` and was not immediately sure what `_module_random` was.)

axelrod/tests/strategies/test_apavlov.py
```diff
@@ -98,64 +111,72 @@ class TestAPavlov2011(TestPlayer):
...
         )
 
+    def test_strategy_random(self):
         opponent = axl.MockPlayer(actions=[C, C, C, C, D, D])
         actions = [(C, C), (C, C), (C, C), (C, C), (C, D), (D, D), (D, C), (D, C)]
         self.versus_test(
             opponent, expected_actions=actions, attrs={"opponent_class": "Random"}
         )
 
+    def test_strategy_random2(self):
```
drvinceknight, line 179: All these small minor cleanups you made along the way are awesome @marcharper :+1:

axelrod/tests/strategies/test_calculator.py
```diff
@@ -22,81 +22,45 @@ class TestCalculator(TestPlayer):
         "manipulates_state": False,
     }
 
-    def test_twenty_rounds_joss_then_defects_for_cyclers(self):
+    def test_twenty_rounds_joss_for_cyclers(self):
         """Uses axelrod.strategies.axelrod_first.Joss strategy for first 20 rounds"""
```
drvinceknight, line 26:
```suggestion
        """Uses axelrod.strategies.axelrod_first.FirstByJoss strategy for first 20 rounds"""
```

axelrod/tests/strategies/test_calculator.py
```diff
@@ -22,81 +22,45 @@ class TestCalculator(TestPlayer):
...
-            axl.Alternator(), expected_actions=twenty_test_actions, seed=seed
-        )
-        self.versus_test(
-            axl.Alternator(), expected_actions=expected_actions, seed=seed
+            axl.Alternator(), expected_actions=match.result, seed=seed
         )
 
-    def test_twenty_rounds_joss_then_tit_for_tat_for_non_cyclers(self):
+    def test_twenty_rounds_joss_then_defects_for_cyclers(self):
         """Uses axelrod.strategies.axelrod_first.Joss strategy for first 20 rounds"""
```
drvinceknight, line 35:
```suggestion
        """Uses axelrod.strategies.axelrod_first.FirstByJoss strategy for first 20 rounds"""
```

axelrod/tests/strategies/test_calculator.py
```diff
@@ -22,81 +22,45 @@ class TestCalculator(TestPlayer):
...
-            (C, D),
-            (D, C),
-            (C, D),
-            (D, C),
-        ]
-
-        opponent_actions = twenty_non_cyclical_actions + subsequent_opponent_actions
-        test_actions = twenty_test_actions + subsequent_test_actions
+    def test_twenty_rounds_joss_for_noncyclers(self):
+        """Uses axelrod.strategies.axelrod_first.Joss strategy for first 20 rounds"""
```
drvinceknight, line 45:
```suggestion
        """Uses axelrod.strategies.axelrod_first.FirstByJoss strategy for first 20 rounds"""
```

axelrod/tests/strategies/test_calculator.py
```diff
@@ -22,81 +22,45 @@ class TestCalculator(TestPlayer):
...
+        match.play()
         self.versus_test(
-            axl.MockPlayer(actions=twenty_non_cyclical_actions),
-            expected_actions=twenty_test_actions,
-            seed=seed,
+            axl.AntiCycler(), expected_actions=match.result, seed=seed
         )
+
+    def test_twenty_rounds_joss_then_tft_for_noncyclers(self):
+        """Uses axelrod.strategies.axelrod_first.Joss strategy for first 20 rounds"""
```
drvinceknight, line 54:
```suggestion
        """Uses axelrod.strategies.axelrod_first.FirstByJoss strategy for first 20 rounds"""
```

axelrod/tests/strategies/test_cycler.py
```diff
@@ -12,6 +12,7 @@
 from .test_player import TestPlayer
 
 C, D = Action.C, Action.D
+random = axl.RandomGenerator()
```
drvinceknight, line 15: Why are we doing this instead of using the runtime `_module_random` (created in `axelrod/__init__.py`) which is what is used in `test_filtering.py`?

axelrod/tests/strategies/test_player.py
```diff
@@ -646,28 +627,53 @@ def versus_test(
...
+        for i, (play, expected_play) in enumerate(zip(player2.history, expected_actions2)):
+            self.assertEqual((i, play), (i, expected_play))
+
+        # Test final player attributes are as expected
+        if attrs:
+            for attr, value in attrs.items():
+                self.assertEqual(getattr(player1, attr), value)
+
+    def search_seeds(self, *args, **kwargs):  # pragma: no cover
+        """Search for a seed that will pass the test. To use to find a new seed
```
drvinceknight, line 659:
```suggestion
        """Search for a seed that will pass the test. Use to find a new seed
```

axelrod/tests/strategies/test_player.py
```diff
@@ -646,28 +627,53 @@ def versus_test(
...
+        for a versus_test, change self.versus_test to self.search_seeds
+        within a TestPlayer or TestMatch class.
```
drvinceknight, line 661: Nice addition. Could we document this a bit more, one of the two options perhaps:

1. Add an example to the docstring.
2. Specifically add a discussion of this to the contributions docs

Certainly doesn't have to be done in this PR: we can open an issue and do that later :)

axelrod/tests/strategies/test_human.py
```diff
@@ -130,3 +130,6 @@ def test_repr(self):
 
     def equality_of_players_test(self, p1, p2, seed, opponent):
         return True
+
+    def test_reproducibility_of_play(self):
```
drvinceknight, line 134: Is this a method in the parent class that needs to be overwritten? (Have looked there but don't seem able to find it.)

axelrod/tests/strategies/test_gambler.py
```diff
@@ -14,6 +13,7 @@
 
 tables = load_pso_tables("pso_gambler.csv", directory="data")
 C, D = axl.Action.C, axl.Action.D
+random = axl.RandomGenerator()
```
drvinceknight, line 16: Same comment here about using `axl._module_random` (or not)?

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -250,3 +252,55 @@ def test_strategy(self):
             expected_actions=actions,
             attrs={"play_as": "ALLD", "shift_counter": -1, "alld_counter": 2},
         )
+
+
+class TestMemoryTwoCooperator(TestCooperator):
+    """Cooperator is equivalent to MemoryOnePlayer((1, 1, 1, 1), C)"""
```
drvinceknight, line 258:
```suggestion
    """Cooperator is equivalent to MemoryTwoPlayer((1, 1, ..., 1), C)"""
```

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -250,3 +252,55 @@ def test_strategy(self):
...
+        "makes_use_of": set(),
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+
+class TestMemoryTwoDefector(TestDefector):
+    """Defector is equivalent to MemoryOnePlayer((0, 0, 0, 0), D)"""
```
drvinceknight, line 273:
```suggestion
    """Defector is equivalent to MemoryTwoPlayer((0, 0, ..., 0), D)"""
```

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -250,3 +252,55 @@ def test_strategy(self):
...
+
+
+def four_vector_to_sixteen_vector(four_vector):
+    a, b, c, d = four_vector
+    sixteen_vector = [a, b, a, b, d, c, d, c] * 2
+    return sixteen_vector
+
+
+class TestMemoryTwoAlternator(TestAlternator):
+    """Alternator is equivalent to MemoryOnePlayer((0, 0, 1, 1), C)"""
```
drvinceknight, line 294:
```suggestion
    """Alternator is equivalent to MemoryTwoPlayer((0, 0, 0, 0, 1, 1, 1, 1, 0, 0, ... 1), C)"""
```

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -250,3 +252,55 @@ def test_strategy(self):
...
+    name = "Generic Memory Two Player: [0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 1, 1, 1, 1], (C, D)"
+    player = lambda x: axl.MemoryTwoPlayer(sixteen_vector=four_vector_to_sixteen_vector((0, 0, 1, 1)), initial=(C, D))
+    # player = lambda x: axl.MemoryTwoPlayer(sixteen_vector=four_vector_to_sixteen_vector((1, 1, 0, 0)), initial=(C, D))
```
drvinceknight, line 297:
```suggestion
```

axelrod/tests/strategies/test_meta.py
```diff
@@ -148,6 +152,42 @@ def test_strategy(self):
...
+        player = axl.MetaWinnerEnsemble(team=team)
+        self.assertTrue(Classifiers["stochastic"](player))
+
+        # Multiplayer teams without repetition are always stochastic
+        team = [axl.Cooperator, axl.Defector]
+        player = axl.MetaWinnerEnsemble(team=team)
+        self.assertTrue(Classifiers["stochastic"](player))
+
+        # If the players are all identical, a multiplayer
+        # team might is in fact deterministic, even though random values
```
drvinceknight, line 184:
```suggestion
        # team might in fact be deterministic, even though random values
```

axelrod/tests/strategies/test_meta.py
```diff
@@ -148,6 +152,42 @@ def test_strategy(self):
...
+        # are being drawn.
```
drvinceknight, line 185: On a more general note: is this wanted behaviour? Not to be done in this PR, but we could add a check to the meta player for this?

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -678,7 +678,7 @@ def test_output_from_literature(self):
         expected_scores = [
             3000.0,
             915.0,
-            2763.0,
```
drvinceknight, line 681: What's happening here? I believe these expected scores are actually from the literature?

drvinceknight: Ignore that, read closer and see that it's just the random opponent (I panicked 😆 ).

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -1056,6 +1015,51 @@ def test_varying_memory_depth(self):
...
+        "makes_use_of": set(),
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+
+class Test1TitsFor2TatsIsTF2T(TestTitFor2Tats):
+    """Tests that for N = 1 = M, all the TFT tests are passed."""
```
drvinceknight, line 1034:
```suggestion
    """Tests that for N = 1,  M = 2, all the TF2T tests are passed."""
```

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -1056,6 +1015,51 @@ def test_varying_memory_depth(self):
...
+        "makes_use_of": set(),
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+
+class Test2TitsFor1TatsIsT2FT(TestTwoTitsForTat):
+    """Tests that for N = 1 = M, all the TFT tests are passed."""
```
drvinceknight, line 1049:
```suggestion
    """Tests that for N = 2,  M = 1, all the 2TFT tests are passed."""
```

axelrod/tests/strategies/test_titfortat.py
```diff
@@ -1056,6 +1015,51 @@ def test_varying_memory_depth(self):
...
+        "stochastic": False,
+        "makes_use_of": set(),
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+
+class Test2TitsFor1TatsIsT2FT(TestTwoTitsForTat):
```
drvinceknight, line 1048:
```suggestion
class Test2TitsFor1TatsIs2TFT(TestTwoTitsForTat):
```

.github/workflows/config.yml
marcharper: Sounds good. It would be good to also have a "presubmit" script to run isort, mypy, etc. locally in line with config.yml.

axelrod/__init__.py
marcharper: I didn't think that hard about the name, we can change it. The idea was that it's a "module level" rng and so reproducibility wouldn't be guaranteed. But I think we could also probably eliminate it altogether, if we're willing to toss a few more warnings / exceptions at the user. The biggest issue in that case is instantiating evolvable players with insufficient parameters (for random initialization).

.github/workflows/config.yml
drvinceknight: Good idea.

axelrod/tests/strategies/test_cycler.py
marcharper: Basically I don't want to use `_module_random` if at all possible. Since I worked on the PR over quite some time, it's probably just inconsistent application of that principle. I'll look into it.

axelrod/tests/strategies/test_player.py
marcharper: I was thinking that we'd update the library to Python 3.7 + 3.8 or start using `attrs`. Then I can add back in the [code](https://github.com/Axelrod-Python/Axelrod/pull/1288/commits/5128e4d8f6762986e5449564349d8874f7a54802) I used to watch `versus_test`, record all the matches, and automatically search for new seeds using this function, which could all be documented in the contributions section.

axelrod/tests/strategies/test_human.py
marcharper: Looks like maybe I renamed it to `test_clone_reproducible_play` (which has an exception for `Human`) so we can delete this

axelrod/tests/strategies/test_cycler.py
drvinceknight: Cool, I thought it might be something like that. I think I didn't comment on all of them.

axelrod/tests/strategies/test_player.py
drvinceknight: Sounds great!

axelrod/tests/strategies/test_gambler.py
marcharper: Same answer. I don't want there to be any "magic" in how the RNGs work to prevent any stealthy reproducibility issues from creeping in, so being explicit here instead of relying on the module random is better.

axelrod/tests/strategies/test_meta.py
marcharper: It's a weird edge case for sure but it makes sense I guess.

axelrod/tests/strategies/test_titfortat.py
marcharper: Oh good catch, I think I got this right in the documentation test. Let me try to reconcile.

axelrod/__init__.py
marcharper: Turns out we can remove the need for a seed from the current filter test and it still works.

axelrod/tests/strategies/test_meta.py
drvinceknight: Yeah I see that :+1:

.github/workflows/config.yml
marcharper: Opened #1360

axelrod/tests/strategies/test_player.py
marcharper: Opened  #1361

marcharper:
Note that I'm assuming we'll run `black` after this goes in, so I didn't fuss about line lengths and other minor formatting issues.

axelrod/tests/strategies/test_meta.py
marcharper: I also made changes to the strategy code to not actually draw any random numbers in this case. `Match` will now only set a seed if a `Player` needs it, determined by whether the strategy is classified as stochastic or not. There are many changes to get the classifiers right, and if it's wrong, there should be a warning or exception thrown as a strategy attempts to use `self._random` when it doesn't exist.

axelrod/player.py
```diff
@@ -122,6 +186,17 @@ def set_match_attributes(self, length=-1, game=None, noise=0):
         self.match_attributes = {"length": length, "game": game, "noise": noise}
         self.receive_match_attributes()
 
+    def set_seed(self, seed):
+        """Set a random seed for the player's random number generator."""
+        if seed is None:
+            warnings.warn(
+                "Initializing player with seed from Axelrod module random number generator."
+                " Results may not be seed reproducible.")
```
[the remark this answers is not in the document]
marcharper: I think we need a space here because the strings get concatenated, so the sentences will run together otherwise. I'll move it to the end of the first string so it doesn't look weird.

axelrod/tests/strategies/test_player.py
```diff
@@ -467,38 +458,34 @@ def test_reset_clone(self):
...
-                axl.seed(seed)
-                m = axl.Match((p, op), turns=turns)
-                m.play()
-            self.assertEqual(len(player1.history), turns)
-            self.assertEqual(player1.history, player2.history)
+            player = self.player()
+            player_clone = player.clone()
+            op = op.clone()
+            op_clone = op.clone()
+            m1 = axl.Match((player, op), turns=turns, seed=seed, noise=noise/100.)
```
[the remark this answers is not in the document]
marcharper: it's an old school python trick :)

axelrod/tests/strategies/test_memorytwo.py
```diff
@@ -83,7 +85,7 @@ class TestMemoryStochastic(TestPlayer):
     )
     player = axl.MemoryTwoPlayer
     expected_classifier = {
-        "memory_depth": 2,  # Memory-two Sixteen-Vector
+        "memory_depth": 0,  # Memory-two Sixteen-Vector
```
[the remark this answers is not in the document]
marcharper: Generically it's depth two, but the default `sixteen_vector` is `[1, 1, ..., 1]`, which as an edge case has depth zero.

axelrod/tests/strategies/test_worse_and_worse.py
```diff
@@ -25,14 +25,15 @@ def test_strategy(self):
         """Test that the strategy gives expected behaviour."""
         # 6 Rounds Cooperate given seed
```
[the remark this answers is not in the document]
marcharper: indeed, removed

axelrod/tests/strategies/test_memorytwo.py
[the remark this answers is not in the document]
Nikoleta-v3: Makes sense, thank you!

axelrod/tests/strategies/test_meta.py
drvinceknight: Nice.

axelrod/tests/unit/test_match.py
```diff
@@ -79,16 +80,15 @@ def test_example_prob_end(self):
         outcomes
         """
         p1, p2 = axl.Cooperator(), axl.Cooperator()
-        match = axl.Match((p1, p2), prob_end=0.5)
-        expected_lengths = [3, 1, 5]
+        expected_lengths = [2, 1, 1]
         for seed, expected_length in zip(range(3), expected_lengths):
-            axl.seed(seed)
+            match = axl.Match((p1, p2), prob_end=0.5, seed=seed)
```
drvinceknight, line 85: Commentary/thinking out loud: This change points out one of the major "fundamental" [great] changes this PR does right? It's confirming that Matches are now also defined by their seeds (so the test is now forced to create a new match for each seed). Cool.

axelrod/tests/unit/test_match.py
```diff
@@ -117,11 +117,8 @@ def test_len_error(self):
         with self.assertRaises(TypeError):
             len(match)
 
-    @given(p=floats(min_value=0, max_value=1))
+    @given(p=floats(min_value=1e-10, max_value=1-1e-10))
```
drvinceknight, line 120: Fancy :)

axelrod/tests/unit/test_match.py
```diff
@@ -360,16 +354,17 @@ def test_sparklines(self):
...
             (1, 0.5, 1),
-            (2, 0.6, 4),
-            (3, 0.4, 1),
+            (2, 0.6, 1),
+            (3, 0.4, 2),
         ]:
-            axl.seed(seed)
-            self.assertEqual(axl.match.sample_length(prob_end), expected_length)
+            rg = RandomGenerator(seed)
+            r = rg.random()
```
drvinceknight, line 363:
```suggestion
            random = RandomGenerator(seed)
            r = random.random()
```

To be consistent with changes elsewhere? 

I'd actually suggest `rag = RandomerGenerator(seed)` and using that elsewhere. 

Not at all vital to change: whatever you prefer @marcharper.

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -121,14 +122,6 @@ def test_StrategyReBuilder_many_decorators(self):
         new_player.__dict__.update(update_dict)
         self.assertEqual(player, new_player)
 
-    def test_all_strategies(self):
```
drvinceknight, line 124: I believe this was added as a regression test because of a bug we'd missed and wanted to make sure it didn't come back with a new strategy. Sure it's worth getting rid of?

axelrod/tests/unit/test_tournament.py
```diff
@@ -745,6 +731,55 @@ def test_write_to_csv_without_results(self):
         expected_df = pd.read_csv(axl_filename(path))
         self.assertTrue(df.equals(expected_df))
 
+    @given(seed=integers(min_value=1, max_value=4294967295))
+    @settings(max_examples=5, deadline=None)
+    def test_seeding_equality(self, seed):
+        """Tests that a tournament with a given seed will return the
+        same results each time."""
```
drvinceknight, line 738:
```suggestion
        """Tests that a tournament with a given seed will return the
        same results each time. This specifically checks when running using multiple cores so as to confirm that https://github.com/Axelrod-Python/Axelrod/issues/1277 is fixed."""
```

(Or something like that?)

axelrod/tests/unit/test_tournament.py
```diff
@@ -745,6 +731,55 @@ def test_write_to_csv_without_results(self):
...
+            name=self.test_name,
+            players=players,
+            game=self.game,
+            turns=3,
+            repetitions=100,
+            seed=seed
+        )
+        results1 = tournament1.play(processes=2)
+        results2 = tournament2.play(processes=2)
+        self.assertEqual(results1.ranked_names, results2.ranked_names)
```
drvinceknight, line 759: Could we pop this in a loop so as to repeat the assertion a few more times? Or perhaps we run 10 tournaments and not just 2?

axelrod/tests/unit/test_tournament.py
```diff
@@ -745,6 +731,55 @@ def test_write_to_csv_without_results(self):
...
+    def test_seeding_equality(self, seed):
+        """Tests that a tournament with a given seed will return the
+        same results each time."""
+        rng = axl.RandomGenerator(seed=seed)
+        players = [axl.Random(rng.random()) for _ in range(8)]
+        tournament1 = axl.Tournament(
+            name=self.test_name,
+            players=players,
+            game=self.game,
+            turns=3,
```
drvinceknight, line 745: Perhaps a few more turns? `20`? (If I'm being over cautious just tell me...)

drvinceknight:
That's all I'm going to have time to do today I'm afraid. I'll be getting back to this on Friday when I'll go over the implemented functionality itself (I might circle back to the tests/docs): really impressive work @marcharper :) :+1:

axelrod/tests/unit/test_match.py
marcharper: That's right. A seed isn't required if the match is deterministic, which means that it's not noisy and both the players are deterministic. Otherwise a seed is required.

axelrod/tests/unit/test_match.py
marcharper: Sure. I've been using the `RandomGenerator` object as a drop-in replacement for the standard `random` module, but it's not quite the same.

axelrod/tests/unit/test_strategy_transformers.py
marcharper: This test was moved into the class that tests the Identity transformer. I did some pretty extensive rearranging of this file. We could consider having a `TestTransformer` class like `TestPlayer` that attempts to decorate every strategy as well, but I've covered a great many unusual cases -- composed transformers, transformed meta strategies, etc. -- elsewhere in this file.

axelrod/tests/unit/test_tournament.py
marcharper: Yeah we could do something like that, or make better use of hypothesis here. I'm open to suggestions, but we also probably want to be careful to avoid a long runtime test.

axelrod/random_.py
```diff
@@ -1,94 +1,147 @@
...
+        -------
+        axelrod.Action
+        """
+        if self.random_choice(threshold) == C:
+            return action.flip()
+        return action
 
+    def randrange(self, a: int, b: int) -> int:
+        """Python 2 / 3 compatible randrange. Returns a random integer uniformly
+        between a and b (inclusive)"""
```
drvinceknight, line 88:
```suggestion
        """Returns a random integer uniformly
        between a and b (inclusive)"""
```

axelrod/tests/unit/test_tournament.py
drvinceknight: Yeah runtime is vital. I'd suggest just throwing it in a loop would be sufficient.

```
for _ in range(4):
     ...
```

Is all I'm thinking.

axelrod/tests/unit/test_strategy_transformers.py
drvinceknight: Great. Thanks for clarifying :+1:

axelrod/evolvable_player.py
```diff
@@ -22,15 +21,26 @@ class EvolvablePlayer(Player):
...
     def overwrite_init_kwargs(self, **kwargs):
         """Use to overwrite parameters for proper cloning and testing."""
         for k, v in kwargs.items():
             self.init_kwargs[k] = v
 
     def create_new(self, **kwargs):
-        """Creates a new variant with parameters overwritten by kwargs."""
+        """Creates a new variant with parameters overwritten by kwargs. This differs from
+        cloning the Player because it propagates a seed forward, and is intended to be
+        used in my the mutation and crossover methods."""
```
marcharper, line 38: Self review: typo "in my" --> "by"

axelrod/player.py
```diff
@@ -59,14 +108,25 @@ def init_params(cls, *args, **kwargs):
...
         self._history = History()
         self.classifier = copy.deepcopy(self.classifier)
         self.set_match_attributes()
 
+    def _post_init(self):
+        """Post initialization tasks such as reclassifying the strategy."""
+        pass
+
+    def _post_transform(self):
+        """Overwrite this function if you need to do a post-transform, post-init
```
marcharper, line 121: Change docstring, no need to override this function now

axelrod/strategies/finite_state_machines.py
```diff
@@ -117,7 +114,8 @@ def __init__(
         initial_state: int = 1,
         initial_action: Action = C
     ) -> None:
-        super().__init__()
+        # super().__init__()
```
marcharper, line 117: remove commented line

axelrod/strategies/meta.py
```diff
@@ -92,8 +93,22 @@ def __repr__(self):
...
+                player.update_history(play, coplay)
+        except TypeError:
+            # If the Meta class is decorated by the Joss-Ann transformer,
+            # such that the decorated class is now deterministic, the underlying
+            # strategy isn't called. In that case, updating the history of all the
+            # team members doesn't matter.
+            # As a sanity check, look for at least one reclassifier, otherwise
+            # this try-except clause could hide a bug.
+            if len(self._reclassifiers) == 0:
+                raise TypeError("MetaClass update_histories issue, expected a transformer.")
```
marcharper, line 107: transformer --> reclassifier

axelrod/strategies/meta.py
```diff
@@ -208,13 +223,35 @@ class MetaWinnerEnsemble(MetaWinner):
 
     name = "Meta Winner Ensemble"
 
+    def __init__(self, team=None):
```
marcharper, line 226: Is this `__init__` doing anything? looks like it can be removed

axelrod/strategies/prober.py
```diff
@@ -341,7 +339,11 @@ def strategy(self, opponent: Player) -> Action:
         if opponent.history[-1] == D:
             return D
         # Otherwise cooperate, defect with probability 1 - self.p
-        choice = random_choice(1 - self.p)
+        if self.p == 0:
```
marcharper, line 342: Maybe add a helper function to `Player` for this common pattern

axelrod/strategies/rand.py
```diff
@@ -38,8 +39,24 @@ def __init__(self, p: float = 0.5) -> None:
...
         self.p = p
-        if p in [0, 1]:
-            self.classifier["stochastic"] = False
 
     def strategy(self, opponent: Player) -> Action:
-        return random_choice(self.p)
+        return self._random.random_choice(self.p)
+
+    def _post_init(self):
+        if self.p in [0, 1]:
```
marcharper, line 47: add `super()._post_init` call

axelrod/tests/unit/test_match.py
marcharper: I changed it to `rng` for now. We can decide on a universal approach throughout pending your documentation effort.

axelrod/strategies/memorytwo.py
```diff
@@ -92,17 +90,50 @@ def set_sixteen_vector(self, sixteen_vector: Tuple):
         self._sixteen_vector = dict(
             zip(states, sixteen_vector)
         )  # type: Dict[tuple, float]
-        self.classifier["stochastic"] = any(0 < x < 1 for x in set(sixteen_vector))
+
+    @staticmethod
+    def compute_memory_depth(sixteen_vector):
```
Nikoleta-v3, line 95: Type hinting

axelrod/tests/property.py
```diff
@@ -144,7 +153,9 @@ def prob_end_tournaments(
     min_repetitions : integer
         The minimum number of repetitions
     max_repetitions : integer
-        The maximum number of repetitions
+        The maximum number of
```
Nikoleta-v3, line 156:
```suggestion
        The maximum number of repetitions
```

axelrod/tests/strategies/test_evolvable_player.py
```diff
@@ -91,30 +97,29 @@ def test_mutate_variations(self):
         """Generate many variations to test that mutate produces different strategies."""
         if not self.init_parameters:
             return
-        axl.seed(100)
         variants_produced = False
-        for _ in range(2, 400):
-            player = self.player()
+        for seed in range(2, 400):
```
Nikoleta-v3, line 101:
```suggestion
        for seed_ in range(2, 400):
```

axelrod/tests/strategies/test_evolvable_player.py
```diff
@@ -91,30 +97,29 @@ def test_mutate_variations(self):
...
+            player = self.player(seed=seed)
```
Nikoleta-v3, line 102:
```suggestion
            player = self.player(seed=seed_)
```

axelrod/tests/unit/test_fingerprint.py
```diff
@@ -269,135 +267,156 @@ def test_plot_figure(self):
...
     def test_pair_fingerprints(self, strategy_pair):
         """
         A test to check that we can fingerprint
         with any two given strategies or instances
         """
-        strategy, probe = strategy_pair
-        af = AshlockFingerprint(strategy, probe)
-        data = af.fingerprint(turns=2, repetitions=2, step=0.5, progress_bar=False)
-        self.assertIsInstance(data, dict)
+        # strategy, probe = strategy_pair
```
Nikoleta-v3, line 384: why are these commented out @marcharper?

axelrod/tests/strategies/test_averagecopier.py
```diff
@@ -109,15 +117,17 @@ class TestNiceAverageCopier(TestPlayer):
         "manipulates_state": False,
     }
 
-    def test_strategy(self):
+    def test_cooperate_if_opponent_always_cooperates(self):
         # Tests that if opponent has played all C then player chooses C.
```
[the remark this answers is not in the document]
Nikoleta-v3: In the tests above you have changed these comments to docstrings

axelrod/tests/strategies/test_axelrod_second.py
```diff
@@ -23,25 +20,20 @@ class TestChampion(TestPlayer):
         "manipulates_state": False,
     }
 
-    def test_strategy(self):
+    def test_initial_rounds(self):
         # Cooperates for first 10 rounds
```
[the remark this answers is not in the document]
Nikoleta-v3: The comment is repeated in the following line.

axelrod/tests/unit/test_fingerprint.py
marcharper: They shouldn't be, it was temporary while debugging

docs/tutorials/further_topics/approximate_moran_processes.rst
```diff
@@ -19,19 +19,19 @@ the cache is built by passing counter objects of outcomes::
...
+`ApproximateMoranProcess` may not always match the results of a standard `MoranProcess`,
+even for the same random seed. We see that, for this random seed, the :code:`Random: 0.5`
+won this Moran process. This is not what happens in a standard Moran process where the
 :code:`Random: 0.5` player will not win::
 
-    >>> axl.seed(3)
-    >>> amp = axl.MoranProcess(players)
-    >>> results = amp.play()
-    >>> amp.population_distribution()
+    >>> mp = axl.MoranProcess(players, seed=2)
```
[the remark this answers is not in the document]
marcharper: Yep! Fixed.

drvinceknight:
What an awesome piece of work @marcharper 👏 👏 :clap:

PR #1289, opened 2020-03-02
Review request

Capybasilisk:
Hi, not ready to merge. Just a request for a review of a new strategy: "Exponential".

Thanks.

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -65,6 +65,7 @@ from .darwin import Darwin
 from .dbs import DBS
 from .defector import Defector, TrickyDefector
 from .doubler import Doubler
+from .exponential import Exponential
 from .finite_state_machines import (
     TF1,
     TF2,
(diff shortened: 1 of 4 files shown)


marcharper:
The code looks mechanically ok. Two comments:
* There are some strategies that use a similar algorithm, see grudger.py and oncebitten.py
* `self._retaliations += self._grudges ** self._grudges - 1` is a pretty harsh escalation. If `self.grudges` gets to 3 then the play is going to defect for 27 turns. Are you sure that's what you wanted? Many opponents will in turn defect indefinitely after 27 defections.

Capybasilisk:
Hi Marc, thanks for feedback.

I've also just seen that I forgot to set grudges and retaliations as private variables in the init method, so I'll change that to match the strategy method.

I could add the strategy in to oncebitten.py or grudger.py as a steep variant of one of those methods.

The steep rise in defections was intended to make the method rapidly turn from an always cooperator to an effectively always defector in the event of defective opponent behavior.

Issue #740, opened 2016-10-09
Code Efficiency of Strategies


meatballs:
It might be interesting to have a measure of how efficient a strategy is in terms of its code complexity. 

e.g. If two strategies have similar scores but one does so in 12 lines of code whereas the other has 250 lines, the first could be considered more efficient.

It's perhaps best to use the inpsect module to look at the amount of byte code produced for a strategy, rather than its actual python code as I wouldn't want to encourage poor readability.

marcharper:
This would be interesting, we could also compare to the long-run time calculations (there's a notebook for it).

BradKML:
Do cognitive complexity as well, it is more reasonable than old metrics https://github.com/rohaquinlop/complexipy

PR #1300, opened 2020-03-28
Move classifiers outside of Players

gaffney2010:
I want to make it so that we can allow classifiers to be a function.

This will allow us to store the result of:
https://github.com/Axelrod-Python/Axelrod/issues/740
to a classifier after we program.

It will also allow us to program in the memory-depth of FSMs that I had done previously; currently, we just ran and copied the values in.

There are probably other places where this functionality will help.

I thought about replacing the default_classifier dict with a dict whose values are lambdas.  But decided against this because 1) it would really clutter the Player code and spread out classifier logic when we want to handle different strategies differently, and 2) it would make memoizing difficult (see below).

I decided to move the classifiers out of the Player class, and keep a table lookup in a singleton class called Classifiers.  I made a function that runs all classifiers on all strategies and saves the result to a yaml function; Classifiers reads from this.  This is necessary for classifiers that take a while to calculate.  (This was the problem with memory-depth on FSMs.)  I did a yaml file so that developers can check the calculated classifiers.

I continue to use the classifier dict in the strategies, where it exists, with the new logic as more of a replacement of the default classifier.  This makes the change backwards compatible, but is also necessary for factory-generated strategies.  I store the classifiers to the yaml also, which isn't necessary, but provides a nice check.

Most of the change here is switching classifier lookups to use the Classifiers class.  This isn't technically necessary, because of the backwards compatibility, but this is how we should lookup.


marcharper:
Do the tests pass locally?

docs/tutorials/contributing/strategy/adding_the_new_strategy.rst
```diff
@@ -5,6 +5,14 @@ To get the strategy to be recognised by the library we need to add it to the
 files that initialise when someone types :code:`import axelrod`.  This is done
 in the :code:`axelrod/strategies/_strategies.py` file.
 
+To classify the new strategy, run rebuild_classifier_table::
+
+    python rebuild_classifier_table.py
+
+This will update :code:`axelrod/strategies/_strategies.py`.  Check that the
+recorded classifications for the strategies are what you expected, then add this
+file to your git commit.
```
drvinceknight:
```suggestion
recorded classifications for the strategies are what you expected.
```

We don't give git instructions elsewhere so let's leave that out.

docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -28,15 +28,30 @@ Note that instances of the class also have this classifier::
...
+some logic that maps classifier definitions to values.  To learn the
+classification of a strategy, we first look in the strategy's classifier
+dictionary, then if the key is not present, then we refer to this logic.
+This logic must be defined for a class, and not specific instances.
+
+To lookup the classifier of a strategy, using the classifier dict, or the
+strategy's logic as default, we use :code:`Classifiers().get(<classifier>,
+<strategy>)`::
+
+    >>> from axelrod.classifier import Classifiers
```
drvinceknight: Could we:

```
import axelrod.classifier as ac
```

and use `ac.Classifiers()`

Or possibly have `Classifiers` loaded in the `axelrod` namespace and:

```
axl.Classifiers()
```

Not entirely sure if either of the two are best solutions (so feel free to go with something else). I'd like to avoid importing in to the main namespace.

docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -28,15 +28,30 @@ Note that instances of the class also have this classifier::
...
+    >>> Classifiers().get('memory_depth', axl.TitForTat())
     1
-    >>> s = axl.Random
-    >>> s.classifier['stochastic']
+    >>> Classifiers().get('stochastic', axl.Random())
     True
 
+Because the logic may take time to run, the script in
+:code:`rebuild_classifier_table.py` pre-calculates these values, and stores to
+:code:`data/all_classifiers.yml`.
```
drvinceknight: I'm not sure this detail needs to be in this part of the documentation (it's potentially confusing to a user).

requirements.txt
```diff
@@ -6,6 +6,7 @@ matplotlib>=2.0.0
 numpy>=1.9.2
 pandas>=0.18.1
 prompt-toolkit>=1.0.7
+pyyaml
```
drvinceknight: Can we add a lower bound.

drvinceknight:
FYI the travis build can be seen here: https://travis-ci.org/github/Axelrod-Python/Axelrod/pull_requests?utm_medium=notification&utm_source=github_status

drvinceknight:
Nice work @gaffney2010. A few mainly documentation suggestion from me for now. Still getting my head around it a bit but I see the advantages :+1:

drvinceknight:
CI is failing due to the following lines not being hit by tests:

```
axelrod/classifier.py                                               75      7    91%   44, 152, 155, 170-174
```

docs/tutorials/advanced/classification_of_strategies.rst
marcharper: I suggested somewhat similarly above that we could initialize `Classifiers()` once somewhere on library load with e.g. `Classifiers = _Classifiers()` and then we could simply use `axl.Classifiers` as needed.

axelrod/classifier.py
```diff
@@ -0,0 +1,205 @@
...
+            raise KeyError("Unknown classifier")
+
+        # Factory-generated players won't exist in the table.  As well, some
+        # players, like Random, may change classifiers at construction time;
+        # this get() function takes a player instance, while the saved-values
+        # are from operations on the player object itself.
+        if classifier in player.classifier:
+            return player.classifier[classifier]
+
+        def return_missing() -> None:
```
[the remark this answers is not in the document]
gaffney2010: I'll inline it.

axelrod/classifier.py
```diff
@@ -0,0 +1,205 @@
...
+        return player_classifiers[classifier]
+
+
+# Strategy classifiers
+
+def is_basic(s):
+    """
+    Defines criteria for a strategy to be considered 'basic'
+    """
+    stochastic = Classifiers().get("stochastic", s)
```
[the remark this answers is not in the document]
gaffney2010: That's a good idea.  I guess that it probably doesn't need to be a singleton in that case.  (Though I'll probably leave it as is.)

axelrod/classifier.py
```diff
@@ -0,0 +1,205 @@
...
+import yaml
+
+from axelrod.player import Player
+
+ALL_CLASSIFIERS_PATH = "data/all_classifiers.yml"
+
+T = TypeVar('T')
+
+
+class Classifier(Generic[T]):
```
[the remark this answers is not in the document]
gaffney2010: I see what you're saying, and I agree that it's usually better to avoid creating a class if we can help it.  However, in this case, I slightly lean towards how I have it, for a few reasons:
1. I like the way we declare type here.
2. We can work with keys and values in a dict, but I kinda like having these as variables named as "player_class_classifier" and "name".
3. It's imaginable that this would be expanded at some point.

docs/tutorials/advanced/classification_of_strategies.rst
gaffney2010: I'm following Marc's advice here, and loading into the axelrod namespace.

gaffney2010:
This is great feedback.  Thanks!

docs/tutorials/advanced/classification_of_strategies.rst
drvinceknight: Sounds good to me :+1:

axelrod/tests/unit/test_classification.py
```diff
@@ -7,7 +7,24 @@
...
+    Classifier,
+    Classifiers,
+    _Classifiers,
+    memory_depth,
+    rebuild_classifier_table,
+)
+from axelrod.player import Player
+
+
+class TestTitForTat(Player):
```
drvinceknight: Any objection to:

```suggestion
class TitForTatWithEmptyClassifier(Player):
```

? (This would need to be carried through.)

axelrod/classifier.py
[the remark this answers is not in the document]
marcharper: ok

axelrod/classifier.py
```diff
@@ -0,0 +1,236 @@
...
+        if not cls.known_classifier(key):
+            raise KeyError("Unknown classifier")
+
+        def classify_player_for_this_classifier(
+            player: Union[Player, Type[Player]]) -> Any:
+            # If the passed player is not an instance, then try to initialize an
+            # instance without arguments.
+            if not isinstance(player, Player):
+                try:
+                    player = player()
```
marcharper: Let's change the name of one of these to make it less confusing in the future. Maybe the function variable name can be `_player`, and then in the except clause here you could use `player = _player` if it can't be instanced without arguments (but I think that should usually work).

axelrod/classifier.py
```diff
@@ -0,0 +1,236 @@
...
+            return player_classifiers.get(key, None)
+
+        return classify_player_for_this_classifier
+
+    @classmethod
+    def is_basic(cls, s: Union[Player, Type[Player]]):
+        """
+        Defines criteria for a strategy to be considered 'basic'
+        """
+        stochastic = cls.__getitem__("stochastic")(s)
```
marcharper: Does `cls["stochastic"](s)` work here?

axelrod/match.py
```diff
@@ -14,7 +14,7 @@
 def is_stochastic(players, noise):
     """Determines if a match is stochastic -- true if there is noise or if any
     of the players involved is stochastic."""
-    return noise or any(p.classifier["stochastic"] for p in players)
+    return noise or any(Classifiers["stochastic"](p) for p in players)
```
marcharper: could use map here

marcharper:
Looks really nice. I left a few minor comments/questions but otherwise I'm happy with it.

axelrod/classifier.py
gaffney2010: No.  That was my first try too.

TypeError: 'type' object is not subscriptable

axelrod/classifier.py
gaffney2010: I'm not sure what to do here.

My intention was to only allow classify_player_for_this_classifier to run on Player instances.  This is because for a strategy, a non-instance classifier dict will be stored in the yaml file, so it should be able to read this anyway.  (Actually I fixed the code just now to actually do this lookup for a non-trivial initializer.)  This code "if not isinstance..." is to simplify a lot of code that was doing trivial initializers, then passing in; it also allows filters and maps in some additional places.

I could do what you're asking for.  This would not change the existing behavior, and not cause a failure if the user passes a strategy without a trivial initializer.  But I almost want to throw a failure there, because it's better to be clear that this function is calling an instance.  We have for example code somewhere that says "[s for s in axelrod.strategies if Classifiers["stochastic"](s)]", but there are lots of strategies (like Random) which are stochastic **sometimes**.  I kind of don't want to give them a True or a False there; neither answer is right, and I want them to rewrite it to pass an instance.

On the other hand, it's weird that I'm essentially swallowing this error when I can trivially initialize.  And if somebody tried to make a strategy without a trivial initializer and add it to all_strategies, this would cause an unexpected error.  I'll probably end up doing what you're suggesting, but want to think about it a little more.

drvinceknight:
Thanks for my requested modifications. Looks good to me :+1:

drvinceknight:
The failing test looks like it's got something to do with the filter set:


```
    self.assertTrue(passes_filterset(self.TestStrategy, full_passing_filterset_1))
164  File "C:\projects\axelrod\axelrod\strategies\_filters.py", line 198, in passes_filterset
165    passes_filters.append(filter_function.function(**kwargs))
166  File "C:\projects\axelrod\axelrod\strategies\_filters.py", line 49, in passes_operator_filter
167    return operator(classifier_value, value)
168TypeError: '>=' not supported between instances of 'NoneType' and 'int'
169
170======================================================================
171FAIL: test_filtered_strategies (axelrod.tests.unit.test_filters.TestFilters)
172----------------------------------------------------------------------
173Traceback (most recent call last):
174  File "C:\projects\axelrod\axelrod\tests\unit\test_filters.py", line 160, in test_filtered_strategies
175    [StochasticTestStrategy, UsesLengthTestStrategy],
176AssertionError: Lists differ: [] != [<class 'axelrod.tests.unit.test_filters.T[177 chars]gy'>]
177
178Second list contains 2 additional elements.
179First extra element 0:
180<class 'axelrod.tests.unit.test_filters.TestFilters.test_filtered_strategies.<locals>.StochasticTestStrategy'>
181
182- []
183+ [<class 'axelrod.tests.unit.test_filters.TestFilters.test_filtered_strategies.<locals>.StochasticTestStrategy'>,
184+  <class 'axelrod.tests.unit.test_filters.TestFilters.test_filtered_strategies.<locals>.UsesLengthTestStrategy'>]
```

axelrod/classifier.py
marcharper: Well maybe the solution is to require trivial initializers throughout the library, and/or throw a warning here that the default initializer was used, which a few strategies like memory-one currently do. Otherwise the user has to initialize their strategies before calling the `Classifier`, we can only read their minds so much.

If a change messes with one of the Meta strategies our tests should catch that. There's a test for example that tries to find violations when strategies claim to be not stochastic.

gaffney2010: I like that.  I implemented it that way.  It will raise an exception if there is not a trivial initializer.  Maybe we can enforce that _before_ Classifiers runs into it.

marcharper:
```
ERROR: test_will_lookup_key_for_classes_that_cant_init (axelrod.tests.unit.test_classification.TestClassification)
  File "C:\projects\axelrod\axelrod\classifier.py", line 192, in classify_player_for_this_classifier
    player = player()
TypeError: __init__() missing 1 required positional argument: 'x'
During handling of the above exception, another exception occurred:
  File "C:\projects\axelrod\axelrod\tests\unit\test_classification.py", line 85, in test_will_lookup_key_for_classes_that_cant_init
    self.assertEqual(Classifiers["memory_depth"](TitForTatWithNonTrivialInitialzer), 1)
  File "C:\projects\axelrod\axelrod\classifier.py", line 201, in classify_player_for_this_classifier
    "Passed player class doesn't have a trivial initializer."
Exception: Passed player class doesn't have a trivial initializer.
```

Also the warning comes up a lot during the tests, perhaps we want to suppress it, or change various invocations to use an instance instead.

gaffney2010:
I don't know how I'm supposed to get that last line of coverage.  I can do something hacky...

axelrod/tests/unit/test_classification.py
```diff
@@ -1,11 +1,106 @@
...
+    classifier = {}
+
+
+class TitForTatWithNonTrivialInitialzer(Player):
+    """
+    Same name as TitForTat, but with empty classifier.
+    """
+
+    def __init__(self, x: Any):
+        pass
```
drvinceknight, line 36: 1 suggestion would be

```suggestion
        pass  # pragma: no cover
```

Instead of doing something hacky we just say we don't need this line to be checked.

drvinceknight:
Thanks @gaffney2010 looks good :+1:

Issue #1286, opened 2020-03-02
ResultsSet __eq__ fails unexpectedly


marcharper:
Specifically I found an example (2 player tournament with 2 turns) that has `nan` eigenmoses results, and the comparison of `nan` to `nan` fails in `__eq__`.

drvinceknight:
Cool, could you paste the inputs here (assuming it's a specific 2 player tournament).

Tagging #1283 so we know that there might be slight overlap.

marcharper:
It was this test but with only two players and two turns:

```
    @given(seed=integers(min_value=1, max_value=4294967295))
    @settings(max_examples=5, deadline=None)
    def test_seeding_equality(self, seed):
        """Tests that a tournament with a given seed will return the
        same results each time."""
        rng = axelrod.RandomGenerator(seed=seed)
        players = [axelrod.Random(rng.random()) for _ in range(8)]
        tournament1 = axelrod.Tournament(
            name=self.test_name,
            players=players,
            game=self.game,
            turns=3,
            repetitions=100,
            seed=seed
        )
        tournament2 = axelrod.Tournament(
            name=self.test_name,
            players=players,
            game=self.game,
            turns=3,
            repetitions=100,
            seed=seed
        )
        results1 = tournament1.play(processes=2)
        results2 = tournament2.play(processes=2)
        self.assertEqual(results1.ranked_names, results2.ranked_names)
```

drvinceknight:
Ah ok cool, hypothesis should print out the specific inputs for which it failed (it stores them in a db ) when run and we should add that in as a test case when we fix.

marcharper:
I think it's not wrong so much as in that case there were not enough vengeful actions to have a well-defined eigenmoses score, hence the NaN values. We'll just need to check for NaN or == instead of just ==.

drvinceknight:
Yeah sounds like it should be relatively easy to sort out :)

PR #1301, opened 2020-03-29
Fix ResultSet __eq__ to handle nans

gaffney2010:
Fixes https://github.com/Axelrod-Python/Axelrod/issues/1286

### DIFF
--- a/axelrod/result_set.py
+++ b/axelrod/result_set.py
@@ -2,6 +2,7 @@ from collections import Counter, namedtuple
 import csv
 import itertools
 from multiprocessing import cpu_count
+from typing import List
 import warnings
 
 import numpy as np
@@ -609,6 +610,17 @@ class ResultSet:
             other : axelrod.ResultSet
                 Another results set against which to check equality
         """
+        def list_equal(v1: List[float], v2: List[float]) -> bool:
+            """Matches lists, accounting for NaNs."""
+            if len(v1) != len(v2):
+                return False
+            for i1, i2 in zip(v1, v2):
+                if np.isnan(i1) and np.isnan(i2):
+                    continue
+                if i1 != i2:
+                    return False
+            return True
+
         return all(
             [
                 self.wins == other.wins,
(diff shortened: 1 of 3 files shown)


axelrod/result_set.py
```diff
@@ -609,6 +610,17 @@ def __eq__(self, other):
             other : axelrod.ResultSet
                 Another results set against which to check equality
         """
+        def list_equal(v1: List[float], v2: List[float]) -> bool:
```
marcharper: Can we say more specifically that nans are considered equal, and rename the function to something like `list_equal_with_nans` or something like that?

axelrod/result_set.py
```diff
@@ -609,6 +610,17 @@ def __eq__(self, other):
...
+            """Matches lists, accounting for NaNs."""
+            if len(v1) != len(v2):
+                return False
+            for i1, i2 in zip(v1, v2):
+                if np.isnan(i1) and np.isnan(i2):
+                    continue
+                if i1 != i2:
+                    return False
```
marcharper: This line is the one not covered by a test

axelrod/result_set.py
```diff
@@ -722,7 +751,8 @@ def summarise(self):
 
         summary_data = []
         for rank, i in enumerate(self.ranking):
-            data = list(summary_measures[i]) + state_prob[i] + state_to_C_prob[i]
+            data = list(summary_measures[i]) + state_prob[i] + state_to_C_prob[
+                i]
```
marcharper: Can you run black on the final commit? Some of the line wraps seem odd.

axelrod/tests/unit/test_resultset.py
```diff
@@ -190,6 +190,52 @@ def test_init(self):
         self.assertEqual(rs.players, self.players)
         self.assertEqual(rs.num_players, len(self.players))
 
+    def _clear_matrix(self, matrix):
+        for i in range(len(matrix)):
```
drvinceknight: FYI, this is sometimes considered an anti pattern and the following is preferred:

```
for i, row in enumerate(matrix):
    for j, _ in enumerate(row):
         matrix[i, j] = 0
```

axelrod/tests/unit/test_resultset.py
```diff
@@ -190,6 +190,52 @@ def test_init(self):
...
+            for j in range(len(matrix[i])):
+                matrix[i][j] = 0
```
drvinceknight:
```suggestion
    def _clear_matrix(self, matrix):
        for i, row in enumerate(matrix):
            for j, _ in enumerate(row):
                 matrix[i, j] = 0
```

drvinceknight:
Looks great TJ, one very minor suggestion from me.

axelrod/result_set.py
gaffney2010: Yep.  I'm using a new IDE, but just added black, so it should be better going forward.

PR #1305, opened 2020-03-31
Fix ResultSet __eq__ to handle nans

gaffney2010:
This is a redo of https://github.com/Axelrod-Python/Axelrod/pull/1301 which was broken and rolled back.

The error is that I was using matrix[i, j] but this is a list of lists and not a numpy-style matrix, so I changed to matrix[i][j].

I will wait for tests and approvals this time.  :D


drvinceknight:
> I will wait for tests and approvals this time. :D

For future reference, we try to not merge our own PRs just to be sure. 2 positive reviews and one of those usually merges.

marcharper:
LGTM, can you rebase on the most recent master?

PR #1307, opened 2020-04-01
Create function to handle relative paths

gaffney2010:
A lot of the file reads and writes are causing me problems because my IDE tries to run logic from the location where the file is located, and the code is written to only be run from the top-level directory.  I could see a new user being thrown by this too.

This change makes paths a little more robust, by accounting for the path the user is actually in.

### DIFF
--- a/axelrod/load_data_.py
+++ b/axelrod/load_data_.py
@@ -1,6 +1,17 @@
-from typing import Dict, List, Tuple
+from typing import Dict, List, Text, Tuple
 
+import os
 import pkg_resources
+
+
+def axl_filename(axl_path: Text) -> Text:
+    """Get the path to Axelrod/<axl_path> from the working directory."""
+    # Working directory
+    dirname = os.path.dirname(__file__)
+
+    # We go up a dir because this code is located in Axelrod/axelrod and
+    # axl_path is from the top-level Axelrod dir.
+    return os.path.join(dirname, "..", axl_path)
 
 
 def load_file(filename: str, directory: str) -> List[List[str]]:
--- a/axelrod/load_data_.py
+++ b/axelrod/load_data_.py
@@ -4,25 +4,21 @@ from typing import Dict, List, Text, Tuple
 import pkg_resources
 
 
-def axl_filename(*axl_path: Text) -> Text:
+def axl_filename(path: pathlib.Path) -> pathlib.Path:
     """Given a path under Axelrod/, return absolute filepath.
 
     Parameters
     ----------
     axl_path
-        Path to file located under top-level Axelrod directory, with file names
-        separated at "/".  For example, to get the path to "Axelrod/xxx/yyy.py"
-        call this function with axl_filename("xxx", "yyy.py").
+        A pathlib.Path object with the relative directory under Axelrod/
 
     Returns
     -------
-        Absolute path to the given path.
+        A pathlib.Path object with the absolute directory.
     """
     # We go up a dir because this code is located in Axelrod/axelrod.
-    path = pathlib.Path(__file__).resolve().parent.parent
-    for p in axl_path:
-        path = path / p
-    return str(path)
+    axl_path = pathlib.Path(__file__).resolve().parent.parent
+    return axl_path / path
 
 
 def load_file(filename: str, directory: str) -> List[List[str]]:
--- a/axelrod/load_data_.py
+++ b/axelrod/load_data_.py
@@ -1,17 +1,28 @@
+import pathlib
 from typing import Dict, List, Text, Tuple
 
-import os
 import pkg_resources
 
 
-def axl_filename(axl_path: Text) -> Text:
-    """Get the path to Axelrod/<axl_path> from the working directory."""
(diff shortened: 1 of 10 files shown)


axelrod/load_data_.py
```diff
@@ -1,8 +1,19 @@
-from typing import Dict, List, Tuple
+from typing import Dict, List, Text, Tuple
 
+import os
 import pkg_resources
 
 
+def axl_filename(axl_path: Text) -> Text:
```
drvinceknight, line 7: Could we add a (simple) test in a  `test_load_data.py` file to the tests. I realise this command is used throughout the tests but it could be helpful at a later stage to know that this function is working as expected in case something else breaks.

marcharper:
Although there's only really one or two places to use it here, +1 to `pathlib` in general. Check it out!

gaffney2010:
I've looked a bit at pathlib, and I'm not seeing a very clean way to do this.

I could do something like Path(__file__), then use other pathlib function to append "../" then axl_path.  That doesn't seem cleaner though...

axelrod/load_data_.py
gaffney2010: Done.

axelrod/load_data_.py
```diff
@@ -1,8 +1,19 @@
...
+    """Get the path to Axelrod/<axl_path> from the working directory."""
+    # Working directory
+    dirname = os.path.dirname(__file__)
+
+    # We go up a dir because this code is located in Axelrod/axelrod and
+    # axl_path is from the top-level Axelrod dir.
+    return os.path.join(dirname, "..", axl_path)
```
marcharper, line 14: If you wanted to use `pathlib`, it would be something like:

```
path = Path(__file__).resolve().parent.parent / axl_path
return str(path)
```

You may be able to simply use `Path.cwd()` to get the current directory as well.

axelrod/tests/integration/test_tournament.py
```diff
@@ -45,7 +46,7 @@ def setUpClass(cls):
     def test_big_tournaments(self, tournament):
         """A test to check that tournament runs with a sample of non-cheating
         strategies."""
-        filename = "test_outputs/test_tournament.csv"
+        filename = axl_filename("test_outputs/test_tournament.csv")
```
marcharper, line 49: Also with pathlib, here you could do something like `filename = axl_filename() / "test_outputs" / "test_tournament.csv`, assuming that axl_filename with no arguments just returns the right directory.

gaffney2010: I like this idea to get it platform independent.  But actually the cast to string needs to happen before applying the / operator.  So this format wouldn't work.

What I did instead was take multiple arguments so that you could do:
axl_filename("test_outputs", "test_tournament.csv")
Then the function would apply the / operator inside.

drvinceknight:
This needs to be rebased on to master after #1308 got merged.

axelrod/load_data_.py
```diff
@@ -1,17 +1,28 @@
...
 
-import os
 import pkg_resources
 
 
-def axl_filename(axl_path: Text) -> Text:
-    """Get the path to Axelrod/<axl_path> from the working directory."""
-    # Working directory
-    dirname = os.path.dirname(__file__)
+def axl_filename(*axl_path: Text) -> Text:
```
drvinceknight, line 7: What about just passing a `pathlib` path object to this function?

So for example the test in `axelrod/tests/integration/test_tournament.py` would become:

```
path = pathlib.Path("test_outputs/test_tournament.csv")
filename = axl_filename(path)
```

The test in the fingerprints file would become:

```
path = pathlib.Path("test_outputs/test_fingerprint.csv")
filename = axl_filename(path)
```

The function code itself would be something like:

```
     axl_path = pathlib.Path(__file__).resolve().parent.parent
     return axl_path/path
```

You could `return str(axl_path/path)` but because of how pathlib handles things I believe you don't need to (I could be wrong).

(I'm also suggesting we change the name of the input variable to be `path`.)

gaffney2010: sgtm

marcharper:
LGTM as well, please rebase so we can merge

gaffney2010:
> LGTM as well, please rebase so we can merge

I merged recently.  I'm seeing "This branch has no conflicts with the base branch".  Is there something more I should do?

drvinceknight:
(Apologies if you know all this:) GitHub allows you to merge PRs in different ways. If we went for a merge commit, the icon looks like:

Screenshot 2020-04-07 at 10 36 30

If we wanted to rebase (which would be tidier, the icon looks like this:

Screenshot 2020-04-07 at 10 37 34

Usually that defaults to the last one you've used.

We don't really have a system/convention (I'd suggest mainly due to my poor knowledge of version control at the start of this project that means our history is pretty messy) but from @marcharper's comment I assumed he wanted a rebase merge :) (In which case if you rebase on master and force push we'll be able to click that.)

marcharper:
Either way is fine with me

PR #1309, opened 2020-04-06
Move to GitHub actions

drvinceknight:

PR #1314, opened 2020-04-08
Fix broken data file imports.

drvinceknight:
This is a bug fix. Will create a new release when it's in.


drvinceknight:
FYI to check that this works manually:

```
cd Axelrod
python setup.py install
cd <anywhere else>
python
>>> import axelrod
```

drvinceknight:
In #1309 I've added a test to the CI that would have caught this.

Nikoleta-v3:
I am merging this!

PR #1321, opened 2020-04-08
Update the nbr of strategies in first page of docs

drvinceknight:
Running `len(axelrod.strategies)` gives 235 :)

Genuinely not sure if I've got my grammar right in this sentence so please do suggest something else. I just wanted to bump the number :)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -8,8 +8,8 @@ Welcome to the documentation for the Axelrod Python library
 
 Here is quick overview of the current capabilities of the library:
 
-* Over 230 strategies including from the literature and some exciting original
-  contributions
+* Over 230 strategies including many from the literature and exciting
+  original contributions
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -8,7 +8,7 @@ Welcome to the documentation for the Axelrod Python library
 
 Here is quick overview of the current capabilities of the library:
 
-* Over 100 strategies from the literature and some exciting original
+* Over 230 strategies including from the literature and some exciting original
```
marcharper, line 11: including many?

I had a thought that we could add a classifier, something like "from a publication" (not necessary for this PR, of course).

drvinceknight: > including many?

Thank you! Much better.

> 
> I had a thought that we could add a classifier, something like "from a publication" (not necessary for this PR, of course).

Not a bad idea and relatively straightforward with @gaffney2010 recent classifier work :+1:

PR #1322, opened 2020-04-08
Add weekly CI run.

### LINKED ISSUE #1317: Add a timed weekly/monthly (?) build?
With GitHub actions it's relatively simply to run a timed build of the CI. This would potentially alert us to bugs that appears as other libraries update. Here is an example of a weekly build on one of 
my class repos: https://github.com/drvinceknight/cfm/blob/master/.github/workflows/config.yml#L7

I don't necessarily think we need this but it's a relatively simple thing to implement. Any thoughts?

drvinceknight:
Closes #1317

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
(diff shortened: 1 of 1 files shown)


Nikoleta-v3:
I will go ahead and merge this!

Issue #1279, opened 2019-12-18
Revise `RevisedDowning` to match k59r


drvinceknight:
It looks like they do not quite match at the moment.

This particular build: https://travis-ci.org/Axelrod-Python/axelrod-fortran/builds/626159813?utm_source=github_status&utm_medium=notification

shows that it failed to reproduce the same behaviour against `Alternator`.

gaffney2010:
I took at look at this.  It looks like when the fortran was translated to python, PAST was changed to self.history[-1].  Because of where it's set in the fortran, it should actually be self.history[-2].

https://github.com/Axelrod-Python/TourExec/blob/master/src/strategies/K59R.f#L35
https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/revised_downing.py#L55

I'm going to run tests to make sure this is the case, and will create a PR.

PR #1323, opened 2020-04-10
Fix RevisedDowning

gaffney2010:
Fixes https://github.com/Axelrod-Python/Axelrod/issues/1279

It looks like when the fortran was translated to python, PAST was changed to self.history[-1]. Because of where it's set in the fortran, it should actually be self.history[-2].

https://github.com/Axelrod-Python/TourExec/blob/master/src/strategies/K59R.f#L35
https://github.com/Axelrod-Python/Axelrod/blob/master/axelrod/strategies/revised_downing.py#L55

This was tested with axelrod_fortran.

### DIFF
--- a/axelrod/strategies/revised_downing.py
+++ b/axelrod/strategies/revised_downing.py
@@ -52,7 +52,7 @@ class RevisedDowning(Player):
 
         # Update various counts
         if round_number > 2:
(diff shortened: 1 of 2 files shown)


drvinceknight:
Nice! Thanks :+1:

PR #1329, opened 2020-04-12
Add documentation about hypothesis

gaffney2010:
After removing from requirements, hypothesis isn't automatically installed.  This may cause confusion, especially since we require a very specific version of hypothesis.  Here I add a note in the Running Tests documentation.


marcharper:
Note that after #1288 we can allow more recent versions of hypothesis.

PR #1330, opened 2020-04-13
Change classifier test to call Classifiers

gaffney2010:
classifier_test in test_player was using dictionaries instead of the Classifiers class.  This error was masked because all classifiers currently match the dictionaries.  In the future, we will use formulas for some of the classifiers.

In addition to the changes on test_player, I removed some lines which altered expected classifiers to incorrect values.  These lines accounted for the fact that the memory_depth was sometimes set on initializing.  The new Classifiers paradigm can account for this.


drvinceknight:
Thanks @gaffney2010 :+1:

PR #1331, opened 2020-04-13
Fix makes_use_of on MindReader

gaffney2010:
The MindReader strategies appear to use game (line 44), so I add this to the classifier.

### DIFF
--- a/axelrod/strategies/mindreader.py
+++ b/axelrod/strategies/mindreader.py
@@ -23,7 +23,7 @@ class MindReader(Player):
     classifier = {
         "memory_depth": float("inf"),
         "stochastic": False,
-        "makes_use_of": set(),
+        "makes_use_of": {"game"},
         "long_run_time": False,
         "inspects_source": True,  # Finds out what opponent will do
         "manipulates_source": False,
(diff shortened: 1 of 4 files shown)


axelrod/strategies/mindreader.py
```diff
@@ -91,7 +91,7 @@ class MirrorMindReader(ProtectedMindReader):
     classifier = {
         "memory_depth": float("inf"),
         "stochastic": False,
-        "makes_use_of": set(),
+        "makes_use_of": {"game"},
```
marcharper, line 94: Technically this one uses the opponent strategy, so whatever the opponent does determines the classifier.

gaffney2010: :|  Okay, I'm going to revert this back to set(), but I guess there's really not a correct value here.

gaffney2010:
> Perhaps we can add a test that tries variations of the game to see if it changes player behavior? I think we're not too concerned about these cheating strategies since the are excluded from everything, but if we can catch this kind of thing from a test in general that's good.

I would like to not do that.  I think it would be difficult to come up with examples.  But the code evidently does use game, and it would be hard to imagine a case where this dependency would break.  I think other strategies that make use of game don't have similar tests, and I think I also saw that cheating strategies are going to be deprecated, so it may not be worth the trouble.

PR #1332, opened 2020-04-14
Add Original Gradual to the list of all strategies.

drvinceknight:
@Nikoleta-v3 ran the full tournament and realised that I'd missed adding `Original Gradual` to the list of strategies.


marcharper:
I'm surprised no doc test failed -- could we think of one, maybe just a count of all the strategies?

drvinceknight:
> I'm surprised no doc test failed -- could we think of one, maybe just a count of all the strategies?

I like the idea of doctesting the number of strategies in the library (I think we've discussed this before). We could have something here https://github.com/Axelrod-Python/Axelrod/blob/master/docs/index.rst just under the "over 230" strategies, we might as well say exactly how many there are? 

FWIW, this isn't the first time this has happened (I usually double check the new strategies are in the results from the last run of the tournament).

drvinceknight:
@marcharper what do you think of 70bffa6 (adding a simple count of all strategies to docs).

gaffney2010:
> I'm surprised no doc test failed -- could we think of one, maybe just a count of all the strategies?

In this case, OriginalGradual was imported and never used.  Maybe we could check for that.  May be overkill, but could check other errors.

pylint --disable=all --enable=unused-import  axelrod/strategies/_strategies.py

[Actually when we start adding logic for classifiers, a test of classifier equality on a strategy would break if the strategy isn't in all_strategies, because it never gets recorded in the yaml.]

drvinceknight:
> In this case, OriginalGradual was imported and never used. Maybe we could check for that. May be overkill, but could check other errors.
> 
> pylint --disable=all --enable=unused-import axelrod/strategies/_strategies.py

Not a bad idea at all, there's no guarantee that there's not another strategy that we've missed in the past. I'll give this a go.

drvinceknight:
@gaffney2010 I've added that check to the CI (I believe the commits above are clear but let me know if I can explain). Note that in some cases there are correct occurrences of classes being imported but not used. Have had to annotate those (and separate their imports), I've documented that in the module doctoring. Let me know what you think :+1:

PR #1333, opened 2020-04-16
Standardize tests for memoryone and memorytwo files.

gaffney2010:
The memoryone and memorytwo strategy files were doing some unusual things in the test files.  It seems like they were trying to duplicate some functionality in Player's repr and in TestPlayers tests.  This was causing problems with Classifiers logic.  Details:

Code was changing name on initialization to include the parameter that the strategy was used with.  name is only used in two places:  for __repr__ and for Classifiers.  __repr__ already attaches the initialization parameters, and Classifiers was getting messed up because instances had a different name than the class.  By making this change we make the strategy behave consistently with other strategies.

I removed test_name from these unittests because this is covered by test_repr in TestPlayer.  It's true that test_play in these tests covers __repr__ for strategies initialized with specific parameters, but we can rely on the unittests that cover a general Player's __repr__ logic.

I removed classifier tests in these files.  We see in https://github.com/Axelrod-Python/Axelrod/pull/1330 that we had to update classifier tests.  We can update here too, but it'd be redundant and hard to maintain to continue to have these tests listed special.

### DIFF
--- a/axelrod/data/all_classifiers.yml
+++ b/axelrod/data/all_classifiers.yml
@@ -1220,6 +1220,14 @@ Opposite Grudger:
   manipulates_state: false
   memory_depth: .inf
   stochastic: false
+Original Gradual:
+  inspects_source: false
+  long_run_time: false
+  makes_use_of: !!set {}
+  manipulates_source: false
+  manipulates_state: false
+  memory_depth: .inf
+  stochastic: false
 PSO Gambler 1_1_1:
   inspects_source: false
   long_run_time: false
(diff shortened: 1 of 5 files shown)


gaffney2010:
Closing for a minute, I need to think about this more.

marcharper:
So IIUC these tests are redundant since 

```
    def test_repr(self):
        """Test that the representation is correct."""
        if self.__class__ != TestPlayer:
            self.assertEqual(str(self.player()), self.name)
```

acts on instances. Is it similarly the case for whether these strategies are stochastic or not?

gaffney2010:
Yes.  The test you quoted tests that a strategy with no initialization parameters will just equal the name.  There logic in the base Player class that appends initialization parameters.  This is tested too, but not for every strategy.  The repr tests I deleted cover the second case for specific strategies.

As well, I deleted tests that check the classifier on these strategies, which is also redundant because of tests in TestPlayer.

PR #1337, opened 2020-05-10
Update ultimatum with new commits from master

gaffney2010:
I started from master and merged in ultimatum.  I'm not sure how to build off of ultimatum otherwise.  The only change is the merge.

### DIFF
--- a/axelrod/__init__.py
+++ b/axelrod/__init__.py
@@ -22,3 +22,4 @@ from axelrod.tournament import Tournament
 from axelrod.result_set import ResultSet
 from axelrod.ecosystem import Ecosystem
 from axelrod.fingerprint import AshlockFingerprint, TransitiveFingerprint
+import axelrod.ultimatum as ultimatum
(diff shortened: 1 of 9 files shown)


marcharper:
Sure, it's really out of date. You could also try rebasing ultimatum onto the latest master, which may be a lot simpler, and if that goes well, we can simply overwrite the ultimatum branch. We have local copies in case anything goes wrong.

gaffney2010:
> You could also try rebasing ultimatum onto the latest master

This is what I'm doing here.  I'm not sure if the PR overwrites the existing ultimatum or not.

marcharper:
Ok, I see (just tried a rebase on #1338 and got essentially the same thing). Since it rebases without any conflicts (as everything is happening in a subdirectory), and the existing tests run fine, I think in this case we should just force push the rebase of ultimatum over the existing ultimatum branch, which effectively fast-forwards the ultimatum branch to the head of the current master. I'm not aware of any way to do this within the github interface.

marcharper:
Ok, so I just pushed the rebase over ultimatum (the original branch is in ultimatum-bak if needed). Let me know if that works for you.

gaffney2010:
That works for me.  I'll close this then.

PR #1339, opened 2020-05-10
Cleanup ultimatum and expand coverage

gaffney2010:
Did a few cleanup tasks and expanded to 100% coverage.

Specially I:
- Added tests
- Deleted set_match_attributes for now.
- Changed name/repr on AcceptanceThresholdPlayer.  It looks like this was a copy-paste error.
- Typed remaining inputs/outputs
- Added default values to strategies, to be consistent with IPD where every strategy is trivially initializable.
- Uncommented distribution strategy
- Changed tests to assertTrue/assertFalse (for boolean tests) and assertAlmostEqual for testing equality of floats.  I think these are best practices.
- Use attrs library instead of namedtuple, to address the commen there - I really love this library.

### DIFF
--- a/axelrod/ultimatum/player.py
+++ b/axelrod/ultimatum/player.py
@@ -1,9 +1,11 @@
 """
-Ultimatum Game .
+Ultimatum Game.
 """
 
-from collections import namedtuple
 from enum import Enum
+from typing import Optional, Tuple
+
+import attr
 from scipy import stats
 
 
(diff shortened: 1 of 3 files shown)


axelrod/ultimatum/player.py
```diff
@@ -12,8 +14,12 @@ class PlayerPosition(Enum):
     DECIDER = 2
 
 
-# Python 3.7 dataclass would be nice here
-Outcome = namedtuple('Outcome', ['position', 'offer', 'decision', 'scores'])
+@attr.s
+class Outcome(object):
+    position: "UltimatumPlayer" = attr.ib()
```
marcharper, line 19: What's going on here with "UltimatumPlayer"? Is it a generic default?

gaffney2010: I'm doing a forward reference to the type that we will declare later.

https://www.python.org/dev/peps/pep-0484/#forward-references

PR #1342, opened 2020-05-17
Add GameParams to make Match generalizable

gaffney2010:
I modify Match to take a "GameParams" object.  A GameParams class is intended to eventually include everything that a Match or Tournament needs to know about a new game type (IPD, ultimatum, etc.).  For now it includes only a few functions used in Match play, described below.  This change is non-API breaking, because Match takes ipd_game_params by default.

GameParams is a class with three elements, all functions passed at initialization:
- generate_play_params: Generates a "PlayParams" object, which tells Match how to play a single round.  (For now, this is players with their positions.)  This is an iterator which should produce all the PlayParams needed for the entire Match run.  It takes in a list of players and the number of desired rounds.
- play_round:  From PlayParams and a Game object (for scoring), play a round of the game.
- result:  This is a translator which for IPD translates Outcome classes to a pair of Actions, which keeps the Match API unchanged.  For new games, we will probably make this a pass-through, an let Match's play function return a list of Outcome classes.

generate_play_params and play_round could be combined, because they get called back-to-back in Match play.  I chose not to do this because: 1) The functionality naturally separates, and 2) For an asymmetric game, like Ultimatum, some research would prefer Matches to alternate roles between the players, while other research would prefer each player to keep a role; between these cases, generate_play_params would change, but not play_round.

In addition to GameParams and it's helper functions, this change generalizes the Outcome class.

In a future change, we can make GameParams class for Ultimatum, and run Match on that game.

### DIFF
--- a/axelrod/game_params.py
+++ b/axelrod/game_params.py
@@ -102,7 +102,7 @@ def symm2p_play_round(
         params.player_positions[Symm2pPosition.POS_1],
         params.player_positions[Symm2pPosition.POS_2],
     )
-    action_1, action_2 = player_1.play(player_2)
+    action_1, action_2 = player_1.play(player_2, noise=noise)
     actions = {
         Symm2pPosition.POS_1: action_1,
         Symm2pPosition.POS_2: action_2,
--- a/axelrod/game_params.py
+++ b/axelrod/game_params.py
@@ -1,0 +1,131 @@
+"""Implements GameParams and other classes.
+
+The GameParams class defines all everything a program should need to understand
+a game (e.g. IPD or ultimatum).
+"""
+
+from enum import Enum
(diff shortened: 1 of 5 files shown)


marcharper:
Hi, looks promising. FYI we should be close to reproducing [this paper](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=2ahUKEwiqkeT2zbzpAhXvJzQIHc34CJUQFjAAegQIBRAB&url=https%3A%2F%2Fwww.pnas.org%2Fcontent%2F110%2F7%2F2581&usg=AOvVaw0GbhzHf5a0FAJU5sKaYPVW) using the Moran process (with "atomic" mutation) and the Ultimatum game.

PR #1346, opened 2020-05-21
Create a scorer for ultimatum

gaffney2010:
Creates a scorer (or "Game") for ultimatum.  This is one of the inputs to Matches and Tournaments, so it's needed to run an ultimatum tournament.

I would have liked to have done a couple of things differently:

1. Move scorer to belong to game_params.  Remove scorer as an input to Matches and Tournaments and Moran processes and Match Generators, preferring instead to pass around game_params.  (game_params would need to be more carefully initialized so that scorer objects aren't shared.)  The advantage of this approach is 1) it makes it impossible to mix-and-match scorers and game types (an issue I've ignored for now), and 2) it centralizes all the _settings_ for a game type.

2. Have the scorer's score function take in a dictionary of Actions keyed by position and return a dictionary of Scores keyed by position.  This more naturally extends to multi-player game, and is consistent with a pattern I've started to use on things like Outcome.  (It's also perhaps more natural for an asymmetric game like Ultimatum vs using order alone to specify position.)

I have restrained from making these changes here, because they are API-breaking.  Aside from causing problems with an eventual merge onto master, it means I have to change A LOT of code to keep the library working (tests passing) on this branch.  At this early stage, where we're still experimenting with API, I want to avoid that.

### DIFF
--- a/axelrod/game_params.py
+++ b/axelrod/game_params.py
@@ -4,17 +4,19 @@ The GameParams class defines all everything a program should need to understand
 a game (e.g. IPD or ultimatum).
 """
 
-from enum import Enum
-from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Text,
+)
 
 import attr
 
-from axelrod.action import Action
-from axelrod.game import Game
-from axelrod.player import Player, PlayerName
-
-Position = Enum
-Score = Union[float, int]
+from .prototypes import BasePlayer, BaseScorer, Position, Score
 
 
 @attr.s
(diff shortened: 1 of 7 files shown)


axelrod/game_params.py
```diff
@@ -4,17 +4,19 @@
 a game (e.g. IPD or ultimatum).
```
marcharper, line 4: It may help, going forward, to give a high level description here of what's happening, and what needs to be defined. E.g.: 

A game is specified by the following information:
- A description of positionality (symmetric, asymmetrics, simulaneous play, etc.)
- A definition of the outcomes of an interaction
- A scoring function that takes (...) as parameters and produces a payoff
- etc.

axelrod/ipd_params.py
```diff
@@ -0,0 +1,21 @@
+"""Implements ipd_params."""
```
marcharper, line 1: Similarly, here it would probably be useful to explicitly describe the define parameters. IPD is a symmetric simultaneously play game with outcome Cooperate (C) and Defect (D). Scoring is determined by a matrix of payoff R, S, T, P ...

axelrod/game_params.py
gaffney2010: Will do

PR #1347, opened 2020-05-22
Add ultimatum game_params, and extend documentation.

gaffney2010:
After defining game_params for ultimatum, we can run a Match.  I added an integration test which does this.

I added some TODOs to track things I want to circle back to, and I added more docstrings to all the new files, following marcharper's request on the last PR.

I added an "adding_games" document in our documentation, which outlines all the things that need to be addressed when we create a new game.  This is a bit of stub and will need to be expanded, but want to use it to keep track of all the changes I'm making.

### DIFF
--- a/axelrod/prototypes.py
+++ b/axelrod/prototypes.py
@@ -1,4 +1,7 @@
-"""Prototypes for the various game components."""
+"""Prototypes for the various game components.
+
+This is used for type annotations and deriving functions.
+"""
 
 from enum import Enum
 from typing import Tuple, Union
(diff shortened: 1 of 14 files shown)


axelrod/prototypes.py
```diff
@@ -1,19 +1,53 @@
...
 Position = Enum
 Score = Union[float, int]
 
 
+# TODO(5.0): Explore what functions should be shared for all players / all
+#  scorers.
 class BasePlayer(object):
-    def strategy(self, opponent: "BasePlayer") -> Action:
-        raise NotImplementedError()  # pragma: no cover
+    """A player for any game.  No gauranteed functionality."""
```
marcharper, line 20: nit: guaranteed

PR #1350, opened 2020-05-31
Move common play functionality

gaffney2010:
We make a BasePlayer, with a common play function.  This play function looks at a GameParams object (passed at initialization) to use get_actions, which says how to get a set of actions from the players playing; it then saves history and returns outcomes.  (To save the GameParams object, we need to change pickling to dilling in many places.)  In order to save history, we need to make a common BaseHistory class, along with a history factory that we store to GameParams.

### DIFF
--- a/axelrod/action.py
+++ b/axelrod/action.py
@@ -10,6 +10,8 @@ from enum import Enum
 from functools import total_ordering
 from typing import Iterable
 
+from .prototypes import BaseAction
+
 
 class UnknownActionError(ValueError):
     """Error indicating an unknown action was used."""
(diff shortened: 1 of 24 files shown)


gaffney2010:
The tests are failing on Windows, because I guess the parallelizations pickle by default and I added a lambda to the Player class.  I could avoid this by, for example, making a get_game_params method that gets overwritten for each player type...

marcharper:
Yeah multiprocessing pickles objects that pass through it. Can we stipulate that each "Player type" specify a history class in its definition, and the abstract base class can just use something generic?

marcharper:
Is there no way to avoid use of dill?

gaffney2010:
Closing this for now.  The work to get it running in Windows may be a little bigger than I thought.

PR #1354, opened 2020-07-05
Modify test to ensure that all strategies have names

caddycarine:
Test 'test_all_strategies_have_names' is always passed because when a strategy has no name, it adopts the name attribute from the 'Player' class.

### DIFF
--- a/docs/tutorials/advanced/classification_of_strategies.rst
+++ b/docs/tutorials/advanced/classification_of_strategies.rst
@@ -43,7 +43,7 @@ strategy's logic as default, we use :code:`Classifiers[<classifier>](
 <strategy>)`::
 
     >>> from axelrod import Classifiers
-    >>> Classifiers['memory_depth'](axl.TitForTat)
+    >>> Classifiers['memory_depth'](axl.TitForTat())
     1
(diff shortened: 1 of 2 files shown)


docs/tutorials/advanced/classification_of_strategies.rst
```diff
@@ -43,7 +43,7 @@ strategy's logic as default, we use :code:`Classifiers[<classifier>](
 <strategy>)`::
 
     >>> from axelrod import Classifiers
-    >>> Classifiers['memory_depth'](axl.TitForTat)
+    >>> Classifiers['memory_depth'](axl.TitForTat())
```
gaffney2010, line 46: Why did this need to be changed?

marcharper: I think it's not necessary per se but there is a warning:
```
/home/user/repos/axelrod/Axelrod/axelrod/classifier.py:196: UserWarning: Classifiers are intended to run on player instances. Passed player Tit For Tat was initialized with default arguments.
```

gaffney2010: Makes sense.

PR #1355, opened 2020-07-05
Implements a make_use_of function that automatically 

### LINKED ISSUE #442: Smarter classification of `makes_use_of`
This is based on #433 and #348 where @marcharper suggested:

> Bonus points:
> 
> Try to detect usage automatically (with the inspect module) as part of another test that checks that every strategy's make_use_of classifier. This way if someone changes a strategy to use (or no longer use) one of the tournament attributes, it should trip the test
> Double bonus points:
> 
> Make the automated bonus test able to accept future new keys of the tournament_attributes dictionary

gaffney2010:
This resolves #442.  The change makes the calculation of the make_use_of classifier easier and less error-prone.

Other changes that are necessary to make this work
- Adds logic to the Classifiers class to calculate when the key is missing from the class.  This should have been there originally.
- Remove makes_use_of from strategies, so that this will calculate from the function.
- Edit transformer to check the transformation and the original strategy, when filling out the makes_use_of classifier.
- Fix tests that had wrong expected makes_use_of.
- Delete special classifier test on memory one strategies.  This is tested by the usual player test suite.


drvinceknight:
:+1:

PR #1359, opened 2020-08-04
Add a long-run strategy to the filtering test

caddycarine:
Currently `test_boolean_filtering` only uses strategies as given by the `strategy_list` hypothesis strategy. This by default only picks short run time strategies so we are not ensuring that the long run filter works.

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -51,8 +51,8 @@ jobs:
         python run_mypy.py
     - name: Check imports are sorted
       run: |
-        python -m pip install isort
-        python -m isort --check-only
+        python -m pip install "isort==4.3.21"
+        python -m isort --check-only --recursive axelrod/.
     - name: Check that all strategies are indexed
       run: |
         python run_strategy_indexer.py
--- a/axelrod/tests/integration/test_filtering.py
+++ b/axelrod/tests/integration/test_filtering.py
@@ -21,6 +21,7 @@ class TestFiltersAgainstComprehensions(unittest.TestCase):
         warnings.simplefilter("default", category=UserWarning)
 
     @given(strategies=strategy_lists(min_size=20, max_size=20))
+    @example(strategies=[axl.DBS, axl.Cooperator])
     def test_boolean_filtering(self, strategies):
 
         classifiers = [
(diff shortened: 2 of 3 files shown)


Nikoleta-v3:
Hello @caddycarine 👋  Thank you for the `PR`. Nice catch that the long run strategies are currently not included in the test. 

The `DBS` strategy is indeed a  long run strategy but `Cooperator` is not. Cooperator is actually included in the `strategy_list`. 

Note that you can see the list of the long run strategies using the following code:

```python 
>>> import axelrod as axl

>>> filterset = {
...    "long_run_time": True
...    }

>>> strategies
```

which is based on this page of the tutorial: https://axelrod.readthedocs.io/en/stable/tutorials/advanced/classification_of_strategies.html?highlight=classifier 😄 hope this helps.

drvinceknight:
I think that's wanted behaviour here @Nikoleta-v3, this test is checking that if you pass a list of strategies, some of which are long run time then the filter will pick the long run time ones out. Whether or not cooperator is long run time is not terribly important, it just needed an example with at least 1 long run time strategy. Am I missing something?

drvinceknight:
@caddycarine the tests are currently failing but this is not anything you've done (probably a dependency that's been upgraded upstream). Could you modify the `/docs/tutorials/advanced/tournament_results.rst` file, specifically where it says:

```
    >>> results.initial_cooperation_count
     [9.0, 0.0, 9.0, 9.0]
```

could you change that to:

```
     [9, 0, 9, 9]
```

Once you've done that, commit the change and push (like we did together), this will update things here.

Nikoleta-v3:
Ah of course 🤦‍♀️ Thank you for the clarification @drvinceknight

marcharper:
This also happened on #1288, it may be a `dask` or `pandas` issue with int vs float types.

drvinceknight:
Thanks for making that change @caddycarine, the tests have failed for another reason but this is not your fault. Can you modify the `.github/workflows/config.yml` file so that:

```
    - name: Check imports are sorted
       run: |
         python -m pip install isort
         python -m isort --check-only
```

becomes:

```
    - name: Check imports are sorted
       run: |
         python -m pip install "isort==4.3.21"
         python -m isort --check-only --recursive axelrod/.
```

Once you've done that commit and push as before :)

marcharper:
Since all the tests pass, I think this is fine to go in as is. We'll rerun `isort` and `black` once #1288 goes in, and we'll have a script to run for all this once #1360 is done.

Issue #1360, opened 2020-08-04
Presumbit script and pin all dependency versions


marcharper:
- [] Create a script or enhancement to [test](https://github.com/Axelrod-Python/Axelrod/blob/master/test) that runs all the same checks as [config.yml](https://github.com/Axelrod-Python/Axelrod/blob/master/.github/workflows/config.yml)
- [] Pin versions of isort, mypy, etc. in [requirements.txt](https://github.com/Axelrod-Python/Axelrod/blob/master/requirements.txt) so they are the same locally as on github

Issue #1352, opened 2020-06-15
Rename the master branch to main


drvinceknight:
I would like to rename the `master` branch to `main`.

(There's a tool here that helps with this: https://github.com/mheap/github-default-branch)

marcharper:
Maybe this is a good time to reconsider the branches "officially". `main` isn't particularly meaningful from a development standpoint -- is it the main working branch? the current release?

How about a structure such as:

* `release` in which the HEAD is the most current tagged release
* `dev` or `develop` which is the "development stable" (i.e. all tests pass) but not yet released, sort of a staging branch I guess
* people can have other auxiliary branches as needed before merging into `dev` (e.g. the current `ultimatum` branch)

When we want to do new releases we just merge `dev` into `release` and tag it. New PRs go into `dev` by default. WDYT?

Next, I'm not sure we need to grant access to a tool for this -- we just push a new branch and tell github that's the new default, right? AFAIK git doesn't have a concept of a default branch (unless that's changed recently), so we just need to set the github default in settings (and delete the old branch).

Also, maybe we want to revisit the contributing instructions and provide a command to setup a venv with the right defaults / dependencies. Seems like we'll need to at least document how people should open PRs and the meaning of the branch names.

drvinceknight:
> 
> How about a structure such as:
> 
> * `release` in which the HEAD is the most current tagged release
> * `dev` or `develop` which is the "development stable" (i.e. all tests pass) but not yet released, sort of a staging branch I guess
> * people can have other auxiliary branches as needed before merging into `dev` (e.g. the current `ultimatum` branch)
> 
> When we want to do new releases we just merge `dev` into `release` and tag it. New PRs go into `dev` by default. WDYT?

That sounds great and makes the release process more straightforward.

> 
> Next, I'm not sure we need to grant access to a tool for this -- we just push a new branch and tell github that's the new default, right? AFAIK git doesn't have a concept of a default branch (unless that's changed recently), so we just need to set the github default in settings (and delete the old branch).

I assumed the tool would take care of any PRs that might be up so that they would automatically "redirect" to the new branch but I'm happy to not use it and just figure out whatever is broken as and when (it won't be anything hard to fix).

> 
> Also, maybe we want to revisit the contributing instructions and provide a command to setup a venv with the right defaults / dependencies. Seems like we'll need to at least document how people should open PRs and the meaning of the branch names.

Yeah great call. This could be the first PR to `dev` :+1:

[...]

PR #1364, opened 2020-08-07
SpitefulCC from [Mathieu2015]

RomeroLaura:
Added new SpitefulCC Player to grudger.py, updated the list of all strategies, and added coverage test in test_grudger.py.
I also run these files by black formatter as specified in the instructions.

Let me know if everything is in order

PR #1367, opened 2020-08-12
Write a new documentation page with branch info

drvinceknight:
This is related to #1352.

Once this PR is merged we should:

- Change github default branch to `dev`
- Delete `master`
- Confirm that read the docs is looking at `release`

Am I missing anything?

We should also remove the release process information from the wiki
(assuming we're happy with what I've written here).

Finally, once #1360 is done we should make sure we update the docs with
the relevant information.

### DIFF
--- a/docs/tutorials/contributing/setting_up_the_environment.rst
+++ b/docs/tutorials/contributing/setting_up_the_environment.rst
@@ -1,0 +1,32 @@
+Setting up the environment
+==========================
+
+Installing all dependencies
+---------------------------
+
+All testing dependencies can be installed by running::
+
+  $ pip install -r requirements.txt
+
+It is recommended to do this using a virtual environment tool of your choice.
+
+The git workflow
+----------------
+
+There are two important branches in this repository:
+
+- :code:`dev`: The most up to date branch with no failing tests.
+  This is the default branch on github.
+- :code:`release`: The latest release.
+
+When working on a new contribution branch from the latest :code:`dev` branch and
+open a Pull Request on github from your branch to the :code:`dev` branch.
+
+The procedure for a new release (this is carried out by one of core maintainers):
+
+1. Create a Pull Request from :code:`dev` to :code:`release` which should
+   include an update to :code:`axelrod/version.py` and :code:`CHANGES.md`
+2. Create a git tag.
+3. Push to github.
+4. Create a release on github.
+5. Push to PyPi: :code:`python setup.py sdist bdist_wheel upload`
--- a/docs/tutorials/contributing/setting_up_the_environment.rst
+++ b/docs/tutorials/contributing/setting_up_the_environment.rst
@@ -14,7 +14,7 @@ For example, when using the virtual environment library :code:`venv`::
 
   $ python -m venv axelrod_development
   $ source axelrod_development/bin/activate
-  $ pip install -r requrirements
+  $ pip install -r requrirements.txt
 
 The git workflow
 ----------------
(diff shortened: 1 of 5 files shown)


drvinceknight:
917eead found a failing example for the reproducible seeding, I fixed that in 3f60d5d (that commit message has more information).

marcharper:
FYI this may cause a problem for PR #1364

docs/tutorials/contributing/setting_up_the_environment.rst
```diff
@@ -0,0 +1,32 @@
...
+==========================
+
+Installing all dependencies
+---------------------------
+
+All testing dependencies can be installed by running::
+
+  $ pip install -r requirements.txt
+
+It is recommended to do this using a virtual environment tool of your choice.
```
marcharper, line 11: In the interest of being beginner friendly, should we add in some venv commands here?

drvinceknight: > In the interest of being beginner friendly, should we add in some venv commands here?

I'm not sure. Depending on the level of beginner what's appropriate? For someone who has just done things using anaconda do we go with a Conda environment (my current setup)? Otherwise do we choose virtualenv, pew, venv (which is relatively standard?), pipenv (which is **really** liked by the people who use it)... ?

Perhaps something like:

```
It is recommended to do this using a virtual environment tool of your choice. For example:

If you use anaconda, create an `axelrod_development` environment by running:

$ conda ...

Otherwise, assuming you have `venv` installed, run:

$ venv ...
```

I'm super open to suggestions although I suspect those who need the documentation instructions might need more explanation as to what/why an virtual env is. Those who don't won't need such instructions as they'll have their own preferred tooling...

drvinceknight:
> FYI this may cause a problem for PR #1364

Yeah. Hopefully easy enough to fix but I'm happy to wait for the right "gap" to get this PR in :)

docs/tutorials/contributing/setting_up_the_environment.rst
marcharper: Having something seems better than nothing and I agree we can't maintain instructions for all the choices. Either `conda` or `venv` are probably fine for complete python newbies. For example take Nik's conda instructions [here](https://github.com/Nikoleta-v3/meta-analysis-of-prisoners-dilemma-tournaments).

Anyone with a strong preference probably doesn't need documentation on this and can use whatever they prefer.

drvinceknight: In 95e2905 I've gone with `venv`, I started writing an example of using `conda` but it gets "slightly" complicated if you wanted to use the available conda installable libraries... Open to suggestions though :)

docs/tutorials/contributing/setting_up_the_environment.rst
```diff
@@ -14,7 +14,7 @@ For example, when using the virtual environment library :code:`venv`::
 
   $ python -m venv axelrod_development
   $ source axelrod_development/bin/activate
-  $ pip install -r requrirements
+  $ pip install -r requrirements.txt
```
marcharper, line 17: requirements.txt (spelling)

drvinceknight: Good catch! (🙄 I'm an idiot.)

marcharper: spelling is hard :)

Nikoleta-v3: > spelling is hard :)
upvoted

Nikoleta-v3 (approved):

PR #1374, opened 2020-09-11
Add a property based test for valid strategies

drvinceknight:
Not sure if this is overkill @marcharper, just thought it could be worth
having.

This is a PR to #1373

### DIFF
--- a/axelrod/tests/strategies/test_meta.py
+++ b/axelrod/tests/strategies/test_meta.py
@@ -2,11 +2,11 @@
 
 import axelrod as axl
 from axelrod.classifier import Classifiers
+from axelrod.tests.property import strategy_lists
 from hypothesis import given, settings
 from hypothesis.strategies import integers
 
 from .test_player import TestPlayer
-from axelrod.tests.property import strategy_lists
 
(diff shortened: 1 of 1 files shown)


marcharper:
Want to move it to `TestPlayer` to run for all strategies? Or maybe update on of the existing tests there to also check that actions are valid?

(BTW this is the kind of thing that type hinting should catch...)

drvinceknight:
> Want to move it to `TestPlayer` to run for all strategies? Or maybe update on of the existing tests there to also check that actions are valid?

I thought about that but thought that it would perhaps add a lot to the run time and given that we always have explicit test for the other players perhaps it wasn't needed? Could limit the turns I suppose... 

> 
> (BTW this is the kind of thing that type hinting should catch...)

After a long journey (that is still interrupted with swearing at mypy) I am 100% on board with type hints :)

marcharper:
It's fine as is if you prefer

drvinceknight:
> It's fine as is if you prefer

I suggest we leave as is but perhaps we open an issue to either specifically target type hinting the strategies strategy method or add a test like this?

PR #1377, opened 2020-10-01
Add docstrings

caddycarine:
Add docstrings to strategy methods so they don't display the default placeholder docstring and simplify chained comparisons in SelfSteem class


.isort.cfg
```diff
@@ -4,4 +4,4 @@ multi_line_output = 3
 include_trailing_comma = True
 force_grid_wrap = 0
 combine_as_imports = True
-line_length = 80
+line_length = 88
```
drvinceknight: This line here is causing a number of other things to break.

caddycarine: `python -m isort --recursive axelrod/.` and `python -m black -l 80` .
seem to do conflicting changes.

drvinceknight: They shouldn't based on the setting in https://github.com/Axelrod-Python/Axelrod/blob/dev/.isort.cfg but it could be that the specific version of `isort` you're using isn't the one we have setup in the CI: https://github.com/Axelrod-Python/Axelrod/blob/dev/.github/workflows/config.yml#L52

(We should clarify that somewhere.)

drvinceknight: Sorry this PR is proving to be a pain for you, if there's something specific I can do to assist please ask. We'll have to have a think afterwards how we can avoid the friction you're experiencing. It hasn't really happened in the past but we've made  a number of small tweaks to our CI recently which might be the cause.

caddycarine: isort
I  can't get both of them to agree on these four files:
1. axelrod\strategies\titfortat.py
2. axelrod\tests\property.py
3. axelrod\tests\strategies\test_evolvable_player.py
4. axelrod\tests\strategies\test_finite_state_machines.py

Would appreciate any suggestions/recommendations.

drvinceknight: Taking a look now :)

drvinceknight: I've opened https://github.com/caddycarine/Axelrod/pull/1 on your fork. This is a PR to this branch so once you merge that it should update this PR and __should__ get the CI to pass.

The main problem seemed to be https://github.com/caddycarine/Axelrod/pull/1/commits/bbe9a7238f75f7362425c2a43aede6f0bbafd87c

caddycarine: Your solution works 😁!
Thanks.

axelrod/tests/unit/test_strategy_transformers.py
```diff
@@ -1001,20 +1001,6 @@ class TestFlippedDualTransformer(TestPlayer):
     }
 
 
-class TestIdentityDualTransformer(TestPlayer):
```
drvinceknight, line 1004: Why remove this?

caddycarine: Repetition; an identical class was defined earlier..

drvinceknight: Great :+1: Thank you.

drvinceknight (approved):
This looks good to me. :+1:

Nikoleta-v3 (approved):

PR #1381, opened 2020-11-26
Fix docstring on Tournament class

ben-harack:
Eliminated a duplicate entry and added a missing one.


Nikoleta-v3:
Nice catch! Thank you for the `PR` @ben-harack 😄

Nikoleta-v3 (approved):

PR #1382, opened 2021-01-11
Update finite_state_machines.py

dashiellfryer:
The Evolved FSM 8 was evolved from a 10 state FSM designed to reduce the Cooperation rate after DD and CD while increasing the Cooperation rate after DC and CC. That strategy scored well in tournaments but after evolving for 500 generations results in this winning strategy.

### DIFF
--- a/axelrod/strategies/finite_state_machines.py
+++ b/axelrod/strategies/finite_state_machines.py
@@ -1024,8 +1024,8 @@ class TF3(FSMPlayer):
         super().__init__(
             transitions=transitions, initial_state=0, initial_action=C
         )
-
- class EvolvedFSM8(FSMPlayer):
+        
+class EvolvedFSM8(FSMPlayer):
     """
     An 8 state FSM player trained with an evolutionary algorithm.
     
--- a/axelrod/strategies/finite_state_machines.py
+++ b/axelrod/strategies/finite_state_machines.py
@@ -1029,30 +1029,7 @@ class EvolvedFSM8(FSMPlayer):
     """
     An 8 state FSM player trained with an evolutionary algorithm.
     
-    Evolved from the 10 state FSM designed by Frederick Vincent with transitions
-    
-    transitions = (
-            (1, C, 2, C),
-            (1, D, 3, D),
-            (2, C, 1, C),
-            (2, D, 4, D),
-            (3, C, 4, C),
-            (3, D, 5, D),
-            (4, C, 5, C),
-            (4, D, 6, D),
-            (5, C, 2, C),
-            (5, D, 9, D),
-            (6, C, 5, D),
-            (6, D, 7, D),
-            (7, C, 5, C),
-            (7, D, 8, D),
-            (8, C, 5, D),
-            (8, D, 5, C),
-            (9, C, 2, C),
-            (9, D, 10, C),
-            (10, C, 2, C),
-            (10, D, 4, D)
-    )
+    Evolved from the 10 state FSM designed by Frederick Vincent.
 
     Names:
 
--- a/axelrod/strategies/finite_state_machines.py
+++ b/axelrod/strategies/finite_state_machines.py
@@ -1024,3 +1024,71 @@ class TF3(FSMPlayer):
         super().__init__(
             transitions=transitions, initial_state=0, initial_action=C
         )
+
+ class EvolvedFSM8(FSMPlayer):
+    """
+    An 8 state FSM player trained with an evolutionary algorithm.
+    
+    Evolved from the 10 state FSM designed by Frederick Vincent with transitions
+    
+    transitions = (
+            (1, C, 2, C),
+            (1, D, 3, D),
+            (2, C, 1, C),
+            (2, D, 4, D),
+            (3, C, 4, C),
+            (3, D, 5, D),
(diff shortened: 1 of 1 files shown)


marcharper:
Hi Dash, the error seems to be unrelated to your new strategy; we'll look into it. You would need to add the strategy name to the [this file](https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/strategies/_strategies.py) though.

marcharper:
@drvinceknight FYI there appears to be a latent docs build error.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1024,3 +1024,71 @@ def __init__(self) -> None:
...
         )
+
+ class EvolvedFSM8(FSMPlayer):
+    """
+    An 8 state FSM player trained with an evolutionary algorithm.
+    
+    Evolved from the 10 state FSM designed by Frederick Vincent with transitions
+    
+    transitions = (
+            (1, C, 2, C),
```
marcharper: Note that this comment is mostly copied from the 10 state strategy and isn't updated for the transitions below.

dashiellfryer: We started with the 10 state FSM, but after 500 generations the output was an 8 state given in the class below. You are right, I did copy the comment for a consistent look.

drvinceknight:
> @drvinceknight FYI there appears to be a latent docs build error.

Looking at the stack race I think it's to do with the doctoring of the strategy:

```
Running Sphinx v3.4.3
100

101
Configuration error:
102
There is a syntax error in your configuration file: unindent does not match any outer indentation level (finite_state_machines.py, line 1028)
103

104
Makefile:53: recipe for target 'html' failed
``` 

This will be a silly thing like a level of indentation or something like that, I see that @dashiellfryer has closed this PR though, if that's a "miss - click" I'll look in to what's wrong with the doctstring and point out exactly what needs to be tweaked :)

marcharper:
@dashiellfryer This is probably easily fixable, did you want to reopen? I'll add a commit to get the tests running correctly.

dashiellfryer:
@marcharper I closed the commit to fix the errors, but if you can fix it easily, please do. I think much of the issue is just my novice status with git and github.

axelrod/strategies/finite_state_machines.py
```diff
@@ -1024,3 +1024,71 @@ def __init__(self) -> None:
...
+            (6, D, 7, D),
+            (7, C, 5, C),
+            (7, D, 8, D),
+            (8, C, 5, D),
+            (8, D, 5, C),
+            (9, C, 2, C),
+            (9, D, 10, C),
+            (10, C, 2, C),
+            (10, D, 4, D)
+    )
```
drvinceknight:
```suggestion
    Evolved from the 10 state FSM designed by Frederick Vincent.
```

drvinceknight (changes requested):
So the documentation fails to build because of something to do with the `transitions = ...`.

I started trying to tweak the indentation to get it to work but actually I suggest removing that altogether from the documentation as it's essentially self documenting in the code itself. I believe that once my suggestion is accepted the docs should build.

Perhaps we could add more about the `10 state FSM designed by Frederick Vincent`:

- is this a particular 10 state FSM that's in a paper somewhere?
- what was the objective function for the evolutionary algorithm?

dashiellfryer (commented):
Frederick and I are writing the paper about the 10 state FSM. I just wanted dibs on top spot!

marcharper:
I'm updating the tests on PR #1383. The tests revealed that two of the states were inaccessible (0 and 1) so I removed them and renamed it "Evolved FSM 6".

Can you give us any more info on the training so we can add it to the doc string?

dashiellfryer:
The objective function used was the default; score.

marcharper:
@dashiellfryer I should have been more explicit. Good to know the objective was the score default. Also -- what version of axelrod did you use? I presume you trained against the short_run_time strategies (the default).

Otherwise PTAL at #1383 and let me know if it looks good to you (the strategy renaming, for example).

dashiellfryer:
I am great with the name change and I should have had the student(s) check for accessible branches. Sorry @marcharper, I was answering @drvinceknight's question. Below, you can find all the specifics.

We used all the default options for fsm_evolve except for the output_file, num_states (10 instead of 8), and generations (500 instead of 50). I am running axelrod-dojo version 0.0.8 and axelrod version 4.10.0. So yes, we trained against short_run_time_strategies for 500 generations, population size of 40, mutation rate at 0.1, bottleneck at 10, processes at 0 (in Ubuntu), objective set to score, 200 turns, 0 noise. 

Last semester we found the easiest way to invoke training with a specific starting point was to add transitions directly to prepare_player_class_kwargs with everything else set in Options and using the EvolvableFSMPlayer provided by @marcharper as the player class. 

fsm_evolve

marcharper:
Thanks! I'll update the docstring and then I think this one is good to go in.

> I should have had the student(s) check for accessible branches

No worries, that's why the tests are there :) We could update the evolutionary algorithms to maintain the number of connected states , or memory depth, if that's of interest. (FYI I'll be folding the evolutionary algorithms into the main library soon.)

PR #1384, opened 2021-03-26
Update citations.md

marcharper:


meatballs (approved):

drvinceknight (approved):
Looks great to me, can't imagine what would break but merge once CI is happy :+1:

drvinceknight:
CI is failing on type checking. We don't pin `mypy` install in the CI:

```
python -m pip install mypy
```

Suggestion:

Change the above (here https://github.com/Axelrod-Python/Axelrod/blob/dev/.github/workflows/config.yml#L48) to:

```
python -m pip install mypy==<version that passes>
```

and open an issue to fix/modify at a later date (perhaps it's time for a `requirements_dev.txt` ?).

marcharper:
It's an unrelated type hint issue due to numpy 1.2.0.

```
axelrod/strategies/ann.py:125: error: Incompatible types in assignment (expression has type "ndarray", variable has type "List[int]")
```

drvinceknight:
I've opened https://github.com/Axelrod-Python/Axelrod/pull/1385 that fixes the issues for me locally.

Issue #1386, opened 2021-05-21
"MetaClass update_histories issue" when running all strategies


xjcl:
How to reproduce:

```bash
git clone https://github.com/Axelrod-Python/Axelrod.git
cd Axelrod
python3.9 setup.py install
git checkout ebcdf99e   # current HEAD on dev branch
python3.9 --version   # Python 3.9.5
```

```python
>>> import axelrod as axl
>>> players = [s() for s in axl.all_strategies]
>>> tournament = axl.Tournament(players, seed=1, turns=10)
>>> results = tournament.play()
```

Issue seems to occur always at 71%, no matter how many turns (so you can try with less for quicker debugging)

```none
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/Axelrod-4.10.0-py3.9.egg/axelrod/strategies/meta.py", line 97, in update_histories
    for player, play in zip(self.team, self._last_results):
TypeError: 'NoneType' object is not iterable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/dist-packages/Axelrod-4.10.0-py3.9.egg/axelrod/tournament.py", line 147, in play
    self._run_serial(build_results=build_results)
  File "/usr/local/lib/python3.9/dist-packages/Axelrod-4.10.0-py3.9.egg/axelrod/tournament.py", line 176, in _run_serial
    results = self._play_matches(chunk, build_results=build_results)
  File "/usr/local/lib/python3.9/dist-packages/Axelrod-4.10.0-py3.9.egg/axelrod/tournament.py", line 456, in _play_matches
    match.play()
  File "/usr/local/lib/python3.9/dist-packages/Axelrod-4.10.0-py3.9.egg/axelrod/match.py", line 195, in play
    plays = self.simultaneous_play(
  File "/usr/local/lib/python3.9/dist-packages/Axelrod-4.10.0-py3.9.egg/axelrod/match.py", line 157, in simultaneous_play
    coplayer.update_history(s2, s1)
  File "/usr/local/lib/python3.9/dist-packages/Axelrod-4.10.0-py3.9.egg/axelrod/strategies/meta.py", line 117, in update_history
    self.update_histories(coplay)
  File "/usr/local/lib/python3.9/dist-packages/Axelrod-4.10.0-py3.9.egg/axelrod/strategies/meta.py", line 107, in update_histories
[...]

PR #1388, opened 2021-05-24
Remove cheating strategies

drvinceknight:
This follows a bug reported here: #1386

As mentioned by @marcharper:

> They are fun but cause much more trouble than they are fun.

It was decided on the issue to move these "cheating" strategies
elsewhere.

Once this PR is merged I will open an issue to capture that.


meatballs (approved):

drvinceknight:
For clarity this ends up removing the following "cheating" files: `mindreader.py`, `mindcontrol.py` and `geller.py` (even though `geller.py` was not causing any of the problems in #1386). I also ended up removing  bunch of functions from `strategy_utils.py` as they were only used by those strategies. Not entirely sure if I've done "too" much removing.

Once this is in I'll open an issue to make sure we keep an eye on what is gone so that we can decide where to put it.

PR #1393, opened 2021-08-03
add a new strategy CAPRI

yohm:
I added a new strategy 'CAPRI', which is proposed in the following paper.
https://www.nature.com/articles/s41598-020-73855-x

Would you please review and let me know if you find something that needs to be revised? Thank you.

### DIFF
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -143,7 +143,7 @@ from .grudger import (
     OppositeGrudger,
     SoftGrudger,
     SpitefulCC,
-    CAPRI,
+    Capri,
 )
 from .grumpy import Grumpy
 from .handshake import Handshake
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -143,6 +143,7 @@ from .grudger import (
     OppositeGrudger,
     SoftGrudger,
     SpitefulCC,
+    CAPRI,
 )
 from .grumpy import Grumpy
 from .handshake import Handshake
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -135,6 +135,7 @@ from .gobymajority import (
 from .gradualkiller import GradualKiller
 from .grudger import (
     Aggravater,
+    Capri,
     EasyGo,
     ForgetfulGrudger,
     GeneralSoftGrudger,
@@ -143,7 +144,6 @@ from .grudger import (
     OppositeGrudger,
     SoftGrudger,
     SpitefulCC,
-    Capri,
 )
 from .grumpy import Grumpy
 from .handshake import Handshake
@@ -304,6 +304,7 @@ all_strategies = [
     Bully,
     BushMosteller,
     Calculator,
+    Capri,
     CautiousQLearner,
     CollectiveStrategy,
     ContriteTitForTat,
--- a/axelrod/strategies/grudger.py
+++ b/axelrod/strategies/grudger.py
@@ -346,16 +346,20 @@ class SpitefulCC(Player):
 
 class CAPRI(Player):
     """
-    CAPRI is a strategy whose behavior is defined by the following five rules.
-    C: Cooperate at mutual cooperation
-    A: Accept punishmentl
-    P: Punish
-    R: Recover
-    I: In all ohter cases, defect.
+    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined by the following five rules.
+    C: Cooperate at mutual cooperation.
+        This rule prescribes c at (ccc, ccc).
+    A: Accept punishment when you mistakenly defected from mutual cooperation.
+        This rule prescribes c at (ccd, ccc), (cdc, ccd), (dcc, cdc), and (ccc, dcc).
+    P: Punish your co-player by defecting once when he defected from mutual cooperation.
+        This rule prescribes d at (ccc, ccd), and then c at (ccd, cdc), (cdc, dcc), and (dcc, ccc).
+    R: Recover cooperation when you or your co-player cooperated at mutual defection.
+        This rule prescribes c at (ddd, ddc), (ddc, dcc), (dcc, ccc), (ddc, ddd), (dcc, ddc), (ccc, dcc), (ddc, ddc), and (dcc, dcc).
+    I: In all the ohter cases, defect.
 
     Names:
 
-    - CAPRI: Original Name by Y. Murase et al. [Murase 2020]
+    - CAPRI: Original Name by Y. Murase et al. [Murase2020]_
     """
 
     name = "CAPRI"
@@ -369,6 +373,7 @@ class CAPRI(Player):
     }
 
     def strategy(self, opponent: Player) -> Action:
+        # initial history profile is full cooperation
         hist = [[C,C],[C,C],[C,C]]
         for i in range(min(len(self.history), 3)):
             hist[-i-1][0] = self.history[-i-1]
--- a/axelrod/strategies/grudger.py
+++ b/axelrod/strategies/grudger.py
@@ -346,7 +346,7 @@ class SpitefulCC(Player):
 
 class Capri(Player):
     """
-    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined by the following five rules.
+    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined in[Murase2020]_ by the following five rules applied to the last 3 moves of the player and the opponent:
     C: Cooperate at mutual cooperation.
         This rule prescribes c at (ccc, ccc).
     A: Accept punishment when you mistakenly defected from mutual cooperation.
--- a/axelrod/strategies/grudger.py
+++ b/axelrod/strategies/grudger.py
@@ -374,7 +374,7 @@ class Capri(Player):
 
     def strategy(self, opponent: Player) -> Action:
         # initial history profile is full cooperation
-        hist = [[C,C],[C,C],[C,C]]
+        hist = [[C, C], [C, C], [C, C]]
         for i in range(min(len(self.history), 3)):
             hist[-i-1][0] = self.history[-i-1]
             hist[-i-1][1] = opponent.history[-i-1]
--- a/axelrod/strategies/grudger.py
+++ b/axelrod/strategies/grudger.py
@@ -342,3 +342,70 @@ class SpitefulCC(Player):
         elif opponent.defections:
             return D
         return C
+
+
+class CAPRI(Player):
+    """
+    CAPRI is a strategy whose behavior is defined by the following five rules.
+    C: Cooperate at mutual cooperation
+    A: Accept punishmentl
+    P: Punish
+    R: Recover
+    I: In all ohter cases, defect.
+
+    Names:
+
+    - CAPRI: Original Name by Y. Murase et al. [Murase 2020]
+    """
+
+    name = "CAPRI"
+    classifier = {
+        "memory_depth": 3,
+        "stochastic": False,
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        hist = [[C,C],[C,C],[C,C]]
+        for i in range(min(len(self.history), 3)):
+            hist[-i-1][0] = self.history[-i-1]
+            hist[-i-1][1] = opponent.history[-i-1]
+
+        if hist == [[C,C],[C,C],[C,C]]:   # Rule: C
+            return C
+        elif hist == [[C,C],[C,C],[D,C]]:  # Rule: A
+            return C
+        elif hist == [[C,C],[D,C],[C,D]]:
+            return C
+        elif hist == [[D,C],[C,D],[C,C]]:
+            return C
+        elif hist == [[C,D],[C,C],[C,C]]:
+            return C
+        elif hist == [[C,C],[C,C],[C,D]]:  # Rule: P
+            return D
+        elif hist == [[C,C],[C,D],[D,C]]:
+            return C
+        elif hist == [[C,D],[D,C],[C,C]]:
+            return C
+        elif hist == [[D,C],[C,C],[C,C]]:
+            return C
+        elif hist == [[D,D],[D,D],[D,C]]:  # Rule: R1
+            return C
+        elif hist == [[D,D],[D,C],[C,C]]:
+            return C
+        elif hist == [[D,C],[C,C],[C,C]]:
+            return C
+        elif hist == [[D,D],[D,D],[C,D]]:  # Rule: R2
+            return C
+        elif hist == [[D,D],[C,D],[C,C]]:
+            return C
+        elif hist == [[C,D],[C,C],[C,C]]:
+            return C
+        elif hist == [[D,D],[D,D],[C,C]]:  # Rule: R3
+            return C
+        elif hist == [[D,D],[C,C],[C,C]]:
+            return C
+        return D
--- a/axelrod/strategies/grudger.py
+++ b/axelrod/strategies/grudger.py
@@ -346,16 +346,24 @@ class SpitefulCC(Player):
 
 class Capri(Player):
     """
-    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined in[Murase2020]_ by the following five rules applied to the last 3 moves of the player and the opponent:
-    C: Cooperate at mutual cooperation.
-        This rule prescribes c at (ccc, ccc).
-    A: Accept punishment when you mistakenly defected from mutual cooperation.
-        This rule prescribes c at (ccd, ccc), (cdc, ccd), (dcc, cdc), and (ccc, dcc).
-    P: Punish your co-player by defecting once when he defected from mutual cooperation.
-        This rule prescribes d at (ccc, ccd), and then c at (ccd, cdc), (cdc, dcc), and (dcc, ccc).
-    R: Recover cooperation when you or your co-player cooperated at mutual defection.
-        This rule prescribes c at (ddd, ddc), (ddc, dcc), (dcc, ccc), (ddc, ddd), (dcc, ddc), (ccc, dcc), (ddc, ddc), and (dcc, dcc).
-    I: In all the ohter cases, defect.
+    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is
+    defined by the following five rules applied to the last 3 moves of the
+    player and the opponent:
+
+    - C: Cooperate at mutual cooperation.  This rule prescribes c at (ccc, ccc).
+    - A: Accept punishment when you mistakenly defected from mutual cooperation.
+      This rule prescribes c at (ccd, ccc), (cdc, ccd), (dcc, cdc), and (ccc,
+      dcc).
+    - P: Punish your co-player by defecting once when he defected from mutual
+      cooperation.  This rule prescribes d at (ccc, ccd), and then c at (ccd,
+      cdc), (cdc, dcc), and (dcc, ccc).
+    - R: Recover cooperation when you or your co-player cooperated at mutual
+      defection.  This rule prescribes c at (ddd, ddc), (ddc, dcc), (dcc, ccc),
+      (ddc, ddd), (dcc, ddc), (ccc, dcc), (ddc, ddc), and (dcc, dcc).
+    - I: In all the other cases, defect.
+
+    The original implementation used in [Murase2020]_ is available at
+    https://github.com/yohm/sim_exhaustive_m3_PDgame
 
     Names:
 
--- a/axelrod/strategies/grudger.py
+++ b/axelrod/strategies/grudger.py
@@ -344,7 +344,7 @@ class SpitefulCC(Player):
         return C
 
 
-class CAPRI(Player):
+class Capri(Player):
     """
     CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined by the following five rules.
     C: Cooperate at mutual cooperation.
--- a/axelrod/strategies/grudger.py
+++ b/axelrod/strategies/grudger.py
@@ -394,7 +394,7 @@ class Capri(Player):
             return C
         if hist == [(D, C), (C, D), (C, C)]:
             return C
-        if hist == [(C, D), (C, C), (C, C)]:
+        if hist == [(C, D), (C, C), (C, C)]:  # Rule: A & R2
             return C
         if hist == [(C, C), (C, C), (C, D)]:  # Rule: P
             return D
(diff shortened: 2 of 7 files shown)


axelrod/strategies/grudger.py
```diff
@@ -342,3 +342,75 @@ def strategy(opponent: Player) -> Action:
         elif opponent.defections:
             return D
         return C
+
+
+class CAPRI(Player):
```
drvinceknight:
```suggestion
class Capri(Player):
```

This follows PEP8: the `name` of the strategy can still be `CAPRI` which is in line with other similar strategies.

axelrod/strategies/grudger.py
```diff
@@ -342,3 +342,75 @@ def strategy(opponent: Player) -> Action:
...
+    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined by the following five rules.
+    C: Cooperate at mutual cooperation.
+        This rule prescribes c at (ccc, ccc).
+    A: Accept punishment when you mistakenly defected from mutual cooperation.
+        This rule prescribes c at (ccd, ccc), (cdc, ccd), (dcc, cdc), and (ccc, dcc).
+    P: Punish your co-player by defecting once when he defected from mutual cooperation.
+        This rule prescribes d at (ccc, ccd), and then c at (ccd, cdc), (cdc, dcc), and (dcc, ccc).
+    R: Recover cooperation when you or your co-player cooperated at mutual defection.
+        This rule prescribes c at (ddd, ddc), (ddc, dcc), (dcc, ccc), (ddc, ddd), (dcc, ddc), (ccc, dcc), (ddc, ddc), and (dcc, dcc).
+    I: In all the ohter cases, defect.
```
drvinceknight: These rules are not immediately clear to me. If it is verbatim from the paper (which I have not read yet) then that should be made clear and some clarification should be added.

axelrod/strategies/grudger.py
```diff
@@ -342,3 +342,75 @@ def strategy(opponent: Player) -> Action:
...
+        "stochastic": False,
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        # initial history profile is full cooperation
+        hist = [[C,C],[C,C],[C,C]]
```
drvinceknight:
```suggestion
        hist = [[C, C], [C, C], [C, C]]
```

PEP8.

axelrod/strategies/grudger.py
```diff
@@ -342,3 +342,75 @@ def strategy(opponent: Player) -> Action:
...
+        elif hist == [[D,D],[D,D],[C,D]]:  # Rule: R2
+            return C
+        elif hist == [[D,D],[C,D],[C,C]]:
+            return C
+        elif hist == [[C,D],[C,C],[C,C]]:
+            return C
+        elif hist == [[D,D],[D,D],[C,C]]:  # Rule: R3
+            return C
+        elif hist == [[D,D],[C,C],[C,C]]:
+            return C
```
drvinceknight:
```suggestion
        if hist == [[C, C], [C, C], [D, C]]:  # Rule: A
            return C
        if hist == [[C, C], [D, C], [C, D]]:
            return C
        if hist == [[D, C], [C, D], [C, C]]:
            return C
        if hist == [[C, D], [C, C], [C, C]]:
            return C
        if hist == [[C, C], [C, C], [C, D]]:  # Rule: P
            return D
        if hist == [[C, C], [C, D], [D, C]]:
            return C
        if hist == [[C, D], [D, C], [C, C]]:
            return C
        if hist == [[D, C], [C, C], [C, C]]:
            return C
        if hist == [[D, D], [D, D], [D, C]]:  # Rule: R1
            return C
        if hist == [[D, D], [D, C], [C, C]]:
            return C
        if hist == [[D, C], [C, C], [C, C]]:
            return C
        if hist == [[D, D], [D, D], [C, D]]:  # Rule: R2
            return C
        if hist == [[D, D], [C, D], [C, C]]:
            return C
        if hist == [[C, D], [C, C], [C, C]]:
            return C
        if hist == [[D, D], [D, D], [C, C]]:  # Rule: R3
            return C
        if hist == [[D, D], [C, C], [C, C]]:
            return C
```

axelrod/tests/strategies/test_grudger.py
```diff
@@ -336,3 +336,69 @@ def test_strategy(self):
         opponent = axl.MockPlayer(actions=opponent_actions)
         actions = [(C, D)] * 2 + [(D, D)] * 18 + [(D, C)] * 20
         self.versus_test(opponent, expected_actions=actions)
+
+
+class TestCAPRI(TestPlayer):
```
drvinceknight:
```suggestion
class TestCapri(TestPlayer):
```

axelrod/tests/strategies/test_grudger.py
```diff
@@ -336,3 +336,69 @@ def test_strategy(self):
...
+
+    name = "CAPRI"
+    player = axl.CAPRI
```
drvinceknight:
```suggestion
    player = axl.Capri
```

axelrod/strategies/grudger.py
drvinceknight:
```suggestion
        if hist == [[C, C], [C, C], [D, C]]:  # Rule: A
            return C
        if hist == [[C, C], [D, C], [C, D]]:
            return C
        if hist == [[D, C], [C, D], [C, C]]:
            return C
        if hist == [[C, D], [C, C], [C, C]]:
            return C
        if hist == [[C, C], [C, C], [C, D]]:  # Rule: P
            return D
        if hist == [[C, C], [C, D], [D, C]]:
            return C
        if hist == [[C, D], [D, C], [C, C]]:
            return C
        if hist == [[D, C], [C, C], [C, C]]:
            return C
        if hist == [[D, D], [D, D], [D, C]]:  # Rule: R1
            return C
        if hist == [[D, D], [D, C], [C, C]]:
            return C
        if hist == [[D, C], [C, C], [C, C]]:
            return C
        if hist == [[D, D], [D, D], [C, D]]:  # Rule: R2
            return C
        if hist == [[D, D], [C, D], [C, C]]:
            return C
        if hist == [[C, D], [C, C], [C, C]]:
            return C
        if hist == [[D, D], [D, D], [C, C]]:  # Rule: R3
            return C
        if hist == [[D, D], [C, C], [C, C]]:
            return C
```

axelrod/strategies/_strategies.py
```diff
@@ -143,6 +143,7 @@
     OppositeGrudger,
     SoftGrudger,
     SpitefulCC,
+    CAPRI,
```
drvinceknight:
```suggestion
    Capri,
```

drvinceknight (changes requested):
Thanks for this, I have made some initial stylistic comments but need to read the paper to offer a comprehensive review (planning on doing that soon as it looks interesting).

A comment: it does look like this strategy is a specific case of a `axl.Lookerup` (https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/strategies/lookerup.py#L236) so it would be possible to implement this as a particular case which could help clear things up.

drvinceknight:
The tests are currently failing as the documentation does not build because of the way you have formatted your doctstring:


```
Warning, treated as error:
172
/home/runner/work/Axelrod/Axelrod/axelrod/strategies/grudger.py:docstring of axelrod.strategies.grudger.CAPRI:3:Unexpected indentation.
173
make: *** [Makefile:53: html] Error 2
```

axelrod/strategies/grudger.py
marcharper: (The rules are from the paper)

yohm:
> Thanks for this, I have made some initial stylistic comments but need to read the paper to offer a comprehensive review (planning on doing that soon as it looks interesting).
> 
> A comment: it does look like this strategy is a specific case of a `axl.Lookerup` (https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/strategies/lookerup.py#L236) so it would be possible to implement this as a particular case which could help clear things up.

Since all the permutations must be specified to use `axl.Lookerup`, can I keep the implementation in the current form?

yohm:
> The tests are currently failing as the documentation does not build because of the way you have formatted your doctstring:
> 
> ```
> Warning, treated as error:
> 172
> /home/runner/work/Axelrod/Axelrod/axelrod/strategies/grudger.py:docstring of axelrod.strategies.grudger.CAPRI:3:Unexpected indentation.
> 173
> make: *** [Makefile:53: html] Error 2
> ```

Thank you for pointing that out. Could you tell me how to test the docstring?

axelrod/strategies/grudger.py
```diff
@@ -342,3 +342,75 @@ def strategy(opponent: Player) -> Action:
...
+    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined by the following five rules.
+    C: Cooperate at mutual cooperation.
+        This rule prescribes c at (ccc, ccc).
+    A: Accept punishment when you mistakenly defected from mutual cooperation.
+        This rule prescribes c at (ccd, ccc), (cdc, ccd), (dcc, cdc), and (ccc, dcc).
+    P: Punish your co-player by defecting once when he defected from mutual cooperation.
+        This rule prescribes d at (ccc, ccd), and then c at (ccd, cdc), (cdc, dcc), and (dcc, ccc).
+    R: Recover cooperation when you or your co-player cooperated at mutual defection.
+        This rule prescribes c at (ddd, ddc), (ddc, dcc), (dcc, ccc), (ddc, ddd), (dcc, ddc), (ccc, dcc), (ddc, ddc), and (dcc, dcc).
+    I: In all the ohter cases, defect.
```
drvinceknight: > Thank you for pointing that out. Could you tell me how to test the docstring?

Of course: top build the documentation, `cd docs` (go to the `docs` directory) and run:

```
make html
```

You might need to `pip` install some dependencies.

I have checked locally and below fixes the sphinx build problems:

```suggestion
    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is
    defined by the following five rules.

    - C: Cooperate at mutual cooperation.  This rule prescribes c at (ccc, ccc).
    - A: Accept punishment when you mistakenly defected from mutual cooperation.
      This rule prescribes c at (ccd, ccc), (cdc, ccd), (dcc, cdc), and (ccc,
      dcc).
    - P: Punish your co-player by defecting once when he defected from mutual
      cooperation.  This rule prescribes d at (ccc, ccd), and then c at (ccd,
      cdc), (cdc, dcc), and (dcc, ccc).
    - R: Recover cooperation when you or your co-player cooperated at mutual
      defection.  This rule prescribes c at (ddd, ddc), (ddc, dcc), (dcc, ccc),
      (ddc, ddd), (dcc, ddc), (ccc, dcc), (ddc, ddc), and (dcc, dcc).
    - I: In all the ohter cases, defect.

    Names:

    - CAPRI: Original Name by Y. Murase et al. [Murase2020]_

```

drvinceknight:
> Since all the permutations must be specified to use `axl.Lookerup`, can I keep the implementation in the current form?

Sure, I don't feel strongly, this could always be refactored at a later date.

axelrod/strategies/grudger.py
```diff
@@ -342,3 +342,75 @@ def strategy(opponent: Player) -> Action:
         elif opponent.defections:
             return D
         return C
+
+
+class Capri(Player):
+    """
+    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined by the following five rules.
```
drvinceknight: I've read the paper now (sorry for taking so long!) which I enjoyed, I suggest the (very minor) following clarification re the description:

```suggestion
    CAPRI is a memory-3 strategy proposed in [Murase2020]_. Its behavior is defined in[Murase2020]_ by the following five rules applied to the last 3 moves of the player and the opponent:
```

I realise this is implied by the `memory-3` comment.

axelrod/tests/strategies/test_grudger.py
```diff
@@ -336,3 +336,69 @@ def test_strategy(self):
...
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+    def assert_prescription(self, player_history, opponent_history, prescription):
+        player = self.player()
+        player._history = player_history
+        opponent = axl.MockPlayer()
+        opponent._history = opponent_history
+        self.assertEqual(player.strategy(opponent=opponent), prescription)
```
drvinceknight: We have moved away from directly setting the `._history` like this in tests would you be able to write this using the `MockPlayer` and a sequence of actions?

This would have the beneficial side effect of identifying if any pairs of histories cannot happen (for some reason or another).

drvinceknight:
Everything is looking really good, I've left a couple of suggestions/comments, let me know if I can assist with anything.

FYI, re my comment about implementing this as a `LookerUp` you could also implement it as a particular `FSM` which would correct to Figure 2 (a) of the paper. Not at all suggesting you need to do this, I just liked that you added that in the paper :) (I like FSMs :))

axelrod/strategies/grudger.py
marcharper: nit: ohter -> other on line 362

marcharper:
[No action required] Based on Figure 2 in the paper, it looks like the strategy could also be implemented as a finite state machine.
[Edit: not sure how I missed @drvinceknight's comment on this above!]

FYI the original implementation is [here](https://github.com/yohm/sim_exhaustive_m3_PDgame), perhaps that's worth documenting. (I don't see any tests though so I'm not sure if the code would be more helpful than the paper descriptions.)

axelrod/tests/strategies/test_grudger.py
yohm: How can we test the states reachable only by noise?
For instance, CAPRI prescribes `C` when the last three moves are `(CCD, CCC)`. However, this history is reachable only when the player erroneously committed defection.

axelrod/strategies/grudger.py
yohm: Thank you for your help. I confirmed that the docstring is compiled without error in my environment.

yohm:
> Everything is looking really good, I've left a couple of suggestions/comments, let me know if I can assist with anything.

Thank you for your help.

> FYI, re my comment about implementing this as a LookerUp you could also implement it as a particular FSM which would correct to Figure 2 (a) of the paper. Not at all suggesting you need to do this, I just liked that you added that in the paper :) (I like FSMs :))

If you like FSM, you might find the following paper interesting :)
A method to translate LookerUp strategies into FSM is proposed.
https://www.nature.com/articles/s41598-020-70281-x

yohm:
> FYI the original implementation is [here](https://github.com/yohm/sim_exhaustive_m3_PDgame), perhaps that's worth documenting. (I don't see any tests though so I'm not sure if the code would be more helpful than the paper descriptions.)

I added a link to the repository in the docstring. Thank you for your suggestion.

axelrod/tests/strategies/test_grudger.py
drvinceknight: That's indeed non trivial but also a good reason to test the "correct way". To get the wanted behaviour use seeded noisy `versus_tests` so something like:

```
self.versus_test(axelrod.Alternator(), expected_actions=actions,
                 match_attributes={"noise": .7}, seed=5)
```

Then it's a question of finding a seed that goes down the correct branch (which I realise is a bit of a pain -- let us know if we can assist).

drvinceknight:
> If you like FSM, you might find the following paper interesting :)
> A method to translate LookerUp strategies into FSM is proposed.
> https://www.nature.com/articles/s41598-020-70281-x

That looks very cool indeed, have added it to my reading list. @marcharper and @gaffney2010 designed an algorithm for finding the memory length of arbitrary FSMs (https://arxiv.org/abs/1912.04493) which is implemented in the library: https://axelrod.readthedocs.io/en/latest/discussion/strategy_archetypes.html?highlight=finite%20state%20machines#finite-state-machines implementing your algorithm to obtain the FSM from the LookerUp could be neat as well.

axelrod/strategies/grudger.py
```diff
@@ -342,3 +342,83 @@ def strategy(opponent: Player) -> Action:
...
+        if hist == [[C, D], [D, C], [C, C]]:
+            return C
+        if hist == [[D, C], [C, C], [C, C]]:
+            return C
+        if hist == [[D, D], [D, D], [D, C]]:  # Rule: R1
+            return C
+        if hist == [[D, D], [D, C], [C, C]]:
+            return C
+        if hist == [[D, C], [C, C], [C, C]]:
+            return C
```
drvinceknight, line 412: The CI is now failing due to coverage (which checks that all lines of code are hit by at least one test):


```
axelrod/strategies/grudger.py                                      160      2    99%   413, 419
```

This line (line 413) is not hit and also line 419.

axelrod/tests/strategies/test_grudger.py
yohm: After I struggled for a while, I find it difficult to find a proper seed. Could you help me find a proper seed?
It would be great if we can specify when the actions are flipped instead of specifying `noise` and `seed` in a future version of Axelrod library. For instance, I would like an interface like

```python
self.versus_test(axelrod.Alternator(), expected_actions=actions, flip=[0,0,0,1,0,0,0,0] )
```
to indicate that the action of the focal player is flipped in the fourth round of the match. It will also improve the readability of the test code.

axelrod/strategies/grudger.py
yohm: These lines are never used because the conditions are duplicates of the previous lines. I commented out these lines.

axelrod/tests/strategies/test_grudger.py
marcharper: I usually use a loop:
```
for seed in range(0, 10000):
  # setup Match with seed
  # check for desired conditions, break and print seed
```

There's a good reason we do things this way -- playing through a `Match` can affect the internal variables of a strategy and lead to subtle bugs.

axelrod/strategies/grudger.py
```diff
@@ -342,3 +342,83 @@ def strategy(opponent: Player) -> Action:
...
+        "stochastic": False,
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        # initial history profile is full cooperation
+        hist = [[C, C], [C, C], [C, C]]
```
marcharper: It might be more readable to do something like:
```
hist = zip(self.history[-3:], opponent.history[-3:])
while len(hist) < 3:
    hist.push(0, (C, C))
```

yohm: Thank you for your suggestion. I polished up the code with 2252c88

axelrod/tests/strategies/test_grudger.py
yohm: Thanks! Actually, I found a useful method `TestMatch#search_seeds` which does exactly what you suggested.
I rewrote the test cases specifying seeds. See 9c2c8f2d

marcharper: Great! I should have remembered that method, I'm pretty sure I wrote it!

Nikoleta-v3:
👋🏻 @yohm! 

The CI is failing due to coverage:

`axelrod/tests/strategies/test_grudger.py                           200      5    98%   356-360`

because you are not using the method `assert_prescription` anymore.

Nikoleta-v3:
Another comment I had was regarding https://github.com/Axelrod-Python/Axelrod/pull/1393/commits/35d1890db1bdb5cd49e581ab33544fc26ab5613e

I would delete these lines instead of commenting them out. Then you could add comments like,

```
if hist == [(D, C), (C, C), (C, C)]:  # Rule P and rule R1
    return(C)
```

or discuss in the docstring that in the implementation of the strategy we don’t need to check all the conditions for each rule. 


p.s. Thank you for sharing this  https://www.nature.com/articles/s41598-020-70281-x!

marcharper:
Almost there @yohm !

yohm:
@Nikoleta-v3 @marcharper Thank you for your comments. The code was revised according to your advice. See 3ff428b1

Nikoleta-v3 (approved):

axelrod/strategies/_strategies.py
```diff
@@ -143,6 +143,7 @@
     OppositeGrudger,
     SoftGrudger,
     SpitefulCC,
+    Capri,
```
drvinceknight: Can you move this so that the tuple is in alphabetical order please and then `Capri` needs to be added to the `all_strategies` list in this file. (This is why the CI is failing).

drvinceknight (changes requested):

axelrod/strategies/_strategies.py
yohm: Fixed.

drvinceknight:
The CI is failing because one of the meta strategies' behaviours has now changed (as it is using `Capri`). Could you change line 635 of `test_meta.py` so that `actions` is:

```python
        actions = [(C, C), (C, D), (C, C), (D, D), (D, C)]
```

(Previously it was:

```python
        actions = [(C, C), (C, D), (D, C), (C, D), (D, C)]
```

)

I have made the change locally and checked that the `test_meta` tests pass.

yohm:
> The CI is failing because one of the meta strategies' behaviours has now changed (as it is using `Capri`). Could you change line 635 of `test_meta.py` so that `actions` is:
> 
> ```python
>         actions = [(C, C), (C, D), (C, C), (D, D), (D, C)]
> ```
> 
> (Previously it was:
> 
> ```python
>         actions = [(C, C), (C, D), (D, C), (C, D), (D, C)]
> ```
> 
> )
> 
> I have made the change locally and checked that the `test_meta` tests pass.

Hopefully, it is fixed now.

marcharper:
A couple of doc tests are failing, just need to bump up the counts:

```
FAIL: ./docs/index.rst
Doctest: index.rst
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/doctest.py", line 2204, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for index.rst
  File "./docs/index.rst", line 0

File "./docs/index.rst", line 55, in index.rst
Failed example:
    len(axl.strategies)
Expected:
    238
Got:
    239


FAIL: ./docs/how-to/classify_strategies.rst
Doctest: classify_strategies.rst
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/doctest.py", line 2204, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for classify_strategies.rst
  File "./docs/how-to/classify_strategies.rst", line 0

File "./docs/how-to/classify_strategies.rst", line 81, in classify_strategies.rst
Failed example:
    len(strategies)
Expected:
    55
Got:
    56


```

marcharper:
Looks like the last failing test is due to `black` -- the formatting `axelrod/tests/strategies/test_grudger.py` is not what it wants. See [here](https://github.com/Axelrod-Python/Axelrod/blob/75ef1f24187350292c43d244370c100c644748bc/docs/how-to/contributing/guidelines.rst) for the command to reformat.

drvinceknight (approved):

drvinceknight:
This looks like everything is fixed now! Thanks for going through it all @yohm :+1:

Issue #1395, opened 2021-08-23
Links to contributing guide broken by docs restructure


RobPDev:
A recent change (#1391) restructured the documentation and including moving the contributing guide from tutorials to how-to. However, the links to the contributing guide (in `README.rst` and `CONTRIBUTING.rst`) weren't updated to match the new location.

PR #1396, opened 2021-08-23
Fix contributing guide links #1395

RobPDev:
I also converted the links to HTTPS and as the `.org` resolved to `.io` I changed the README to point to `.io`.

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -45,7 +45,7 @@ jobs:
         python doctests.py
     - name: Run static type checker
       run: |
-        python -m pip install mypy
+        python -m pip install mypy types-setuptools
         python run_mypy.py
     - name: Check imports are sorted
       run: |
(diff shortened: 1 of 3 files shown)


RobPDev:
I've tried running the failing test locally, using python 3.8 it passes if `types-setuptools` is installed.

drvinceknight:
Thanks for this, your fixes certainly shouldn't affect the type hints so I'm guessing this might be due to an update of `mypy` (?). Could you add `types-setuptools` to this line of the workflow config: https://github.com/Axelrod-Python/Axelrod/blob/dev/.github/workflows/config.yml#L48 to check if that fixes things?

RobPDev:
I've done a bit of testing, `mypy==0.812` passes, `mypy==0.900` requires `types-setuptools`.

`mypy` 0.900 [removed library stubs](https://mypy-lang.blogspot.com/2021/06/mypy-0900-released.html).

drvinceknight:
This looks like it's sorted the problem. Thanks @RobPDev :+1:

drvinceknight (approved):

RobPDev:
Just noticed this PR hasn't linked to the issue.
Fixes #1395

PR #1399, opened 2022-02-27
Add a thesis to citations

marcharper:


marcharper:
Error appears unrelated (black wants to reformat some files).

Nikoleta-v3:
For some citations we use brackets for the publication year and for some we don't 🤷🏻 Do we care about this? Do we want to be consistent?

drvinceknight:
> Error appears unrelated (black wants to reformat some files).

This could be due to an update of black.

drvinceknight:
Have opened https://github.com/Axelrod-Python/Axelrod/pull/1400 which is a PR to this branch with an updated version of black. (Do we want to pin black?)

drvinceknight:
This looks like hypothesis has stumbled on an edge case?

drvinceknight:
I believe this is good to go.

PR #1400, opened 2022-03-01
Run black.

drvinceknight:
This is a PR to https://github.com/Axelrod-Python/Axelrod/pull/1399 just
to check if running an updated version of black (22.1.0) fixes the CI.


marcharper:
Seems an odd change, but I guess we're letting black make the formatting decisions, eh?

drvinceknight:
> Seems an odd change, but I guess we're letting black make the formatting decisions, eh?

I think so yeah 🤷‍♂️ , an alternative would be to pin `black` to a prior version. I don't care either way :)

marcharper (approved):

Nikoleta-v3 (approved):

PR #1403, opened 2022-07-03
attempting to fix docutils version issue with sphinx

marcharper:
CI is broken because sphinx requires docutils >= 0.14 now but we're pinning == 0.12 in config.yaml (and also docutils >= 0.17 in requirements).

### DIFF
--- a/axelrod/strategies/axelrod_second.py
+++ b/axelrod/strategies/axelrod_second.py
@@ -1661,7 +1661,7 @@ class SecondByBlack(Player):
         recent_history = opponent.history[-5:]
 
         did_d = np.vectorize(lambda action: int(action == D))
-        number_defects = np.sum(did_d(recent_history))
+        number_defects = sum(did_d(recent_history))
 
         return self._random.random_choice(self.prob_coop[number_defects])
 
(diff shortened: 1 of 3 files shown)


marcharper:
As a result of the changes above (version restrictions for docutils removed completely), docutils version 0.18.1 is initially installed and then replaced by 0.17.1 because sphinx_rtd_theme requires < 0.18

axelrod/strategies/axelrod_second.py
```diff
@@ -1661,7 +1661,7 @@ def strategy(self, opponent: Player) -> Action:
         recent_history = opponent.history[-5:]
 
         did_d = np.vectorize(lambda action: int(action == D))
-        number_defects = np.sum(did_d(recent_history))
+        number_defects = sum(did_d(recent_history))
```
drvinceknight, line 1664: Out of curiosity: why this tweak?

drvinceknight:
Thanks for hunting this down @marcharper, looks good. :+1:

axelrod/strategies/axelrod_second.py
marcharper: The type checker was complaining and breaking CI, my guess is that numpy started annotating more things and so it wanted the types to propagate. This was the easiest quick fix.

drvinceknight: Cool 👍

PR #1404, opened 2022-10-27
Bump Python version up to 3.11

marcharper:
It seems that Python 3.11 isn't available in the testing environment yet


drvinceknight:
Nice!

drvinceknight (approved):

PR #1407, opened 2022-12-31
Split requirements.txt to factor out hypothesis and prompt_toolkit

### LINKED ISSUE #1379: Axelrod pip install in Colab fails due to prompt_toolkit
It appears a version mismatch for prompt_toolkit causes problems for pip installing Axelrod in Colab.

`!pip install -q axelrod`

results in

`ERROR: jupyter-console 5.2.0 has requirement prompt-toolkit<2.0.0,>=1.0.0, but you'll have prompt-toolkit 3.0.8 which is incompatible.`
`ERROR: ipython 5.5.0 has requirement prompt-toolkit<2.0.0,>=1.0.4, but you'll have prompt-toolkit 3.0.8 which is incompatible.`

Please see a reproduction of this error in [this notebook](https://colab.research.google.com/drive/1hJa3gStIK9brVW3tACczOvgE0EIe7uJC?usp=sharing)

marcharper:
To address #1379, this PR defines three installation options for requirements:
* base/default -- everything except hypothesis and prompt_toolkit
* human -- includes prompt_toolkit
* development -- includes hypothesis and prompt_toolkit

Once merged and released, the different variants can be installed like so:

```pip install axelrod[Human]```

with the default case not requiring specification (e.g. `pip install axelrod`). I tested the variants and warnings locally in a venv but didn't add explicit tests to CI.

Strategy import is updated to check for the presence of prompt_toolkit and include the Human strategy or not accordingly (and with a warning to the user that it's not available, when applicable).

I wasn't sure whether to include prompt_toolkit by default (the Human strategy is nice for live demos). We could go that route and instead have
* base/default -- everything except hypothesis
* non-human (or minimal?) -- everything except hypothesis and prompt_toolkit
* development -- everything

We could define further variants (e.g. don't import any plotting functions) if desired.

Fixes #1379

### DIFF
--- a/docs/tutorials/new_to_game_theory_and_or_python/installation.rst
+++ b/docs/tutorials/new_to_game_theory_and_or_python/installation.rst
@@ -10,6 +10,9 @@ repository::
 
     $ pip install axelrod
 
+If you want to have access to the manual Human strategy for interactive play, use the following command to also install `prompt_toolkit`::
+
+    $ pip install axelrod[Human]
 
 You can also build it from source if you would like to::
 
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install -r requirements.txt
+        python -m pip install -r requirements/development.txt
     - name: Check that documentation builds
       run: |
         python -m pip install sphinx
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -151,10 +151,10 @@ from .grumpy import Grumpy
 from .handshake import Handshake
 from .hmm import EvolvedHMM5
 from .hmm import EvolvableHMMPlayer, HMMPlayer  # pylint: disable=unused-import
-try: # pragma: no cover
+try:
     import prompt_toolkit
     from .human import Human  # pylint: disable=unused-import
-except ImportError:
+except ImportError: # pragma: no cover
     warnings.warn("Human strategy not available because python package prompt_toolkit is not available.")
 from .hunter import (
     AlternatorHunter,
--- a/axelrod/strategies/_strategies.py
+++ b/axelrod/strategies/_strategies.py
@@ -151,12 +151,15 @@ from .grumpy import Grumpy
 from .handshake import Handshake
 from .hmm import EvolvedHMM5
 from .hmm import EvolvableHMMPlayer, HMMPlayer  # pylint: disable=unused-import
+
 try:
     from .human import Human  # pylint: disable=unused-import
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Review by Nikoleta-v3 — COMMENTED** [review]
Hey 👋🏻 thank you for addressing the issue.

I tried to test the changes locally but when I run `python setup.py install` I get the following error: 

```shell
  File "/Users/glynatsi/src/AllAxelrod/Axelrod/setup.py", line 7, in <module>
    with os.listdir("requirements/") as filenames:
AttributeError: __enter__
```

`os.listdir("requirements/")` returns a list. I can tweak the file and get to it run; but why the codes work for you but not for me?

marcharper:
> os.listdir("requirements/") returns a list. I can tweak the file and get to it run; but why the codes work for you but not for me?

There were a couple of issues actually. I was using pip install instead of setup.py, but it seems like it should not have worked for me locally anyway (given the list vs context manager error you got). I rewrote setup.py to use pathlib and fixed that issue.

Also, there's a difference in the pip command for the package on pypy and what you'll need to use locally. I updated the documentation with the correct command.

I think it's possible to use setup.py directly, with a command like the following, but it didn't work for me locally, and easy_install is deprecated, so I'm sticking with pip. 

```
python setup.py install easy_install "axelrod[human]"
```

Nikoleta-v3:
Perfect 🚀 The `setup.py` file now works. 

Regarding `pip`, since I am using `zsh` I have to run,

 ```
pip install .\[development\]
# or
pip install ".[development]"
```

 instead of `pip install .[development]`. Not sure if we want to mention it in the docs.

File: docs/tutorials/new_to_game_theory_and_or_python/installation.rst [hunk 1]
```diff
@@ -10,6 +10,9 @@ repository::
 
     $ pip install axelrod
 
+If you want to have access to the manual Human strategy for interactive play, use the following command to also install `prompt_toolkit`::
```
drvinceknight, line 13: Love this.

drvinceknight:
The failure of the CI looks like a specific check we had implemented to make sure no strategies were forgotten along the way:

```
Run python -m pip install pylint
[2](https://github.com/Axelrod-Python/Axelrod/actions/runs/3864735683/jobs/6587748984#step:11:2)
  python -m pip install pylint
[3](https://github.com/Axelrod-Python/Axelrod/actions/runs/3864735683/jobs/6587748984#step:11:3)
  python -m pylint --disable=all --enable=unused-import axelrod/strategies/_strategies.py
[4](https://github.com/Axelrod-Python/Axelrod/actions/runs/3864735683/jobs/6587748984#step:11:4)
  shell: /usr/bin/bash -e {0}
```

I'm guessing this is something that can be addressed by including a pragma somewhere.

Happy to take a look if that would be helpful.

Nikoleta-v3 (approved):

Issue #1411, opened 2023-03-14
Implement asymmetric games


alexhroom:
@drvinceknight suggested I open an issue sharing some code for asymmetric games ([see asymmetric_game.py and trade_tournament.py here](https://github.com/alexhroom/gametheory/tree/main/code)) which I have been using to run asymmetric tournaments in Axelrod. In particular, it generalises the Game object to take two matrices, rather than just one.

This could be more fully implemented into Axelrod (and I'm happy to do so after discussing it here) in one of two ways:
1. Create the AsymmetricGame as a separate object, with AsymmetricTournament class etc. Via duck typing it would still be able to be put into Match objects and so on.
2. Implement the AsymmetricGame as the 'base' game class, and then refactor the symmetric game code to inherit from there: e.g. in (truncated) code form:
```
class AsymmetricGame():
    def __init__(self, A, B): # where A and B are the game matrices
    # ... so on...

class Game(AsymmetricGame):
    def __init__(self, A): # where A is the shared game matrix
        super().__init__(A, A)
    # ... the regular Game we all know and love...
```
Similarly, an asymmetric tournament takes two groups (one of row players, one of column players), and then a regular Tournament would become one where the same list of players is passed to both groups. This would build asymmetric Axelrod 'on top' of the existing code with minimal damage to the existing code.

One potential issue I've identified is with the [RPST](https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/game.py#LL42C3-L42C3) method, which doesn't make sense for asymmetric games; if any important code requires RPST this would need some deeper refactoring (to a 'self' and 'opponent' RPST), but otherwise this could just throw an error if used on an asymmetric game.

Finally, I had written a [Paired Moran](https://github.com/alexhroom/gametheory/blob/main/code/paired_moran.py) algorithm which generalises the Moran process to asymmetric games (which takes two populations, and uses fitness functions based on each players' performance against the *other* population), but this would be a separate PR. @marcharper Vince said your thoughts on this would be valuable.

I would angle towards implementation 2 if given full choice as it reduces code complexity and builds asymmetric games directly into Axelrod. However, it would be a deeper refactoring of the library, so I understand if that is undesirable :-)

drvinceknight:
I'd be keen to hear @marcharper's thoughts on this in terms of how this aligns with other theoretic constructs.

[...]

PR #1412, opened 2023-03-16
Pin black to 22.10.0

marcharper:

PR #1413, opened 2023-03-20
Refactored games to generalise & add asymmetric games

alexhroom:
This PR adds asymmetric games and refactors them as the base game class, with the original `Game` sitting on top. This is the simplest implementation that does not break any existing behaviour, and when we are willing to make breaking changes (for 5.0.0) the implementation can be simplified as specified [here](https://gist.github.com/alexhroom/f958e1b614e5adee462e1fe383e66cec) (which I will create as an issue once this is merged)

Solves #1411.

The docs currently do not build, but the unit tests work locally as far as I can tell - @drvinceknight kindly offered to look at the docs issue.

Changes:
- Creates the `AsymmetricGame` class, which takes two payoff matrices `A` and `B` as input. This can be of any size. (see the new documentation, which shows how to create Rock Paper Scissors with the `AsymmetricGame` class; if the simpler implementation is added, this can be done with the regular `Game` class)
- The Game class now creates a matrix from the `r, p, s, t` values and creates an AsymmetricGame with matrices `A` and `A.T`.
- The score method now takes two integers `(r, c)` and returns the value of the matrices at row r, column c - this is an inversion from the initial 'single point of truth' being `self.scores` (so now it is `self.A` and `self.B`, the matrices).
- Changed `Action` from an `Enum` to an `IntEnum` so they can be used as array indices in the `score` method (e.g. for the Prisoner's Dilemma, `A[C][C]` = `3`, equivalent to the score of `(C, C)`)
- Added documentation on creating asymmetric and larger games.
- Removes numpy mocking (as discussed below) so that the docs build with the new Game class.

Follow-up:
- create asymmetric tournaments (which are quite a bit more work to implement)
- create [issue for breaking changes](https://gist.github.com/alexhroom/f958e1b614e5adee462e1fe383e66cec)
- create follow-up for generalising Actions to other games
- create follow-up for categorising strategies by what game they're for


axelrod/game.py
```diff
@@ -16,9 +18,77 @@ class Game(object):
...
+
+        Parameters
+        ----------
+        A: np.array
+            the payoff matrix for player A.
+        B: np.array
+            the payoff matrix for player B.
+        """
+
+        self.A = A
```
marcharper, line 41: Can we add an error / warning if A and B.transpose do not have the same dimensions?

marcharper:
There are some failing tests for me locally. I recommend fixing those first before the docs build issue. The tests for Game directly seem ok, but other tests fail. For example, try:

```
python3 -m unittest axelrod/tests/unit/test_match.py
```

This throws a very similar error as the one for docs building.

```
  File "/home/user/repos/axelrod/Axelrod/axelrod/tests/unit/test_match.py", line 100, in test_non_default_attributes
    @example(turns=5, game=axl.DefaultGame)
  File "/home/user/repos/axelrod/Axelrod/axelrod-venv/lib/python3.7/site-packages/hypothesis/core.py", line 1141, in wrapped_test
    raise the_error_hypothesis_found
  File "/home/user/repos/axelrod/Axelrod/axelrod/tests/property.py", line 382, in games
    game = axl.Game(r=r, s=s, t=t, p=p)
  File "/home/user/repos/axelrod/Axelrod/axelrod/game.py", line 107, in __init__
    super().__init__(A, A.transpose())
  File "/home/user/repos/axelrod/Axelrod/axelrod/game.py", line 38, in __init__
    pair: self.score(pair) for pair in ((C, C), (D, D), (C, D), (D, C))
  File "/home/user/repos/axelrod/Axelrod/axelrod/game.py", line 38, in <dictcomp>
    pair: self.score(pair) for pair in ((C, C), (D, D), (C, D), (D, C))
  File "/home/user/repos/axelrod/Axelrod/axelrod/game.py", line 67, in score
    return (self.A[row][col].item(), self.B[row][col].item())
AttributeError: 'int' object has no attribute 'item'

```

marcharper:
@drvinceknight #1412 also moves github's test ordering to run the unittests before the doc build, so it will give more useful errors in a case like this going forward.

alexhroom:
@marcharper my bad, that test is failing for me now too - not sure how I missed it originally, will fix. I also don't think it's the same error as the one in the docs build - when @drvinceknight and I looked at it we found the docs error didn't occur when the mocked modules were changed to (subscriptable) MagicMock objects, although that fix then caused errors elsewhere. (I think this one is due to how Numpy tracks the datatypes of items in arrays!)

I'll also add some tests for asymmetric games (naturally, the implementation is already tested through the existing tests, but some for asymmetric games specifically wouldn't hurt!)

marcharper:
> @marcharper my bad, that test is failing for me now too - not sure how I missed it originally, will fix.

No worries, there's always some failing tests when you change a low level library detail like this. Locally running `./test` will run the entire test suite and often turns up some subtle bugs.

@drvinceknight Maybe we should split up the tests on github a bit so that the documentation and doc tests also run independently, that way each CI run will give more info if there's an early failing test. We could probably just build the docs on linux as with the format checker.

> I think this one is due to how Numpy tracks the datatypes of items in arrays!)

That was my suspicion as well

> I'll also add some tests for asymmetric games (naturally, the implementation is already tested through the existing tests, but some for asymmetric games specifically wouldn't hurt!)

Agreed, I'd suggest at least one test where `B != A.transpose()` and at least one matrix that isn't square (maybe the same test?). Otherwise this code will be called *many* times in the rest of the test suite so it should be very well covered.

alexhroom:
Removing the change of type to Python types seems to fix matches (and not cause additional issues - i'm not too sure what I originally put it there to fix!). 

However there's now a weird issue with strategy transformers - when creating a Joss-Ann transformer, sometimes the transformed player gives a 0 or 1 to the strategy which causes an error when `proposed_action.flip()` is called. I've not been able to track down where this happens, but I've only seen 0's and 1's (both in the np.int64 datatype) so I'm guessing somewhere the Action values are being cast to np.int64 - I'd also assume this is the issue as it doesn't happen when the Action class is an `Enum` rather than an `IntEnum` (although this creates messier design problems elsewhere).

Currently trying to digest how the Joss-Ann transformer works in order to see where this might be happening!

drvinceknight:
> @drvinceknight Maybe we should split up the tests on github a bit so that the documentation and doc tests also run independently, that way each CI run will give more info if there's an early failing test. We could probably just build the docs on linux as with the format checker.

Yeah that's a good idea. I can pick that up in a separate PR (or if it makes things easier here I can help @alexhroom)

alexhroom:
should all be working now (apart from the docs issue) - essentially some functions were casting Actions to Ints and then failing because they used methods such as .flip(). the fix was simply ensuring they were cast to Action before being flipped.

I also added some unit tests for asymmetric games; these test arbitrary shapes via Hypothesis.

If you'd like to review the code, I think my implementation is now finished unless anything weird turns up!

as for the docs issue - I see two potential solutions;
1. figure out how to run the mock classes for numpy such that it is a much more accurate 'mock' of numpy arrays. i can't seem to find a way of doing this, but there might be a creative solution here.
2. @drvinceknight I think you said that we mock the modules for the docs so that the user does not need the modules installed in order to build the docs? perhaps mocking every module *except* numpy would be a compromise that fixes this pretty quickly

axelrod/history.py
```diff
@@ -33,6 +33,8 @@ def __init__(self, plays=None, coplays=None):
     def append(self, play, coplay):
         """Appends a new (play, coplay) pair an updates metadata for
         number of cooperations and defections, and the state distribution."""
+        # casts plays sent as integers into Action objects
+        play, coplay = Action(play), Action(coplay)
```
marcharper: This feels unnecessary to me, which strategies are not returning proper `Action`s?

axelrod/history.py
```diff
@@ -130,12 +134,13 @@ def flip_plays(self):
...
```
marcharper: I see that there are several castings to `Action`, but this is inefficient if rarely needed, and smells like something upstream is broken. Some of the strategies below have been modified, which seems like the right thing to do, so is this still needed? If it's just the transformer, can we fix it instead?

alexhroom: hmm, you’re right that there may be a better solution - it’s more than just the transformer (there were also issues in history). they’re caused by `Action` objects being automatically cast by some functions due to being the `IntEnum` type. I might just change it back to a regular `Enum` and then cast it to an integer purposefully when getting the score (this would also allow it to work for other `Enum`-style `Action` classes, e.g. if someone wanted to make a Rock Paper Scissors `Action` class with `R`, `P`, `S`)

drvinceknight:
> as for the docs issue - I see two potential solutions;

> figure out how to run the mock classes for numpy such that it is a much more accurate 'mock' of numpy arrays. i can't seem to find a way of doing this, but there might be a creative solution here.
> @drvinceknight I think you said that we mock the modules for the docs so that the user does not need the modules installed in order to build the docs? perhaps mocking every module except numpy would be a compromise that fixes this pretty quickly

The reason for this mocking might be outdated: this was mainly to ensure read the docs didn't need to install them. BUT I think that's no longer necessary. I could take a look at this if you want (I think it would require removing all mocking, and changing a setting on read the docs), otherwise I'm fine with either of your two solutions (whatever makes it work to be honest 😆 ).

alexhroom:
Now passes CI and uses a system for turning `Action`s into integers with far fewer side effects. :-)

drvinceknight:
> Now passes CI and uses a system for turning `Action`s into integers with far fewer side effects. :-)

Nice work! Will review this afternoon.

axelrod/action.py
```diff
@@ -63,9 +63,9 @@ def from_char(cls, character):
         UnknownActionError
             If the input string is not 'C' or 'D'
         """
-        if character == "C":
+        if character == "C" or character == "0":
             return cls.C
-        if character == "D":
+        if character == "D" or character == "1":
             return cls.D
         raise UnknownActionError('Character must be "C" or "D".')
```
drvinceknight:
```suggestion
        raise UnknownActionError('Character must be from {"C", "D", "0" or "1"}.')
```

docs/how-to/use_different_stage_games.rst
```diff
@@ -47,3 +47,28 @@ The default Prisoner's dilemma has different results::
...
     ['Defector', 'Tit For Tat', 'Cooperator']
+
+Asymmetric games can also be implemented via the AsymmetricGame class 
+with two Numpy arrays for payoff matrices::
+
+    >>> import numpy as np
+    >>> A = np.array([[3, 1], [1, 3]])
+    >>> B = np.array([[1, 3], [2, 1]])
+    >>> asymmetric_game = axl.AsymmetricGame(A, B)
+    >>> asymmetric_game  # doctest: +NORMALIZE_WHITESPACE
```
drvinceknight, line 60: I've had a read of https://docs.python.org/3/library/doctest.html but don't fully understand what `NORMALIZE_WHITESPACE` does. For my benefit (not strictly related to reviewing your PR) could you explain what it's needed for here?

docs/how-to/use_different_stage_games.rst
```diff
@@ -47,3 +47,28 @@ The default Prisoner's dilemma has different results::
...
+    Axelrod game with matrices: (array([[3, 1],
+                                        [1, 3]]),
+                                 array([[1, 3],
+                                        [2, 1]]))
+
+Asymmetric games can also be different sizes, such as Rock Paper Scissors::
```
drvinceknight: Can we add a note here along the lines of:

```suggestion
Asymmetric games can also be different sizes, such as Rock Paper Scissors. Note that this game could not currently be used with the IPD strategies::
```

axelrod/action.py
```diff
@@ -63,9 +63,9 @@ def from_char(cls, character):
         UnknownActionError
             If the input string is not 'C' or 'D'
         """
-        if character == "C":
+        if character == "C" or character == "0":
```
drvinceknight: Could you add a test or two to `tests/test_action.py` to isolate this?

axelrod/tests/property.py
```diff
@@ -381,3 +382,16 @@ def games(draw, prisoners_dilemma=True, max_value=100):
 
     game = axl.Game(r=r, s=s, t=t, p=p)
     return game
+
+
+@composite
+def asymmetric_games(draw, valid=True):
```
drvinceknight, line 388: nice

drvinceknight (changes requested):
Everything looks great to me. A couple of minor requests

Do you have capacity to add to the how to an example of running a tournament or match  (and interpreting outcome) with the asymmetric game?

And also, ideally add a unit test for a match, tournament and moran process with the asymmetric game. So we know it all works.

docs/how-to/use_different_stage_games.rst
alexhroom: It means that doctest will test that the strings are the same up to blocks of ` ` or `\n` characters; e.g. with `NORMALISE_WHITESPACE` the strings

```
Game with array: [0, 1]
[1, 2]
```
and
```
Game with array: [0, 1]
                 [1, 2]
```
are equivalent. I added it because doctest was being picky about how much whitespace the `__repr__` output had for the matrices.

docs/how-to/use_different_stage_games.rst
alexhroom: sure, but it'll be a slightly different comment - this game *can* be used with the IPD strategies, although those IPD strategies will only ever choose actions 1 or 2 (and thus will be playing 'Rock-Paper') - so it is possible but not recommended

docs/how-to/use_different_stage_games.rst
drvinceknight: OOO that's lovely. Thanks: great to know.

docs/how-to/use_different_stage_games.rst
drvinceknight: Yup great. :)

marcharper: Perhaps we should hold off on mentioning non 2x2 games, other stuff may break like the mapping of "2" to a non-existent action. Maybe in a follow-up PR a full rock-paper-scissors example could be added, with a few actual strategies (all R, all P, all R, random, etc.).

axelrod/action.py
marcharper: This isn't going to work for games with > 2 possible actions.

axelrod/history.py
```diff
@@ -130,12 +131,12 @@ def flip_plays(self):
     def append(self, play, coplay):
         """Appends a new (play, coplay) pair an updates metadata for
         number of cooperations and defections, and the state distribution."""
+        # casts plays sent as integers into Action objects
```
marcharper: Is this comment correct? I don't see a change that would cause a cast. (same below)

axelrod/action.py
marcharper: Maybe a MockPlayer or transformer that translates C, D into 0, 1 would help sus out any weird edge cases.

marcharper (changes requested):

marcharper:
Almost there! thanks for sticking with it!

docs/how-to/use_different_stage_games.rst
drvinceknight: > Maybe in a follow-up PR a full rock-paper-scissors example could be added, with a few actual strategies (all R, all P, all R, random, etc.).

Yeah, good call.

axelrod/action.py
drvinceknight: Ah. Yeah, good catch. How about: 

```
raise UnknownActionError(f'{character} is not a valid Action')
```

axelrod/action.py
alexhroom: I think this is leftover from before I re-implemented the integers to Actions casting; it may no longer be a relevant edge case! Will remove and see if that fixes things

axelrod/history.py
alexhroom: whoops! forgot to remove this.

docs/how-to/use_different_stage_games.rst
alexhroom: I'd prefer to leave it in with a health warning (that it's experimental and some things break) because as it stands, creating larger games, strategies for those larger games, and matches and tournaments between those strategies also work. What doesn't work (such as transformers) are algorithms and objects particularly created for the study of the Iterated Prisoner's Dilemma, which isn't a fault of either the algorithm, strategy or game, just inappropriate application of said algorithms to those strategies and games.

The wider solution (which I'd recommend doing in a follow-up PR) would be to classify strategies by what game they are created for, and raise a warning/error if an inappropriate algorithm or procedure is being used (for example, attempting to use a transformer that flips strategies on a rock-paper-scissors strategy)

axelrod/action.py
alexhroom: This `Action` class doesn't make *sense* for games with >2 possible actions; it's an `Action` class specifically created for the Prisoner's Dilemma (if there are 3 actions, what does 'flipping' an action even mean?)

A game with >2 possible actions would need its own Action class, for example:
```
class RPSActions(Enum):
    R = 0
    P = 1
    S = 2
```
As said in another comment also, I think the "0" and "1" cases are leftover from my original fix of the integer casting. I don't think this Action class needs to be equipped to deal with arbitrary actions for any game.

docs/how-to/use_different_stage_games.rst
drvinceknight: I'm fine with it if you want to leave it in but yeah as long as there's an explanation.

alexhroom: it was a lot simpler than i thought to put together a little mock rock-paper-scissors with strategies & actions, and this is capable of running a tournament: https://gist.github.com/alexhroom/bea19c0bec62a08838bb6c8c4c949a6a

should I put this somewhere in the docs as a tutorial on adding different games to axelrod?

axelrod/action.py
alexhroom: yep - tests run fine without it. no longer relevant

docs/how-to/use_different_stage_games.rst
marcharper: A tutorial would be great. I just don't want someone to search for a library that can play RSP or NxN games for N > 2 and then be really disappointed to see that you can't really do it with axelrod. You *technically* could but not really since we only have IPD strategies so they'll only use an 2x2 submatrix.

More generally, and going back to the discussion on the issue about the library's abstractions, we probably need to add some way to filter strategies for compatibility with various games.

alexhroom: Added a 'health warning' to the page on using different stage games and a full tutorial on implementing a new game with actions and strategies.

marcharper (approved):

drvinceknight (approved):

drvinceknight:
Merging. Thanks for this @alexhroom, I'll try and put a new release out tomorrow (time permitting).

Issue #1353, opened 2020-07-02
Idea: New way to organize many tests


marcharper:
In the course of finding new seeds for many tests for #1288 it occurs to me that we can probably organize these tests in a more useful way. There are many tests in the library like the following:

```       
 actions = [(C, C), (C, D), (D, C), (D, D), (C, C), (C, D), (C, C), (D, D), (D, C), (C, D)]
        self.versus_test(axelrod.Alternator(), expected_actions=actions, seed=887)
```
where the player is given by `self.Player()` from `TestPlayer`. Sometimes these tests use `MockPlayer` as an opponent rather than an actual opponent. Searching for a new seed involves manually extracting the info (opponent, histories, etc.) and looping over seeds until a new one producing the same behavior is found.

I think a better approach would be to have a large file of expected matches, encoding the range of expected behaviors of every strategy, with rows of the form
```Player1Class, Player2Class, expected_history_1, expected_history_2, seed, other params (like noise), ...```
e.g.
```
Cooperator, Defector, CCCCC, DDDDD, None, ...
```

Essentially it's a dataframe of tests. We could include a description of the test and other metadata. Maybe some other format would be better but hopefully you get the idea.

Such a structure has a few benefits:

1) The file can encode all expected behaviors of a strategy by the histories it should at some point yield in a simple way, rather than being scattered across various tests as is currently, reducing a lot of redundant code (regardless of whether a seed is required)

2) It's easier to find new seeds when we need them.  An auxiliary script can easily scan for new seeds if we change something about how seeding works, or how a strategy works, etc. Right now there's no easy way to extract all the expected tests to systematically find new seeds because the necessary data is hard-coded into functions, and often there is more than one "row" per test function.

3) Similarly, when adding a new strategy, generic search functions can look for an opponent, a seed, etc. that generates a specific sequence of outcomes. I think we're all currently doing these as one-offs, with MockPlayers, etc. 

4) The associated tests will be more single issue now rather than some of the `test_strategy` functions we have that test several different things at once. This will show all the failures rather than failures one at time now for the compound tests as each subtest fails.

5) The collection of expected matches might itself be useful somehow

Similarly, we have a lot of example tournaments and Moran processes with expected outputs that are seed dependent; perhaps they could be encoded in a similar manner. Not every test can be written as so but I would guess that the majority of tests could be done this way. This bumps up against #421, having a way to configure a tournament in a code-free way.

Thoughts?

drvinceknight:
[...]

PR #1417, opened 2023-04-25
Jodoyle29 patch 1

jodoyle29:
Adding a probabilistic hill climbing algorithm.

This strategy is based on the following assumption: if the opponent is going to defect, it is better to defect. If the opponent is going to confess, it is better to confess. Using a simple probabilistic approach, we can predict the opponents next move and chose to confess/defect accordingly.

### DIFF
--- a/axelrod/strategies/probabilistic_hill_climb.py
+++ b/axelrod/strategies/probabilistic_hill_climb.py
@@ -5,16 +5,22 @@ C, D = Action.C, Action.D
 
 class ProbabilisticHillClimb(Player):
 	"""
-	Defects with probability of 50%.
+	Defects with initial probability of 50%.
 	Every time the oppenent deffects, probability becomes '(100 + 1)/100', increasing by 1%.
-	Every time the opponent confesses, probability becomes '(100 - 1)/100', decreasing by 1%.
+	Every time the opponent cooperates, probability becomes '(100 - 1)/100', decreasing by 1%.
 	In case of error (conditions aren't triggered, repeat last move)
 
-	This strategy is based on the following assumption: if the opponent is going to deffect,
-	it is better to deffect. If the opponent is going to confess, it is better to confess. 
-	Using a simple probabilistic approach, we can predict the opponents nexzt move and chose
-	to confess/deffect accordingly.
+	This strategy is based on the following assumption: if the opponent is going to defect,
+	it is better to defect. If the opponent is going to cooperate, it is better to cooperate. 
+	Using a simple probabilistic approach, we can predict the opponents next move and chose
+	to cooperate/defect accordingly.
 	
+    Hill climbing algorithms can be prone to being 'stuck' in local minima. To avoid this we 
+    introduce some randomness, that is, if the probability of defection becomes equal to or 
+    greater than 1 ( >= 100%), the probability of defection will be reset to 50% to avoid this. 
+    For example: Think of this strategy playing against another strategy such as Tit-For-Tat. 
+    If the probability of defection becomes too high, then (without adding radomness) both 
+    ProbabilistciHillClimbing and Tit-For-Tat will infinitely defect, leading to a worse outcome.
 	"""
 	
 	name = "Hill Climb"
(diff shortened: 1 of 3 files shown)


axelrod/strategies/probabilistic_hill_climb.py
```diff
@@ -0,0 +1,55 @@
+from axelrod.action import Action
+from axelrod.player import Player
+
+C, D = Action.C, Action.D
+
+class ProbabilisticHillClimb(Player):
+	"""
+	Defects with probability of 50%.
+	Every time the oppenent deffects, probability becomes '(100 + 1)/100', increasing by 1%.
+	Every time the opponent confesses, probability becomes '(100 - 1)/100', decreasing by 1%.
```
drvinceknight: The `C` in the IPD stands for cooperate (confessing is technically the defection).
```suggestion
	Every time the opponent cooperates, probability becomes '(100 - 1)/100', decreasing by 1%.
```

axelrod/strategies/probabilistic_hill_climb.py
```diff
@@ -0,0 +1,55 @@
+from axelrod.action import Action
+from axelrod.player import Player
+
+C, D = Action.C, Action.D
+
+class ProbabilisticHillClimb(Player):
+	"""
+	Defects with probability of 50%.
```
drvinceknight: This strategy defects with a given probability that changes over time. It's initial probability of defection is 0.5.

axelrod/strategies/probabilistic_hill_climb.py
drvinceknight: (This will need to be modified throughout.)

axelrod/strategies/probabilistic_hill_climb.py
```diff
@@ -0,0 +1,55 @@
...
+		"stochastic": True,
+		"long_run_time": False,
+		"inpects_source": False,
+		"manipulates_source": False,
+		"manipulates_state": False,
+	}
+	
+	def __init__(self) -> None:
+		super().__init__()
+		self.probability = 0.5
```
drvinceknight:
```suggestion
		self.probability_of_defection = 0.5
```

axelrod/strategies/probabilistic_hill_climb.py
```diff
@@ -0,0 +1,55 @@
...
+	
+	def strategy(self, opponent: Player) -> Action:
+		"""Actual strategy definition that determines player's action."""
+		MAX = 100
+		if not len(opponent.history): # if opponent has no previous moves, confess on first move
```
drvinceknight:
```suggestion
		if len(opponent.history) == 0: # if opponent has no previous moves, cooperate on first move
```

axelrod/strategies/probabilistic_hill_climb.py
```diff
@@ -0,0 +1,55 @@
...
+		"manipulates_state": False,
+	}
+	
+	def __init__(self) -> None:
+		super().__init__()
+		self.probability = 0.5
+	
+	def strategy(self, opponent: Player) -> Action:
+		"""Actual strategy definition that determines player's action."""
+		MAX = 100
```
drvinceknight:
```suggestion
```

Based on the doctoring and that this is not a usable parameter I'd suggest just removing it.

If you did want to include it then you'd need to change it to match PEP8.

```suggestion
		denominator = 100
```

axelrod/strategies/probabilistic_hill_climb.py
```diff
@@ -0,0 +1,55 @@
...
+		if not len(opponent.history): # if opponent has no previous moves, confess on first move
+			return C
+		
+		else:
+			if opponent.history[-1] == D:
+				self.probability += 1/MAX
```
drvinceknight:
```suggestion
				self.probability += 1 / MAX
```

axelrod/strategies/probabilistic_hill_climb.py
```diff
@@ -0,0 +1,55 @@
...
+				if(self.probability >= 1):
+					self.probability = 0.5 # to excape local maxima
```
drvinceknight: I'm not sure the comment is precise here. Could you describe the behaviour in the doctoring please.

axelrod/strategies/probabilistic_hill_climb.py
```diff
@@ -0,0 +1,55 @@
...
+				
+			if opponent.history[-1] == C:
+				self.probability -= 1/MAX
+				if(self.probability <= 0):
+					self.probability = 0.1 # avoid crash
+				return self._random.random_choice(self.probability)
+			
+			else:
+				print("There has been an error")
+				return self.history[-1]
```
drvinceknight:
```suggestion
    return self.history[-1]
```

drvinceknight (changes requested):
Interesting strategy.

I've left some documentation suggestions and minor PEP8 requirements.

There are no tests for your strategy which will be needed to merge any PR. There is some information on this here: https://axelrod.readthedocs.io/en/stable/how-to/contributing/strategy/writing_test_for_the_new_strategy.html

If you need any guidance let us know.

jodoyle29:
Thank you for the feedback. I will make these modifications and work on some tests within the next few days.

jodoyle29:
I believe I will need some guidance for writing the tests. How can I test for expected behaviour if the behaviour is probability based (I don't know exactly what my strategy will do at each move, I only know what it will probably do)?

marcharper:
You can fix a random seed (or search for seeds) that trigger the expected behaviors. We strive for 100% coverage of the possible behaviors, if possible.

drvinceknight:
> You can fix a random seed (or search for seeds) that trigger the expected behaviors. We strive for 100% coverage of the possible behaviors, if possible.

You can see an example of this here: https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/tests/strategies/test_memorytwo.py#L99 (there are a number of other stochastic strategies throughout the library).

marcharper:
See also the code snippet in #1353

drvinceknight:
> See also the code snippet in #1353

Gosh I'd forgotten about that! Yes!!!

PR #1418, opened 2023-05-03
Added player assumption classification

### LINKED ISSUE #1414: Game classification
Following up from #1413, to ensure strategies and algorithms are being used with compatible games there should be a method of classifying and organising games.

Some ideas:
- a 'games' directory in the package, which contains a sort of 'mini-package' for a specific game. For example, the import path `axelrod.games.rockpaperscissors` would contain a `Game`, `Action`s and `Strategy` objects for rock-paper-scissors.
- Maybe for convenience, the `axelrod.games` directory could contain a function where a user enters a game name and it returns a dict with the various components of the game, e.g.:
```
>>> from axelrod.games import get_game
>>> get_game("rock_paper_scissors")
{"game": axl.AsymmetricGame([RPS MATRICES HERE]),
 "actions": [R, P, S],
 "strategies": [[RPS STRATEGIES HERE]]
}
```
- Strategies themselves should have classifications for what action set size they're for, and raise an error or warning if used on an inappropriate game size.
- If we're worried about the package getting too big with more games/strategies it may be a good idea to use [namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) so that other games can be installed separately but used from the axelrod namespace - particularly if a few games are added before the release of 5.0.0 this would be a good way to "generalise" the library completely.

alexhroom:
Fixes #1414 by adding a new 'assumptions' feature to strategy classifiers, which check the strategy's assumptions against 'attributes' defined for each game.

Changes:
~~- `Player`s now have an `'actions_size'` classifier, which is an integer, and defaults to 2.~~
~~- The `Player` class now has a method `check_actions_size` which compares their actions size to a given integer. If the game is too small for their actions size, an error is raised (as otherwise it may cause an `IndexError` if it plays an out-of-bounds action); if it is too big, a warning is raised (as it won't cause errors, but may cause unexpected behaviour).~~
~~- The `Match` class now checks these actions sizes against the game matrices when initialising.~~
**Note to reader: This PR originally just implemented game size checking, but as seen in the discussion below this was decided to be insufficient. The PR was then redone to introduce the following 'assumptions' model:**

- When creating a strategy, the classifier can now contain 'assumptions' like game size, `Action` class, things like RPST existing, or even specifically what the game is
- When creating a game, the game now has 'attributes' which identify features that strategies can make assumptions about
- Then when a match is created, the player's assumptions are checked against the game's attributes. If the player makes an assumption that the game doesn't have as an attribute, or if the values mismatch, an error is raised
  - Matches also now have a `strict_player_compatibility` parameter which, if set to False, downgrades these errors to warnings
- This allows each individual strategy to have as tight or loose compatibility with games as it needs, with automatic support for new games whether they're added 'officially' or for personal/individual use. 
- The `MockPlayer` accepts classifiers as an initialisation parameter to aid in the testing of this.

Some examples:
- `axl.Cooperator` would not need to assume anything! It always chooses action 1, which always exists.
- `axl.Adaptive` would assume its action set has size 2 exactly, as it is adapting according to that.
- `axl.Cycler`'s current implementation would need to assume that the actions are {C, D} as it specifically reads a string of C's and D's. (however it doesn't make assumptions on the actual game matrix itself!)
- `axl.FirstByDowning` would need to assume the game has RPST, since it uses that in its implementation.
- `axl.EvolvedANN` would assume *precisely* that the game is Prisoners' Dilemma, as it is trained on that game.

### DIFF
--- a/axelrod/game.py
+++ b/axelrod/game.py
@@ -20,7 +20,7 @@ class AsymmetricGame(object):
     """
 
     # pylint: disable=invalid-name
-    def __init__(self, A: np.array, B: np.array) -> None:
+    def __init__(self, A: np.array, B: np.array, **attributes) -> None:
         """
         Creates an asymmetric game from two matrices.
 
@@ -30,6 +30,9 @@ class AsymmetricGame(object):
             the payoff matrix for player A.
         B: np.array
             the payoff matrix for player B.
+        **attributes
+            optional attributes for the game. Used
+            to ensure strategies used with the game are valid.
         """
 
         if A.shape != B.transpose().shape:
@@ -40,6 +43,8 @@ class AsymmetricGame(object):
 
         self.A = A
         self.B = B
+
+        self.attributes = attributes
 
         self.scores = {
             pair: self.score(pair) for pair in ((C, C), (D, D), (C, D), (D, C))
@@ -75,6 +80,24 @@ class AsymmetricGame(object):
 
         return (self.A[row][col], self.B[row][col])
 
+    @property
+    def attributes(self):
+        return self._attributes
+
+    @attributes.setter
+    def attributes(self, attributes):
+        """
+        Adds or changes game attributes.
+
+        Parameters
+        ----------
+        attributes: dict
+            Attributes to add to the game. If the added
+            attribute already exists, it will overwrite the
+            previous value.
+        """
+        self._attributes = {**self._attributes, **attributes}
+
     def __repr__(self) -> str:
         return "Axelrod game with matrices: {}".format((self.A, self.B))
 
@@ -97,7 +120,7 @@ class Game(AsymmetricGame):
         The numerical score attribute to all combinations of action pairs.
     """
 
-    def __init__(self, r: Score = 3, s: Score = 0, t: Score = 5, p: Score = 1) -> None:
+    def __init__(self, r: Score = 3, s: Score = 0, t: Score = 5, p: Score = 1, **attributes) -> None:
         """Create a new game object.
 
         Parameters
@@ -110,10 +133,19 @@ class Game(AsymmetricGame):
             Score obtained by a player for defecting against a cooperator.
         p: int or float
             Score obtained by both player for mutual defection.
+        **attributes
+            optional attributes for the game. Used
+            to ensure strategies used with the game are valid.
+
         """
         A = np.array([[r, s], [t, p]])
 
-        super().__init__(A, A.transpose())
+        default_attributes = {
+            'has_RPST': True
+        }
+        attributes = {**default_attributes, **attributes}
+
+        super().__init__(A, A.transpose(), **attributes)
 
     def RPST(self) -> Tuple[Score, Score, Score, Score]:
         """Returns game matrix values in Press and Dyson notation."""
@@ -132,4 +164,4 @@ class Game(AsymmetricGame):
         return self.RPST() == other.RPST()
 
 
-DefaultGame = Game()
+DefaultGame = Game(3, 0, 5, 1, game_type='prisoners_dilemma')
--- a/axelrod/game.py
+++ b/axelrod/game.py
@@ -17,6 +17,9 @@ class AsymmetricGame(object):
     ----------
     scores: dict
         The numerical score attribute to all combinations of action pairs.
+    attributes: dict
+        A dictionary of attributes of the game. Used to ensure strategies
+        used with the game are valid.
     """
 
     # pylint: disable=invalid-name
@@ -96,7 +99,10 @@ class AsymmetricGame(object):
             attribute already exists, it will overwrite the
             previous value.
         """
-        self._attributes = {**self._attributes, **attributes}
+        try:
+            self._attributes = {**self._attributes, **attributes}
+        except AttributeError:
+            self._attributes = attributes
 
     def __repr__(self) -> str:
         return "Axelrod game with matrices: {}".format((self.A, self.B))
--- a/axelrod/game.py
+++ b/axelrod/game.py
@@ -23,7 +23,7 @@ class AsymmetricGame(object):
     """
 
     # pylint: disable=invalid-name
-    def __init__(self, A: np.array, B: np.array, **attributes) -> None:
+    def __init__(self, A: np.array, B: np.array, **characteristics) -> None:
         """
         Creates an asymmetric game from two matrices.
 
@@ -33,8 +33,8 @@ class AsymmetricGame(object):
             the payoff matrix for player A.
         B: np.array
             the payoff matrix for player B.
-        **attributes
-            optional attributes for the game. Used
+        **characteristics
+            optional characteristics detailing features of the game. Used
             to ensure strategies used with the game are valid.
         """
 
@@ -47,7 +47,7 @@ class AsymmetricGame(object):
         self.A = A
         self.B = B
 
-        self.attributes = attributes
+        self.characteristics = characteristics
 
         self.scores = {
             pair: self.score(pair) for pair in ((C, C), (D, D), (C, D), (D, C))
@@ -84,25 +84,25 @@ class AsymmetricGame(object):
         return (self.A[row][col], self.B[row][col])
 
     @property
-    def attributes(self):
-        return self._attributes
+    def characteristics(self):
+        return self._characteristics
 
-    @attributes.setter
-    def attributes(self, attributes):
+    @characteristics.setter
+    def characteristics(self, characteristics):
         """
-        Adds or changes game attributes.
+        Adds or changes game characteristics.
 
         Parameters
         ----------
-        attributes: dict
-            Attributes to add to the game. If the added
-            attribute already exists, it will overwrite the
+        characteristics: dict
+            characteristics to add to the game. If the added
+            characteristic already exists, it will overwrite the
             previous value.
         """
         try:
-            self._attributes = {**self._attributes, **attributes}
+            self._characteristics = {**self._characteristics, **characteristics}
         except AttributeError:
-            self._attributes = attributes
+            self._characteristics = characteristics
 
     def __repr__(self) -> str:
         return "Axelrod game with matrices: {}".format((self.A, self.B))
@@ -126,7 +126,7 @@ class Game(AsymmetricGame):
         The numerical score attribute to all combinations of action pairs.
     """
 
-    def __init__(self, r: Score = 3, s: Score = 0, t: Score = 5, p: Score = 1, **attributes) -> None:
+    def __init__(self, r: Score = 3, s: Score = 0, t: Score = 5, p: Score = 1, **characteristics) -> None:
         """Create a new game object.
 
         Parameters
@@ -146,12 +146,12 @@ class Game(AsymmetricGame):
         """
         A = np.array([[r, s], [t, p]])
 
-        default_attributes = {
-            'has_RPST': True
+        default_characteristics = {
+            'has_RPST': hasattr(self, 'RPST')
         }
-        attributes = {**default_attributes, **attributes}
+        characteristics = {**default_characteristics, **characteristics}
 
-        super().__init__(A, A.transpose(), **attributes)
+        super().__init__(A, A.transpose(), **characteristics)
 
     def RPST(self) -> Tuple[Score, Score, Score, Score]:
         """Returns game matrix values in Press and Dyson notation."""
--- a/axelrod/game.py
+++ b/axelrod/game.py
@@ -146,11 +146,6 @@ class Game(AsymmetricGame):
         """
         A = np.array([[r, s], [t, p]])
 
-        default_characteristics = {
-            'has_RPST': hasattr(self, 'RPST')
-        }
-        characteristics = {**default_characteristics, **characteristics}
-
         super().__init__(A, A.transpose(), **characteristics)
 
     def RPST(self) -> Tuple[Score, Score, Score, Score]:
--- a/axelrod/mock_player.py
+++ b/axelrod/mock_player.py
@@ -26,9 +26,12 @@ class MockPlayer(Player):
         super().__init__()
         if not actions:
             actions = []
-        if classifier:
-            self.classifier = classifier
         self.actions = cycle(actions)
+
+        if not classifier:
+            self.classifier = {}
+        else:
+            self.classifier = classifier
 
     def strategy(self, opponent: Player) -> Action:
         # Return the next saved action, if present.
--- a/axelrod/mock_player.py
+++ b/axelrod/mock_player.py
@@ -26,8 +26,7 @@ class MockPlayer(Player):
         super().__init__()
         if not actions:
             actions = []
-        if classifier:
-            self.classifier = classifier
+        self.classifier = classifier
         self.actions = cycle(actions)
 
     def strategy(self, opponent: Player) -> Action:
--- a/axelrod/mock_player.py
+++ b/axelrod/mock_player.py
@@ -21,14 +21,13 @@ class MockPlayer(Player):
     """
 
     name = "Mock Player"
-    attributes = {}
 
-    def __init__(self, actions: List[Action] = None, attributes: dict = None) -> None:
+    def __init__(self, actions: List[Action] = None, classifier: dict = None) -> None:
         super().__init__()
         if not actions:
             actions = []
-        if attributes:
-            self.attributes = attributes
+        if classifier:
+            self.classifier = classifier
         self.actions = cycle(actions)
 
     def strategy(self, opponent: Player) -> Action:
--- a/axelrod/mock_player.py
+++ b/axelrod/mock_player.py
@@ -8,8 +8,8 @@ C, D = Action.C, Action.D
 
 
 class MockPlayer(Player):
-    """Creates a mock player that plays a given sequence of actions. If
-    no actions are given, or it runs out of actions,
+    """Creates a mock player that cycles through a given 
+    sequence of actions. If no actions are given, 
     plays like Cooperator. Used for testing.
 
     Parameters
--- a/axelrod/mock_player.py
+++ b/axelrod/mock_player.py
@@ -9,15 +9,26 @@ C, D = Action.C, Action.D
 
 class MockPlayer(Player):
     """Creates a mock player that plays a given sequence of actions. If
-    no actions are given, plays like Cooperator. Used for testing.
+    no actions are given, or it runs out of actions,
+    plays like Cooperator. Used for testing.
+
+    Parameters
+    ----------
+    actions: List[Action], default []
+        The sequence of actions played by the mock player.
+    attributes: dict, default {}
+        A dictionary of player attributes.
     """
 
     name = "Mock Player"
+    attributes = {}
 
-    def __init__(self, actions: List[Action] = None) -> None:
+    def __init__(self, actions: List[Action] = None, attributes: dict = None) -> None:
         super().__init__()
         if not actions:
             actions = []
+        if attributes:
+            self.attributes = attributes
         self.actions = cycle(actions)
 
     def strategy(self, opponent: Player) -> Action:
--- a/axelrod/mock_player.py
+++ b/axelrod/mock_player.py
@@ -26,7 +26,8 @@ class MockPlayer(Player):
         super().__init__()
         if not actions:
             actions = []
-        self.classifier = classifier
+        if classifier:
+            self.classifier = classifier
         self.actions = cycle(actions)
 
     def strategy(self, opponent: Player) -> Action:
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -258,6 +258,28 @@ class Player(object, metaclass=PostInitCaller):
     def update_history(self, play, coplay):
         self.history.append(play, coplay)
 
+    def check_actions_size(self, game_size: int):
+        """
+        Compares the actions set size of the player to a given game size.
+        If the game is too small, throws an error; if the game is too big,
+        creates a warning that the player may not use certain actions.
+
+        Parameters
+        ----------
+        game_size: int
+            The size of the game to compare player action size with.
+        """
+        
+        actions_size = self.classifier.get('actions_size', 2)
+        if actions_size > game_size:
+            raise IndexError("The action set of player {} is larger "
+                             "than the number of possible actions "
+                             "in the game.".format(self.name))
+        if actions_size < game_size:
+            warnings.warn("The action set of player {} is smaller "
+                          "than the size of the game; they may "
+                          "not play certain actions.")
+
     @property
     def history(self):
         return self._history
--- a/axelrod/player.py
+++ b/axelrod/player.py
@@ -258,14 +258,14 @@ class Player(object, metaclass=PostInitCaller):
     def update_history(self, play, coplay):
         self.history.append(play, coplay)
 
-    def check_assumptions(self, characteristics: dict, raise_error: bool=True):
+    def check_assumptions(self, game_characteristics: dict, raise_error: bool=True):
         """
         Compares the player assumptions to a dictionary of game characteristics.
         Generates a warning or error if an assumption is not fulfilled.
 
         Parameters:
         -----------
-        characteristics: dict
+        game_characteristics: dict
             The dictionary of game characteristics to compare the player's assumptions to.
         raise_error: bool, default True
             If True, raises an error if the assumption is violated. Else,
@@ -274,22 +274,22 @@ class Player(object, metaclass=PostInitCaller):
 
         for key, value in self.classifier.get('assumptions', {}).items():
             msg = None
-            if key not in characteristics.keys():
+            if key not in game_characteristics.keys():
                 msg = ("Player {} assumes that "
                        "the game has the attribute {}, "
                        "but the game does not declare this attribute."
                        "".format(self.name, key))
-            elif value != characteristics[key]:
+            elif value != game_characteristics[key]:
                 msg = ("Player {} assumes that the game attribute "
                        "{} is set to {}, but it is actually set to {}."
-                       "".format(self.name, key, value, characteristics[key]))
+                       "".format(self.name, key, value, game_characteristics[key]))
 
             if msg is not None:
                 if raise_error:
                     raise RuntimeError(msg)
                 warnings.warn(msg + " The strategy may not behave as expected.")
 
-    def assumptions_satisfy(self, characteristics: dict) -> bool:
+    def assumptions_satisfy(self, game_characteristics: dict) -> bool:
         """
         Compares the player assumptions to a dictionary of game characteristics.
         Returns True if the player assumptions are all satisfied by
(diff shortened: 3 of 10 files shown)


marcharper:
Is the size of the action set the right thing to key on? Did you consider having tighter agreement between the player's outputs actions and the expectations of the Game rather than just mathematical compatibility?

I.e. IPD_actions = {C, D} (where C, D come from the IPD Game), then we could programmatically check if the Game's domain matches the Player's range. As other games are added, we could do something similar. For a game like Hawk-Dove, the actions would be {escalate, withdraw}, for rock paper scissors they would be {R, P, S}, and so on.

alexhroom:
@marcharper I did consider this; however my main concern was a use case e.g. if a user wanted to play around with the Hawk-Dove game but still use strategies like `axl.Cooperator` or `axl.TitForTat` - having to make new `axl.hawkdove.Cooperator` or `axl.hawkdove.TitForTat`, both identical save for the actual action objects, would cause massive amounts of code duplication.

For strategies specific to one game, I trust that when the user is importing some `axl.rockpaperscissors.Strategy` that in the very act of typing "rockpaperscissors" they're aware that they're importing a rock-paper-scissors strategy. To me, warning the user because because they used an IPD strategy on a Hawk-Dove matrix would be adding safeguards that frustrate users more often than they're actually helpful - warnings should be saved for times when the user may be surprised by how the code interprets their input. The user is arguably correct to expect that "play action 2 on the Hawk-Dove game" means the same thing as "play action 2 on the Prisoners' Dilemma"; what may be more unexpected is playing `axl.Cycler` on rock-paper-scissors and discovering it just cycles between actions 1 and 2, not all 3 actions, or playing a rock-paper-scissors strategy on the IPD and having their tournament fall over when the strategy tries to play scissors.

Also note that (contrary to what I said in the docs - I made a mistake on the rock-paper-scissors docs and have submitted #1419 to fix it) because of how the ordering of actions is defined (two actions are equal if their values are equal), any set of actions with a total ordering will be interoperable. For example, `axl.hawkdove.Action.Escalate == axl.Action.D` is `True`.

marcharper:
Let's think carefully about this. A strategy could be technically playable with any 2x2 game but designed for / intended for a specific game. For example, many of the IPD strategies assume specific values of RPST. It's not clear to me that game action size is the determining factor, nor should we rely on a total ordering and equate axl.hawkdove.Action.Escalate == axl.Action.D. Different libraries map C and D to 0 and 1 differently, and one can swap rows in a game matrix to effectively change the specific ordering (total orders are not unique for finite indexing sets).

It would be better for strategies to declare the games they are intended for, and perhaps also the games they are technically playable with, and we should warn the user when the player and the game are not aligned. A common rule of software development is that users can't be trusted to know what they are doing! In other words, we need to think carefully about the proper abstractions for the library rather than just push through what's technically feasible.

> having to make new axl.hawkdove.Cooperator or axl.hawkdove.TitForTat, both identical save for the actual action objects, would cause massive amounts of code duplication.

This isn't necessarily true. Hypothetically, with better abstractions, both strategies could derive from a "Play index 1 always of a finite game" strategy, retaining proper type alignment between Player and Game. Also, this is a research library focusing on usability and reproducibility, so minimizing code duplication isn't necessarily a privileged optimization metric.

alexhroom:
_(Sorry! This ended up being a long comment, since I wanted to fully explain my thoughts on the design of this. **TL;DR: Bullet point 4, and the final paragraph**)_

I understand your point, and it may be worth adding a classifier for if a game *really* depends on the specific game being played.  However, I can see a few points where I'm not sure that using this as our main classifier would really increase usability, and it would add complexity to development at the same time:

- firstly just to note: the library *already* depends on the total ordering of actions and *already* bases this ordering on their values. Look at TitForTat; if I've created a different library `mylib` which has `mylib.Action.D=0` and `mylib.Action.C=1`, then TitForTat would *already* evaluate `mylib.Action.C == axl.Action.D` as `True` and defect in that case.
- Strategies which rely on the specific values of RPST get said values from the `Game` object's `RPST` attribute. If the behaviour of these strategies is drastically different because the RPST values are drastically different, this would be an issue that's always existed (the `Game` class has always had the ability to modify RPST, by the looks of git blame), but again that may be the best place to add the safeguards.
- ~~currently, games aren't 'coupled' to actions in any way; `Game`s just treat actions as integers and return scores. Adding this coupling and then attaching all existing strategies to it would *stop* existing strategies from working with other games, and then going back through to generalise them again would be a huge piece of work modifying hundreds of strategies just to get back to where we started~~ _(update: semi-avoided this with the new implementation)._ I do definitely agree that generalising strategies which essentially just run a sequence of actions with no assumptions on the size of the game (e.g. `Random`, `Cycler`) would be a good idea, as they don't currently work with larger games and a user would likely expect that they should.
- if strategies are coded to be tied to specific games, then every time a new game is added it'd be necessary to go through the *entire* strategy index to update what games they're "technically playable with", adding a huge piece of tedious busywork to any contributor who wants to add a new game; likewise someone adding a new strategy would have to figure out which games it does and doesn't work with.
  - Note here also that if a user creates a custom game that they don't want to add to the library itself (e.g. for a research project) then they'd either have to create every strategy they want to use for their project from scratch, or put up with endless warnings; they *can't* go through modifying library source code to flag up that the strategies are compatible with their game.
- For usability, I agree that maybe code duplication in implementation isn't as important but *strategy* duplication absolutely is. If I pick up Axelrod for the first time and then have to figure out for my code whether "play action 1 unless you last played action 2, then play action 2" should use `axl.TitForTat()` or `axl.hawkdove.TitForTat()` or `axl.GeneralisedTitForTat(actions={myaction.A, myaction.B})` etc. etc. would make my head spin. Fundamentally, the abstraction for a `Player` is "a mapping from play history onto the game's actions {1, 2, ..., n}" and the simplest path for the user is if strategies are designed that way.

I just think that the incidence of "someone accidentally uses a strategy which doesn't work on their custom game" is lower than that of "someone wants to use an existing strategy on a different game of the same size", and gearing our classification system towards the first one at the expense of the latter just feels like it'll create more ignorable warnings than actually informative ones. In my experience with developing research software, when a less code-savvy user gets a warning then they usually treat it like the check engine light on a car! With this in mind I think a warning that's not relevant more often than it's relevant is problematic.

As for development, my main rub is related to point 4. Enumerating exactly "this strategy works with game X, Y, Z and doesn't work with any other game" and then having to update it every time a game is added is almost certainly not the best abstraction (one could argue it isn't even an abstraction!) Perhaps for strategies relying on specific assumptions of a game, a better solution would be to add an "attributes" dict to the `Game`, and then an "assumptions" attribute to the strategy classifier. If the `Game` attributes disagree with the `Player` assumptions (e.g. the `Player` assumes something that isn't in the `Game` attributes, or the `Game` attribute value is not equal to the equivalent `Player` assumption attribute), then a warning is raised. This allows us to account for strategies which *really* rely on the game being a certain way, while leaving the rest of the library completely plug-and-play, and hugely reducing the work required to add a new game to the library; a new game could add its own assumptions and have that as a single point of truth for what strategies do and don't work with it, and vice versa with new strategies and valid games.

alexhroom:
Reimplemented as a more general 'assumptions' model:
- When creating a strategy, the classifier can now contain 'assumptions' like game size, `Action` class, things like RPST existing, or even specifically what the game is
- When creating a game, the game now has 'attributes' which identify features that strategies can make assumptions about
- Then when a match is created, the player's assumptions are checked against the game's attributes. If the player makes an assumption that the game doesn't have as an attribute, or if the values mismatch, an error is raised
  - Matches also now have a `strict_player_compatibility` parameter which, if set to False, downgrades these errors to warnings
- This allows each individual strategy to have as tight or loose compatibility with games as it needs, with automatic support for new games whether they're added 'officially' or for personal/individual use. 

Some examples:
- `axl.Cooperator` would not need to assume anything! It always chooses action 1, which always exists.
- `axl.Adaptive` would assume its action set has size 2 exactly, as it is adapting according to that.
- `axl.Cycler`'s current implementation would need to assume that the actions are {C, D} as it specifically reads a string of C's and D's. (however it doesn't make assumptions on the actual game matrix itself!)
- `axl.FirstByDowning` would need to assume the game has RPST, since it uses that in its implementation.
- `axl.EvolvedANN` would assume *precisely* that the game is Prisoners' Dilemma, as it is trained on that game.

does this address your concerns @marcharper? I think it's a much more flexible model for maintaining alignment between `Player` and `Game` without having to be fully locked-down (and without the greater quantity of busywork for future contributors described in the previous comment). If we wanted to then avoid using mismatched action sets, strategies could be generalised with an `action_set` parameter to their `__init__` method to take an action set - this generalisation could be done alongside actually implementing the assumptions to all of the strategies (in a separate PR to this!)

drvinceknight:
I think I like the look of this @alexhroom, using the basic definition of a strategy as a way of mapping "information" to actions.

I understand your point @marcharper though and really trust your instincts here.

Would creating a specific kind of `AxelrodError Invalid Game` be a good thing to implement? So if a strategy requires something that's not in the game it raises a verbose error?

alexhroom:
@drvinceknight one can see here https://github.com/Axelrod-Python/Axelrod/pull/1418/files#diff-0e43e29ffb206bbff9d1780ef3d83a2407c3ec58e893fd93c06bbcdd478a1278R275-R290 that currently a RuntimeError with a verbose message is raised, stating what assumption was violated. as far as I can see the only real benefit to using a unique exception type here is to be able to suppress it, but the `raise_error=False` option allows code to bypass it anyway

axelrod/game.py
```diff
@@ -110,10 +139,19 @@ def __init__(self, r: Score = 3, s: Score = 0, t: Score = 5, p: Score = 1) -> No
...
+        **attributes
+            optional attributes for the game. Used
+            to ensure strategies used with the game are valid.
+
         """
         A = np.array([[r, s], [t, p]])
 
-        super().__init__(A, A.transpose())
+        default_attributes = {
+            'has_RPST': True
```
marcharper: This could be explicitly tested for, either with a try-catch blog or by inspecting the class for the presence of an RPST() function.

That said, is RPST synonymous with the game being IPD?

marcharper:
This is better, thanks. Some notes:

* It's true that it's always been possible to make strategies behave oddly with unusual RPST. Similarly, some of the older strategies hard code values that may be based on RPST implicitly. These will probably behave poorly on other games. It's always been possible to change the game matrix, but it's not been a prominent feature until recently.
* It's not true that the library depended on the order of {C, D} -- we've always used either the characters 'C' and 'D' or the enum (matching explicitly on action.C or action.D) once Python supported enums. (The enum values 0 and 1 are only there because the Python enum syntax requires something.) I'm not necessarily opposed to making such a change, but currently Cooperator isn't synonymous with "play first action" and Defector could be any of "play the opposite of C" or "play index 2 action" or "play last action"... We can change this but you can see how it doesn't uniquely generalize to nxn games canonically.
* I certainly agree that we don't want to make contributing a new strategy more tedious or necessitating many changes. As we start supporting a larger variety of games, there's more chance for confusion, as existing parts of the library like `axelrod.strategies` change semantics from "strategies usable with IPD" to a mix of strategies intended for different games that now need to be filtered somehow for use with any specific game without producing errors or non-intuitive results. (Imagine a RPS strategy all-Scissors strategy -- it won't work with IPD.) The classifier approach seems like a good way to handle this, we already have some filters for "obeys [the] axelrod [tournament rules]". However we might want to take a different approach, making axelrod.strategies a more complex datastructure, or splitting it per game, or something else.

axelrod/game.py
alexhroom: i might play around with this kind of thing - what would be nice is if the assumptions could take a predicate function and 'query' the game with it. 

Currently any game created through the Game class has an RPST method, but when the `AsymmetricGame` and `Game` classes are just collapsed into one class for 5.0.0 this will have to be changed (either being removed entirely (and strategies using RPST just getting this from the game array data instead) or throwing a NotImplementedError if the game is a different size).

I guess on an 'academic' level for the study of game theory, RPST is synonymous with the game being IPD, but on a codebase level, a strategy needing to access RPST just means "this strategy does some math to the game matrix values". look at `axl.FirstByDowning` for a good example:
```
On the second page of the appendix, figure 4 (page 390)
    identifies an expression for EV_TOT.
    A specific term is made to disappear in
    the case of T - R = P - S (which is not the case for the standard
    (R, P, S, T) = (3, 1, 0, 5)):
```
that is, it uses RPST *without* the assumption that the game is exactly our implementation of the IPD, i.e. without assuming the payoffs are (3, 1, 0, 5). This could be refactored for the strategy to just take the values from the game matrix and then this strategy would only rely on the game being 2x2 instead.

alexhroom:
> We can change this but you can see how it doesn't uniquely generalize to nxn games canonically.

I understand - maybe better to keep those strategies as 2x2 for now. Like I said, this would be done in a different PR (i'd prefer this PR to just focus on implementing the framework) so the details of that can be hashed out then.

> However we might want to take a different approach, making axelrod.strategies a more complex datastructure, or splitting it per game, or something else.

See the discussion in #1414. Maybe it'd be useful here to implement a helper function which takes a `Game` and a strategy or list of strategies and returns a bool or some other information on whether or not the strategies work with the game? this could then be used with a list comprehension in the same way as we can already classify strategies; then akin to `axl.demo_strategies` or `axl.basic_strategies` it's not too difficult to take `axl.ipd_strategies` etc.

drvinceknight:
> * However we might want to take a different approach, making axelrod.strategies a more complex datastructure, or splitting it per game, or something else.

Not against a classifier approach but I do like the sound of `axelrod.strategies.rps`, `axelrod.strategies.ipd`, `axelrod.strategies.ultimatum_game` etc... I feel it would allow for a more "automatic"/easier classification.

alexhroom:
I guess my issue with `axelrod.strategies.rps`, `axelrod.strategies.ipd` etc. would be where we put generic strategies; e.g. if we created some generic strategy `axl.Static` which just takes an action and plays it every turn (this is a generalisation of Cooperator, Defector, etc) or even specifically IPD strategies like `axl.TitForTat` which would be a valid strategy on, e.g. the Hawk-Dove game (where it withdraws on turn 1, then copies the opponent's previous move in future). maybe there's something cool with namespaces that would get the best of both worlds - i.e. we take a classifier approach but use some Python import wizardry to make it accessible both from `axl.TitForTat` or `axl.strategies.ipd.TitForTat`, where axl.strategies.ipd provides a namespace for both strategies *specific* to the IPD (which are in a specific strategies.ipd folder) and strategies that *work* with the IPD. i think i've seen similar before in another package (but I'm not sure where). that way users can get strategies from `axl.strategies.ipd` for a safety guarantee that their strategies will work with the game, or just from `axl.` if they want to mix and match some generic strategies, or borrow the strategy to add it to their own new game without having to remember what game it's "originally for".

drvinceknight:
> I guess my issue with `axelrod.strategies.rps`, `axelrod.strategies.ipd` etc. would be where we put generic strategies; e.g. if we created some generic strategy `axl.Static` which just takes an action and plays it every turn (this is a generalisation of Cooperator, Defector, etc) or even specifically IPD strategies like `axl.TitForTat` which would be a valid strategy on, e.g. the Hawk-Dove game (where it withdraws on turn 1, then copies the opponent's previous move in future). maybe there's something cool with namespaces that would get the best of both worlds - i.e. we take a classifier approach but use some Python import wizardry to make it accessible both from `axl.TitForTat` or `axl.strategies.ipd.TitForTat`, where axl.strategies.ipd provides a namespace for both strategies _specific_ to the IPD (which are in a specific strategies.ipd folder) and strategies that _work_ with the IPD. i think i've seen similar before in another package (but I'm not sure where). that way users can get strategies from `axl.strategies.ipd` for a safety guarantee that their strategies will work with the game, or just from `axl.` if they want to mix and match some generic strategies, or borrow the strategy to add it to their own new game without having to remember what game it's "originally for".

Like the idea of something smart with namespaces (with the caveat for ease of maintenance...) -- I think the flat `axelrod.` namespace was something we would/should have done differently so would be in favour of taking advantage of a `5.0.0` to do that.

A thought: would a `axl.strategies.generic` namespace be helpful? For things like `axl.strategies.generic.Static`?

alexhroom:
> A thought: would a axl.strategies.generic namespace be helpful? For things like axl.strategies.generic.Static?

maybe - if we're trashing the flat `axl.` namespace (good idea) then that'd be useful. what I had in mind for the specific game strategies is something like the following:
- generic strategies by default just take integers, but they can be given an action set. This can be done programmatically on the user end (e.g. axl.strategies.generic.Static(actions={C, D})
- we add some function `equip_with_actions()` which returns a class but equipped with the action set. e.g. 
  `equip_with_actions(axl.strategies.generic.Static, {C, D})` returns an `axl.strategies.generic.Static` *class* object (uninstantiated) but where the action set for the class has been set to {C, D}. This should be easy enough with class methods.
- Then, for example, the `strategies/ipd/__init__.py` file contains something like the following (**note this isn't working python code, i don't think list comprehensions work on modules like this!**):
```
from . import *
from ..generic import [equip_with_actions(strategy, {C, D}) for strategy in generic if strategy.assumptions_satisfy(ipd_game_attributes)]
```
i.e. the generic strategy `axl.strategies.generic.Static` is a static strategy that plays a given integer action, whereas `axl.strategies.ipd.Static` is a static strategy that plays *specifically* `C` or `D`, i.e. it contains all generic strategies that work with the IPD and equips them with the correct action set too.

However this is all a different PR to this one. I think this discussion should be moved to a relevant issue instead! (#1420)

drvinceknight:
> However this is all a different PR to this one. I think this discussion should be moved to a relevant issue instead! 

:+1:

Minor point: Did you want to grab the commit with the change to fix the readthedocs build?

alexhroom:
> Minor point: Did you want to grab the commit with the change to fix the readthedocs build?

I'll just merge my fork's dev into this branch when #1419 is merged - trying to cherry-pick the commit without messing anything up is a nightmare in my experience (and suspiciously the read the docs build seems to now be succeeding without it)

axelrod/player.py
```diff
@@ -258,6 +258,37 @@ def reset(self):
...
         self.history.append(play, coplay)
 
+    def check_assumptions(self, attributes, raise_error=True):
+        """
+        Compares the player assumptions to a dictionary of attributes.
+        Generates a warning or error if an assumption is not fulfilled.
+
+        Parameters:
+        -----------
+        attributes: dict
```
marcharper: Can we call this something less generic than `attributes`? Maybe `game_assumptions` or maybe just `assumptions` (similarly elsewhere)

axelrod/mock_player.py
```diff
@@ -8,16 +8,26 @@
...
     """
 
     name = "Mock Player"
 
-    def __init__(self, actions: List[Action] = None) -> None:
+    def __init__(self, actions: List[Action] = None, classifier: dict = None) -> None:
         super().__init__()
         if not actions:
             actions = []
+        if classifier:
```
marcharper: Should this be attributes?

axelrod/player.py
marcharper: It's just that attributes is a common python term (there's a library called `attrs`).

axelrod/mock_player.py
marcharper: Also, not sure you need the if here?

axelrod/player.py
alexhroom: yeah i did think maybe to change it - 'assumptions' doesn't feel right because the Game doesn't assume anything, it either does or doesn't have something. also i wouldn't want the player's `assumptions` getting mixed up with the game's `attributes`/features. maybe `characteristics` or `game_characteristics`?

axelrod/mock_player.py
alexhroom: nope! it's the player classifier. you could add other classifier items like `'memory_depth'` etc to the mock player through this interface too. will remove the if, good point

alexhroom: turns out the `if` is load bearing - the `super().__init__()` sets classifier to `{}` so without the if statement it already takes a default there (as opposed to a usual construct like defaulting it to None and saying `if classifier is None; classifier = {}`). reverted

drvinceknight:
Failing on coverage:

```
TOTAL                                                            17952      5    99%
Coverage failure: total of 99 is less than fail-under=100
Error: Process completed with exit code 2.
```

axelrod/player.py
marcharper: Yeah maybe `game_characteristics` or `game_requirements` ?

alexhroom: done

axelrod/game.py
```diff
@@ -110,10 +139,19 @@ def __init__(self, r: Score = 3, s: Score = 0, t: Score = 5, p: Score = 1) -> No
...
+        **attributes
+            optional attributes for the game. Used
+            to ensure strategies used with the game are valid.
+
         """
         A = np.array([[r, s], [t, p]])
 
-        super().__init__(A, A.transpose())
+        default_characteristics = {
+            'has_RPST': hasattr(self, 'RPST')
```
marcharper: We don't really need a distinguished `has_RPST` characteristic since R, P, S, and T are just the values of the game matrix and so any 2x2 matrix game provides them. For clarity in the calling strategies I think we should still have an RPST function though. The difference is that the strategies need the values to be well-defined but the game itself always has them, so the requirement is really "game can provide RPST values", which wouldn't be possible for say non-linear payoffs.

axelrod/mock_player.py
marcharper: I still don't quite understand this one. `classifier` is local so how can `Player.__init__` set it? I could see `self.classifier` being set by `Player.__init__`, and if that's the case, maybe we should be passing `classifier` in as a kwarg to `Player`?

axelrod/player.py
```diff
@@ -258,6 +258,64 @@ def reset(self):
     def update_history(self, play, coplay):
         self.history.append(play, coplay)
 
+    def check_assumptions(self, game_characteristics: dict, raise_error: bool=True):
```
marcharper, line 261: Should the assumptions check be here or in the `Match` class? It's in a `Match` that `Player`s and `Game`s are combined (and also in the Moran process class, which uses the `Match` class). The `Player` class just needs to be initialized properly and be able to return values when `.strategy()` is called, it's not involved in the scoring process and so it doesn't need to care about the `Game` usually (unless it requires info about the game). Note that [it's in](https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/match.py#L115) a `Match` that `Player` instances are currently given info about the game or match.

marcharper (changes requested):

axelrod/player.py
alexhroom: checking assumptions has use cases outside of matches - if it's in the Player class we can use it for filtering, e.g. if we wanted all strategies from a list `strategies` that work on a game with assumptions `my_assumptions`, we could do

`[strategy() for strategy in strategies if strategy().assumptions_satisfy(my_assumptions)]`.

Note that in general, the methods just take a dict, not a `Game` object specifically; the `Game` itself is still completely hidden from the `Player` class. [In the Match class, the `Player`s and `Game` are combined in the way that you want, without either of them having their whole class exposed to the other.](https://github.com/Axelrod-Python/Axelrod/pull/1418/files#diff-78d3d04c3d62ef794070aad3a16585106965b56363df56323b5854ff1f2dfa64R119-R126)

> it doesn't need to care about the Game usually (unless it requires info about the game).

"how should strategies handle requiring info about the game" is really the issue that we want to solve in this PR! :)

axelrod/mock_player.py
alexhroom: I checked because I wasn't sure either, but it turns out the local variable `classifier` in the body of the class and the attribute `self.classifier` in the ~~instance~~ _(edit: [class](https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables))_ attributes are *the same thing*! I was surprised to see it works like this. Here's an example:

```python
>>> class ExampleClass:
... 
...     attribute1 = "Hello, World!"
...
>>> example = ExampleClass()
>>> print(example.attribute1)
"Hello, World!"
```
The main difference, of course, is that local variables can be set externally through instantiation parameters in `__init__` but can only be set in the body of the class when the class source code is written. I'd argue this is a good thing (I don't think classifiers should be/need to be mutable by the user). As far as I can tell, all `instance.attribute` does is get the local variable `attribute` from the scope of `instance` [(see the Python docs on classes!)](https://docs.python.org/3/tutorial/classes.html#python-scopes-and-namespaces)

axelrod/game.py
alexhroom: fine by me - in that case I'd suggest refactoring RPST in a different PR, and I'll remove the default characteristic for now.

axelrod/mock_player.py
alexhroom: actually, looking further into these docs, [the attribute in the body is shared by all instances](https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables); so turns out that not only is the if statement necessary it's actually insufficient (e.g. we need the `if Classifier is None` etc. etc.)

alexhroom:
as far as I can tell the failure in the CI isn't caused by anything i've done - it seems to always almost pass except randomly fail sometimes (looking at both the test logs and running it locally a few times), unrelated to this PR

drvinceknight:
> as far as I can tell the failure in the CI isn't caused by anything i've done - it seems to always almost pass except randomly fail sometimes (looking at both the test logs and running it locally a few times), unrelated to this PR

This happens sometimes with hypothesis. Have kicked the CI, hopefully that sorts it but if not will look in to it.

axelrod/game.py
```diff
@@ -132,4 +165,4 @@ def __eq__(self, other):
         return self.RPST() == other.RPST()
 
 
-DefaultGame = Game()
+DefaultGame = Game(3, 0, 5, 1, game_type='prisoners_dilemma')
```
marcharper, line 168: Is it wise to use a string to indicate the type rather than the typing system, or inheritance from an abstract base game?

marcharper:
Do I understand correctly: with this PR every existing strategy assumes that the action set size is 2 and there are no other specified assumptions?

alexhroom:
@marcharper indeed - my aim for this PR is just to add the actual framework for strategy classification (which is not a breaking change so could be merged to dev) and then a follow-up issue would be to actually go through all the strategies and add assumptions, probably alongside a bigger refactoring of the strategies folder to account for non-ipd games - this would be a breaking API change (so would be part of 5.0.0).

axelrod/game.py
alexhroom: how do you suggest we use the typing system? i'd be up for that once we have a better idea of how to handle different games (at which point prisoners' dilemma ceases to be the `DefaultGame`) but currently it's just a Game class with certain attributes.

marcharper: Well, one approach would be to have a hierarchy of games with different Types, e.g. `IPD(Game)` and then one could use `isinstance(IPD, game)` to determine capabailities and compatibilities. We can discuss in future issues/PRs.

marcharper:
Should this go into the 5.0.0 branch now then, since we anticipate breaking behavior soon?

alexhroom:
@marcharper sure! changed base branch (note that means the most recent commit to dev has been added to this pr!)

axelrod/game.py
alexhroom: okay - will leave for now

marcharper (approved):

drvinceknight:
Thanks for all the work on this @alexhroom

PR #1419, opened 2023-05-06
fixed the rock-paper-scissors docs example

alexhroom:
Fixes a mistake in the doc page added by #1413 - it turns out that strategies *can* recognise actions from different games corresponding to the same row/column, so long as both of the `Action` classes have a total ordering. That is:

Using the example code from before this PR:
`axl.rockpaperscissors.RPSAction.P == axl.Action.D` -> `False`
After this PR:
`axl.rockpaperscissors.RPSAction.P == axl.Action.D` -> `True`

This is good news for the interoperability of different games! :)

### DIFF
--- a/docs/tutorials/implement_new_games/index.rst
+++ b/docs/tutorials/implement_new_games/index.rst
@@ -29,14 +29,9 @@ corresponding to rows 1 and 2 respectively). If we tried to use it on rock-paper
 interpret the game in the following way:
 
 1. On the first turn, choose rock (option 1, cooperate)
-2. If the opponent's last move is the Python object
-   :code:`axl.Action.D` (which it may never be unless the opponent also thinks it's playing IPD!), 
-   then choose paper (option 2, defection)
+2. If the opponent's last move is action 2, then choose paper (action 2, defect)
 
-and so as we see, :code:`Tit-For-Tat` would simply play Rock every turn, unless it
-were playing against another Prisoners' Dilemma strategy (then it
-plays rock unless the opponent last played paper, in which case it plays paper). In
-particular, it would never play scissors - it does not know that Scissors is something
+In particular, it would never play scissors - it does not know that Scissors is something
 it can even do. This is not a bug, or an issue with the strategy itself; 
 simply that :code:`Tit-For-Tat` *thinks* it is playing the Iterated Prisoners' Dilemma
 and its :code:`Action` set, regardless of what the actual game is.
@@ -53,11 +48,16 @@ has :code:`flip`, which flips a :code:`C` to a :code:`D` and vice versa!)
 A simple rock-paper-scissors action class would look like so::
 
     >>> from enum import Enum
+    >>> from functools import total_ordering
+    >>> @total_ordering
     >>> class RPSAction(Enum):
     ...     """Actions for Rock-Paper-Scissors."""
     ...     R = 0  # rock
     ...     P = 1  # paper
     ...     S = 2  # scissors
+    ...     
+    ...     def __lt__(self, other):
+    ...         return self.value < other.value
     ...     
     ...     def __repr__(self):
     ...         return self.name
@@ -77,6 +77,11 @@ A simple rock-paper-scissors action class would look like so::
     ...         }
     ...         
     ...         return rotations[self]
+
+Note that the :code:`total_ordering` decorator and the :code:`__lt__` method are required for
+Python to equate different types of Action sets based on what action they correspond
+to; without it, strategies on two separate games of the same size wouldn't be
+interoperable!
 
 We can then implement some strategies. Below we have the implementation of an
 Axelrod strategy into Python. These follow the same format;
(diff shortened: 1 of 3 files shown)


docs/tutorials/implement_new_games/index.rst
```diff
@@ -53,12 +48,17 @@ has :code:`flip`, which flips a :code:`C` to a :code:`D` and vice versa!)
...
-    >>> class RPSAction(Enum):
+    >>> from functools import total_ordering
+    >>> @total_ordering
+    ... class RPSAction(Enum):
     ...     """Actions for Rock-Paper-Scissors."""
     ...     R = 0  # rock
     ...     P = 1  # paper
     ...     S = 2  # scissors
     ...     
+    ...     def __lt__(self, other):
```
marcharper, line 59: Why do we need this? Are we ever relying on R < P < S ? Also it's technically a circular ordering, since S < R.

alexhroom: I mentioned a more full reason here: https://github.com/Axelrod-Python/Axelrod/pull/1418#issuecomment-1537077559 and also in the paragraph below this code block in the docs page itself.

but outside of interoperability between games (which we may or may not want) it's there for best practice (and to avoid user confusion if someone creates a game with a strategy containing a line like `if action == P`), and to keep consistency with the library itself [it has been implemented in exactly the same way as the IPD `Action` class implements ordering of actions](https://github.com/Axelrod-Python/Axelrod/blob/784cdd361bc40119b828e97801d0bb58db3a5669/axelrod/action.py#L21-L33). I can change it to just an `__eq__` method if you prefer and remove the `@total_ordering`.

drvinceknight:
Not actually sure what's causing the sphinx failure on read the docs. Here's the stack trace:

```
raceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/axelrod/envs/1419/lib/python3.7/site-packages/sphinx/registry.py", line 472, in load_extension
    mod = __import__(extname, None, None, ['setup'])
  File "/home/docs/checkouts/readthedocs.org/user_builds/axelrod/envs/1419/lib/python3.7/site-packages/sphinx/builders/linkcheck.py", line 19, in <module>
    from requests.exceptions import HTTPError
  File "/home/docs/checkouts/readthedocs.org/user_builds/axelrod/envs/1419/lib/python3.7/site-packages/requests/__init__.py", line 43, in <module>
    import urllib3
  File "/home/docs/checkouts/readthedocs.org/user_builds/axelrod/envs/1419/lib/python3.7/site-packages/urllib3/__init__.py", line 39, in <module>
    "urllib3 v2.0 only supports OpenSSL 1.1.1+, currently "
ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168

During handling of the above exception, another exception occurred:

  File "/home/docs/checkouts/readthedocs.org/user_builds/axelrod/envs/1419/lib/python3.7/site-packages/sphinx/cmd/build.py", line 303, in build_main
    args.tags, args.verbosity, args.jobs, args.keep_going)
  File "/home/docs/checkouts/readthedocs.org/user_builds/axelrod/envs/1419/lib/python3.7/site-packages/sphinx/application.py", line 224, in __init__
    self.setup_extension(extension)
  File "/home/docs/checkouts/readthedocs.org/user_builds/axelrod/envs/1419/lib/python3.7/site-packages/sphinx/application.py", line 449, in setup_extension
    self.registry.load_extension(self, extname)
  File "/home/docs/checkouts/readthedocs.org/user_builds/axelrod/envs/1419/lib/python3.7/site-packages/sphinx/registry.py", line 475, in load_extension
    raise ExtensionError(__('Could not import extension %s') % extname, err)
sphinx.errors.ExtensionError: Could not import extension sphinx.builders.linkcheck (exception: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168)

Extension error:
Could not import extension sphinx.builders.linkcheck (exception: urllib3 v2.0 only supports OpenSSL 1.
```

I believe this is fixed here: https://github.com/readthedocs/readthedocs.org/issues/10290. Will open a PR (hopefully) a fix to your branch @alexhroom.

drvinceknight:
I've opened https://github.com/alexhroom/Axelrod/pull/1

alexhroom:
@drvinceknight looks like it's working - i also had to add numpy to the doc requirements.txt since it isn't being mocked.

marcharper:
> axl.rockpaperscissors.RPSAction.P == axl.Action.D -> True

I'm still not convinced we want this to be true -- it feels like a type mismatch. axl.action.D could reasonably equal RPSAction.S as well (being the last play rather than the second play). We're forcing a specific permutation of the actions for not a lot of benefit IMO.

Regardless of the mapping chosen, a 2x2 player like TFT may be able to technically play a game of higher dimension, but only in a degenerate way (restricted to a subset of actions). I don't see that this is really useful, other than perhaps refactoring some of the (degenerate) pure strategies like Cooperator and Defector to be more generic.

alexhroom:
@marcharper it matters a lot less now we've got a better idea on how we're going to redesign/refactor strategies - i'll leave this for now and create a separate PR just with the RTD fix.

PR #1428, opened 2023-10-17
Bump Supported Python versions up to 3.12

marcharper:
Python 3.12 doesn't seem to be available yet in CI, will re-run tests later.

Update: some packages like distutils are no longer available in Python 3.12 so I had a make a few small changes. In one case we had a check for matplotlib versions below 2, but we're already requiring 3+, so there's no change in functionality.

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -14,7 +14,7 @@ jobs:
       max-parallel: 4
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.10", "3.11", "3.12"]
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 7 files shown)


Nikoleta-v3 (approved):

drvinceknight:
Looks like this isn't passing because of some downstream libraries that aren't on 3.12 yet? I tried restarting the tests just now and it didn't make much difference.

drvinceknight:
@marcharper I've opened #1435 with a PR to your branch that I think cleans up the doctests.

marcharper:
`setuptools` is deprecated and removed in Python 3.12, so we have to either use a third party version or change how we do packaging :/

This link is helpful: https://gregoryszorc.com/blog/2023/10/30/my-user-experience-porting-off-setup.py/

IIUC we need to migrate from setup.py to pyproject.toml. But there are some choices that need to be made (e.g. build tool, setuptools is still an option).

Some options here: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

marcharper:
I pulled out the commits after the first into #1436 so we can merge those. I'll rebase the first commit onto a python-3.12 branch and we can figure out the packaging issues there.

marcharper:
Closing, let's discuss at #1437

PR #1433, opened 2024-02-07
ModalTFT, ModalDefector strategies - Ben McGuffog

benjjo:
Hi, I've added two strategies that are interesting (well at least I think). Here are their descritpions:

ModalTFT
    A player starts by cooperating and then analyses the history of the opponent. If the opponent Cooperated in the
    last round, they are returned with a Cooperation. If the opponent chose to Defect in the previous round,
    then this strategy will return with the mode of the previous opponent responses.
 
ModalDefector
    A player starts by Defecting and then analyses the history of the opponent. If the opponent Cooperated in the
    last round, they are returned with a Defection. If the opponent chose to Defect in the previous round,
    then this strategy will return with the mode of the previous opponent responses.
 
Interestingly, when pitted against a few of the classics from the orignal Axelrod tournement, they do quite well. When tested agianst all of the strategies in this repository, they rank 103 and 200 respectively. 

I have never written code for other people in my life. This would be the first. My tests, while simple, seem to pass though. 
This is also my first pull request ever so I'm probably doing a number on it. 
I have tried searching through your code to see if these strategies already exist and I haven't found any like them. It may be that I've simply missread the code and im wasting your time right now, for which I appologise in advance. 
It would be cool to have the contributors badge on my account though.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,7 +79,7 @@ The simplest way to install is::
 
 To install from source::
 
-    $ git clone https://github.com/Axelrod-Python/Axelrod.git
+    $ git clone https://github.com/benjjo/Axelrod.git
     $ cd Axelrod
     $ python setup.py install
 
--- a/axelrod/strategies/defector.py
+++ b/axelrod/strategies/defector.py
@@ -1,5 +1,6 @@
 from axelrod.action import Action
 from axelrod.player import Player
+import statistics
 
 C, D = Action.C, Action.D
 
@@ -61,3 +62,35 @@ class TrickyDefector(Player):
         ):
             return C
         return D
+
+
+class ModalDefector(Player):
+    """
+    A player starts by Defecting and then analyses the history of the opponent. If the opponent Cooperated in the
+    last round, they are returned with a Defection. If the opponent chose to Defect in the previous round,
+    then this strategy will return with the mode of the previous opponent responses.
+    """
+
+    # These are various properties for the strategy
+    name = "Modal Defector"
+    classifier = {
+        "memory_depth": float("inf"),
+        "stochastic": False,
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+    def strategy(self, opponent: Player) -> Action:
+        """This is the actual strategy"""
+        # First move
+        if not self.history:
+            return D
+        # React to the opponent's historical moves
+        if opponent.history[-1] == C:
+            return D
+        else:
+            # returns with the mode of the opponent's history.
+            return statistics.mode(opponent.history)
+
--- a/axelrod/strategies/titfortat.py
+++ b/axelrod/strategies/titfortat.py
@@ -956,8 +956,9 @@ class BurnBothEnds(Player):
 
 class ModalTFT(Player):
     """
-    A player starts by cooperating and then analyses the history of the opponent. From this information it will
-    respond with the mode of the opponents historical responses.
+    A player starts by cooperating and then analyses the history of the opponent. If the opponent Cooperated in the
+    last round, they are immediately returned with a Cooperation. If the opponent chose to Defect in the previous round,
+    then this strategy will return with the mode of the previous opponent responses.
     """
 
     # These are various properties for the strategy
@@ -977,5 +978,8 @@ class ModalTFT(Player):
         if not self.history:
             return C
         # React to the opponent's historical moves
-        return statistics.mode(opponent.history)
+        print(statistics.mode(opponent.history))
+        return C if opponent.history[-1] == C else statistics.mode(opponent.history)
+
+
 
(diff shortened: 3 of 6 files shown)


marcharper:
Hi, there's one doc test failing: 
```
File "/home/runner/work/Axelrod/Axelrod/./docs/index.rst", line 55, in index.rst
Failed example:
    len(axl.strategies)
Expected:
    240
Got:
    242


Ran 65 tests in 76.362s

FAILED (failures=1)
```

It's because you added two new strategies.

Also, are you sure there's no similar strategies in the library? Many are TFT with a tweak like what you've done.

benjjo:
Thanks Marc,

I'll go through them all again. I'm a little more familiar with the whole
package now.

README.rst
```diff
@@ -79,7 +79,7 @@ The simplest way to install is::
 
 To install from source::
 
-    $ git clone https://github.com/Axelrod-Python/Axelrod.git
+    $ git clone https://github.com/benjjo/Axelrod.git
```
marcharper, line 82: We'll want to keep this as is for others.

axelrod/strategies/defector.py
```diff
@@ -61,3 +62,35 @@ def strategy(self, opponent: Player) -> Action:
...
+        """This is the actual strategy"""
+        # First move
+        if not self.history:
+            return D
+        # React to the opponent's historical moves
+        if opponent.history[-1] == C:
+            return D
+        else:
+            # returns with the mode of the opponent's history.
+            return statistics.mode(opponent.history)
```
marcharper, line 95: Rather than use the statistics library, the history class already keeps a running count of C and D, so you can do something like `if opponent.history.cooperations > opponent.history.defections return C else return D`. That saves a library import and is more efficient (mode is going to be O(n^2) over the entire game, this way is linear).

axelrod/strategies/titfortat.py
```diff
@@ -945,9 +946,41 @@ def strategy(self, opponent):
...
+        """This is the actual strategy"""
+        # First move
+        if not self.history:
+            return C
+        # React to the opponent's historical moves
+        if opponent.history[-1] == C:
+            return C
+        else:
+            # returns with the mode of the opponent's history.
+            return statistics.mode(opponent.history)
```
marcharper, line 985: Same comment re: statistics.mode

axelrod/tests/strategies/test_defector.py
```diff
@@ -61,3 +61,21 @@ def test_defects_if_opponent_last_three_are_not_D(self):
...
+    expected_classifier = {
+        "memory_depth": float("inf"),
+        "stochastic": False,
+        "makes_use_of": set(),
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+    def test_strategy(self):
```
marcharper, line 78: Try to add a few more tests that single out the specific behaviors like the first move and that the mode works for both C and D. If possible try to use a real strategy like Alternator rather than mock players, but mocks are ok if it's hard to target a particular case.

marcharper (commented):
Left some feedback, should you find that these are new strategies.

axelrod/strategies/defector.py
```diff
@@ -61,3 +62,35 @@ def strategy(self, opponent: Player) -> Action:
...
+
+
+class ModalDefector(Player):
+    """
+    A player starts by Defecting and then analyses the history of the opponent. If the opponent Cooperated in the
+    last round, they are returned with a Defection. If the opponent chose to Defect in the previous round,
+    then this strategy will return with the mode of the previous opponent responses.
+    """
+
+    # These are various properties for the strategy
```
Nikoleta-v3, line 74: I believe this comment/line is from our contributing guidelines. You can remove it.

axelrod/strategies/titfortat.py
```diff
@@ -945,9 +946,41 @@ def strategy(self, opponent):
...
+
+
+class ModalTFT(Player):
+    """
+    A player starts by cooperating and then analyses the history of the opponent. If the opponent Cooperated in the
+    last round, they are returned with a Cooperation. If the opponent chose to Defect in the previous round,
+    then this strategy will return with the mode of the previous opponent responses.
+    """
+
+    # These are various properties for the strategy
```
Nikoleta-v3, line 964: Same here.

Nikoleta-v3 (commented):
@benjjo thank you very much for your contribution! I agree with `marcharper` that you should add a few more tests for the strategies.

I am assuming you designed these strategies? If that's the case, could you add a line in the docstring like this:

```
     Names: 
  
     - Modal TFT: Original name by benjjo or Ben McGuffog
```

Here's an example: https://github.com/Axelrod-Python/Axelrod/blob/fa748627cd4f0333bb2dbfcb1454372a78a9098a/axelrod/strategies/qlearner.py#L12-L22

PR #1435, opened 2024-03-08
Simplify the doctest.

drvinceknight:
The tests were failing on ubuntu
https://github.com/Axelrod-Python/Axelrod/actions/runs/8198844499/job/22423011180?pr=1428

This was caused by some rounding error.

I think this cleans things up.


drvinceknight:
I do not know what's causing the rtd failure  😱🫠

PR #1438, opened 2024-03-17
Greedy and EpsilonGreedy strategies, using Multi-armed Bandits algorithms

bing-j:
Hello! I wrote some strategies that use armed bandit algorithms. Originally, I only wanted to implement the epsilon-greedy strategy, but I now plan on extending this effort and implementing all the algorithms mentioned in the multi-armed bandit chapter of Sutton's Reinforcement Learning: an Introduction (I added the reference to the bibliography). So the branch name is no longer very representative; I added both Greedy and EpsilonGreedy on this branch.

Greedy:
Always chooses the action that has the highest average/expected "reward" (score), calculated from its own previous turns. The reward function is updated incrementally and optionally recency weighted, and initial expected rewards of each action default to zero if not modified through a parameter.

EpsilonGreedy:
Mostly works like Greedy (with p=1-e), but sometimes acts randomly (with p=e).

These strategies are described in detail in the textbook mentioned above as well.

As I've mentioned on gitter, I was unable to find any strategies that implement these algorithms, although I did find some similar ones. For example, Adaptive() works similarly to Greedy() without weights, but has a hard coded initial sequence, and uses raw sum of scores to choose the optimal play instead of average score. (Side note: the comments in Adaptive().strategy() indicate that it was intended to use the highest average; this may be an error in the code!) If similar strategies already exist, and/or there's any modifications I need to make in the code, please let me know!

Cheers :)


axelrod/strategies/_strategies.py
```diff
@@ -89,6 +89,7 @@
 from .dbs import DBS
 from .defector import Defector, TrickyDefector
 from .doubler import Doubler
+from .armed_bandits import Greedy, EpsilonGreedy
```
marcharper: Maybe just `bandits`

axelrod/strategies/armed_bandits.py
```diff
@@ -0,0 +1,156 @@
...
+    classifier = {
+        "memory_depth": float("inf"),
+        "stochastic": False,
+        "long_run_time": False,
+        "inspects_source": False,
+        "manipulates_source": False,
+        "manipulates_state": False,
+    }
+
+    UNIFORM = np.inf  # constant that replaces weight when rewards aren't weighted
```
marcharper: Is there another conceivable value?

axelrod/strategies/armed_bandits.py
```diff
@@ -0,0 +1,156 @@
...
+        recency_weight
+            0.0 <= recency_weight <= 1.0
+            The exponential recency weight used in calculating the average reward.
+            If this argument is not provided, the player will not weigh rewards based on recency.
+        """
+        super().__init__()
+        self._rewards = {C: init_c_reward, D: init_d_reward}
+        self.weight = recency_weight
+
+        # treat out of range values as extremes
```
marcharper: Rephrase the comment to something like "limit parameter value range"

axelrod/strategies/armed_bandits.py
```diff
@@ -0,0 +1,156 @@
...
+            )
+        else:
+            weight = self.weight
+
+        self._rewards[last_play] = self._rewards[last_play] + weight * (
+            last_score - self._rewards[last_play]
+        )
+
+    def strategy(self, opponent: Player) -> Action:
+        """Actual strategy definition that determines player's action."""
```
marcharper: You can remove this comment; for the others try to use complete sentences.

axelrod/strategies/armed_bandits.py
```diff
@@ -0,0 +1,156 @@
...
+        if len(self.history) != 0:
+            self.update_rewards(opponent)
+
+        # select the optimal play
+        return max(self._rewards, key=self._rewards.get)
+
+
+class EpsilonGreedy(Greedy):
+    """
+    Has a 1 - epsilon probability of behaving like Greedy(), and plays randomly otherwise.
```
marcharper: Greedy() --> Greedy

Can you elaborate more on "plays randomly otherwise"?

axelrod/tests/strategies/test_armed_bandits.py
```diff
@@ -0,0 +1,123 @@
...
+        # never explores/always exploits
+        actions = [(C, C), (C, C), (C, C)]
+        self.versus_test(
+            axl.TitForTat(),
+            expected_actions=actions,
+            attrs={"_rewards": {C: 3, D: 0}},
+            seed=1,
+        )
+
+    # temporary overriding function used to search for seeds
```
marcharper: Remove this if it's not needed at runtime.

marcharper (changes requested):
Thanks for the contribution, looks interesting! Most of the feedback is just on matching style and improving the comments.

marcharper:
Looks like we broke the test invocator with some recent commits, I'll try to fix it. You'll need to update one of the doc tests to indicate that two new strategies have been added.

axelrod/strategies/armed_bandits.py
bing-j: I've changed this to -1.0, and changed other places in the code to refer to this constant for consistency. This does mean that if a user uses recency_wieght=-1.0 at time of creation, it will be treated as not recency weighted (instead of an out of range value limited to 0.0 as in previous implementations).

axelrod/strategies/armed_bandits.py
bing-j: Changed to "Has a 1 - epsilon probability of behaving like Greedy; otherwise, randomly choose to cooperate or defect."

marcharper:
Thanks for the updates. The test that's failing is:

```
FAIL: test_strategy (axelrod.tests.strategies.test_meta.TestNMWEDeterministic.test_strategy)
  File "/home/runner/work/Axelrod/Axelrod/axelrod/tests/strategies/test_meta.py", line 636, in test_strategy
    self.versus_test(
  File "/home/runner/work/Axelrod/Axelrod/axelrod/tests/strategies/test_player.py", line 580, in versus_test
    test_match.versus_test(
  File "/home/runner/work/Axelrod/Axelrod/axelrod/tests/strategies/test_player.py", line 665, in versus_test
    self.assertEqual((i, play), (i, expected_play))
AssertionError: Tuples differ: (2, D) != (2, C)

First differing element 1:
D
C

- (2, D)
?     ^ 

+ (2, C)
?     ^
```

This is happening because there are some ensemble strategies and the behavior of one of them has changed with the addition of these new strategies. You can run these tests with something like

```
python -m unittest axelrod.tests.unit.test_meta.py
```

I think in this case you just need to update the expected output that has changed now.

marcharper:
Hi @bing-j, if you rebase onto the dev branch now the failing test should pass.

PR #1439, opened 2024-03-19
Move to modern packaging setup

drvinceknight:
@marcharper this is a lazy move to using pyproject.toml with [flit](https://flit.pypa.io/).
This is the same setup I use for Nashpy so it wasn't much work to throw
the files around.

This will in no way pass as is: it's a WIP (as you mentioned in our
call, this needs the static files to be plugged in etc...).

You mentioned that you were going to work on this so happy for what I've
thrown here to be ignored (it really just took me 3 minutes) but if
you/we/us wanted to continue with this it would still need:

- Actually getting the build to work (I'm sure there are broken static
  links etc);
- Documentation:
  - the development suite uses [`tox`](https://tox.readthedocs.io/) here so we'd need to document that.
  - there might be some links in the documentation that are broken as
    I've moved the files to `src/axelrod` and moved the `tests` to their
    seperate dir.

Let me know what you think, I can probably keeping banging rocks
together to get the above done :)


marcharper:
Happy to let you do it.

If we're going to mess with a lot of links, should we consider swapping axelrod and axelrod-python as the repository names? Or is that likely to break a lot of other things?

drvinceknight:
> Happy to let you do it.
> 
> If we're going to mess with a lot of links, should we consider swapping axelrod and axelrod-python as the repository names? Or is that likely to break a lot of other things?

That's a good idea! Hadn't thought about that. I don't think that would break much stuff at all...

drvinceknight:
Closing this as I realised I was just creating too big a diff this way. Have opened #1440 which keeps the structure the same and just moves to `pyproject.toml` and `tox`

PR #1440, opened 2024-03-25
Move to pyproject.toml.

drvinceknight:
This uses hatch.

Note I've removed the human player. I don't believe we ever use it. If we want to we can add it back in.


marcharper:
Looks good. I don't personally use the Human strategy but maybe others do. Is there an issue with prompt-toolkit or some other reason it was removed?

drvinceknight:
> Looks good. I don't personally use the Human strategy but maybe others do. Is there an issue with prompt-toolkit or some other reason it was removed?

It has always required a bit more maintenance work in terms of compatibility (we've always had to pin it to some older version which sometimes causes problems -- I've often had to fix things on my students' machine to essentially break the human player quietly when they install Axelrod). 

I'd prefer to remove it, we can always add it back in (or having a separate human lib?) if someone really wants it (and we could take a look at using a more up to date version of `prompt_toolkit` and/or use something like `textual`: https://www.textualize.io).

marcharper:
Ok, let's maybe open an issue saying the human strategy is removed so people can let us know if they use it. If so we can reconsider, and a separate library makes sense to compartmentalize the downsides. Probably we should bump up the version number as well since it's technically removing functionality and people can pin an older version for now if they need it.

marcharper (approved):

drvinceknight:
So bump to `4.

> Ok, let's maybe open an issue saying the human strategy is removed so people can let us know if they use it. If so we can reconsider, and a separate library makes sense to compartmentalize the downsides. Probably we should bump up the version number as well since it's technically removing functionality and people can pin an older version for now if they need it.

Good call, [1e22e0f](https://github.com/Axelrod-Python/Axelrod/pull/1440/commits/1e22e0f6c4e3144c1e261c2bd602377f0993c3fc) bumps to `4.14.0`

drvinceknight:
@Nikoleta-v3 if you get a chance can you take a look at this?

Nikoleta-v3 (approved):

PR #1445, opened 2024-07-10
Update citations.md

marcharper:

PR #1444, opened 2024-07-05
add freqAnalyzer strategy

miller-ian:

PR #1446, opened 2024-07-11
bump deps

miller-ian:
numpy + tox issues resolved with upgrade to numpy version 1.26.4

Tox py312 defaults to using numpy 2.0. Some numpy issues have been identified [here](https://github.com/Axelrod-Python/Axelrod/pull/1444) and [here](https://github.com/Axelrod-Python/Axelrod/pull/1445). Dictating numpy version 1.26.4 (most recent version prior to 2.0) in tox.ini file resolves issues


marcharper:
Some of the types have changed, causing test failures like so:
```
  037     >>> import axelrod
  038     >>> axelrod.game.DefaultGame.RPST()
  Expected:
      (3, 1, 0, 5)
  Got:
      (np.int64(3), np.int64(1), np.int64(0), np.int64(5))
```

To be clear this appears to be due to numpy changes, not any of our type hints.

miller-ian:
> Some of the types have changed, causing test failures like so:
> 
> ```
>   037     >>> import axelrod
>   038     >>> axelrod.game.DefaultGame.RPST()
>   Expected:
>       (3, 1, 0, 5)
>   Got:
>       (np.int64(3), np.int64(1), np.int64(0), np.int64(5))
> ```
> 
> To be clear this appears to be due to numpy changes, not any of our type hints.

yep, this is an issue with tox + numpy. tox appears to default to using latest numpy version which, as you mention, gives us some problems. Reverting to numpy 1.26.4 seems to solve issue

tox.ini
```diff
@@ -27,6 +27,7 @@ deps =
     pytest-sugar
     isort
     black
+    numpy==1.26.4
```
drvinceknight, line 30: Should we pin this in `pyproject.toml` instead of here?

drvinceknight:
Thanks for this: it looks good to me.

tox.ini
miller-ian: its in there as well!

drvinceknight: In there it's not hard pinned. Just a lower bound. I wonder if it's worth hard pinning it.

drvinceknight: We have avoided doing this in the past (I think mainly on my request?) but I wonder if it's worth starting to do (in our installation docs we do recommend using a venv etc...)...

We shouldn't delay merging this though, @miller-ian's contribution is a super helpful bug fix that lets other PRs get through :)

drvinceknight (approved):

PR #1453, opened 2024-11-05
Change cache pattern

gaffney2010:
Switches caching pattern to check-cache-and-early-exit to make code slightly more readable.

Previously, caching was only read if the strategy was not stochastic.  This new pattern doesn't save caches in the stochastic case.


gaffney2010:
I'll figure out why these tests are failing.

marcharper (approved):

PR #1454, opened 2024-11-05
Reduce match init complexity

gaffney2010:
A few minor changes to improve readability on Match __init__:

- Set fields in the same order as the function arguments.
- Reduce unnecessary elses.
- Change difficult-to-parse default dict.

### DIFF
--- a/axelrod/match.py
+++ b/axelrod/match.py
@@ -56,7 +56,6 @@ class Match(object):
             Random seed for reproducibility
         """
 
-
         self.turns, self.prob_end = turns, prob_end
         if prob_end is None:
             self.prob_end = 0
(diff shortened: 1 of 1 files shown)


gaffney2010:
I'll figure out why the test is failing.

drvinceknight:
Looks good to me.

drvinceknight (approved):

axelrod/match.py
```diff
@@ -56,42 +56,40 @@ def __init__(
...
 
+        self._cache = deterministic_cache
         if deterministic_cache is None:
             self._cache = DeterministicCache()
-        else:
-            self._cache = deterministic_cache
 
+        self.match_attributes = match_attributes
         if match_attributes is None:
+            # known_turns = inf if both prob_end and turns are None, else turns
```
marcharper, line 80: This seems inconsistent with the above, no? If both are `None` then it should be `DEFAULT_TURNS`.

Also, do we need to repeat the logic here or can we just use `self.turns`?

PR #1455, opened 2024-11-13
Fix Graph vertices logic and use Graph in MatchGenerator

gaffney2010:
`complete_graph` already exists in the graph.py file, so change match_generator.py to use this logic.

This change uncovered a bug in `Graph` where vertices only includes source nodes in a connected graph.  So I fixed this real quick.


gaffney2010:
This is changing output order on a bunch of tests.  I'll just leave this closed.

PR #1456, opened 2024-11-13
Create MatchChunk dataclass

gaffney2010:
This avoids necessitating agreement on ordering between setting and reading.  This is slightly more readable and maintainable.

This doesn't break API, because match_generator isn't imported on __init__.


gaffney2010:
Need to resort imports, then will reopen.

drvinceknight (approved):

drvinceknight:
lgtm

PR #1457, opened 2024-11-16
Get mypy working again

### LINKED ISSUE #1406: CI failing due to typing issues
Hello everyone 👋🏻  I hope you are enjoying the holidays. 

Currently the CI is failing https://github.com/Axelrod-Python/Axelrod/actions/runs/3773670578/jobs/6415295764, because of typing issues.  I had a look and apparently there are two types of issues: 

1. “Incompatible default for argument”

For example this error occurs in `finite_state_machines.py` at the `__init__` of class `EvolvableFSMPlayer`: 

```python
class EvolvableFSMPlayer(FSMPlayer, EvolvablePlayer):
    """Abstract base class for evolvable finite state machine players."""
   ...
    def __init__(
        self,
        transitions: tuple = None,
        initial_state: int = None,
        initial_action: Action = None,
        num_states: int = None,
        mutation_probability: float = 0.1,
        seed: int = None,
    ) -> None:
```

Here we declare that `transitions` is a `tuple` but set `None` as a default value and `mypy` is not happy with this. I believe there are two ways to fix the above, we either give some default values that have are of correct type or we can use the `Optional`  type modifier as described on the `mypy` documentation: https://mypy.readthedocs.io/en/stable/kinds_of_types.html. 

For example if we tweak the `__init__` class to be as follows:

```python
    def __init__(
        self,
        transitions: Optional[tuple] = None,
        initial_state: Optional[int] = None,
        initial_action: Optional[Action] = None,
        num_states: Optional[int] = None,
        mutation_probability: float = 0.1,
        seed: Optional[int] = None,
    ) -> None:
```

the tests pass.  

2. "Missing return statement"

These errors occur because all the return statements are under `if` statements, and  `mypy` is complaining that there is no return statement in case all of the `if` statements fail.  This can be fixed by slightly changing the code. 

I am happy to work on this 👍🏻 could you please let me know what you prefer regarding the errors “Incompatible default for argument”?

gaffney2010:
Resolves #1406

### DIFF
--- a/axelrod/deterministic_cache.py
+++ b/axelrod/deterministic_cache.py
@@ -14,7 +14,7 @@ else:
 
 import pickle
 from collections import UserDict
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 from axelrod import Classifiers
 
@@ -104,7 +104,7 @@ class DeterministicCache(UserDict):
     methods to save/load the cache to/from a file.
     """
 
-    def __init__(self, file_name: str = None) -> None:
+    def __init__(self, file_name: Optional[str] = None) -> None:
         """Initialize a new cache.
 
         Parameters
--- a/axelrod/ecosystem.py
+++ b/axelrod/ecosystem.py
@@ -12,7 +12,7 @@ ecosystem.reproduce(100)
 """
 
 import random
-from typing import Callable, List
+from typing import Callable, List, Optional
 
 from axelrod.result_set import ResultSet
 
@@ -29,8 +29,8 @@ class Ecosystem(object):
     def __init__(
         self,
         results: ResultSet,
-        fitness: Callable[[float], float] = None,
-        population: List[int] = None,
+        fitness: Optional[Callable[[float], float]] = None,
+        population: Optional[List[int]] = None,
     ) -> None:
         """Create a new ecosystem.
 
--- a/axelrod/fingerprint.py
+++ b/axelrod/fingerprint.py
@@ -1,7 +1,7 @@
 import os
 from collections import namedtuple
 from tempfile import mkstemp
-from typing import Any, List, Union
+from typing import Any, List, Optional, Union
 
 import dask.dataframe as dd
 import matplotlib.pyplot as plt
@@ -280,10 +280,10 @@ class AshlockFingerprint(object):
         turns: int = 50,
         repetitions: int = 10,
         step: float = 0.01,
-        processes: int = None,
-        filename: str = None,
+        processes: Optional[int] = None,
+        filename: Optional[str] = None,
         progress_bar: bool = True,
-        seed: int = None,
+        seed: Optional[int] = None,
     ) -> dict:
         """Build and play the spatial tournament.
 
@@ -358,7 +358,7 @@ class AshlockFingerprint(object):
         self,
         cmap: str = "seismic",
         interpolation: str = "none",
-        title: str = None,
+        title: Optional[str] = None,
         colorbar: bool = True,
         labels: bool = True,
     ) -> plt.Figure:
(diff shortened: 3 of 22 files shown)


gaffney2010:
Need to test adding new exception handling.  Will reopen.

drvinceknight:
Looks like this is failing on coverage:

```
axelrod/load_data_.py                                               37      1    97%   30
```

gaffney2010:
> Looks like this is failing on coverage:
> 
> ```
> axelrod/load_data_.py                                               37      1    97%   30
> ```

But why?  That line is tested.  Any idea?

drvinceknight:
> But why? That line is tested. Any idea?

Could it be the `assertRaisesRegex` doing something funky? Would it be worth trying `assertRaises`? That's a very underwhelming idea that I'm not convinced by myself.

gaffney2010:
> Could it be the `assertRaisesRegex` doing something funky? Would it be worth trying `assertRaises`? That's a very underwhelming idea that I'm not convinced by myself.

I actually had assertRaises that I changed thinking perhaps it wanted to test the text of the error message.

gaffney2010:
Ah, so get_data raises FileNotFoundError in my test case, but returns None for some other kinda failure.  I'll figure it out.

gaffney2010:
I chose to inject the dependency on `get_data` for the sake of testing because:
1.  It's a little hard to figure out [this failure path](https://github.com/python/cpython/blob/5eb7fd4d0fc37b91058086181afebec41e66e5ad/Lib/pkgutil.py#L382-L392)
2. It seems to be [missing tests](https://github.com/python/cpython/blob/5eb7fd4d0fc37b91058086181afebec41e66e5ad/Lib/pkgutil.py#L388) even in the core library.
3. It may require the library to be formatted different (whatever PEP 302 is)

gaffney2010:
Oh, mypy is giving a lot more errors.  I'm not sure why I thought it was resolved.  Will reopen after addressing the others.

gaffney2010:
> Oh, mypy is giving a lot more errors. I'm not sure why I thought it was resolved. Will reopen after addressing the others.

Ah, this was passing locally.  I'm gonna continue to test via CI / build

gaffney2010:
Everything seems to be passing now.

drvinceknight:
I can't seem to get any information from read the docs about the failed build there.

CleanShot 2024-12-10 at 12 48 21@2x


Do the docs build locally for you? If they do it could be a read the docs thing instead of a sphinx thing...

gaffney2010:
> Do the docs build locally for you? If they do it could be a read the docs thing instead of a sphinx thing...

Yes.

image

gaffney2010:
> Do the docs build locally for you? If they do it could be a read the docs thing instead of a sphinx thing...

`python doctest.py` runs without errors

image

Is there a different command to build?

marcharper:
That's just the doc tests, you need to also build the docs by running `make`. I thought we had this documented but I wasn't easily able to find it just now.

drvinceknight:
> That's just the doc tests, you need to also build the docs by running `make`. I thought we had this documented but I wasn't easily able to found it just now.

You're right Marc, that's not documented at all. We should put it here right? https://axelrod.readthedocs.io/en/stable/how-to/contributing/running_tests.html#testing-the-documentation.

I believe:

```
cd docs
make html
```

is what is needed.

gaffney2010:
`make html` runs locally and creates no new files.

image

When we update the docs for `make html`, can we specify which version of Sphinx we're using - I had compatibility issues.

marcharper:
@drvinceknight yes that seems like a good place.

Where are the commands that CI is running to build the docs?

drvinceknight:
It's using the readthedocs config file which is here: https://github.com/Axelrod-Python/Axelrod/blob/dev/.readthedocs.yaml

Here's info about the file itself: https://docs.readthedocs.io/en/stable/config-file/v2.html

The rest is essentially done through the readthedocs configuration panel. As this isn't changing anything relevant to the docs how about we merge it, then I'll open a PR to fix the docs build?

drvinceknight:
The fact that the other recent PRs pass rtd but not this one is weird...

PR #1460, opened 2025-01-02
Simplify moran mutate function

gaffney2010:
Makes a few changes for readability:
1.  Handle the non-mutation case as an early exit.
2. Don't handle the 0-mutation-rate case separately, as this will simply always result in an early exit.
3. Draw directly from the initial population, as commented, instead of making `mutation_targets` in the init.

Many tests are failing because we've changed how many random numbers get drawn and when.  If the code looks good to a reviewer, I'll update the tests.

### DIFF
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -122,6 +122,12 @@ class MoranProcess(object):
         self._random = RandomGenerator(seed=seed)
         self._bulk_random = BulkRandomGenerator(self._random.random_seed_int())
         self.set_players()
+        
+        # Dedupe initial players as mutation targets
+        initial_players_by_name = {
+            player.name: player for player in self.initial_players
+        }
+        self.mutation_targets = [pi for pi in initial_players_by_name.values()]
 
         if interaction_graph is None:
             interaction_graph = complete_graph(len(players), loops=False)
@@ -211,7 +217,7 @@ class MoranProcess(object):
         # Choose another strategy at random from the initial population
         player = None
         while player is None or str(player) == str(self.players[index]):
-            player = self._random.choice(self.initial_players)
+            player = self._random.choice(self.mutation_targets)
         return player.clone()
 
     def death(self, index: int = None) -> int:
--- a/axelrod/moran.py
+++ b/axelrod/moran.py
@@ -122,20 +122,6 @@ class MoranProcess(object):
         self._random = RandomGenerator(seed=seed)
         self._bulk_random = BulkRandomGenerator(self._random.random_seed_int())
         self.set_players()
-        # Build the set of mutation targets
-        # Determine the number of unique types (players)
-        keys = set([str(p) for p in players])
-        # Create a dictionary mapping each type to a set of representatives
-        # of the other types
-        d = dict()
-        for p in players:
-            d[str(p)] = p
-        mutation_targets = dict()
-        for key in sorted(keys):
-            mutation_targets[key] = [
-                v for (k, v) in sorted(d.items()) if k != key
-            ]
-        self.mutation_targets = mutation_targets
 
         if interaction_graph is None:
             interaction_graph = complete_graph(len(players), loops=False)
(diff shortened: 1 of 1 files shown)


marcharper:
In this case I think the comment is wrong, it should be drawing from the mutation targets, not the initial population.

axelrod/moran.py
```diff
@@ -218,16 +204,15 @@ def mutate(self, index: int) -> Player:
...
-            # Choose another strategy at random from the initial population
-            r = self._random.random()
-            if r < self.mutation_rate:
-                s = str(self.players[index])
-                j = self._random.randrange(0, len(self.mutation_targets[s]))
-                p = self.mutation_targets[s][j]
-                return p.clone()
-        # Just clone the player
-        return self.players[index].clone()
+        if self._random.random() > self.mutation_rate:
```
drvinceknight, line 213: I believe but I might be remembering incorrectly that we had it this way to avoid sampling a random number if no such sample was needed. I think this helped with performance.

gaffney2010: Did some quick tests, and I'm seeing that drawing a random number takes 1.4 microseconds, while an if takes 2.1 microseconds.  We actually reduce an if when mutation_rate = 0, so we may be saving time.  But moreover, it doesn't seem like a meaningful optimization.

I'm not sure, though.  Should we add a performance test?

drvinceknight: > I'm not sure, though. Should we add a performance test?

Yeah sounds like a good idea to be sure. Running a reasonably sized tournament perhaps? (I have played around with https://pypi.org/project/pytest-benchmark/ which is a neat way to get benchmarks but I have no idea if this is the sort of thing you were thinking: whatever you think is best :) -- no need to do a huge amount for this PR)

axelrod/moran.py
```diff
@@ -122,20 +122,6 @@ def __init__(
         self._random = RandomGenerator(seed=seed)
         self._bulk_random = BulkRandomGenerator(self._random.random_seed_int())
         self.set_players()
-        # Build the set of mutation targets
```
marcharper, line 125: This code was first added [here](https://github.com/Axelrod-Python/Axelrod/commit/4932c987a2660f258cca4b300c02f686f85a8088). The idea is that a player can't mutate into the same type if the random pull dictates a mutation (that's the case if the if clause fails). Typically in the literature for the Moran process when mutating one only chooses from the _set_ of initial types (regardless of their prevalence). So the point of the `mutation_targets` dict is to (1) remove duplicate types from the initial population, and (2) remove the current type. Later, when you want to mutate a player, you can just choose uniformly at random from the mutation dictionary's possibilities, i.e. we've cached all the logic for the choices. (Note this assumes that no new player types are formed, that's a different kind of mutation.)

This is not well explained in the comments, and also `mutation_targets` might be better named `mutation_possibilities` or something else. In any case I think the logic of the current code is correct.

axelrod/moran.py
marcharper: Note that the class now uses the bulk random number generator, which should be faster at scale. This came later and may have altered the tradeoffs.

axelrod/moran.py
gaffney2010: Okay, I missed a subtlety that this was deduping initial_strategies.  I changed the code to create a single list of `mutation_targets` and leave the responsibility of choosing a different type to the `mutate` function.  A little closer now to the original code, but still (imo) more easily to understand.

axelrod/moran.py
gaffney2010: I will add a benchmark in another change, coming soon.

axelrod/moran.py
marcharper: But isn't this less efficient? If the number of initial types is small (which is commonly the case) then the updated code may call for a lot of extra random numbers. The existing code removes the need to check for the same type. Agreed that it could use some cleaning up and clarification though.

The algorithm is supposed to, for given mutation rate \mu, chose another type at random with probability \mu / (n-1) for the other (n-1) types and (1-\mu) for the same type. Maybe there's a simpler/clearer way to do this without sacrificing efficiency.

gaffney2010: At most it would double (on avg) the number of randoms drawn.  My gut says this is insignificant.  I will add a performance test to see.

gaffney2010:
Top table before change, bottom table after change.
image

Looks like there is significant performance regression.  I will close the PR.  Will follow-up with:
- Performance script
- Potentially some comments here.

marcharper:
Thanks for looking into it, let's definitely update the comments to better explain what's happening and why. And maybe exiting early on `mutation_rate == 0` will improve the readability a bit.

PR #1464, opened 2025-01-08
Add/update a few citations

marcharper:


Nikoleta-v3:
I just noticed that we are not using any specific style when listing these citations. Is this something we should care about?

marcharper:
I just copy the Google Scholar citation text and add a direct link. We could choose a different or consistent format, if you think it's worth it. I update the citations occasionally so people can see that the library is actively used so I haven't been particularly concerned with the format.

PR #1467, opened 2025-01-11
Build moran benchmark

gaffney2010:
This benchmark was used to analyze #1460.

There's a way to integrate pytest-benchmark with tox.  This PR doesn't do that, but may be a good starting point.

### DIFF
--- a/docs/how-to/contributing/running_tests.rst
+++ b/docs/how-to/contributing/running_tests.rst
@@ -58,3 +58,21 @@ You can also run the doctests on any given file. For example, to run the
 doctests for the :code:`docs/tutorials/getting_started/match.rst` file::
 
     $ python -m doctest docs/tutorials/getting_started/match.rst
+
+Performance testing
+-------------------
+
+Performance is not automatically tested as of yet.  To check the performance
+of a change, first install pytest-benchmark::
+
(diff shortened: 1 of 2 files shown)


docs/how-to/contributing/running_tests.rst
```diff
@@ -58,3 +58,21 @@ You can also run the doctests on any given file. For example, to run the
...
 
     $ python -m doctest docs/tutorials/getting_started/match.rst
+
+Performance testing
+-------------------
+
+Performance is not automatically tested as of yet.  To check the performance
+of a change, first install pytest-benchmark::
+
+    $ pip install pytest-benchmark
```
drvinceknight:
```suggestion
    $ python -m pip install pytest-benchmark
```

(I'm sure we're not consistent yet here throughout the docs but nice to catch them when we can :)).

drvinceknight:
Ok so the CI is failing for an upstream change in pandas (I THINK). The fix is to remove an unused line in the result_set file:. This commit [9e8092c](https://github.com/Axelrod-Python/Axelrod/pull/1469/commits/9e8092c672455d03004cd746c4a3ae4e3d1e855c) on #1469 shows what's needed.

PR #1469, opened 2025-01-22
This PR adds the Momentum strategy

dongwonmoon:
Summary: This pull request introduces a new strategy, Momentum, inspired by the concept of Gradual and the Momentum optimizer used in deep learning. The strategy models the dynamics of trust evolution in a repeated game setting, with momentum reflecting how trust shifts based on previous interactions. The momentum formula used in this strategy is derived from the Momentum optimizer in deep learning, which adjusts the velocity of a model's parameters based on gradients and previous updates. Similarly, the Momentum strategy adapts based on the opponent's previous actions.

Key Features:

Momentum Formula: The strategy uses an alpha parameter to weight the influence of the previous momentum value and the current opponent's action (cooperate or defect). This formula is inspired by the Momentum optimizer in deep learning, which accelerates convergence by adjusting updates with a combination of previous gradients and the current gradient.
Threshold: A threshold value determines the decision to cooperate or defect. If the momentum is above the threshold, the strategy cooperates; otherwise, it defects.
Initial Momentum: When the history is empty (first round), the momentum is set to 1.0, and cooperation is guaranteed.
Implementation Details:

The strategy uses an alpha (momentum decay factor) and a threshold to determine the behavior.
The momentum is updated after each round based on the opponent's previous action, using the same principle as the Momentum optimizer.
The strategy's behavior evolves over time, with a more cooperative stance when momentum is high and a defection tendency when momentum is low.

### DIFF
--- a/axelrod/strategies/momentum.py
+++ b/axelrod/strategies/momentum.py
@@ -14,9 +14,9 @@ class Momentum(Player):
     responds to gradients in optimization.
 
     Parameters:
-     - alpha:
-     - threshold:
-     - momentum:
+     - alpha: Momentum decay factor that determines the rate of trust reduction. A higher value leads to slower decay, and the opponent's Defect acts as a trigger. (Optimized by Genetic Algorithm)
+     - threshold: The minimum momentum required to continue cooperation. If momentum falls below this value, the strategy switches to Defect as punishment. (Optimized by Genetic Algorithm)
+     - momentum: Represents the inertia of trust, dynamically changing based on past cooperation.
 
     Names:
      - Momentum: Original name by Dong Won Moon
@@ -35,8 +35,8 @@ class Momentum(Player):
 
     def __init__(
         self,
-        alpha=0.9914655399877477,  # Optimized by Genetic Algorithm. You can try to adapt it to any Env.
-        threshold=0.9676595613724907,  # This one too
+        alpha=0.9914655399877477,
+        threshold=0.9676595613724907,
     ) -> None:
         super().__init__()
         self.alpha = alpha
--- a/axelrod/strategies/momentum.py
+++ b/axelrod/strategies/momentum.py
@@ -16,12 +16,6 @@ class Momentum(Player):
     Names:
      - Momentum: Original name by Dong Won Moon
 
-    Notes:
-     - While I am an undergraduate student with limited experience in game theory, I
-       believe this strategy has potential in various scenarios.
-     - I encourage experts to explore and extend this idea in other contexts, such as
-       environments with noise, one-hot vectorization approaches at multiple actions.
-
     """
 
     name = "Momentum"
--- a/axelrod/strategies/momentum.py
+++ b/axelrod/strategies/momentum.py
@@ -13,6 +13,11 @@ class Momentum(Player):
     create significant and rapid changes in the player's behavior, much like how momentum
     responds to gradients in optimization.
 
+    Parameters:
+     - alpha:
+     - threshold:
+     - momentum:
+
     Names:
      - Momentum: Original name by Dong Won Moon
 
@@ -42,8 +47,8 @@ class Momentum(Player):
         return f"Momentum: {self.alpha}, {self.threshold}"
 
     def update_momentum(self, opponent_action):
-        action_value = 1 if opponent_action == C else 0
         # If the opponent defects, the momentum decreases, reflecting a loss of trust.
+        action_value = 1 if opponent_action == C else 0
         self.momentum = (
             self.alpha * self.momentum + (1 - self.alpha) * action_value
         )
--- a/axelrod/strategies/momentum.py
+++ b/axelrod/strategies/momentum.py
@@ -1,0 +1,64 @@
+from axelrod.action import Action
+from axelrod.player import Player
+
+C, D = Action.C, Action.D
+
+
+class Momentum(Player):
+    """
+    This strategy is inspired by the concept of Gradual and the mathematical foundation of
+    the Momentum optimizer used in deep learning.
+
+    The idea is that trust (or cooperation) evolves dynamically. A shift in trust can
+    create significant and rapid changes in the player's behavior, much like how momentum
+    responds to gradients in optimization.
+
+    Names:
+     - Momentum: Original name by Dong Won Moon
+
+    Notes:
+     - While I am an undergraduate student with limited experience in game theory, I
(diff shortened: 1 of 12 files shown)


drvinceknight:
Thanks for this @dongwonmoon.

The CI is failing with:

```
ImportError: cannot import name 'DataFrame' from 'dask.dataframe.core' 
```

I have been able to recreate this locally by rebuilding the tox environments:

```
$ python -m tox -r
```

The error is not caused by anything you're adding, it's something to do with an upstream update to task (I think!).

I have fixed the error by removing [this line](https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/tests/unit/test_resultset.py#L7) in `test_resultset.py`:

```
from dask.dataframe.core import DataFrame
```

I also noted when running things locally that you're going to need to run `black` on the two new files. If you need a hand with any of this let me know.

dongwonmoon:
Thank you for the help with resolving the previous CI issue. However, a new issue has emerged.
It seems to be related to invalid escape sequences in the strings.

How sould i go about fixing this? @drvinceknight

Nikoleta-v3:
Hey 👋🏻 @dongwonmoon,  

This isn’t exactly where the tests are failing. These are just warnings that invoke a warning message but don’t raise an error.  

The tests are actually failing because of some other code that was recently added to the codebase. More specifically:  

```
axelrod/strategies/frequency_analyzer.py:65: error: Need type annotation for "frequency_table" (hint: "frequency_table: dict[<type>, <type>] = ...")  [var-annotated]
Found 1 error in 1 file (checked 1 source file)
```

In another strategy that was introduced recently, one of the attributes’ types is not specified, and this is causing the MyPy tests to fail. Could you tweak line 65 in `strategies/frequency_analyzer.py` (see: [link](https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/strategies/frequency_analyzer.py)) to:  

```python
self.frequency_table: dict = dict()
```

The tests should pass after that 😅 (I’m about 90% sure). I assume it’ll be okay with the rest of the team if you include this fix in this pull request.

dongwonmoon:
Thank you for the detailed explanation. @Nikoleta-v3 

This is my first PR, so I wanted to check—am I allowed to fix the error myself, or should I leave it for the maintainers to handle?

drvinceknight:
> This is my first PR, so I wanted to check—am I allowed to fix the error myself, or should I leave it for the maintainers to handle?

I can do it if you'd like me to do it BUT usually what happens is you would make the change on the same branch you are already have (just adding a new commit). Then when you push it here it will just update.

Let me know if I can help with anything.

dongwonmoon:
Tests have been completed. Please let me know if there's anything else i should do.

axelrod/strategies/momentum.py
```diff
@@ -0,0 +1,58 @@
...
+class Momentum(Player):
+    """
+    This strategy is inspired by the concept of Gradual and the mathematical foundation of
+    the Momentum optimizer used in deep learning.
+
+    The idea is that trust (or cooperation) evolves dynamically. A shift in trust can
+    create significant and rapid changes in the player's behavior, much like how momentum
+    responds to gradients in optimization.
+
+    Names:
```
drvinceknight: Could you add a few sentences about the role of the alpha, momentum and the threshold here please.

Then you could add a quick description of their default values which means we can remove the comments in the `init` method.

axelrod/strategies/momentum.py
```diff
@@ -0,0 +1,58 @@
...
+        self.threshold = threshold
+        self.momentum = 1.0
+
+    def __repr__(self):
+        return f"Momentum: {self.alpha}, {self.threshold}"
+
+    def update_momentum(self, opponent_action):
+        action_value = 1 if opponent_action == C else 0
+        # If the opponent defects, the momentum decreases, reflecting a loss of trust.
+        self.momentum = (
```
drvinceknight:
```suggestion
    def update_momentum(self, opponent_action):
        """If the opponent defects, the momentum decreases, reflecting a loss of trust."""
        action_value = 1 if opponent_action == C else 0
        self.momentum = (
```

drvinceknight (changes requested):
This looks good to me, I've made some minor stylistic suggestions as well as a request for a bit more documentation.

dongwonmoon:
Edited. I would appreciate it if you could check it out.
Also, can you give me a contributor badge?

dongwonmoon:
I tried running black and isort locally, but nothing changed.
Can you give me some hint about this problem?

drvinceknight:
> Can you give me some hint about this problem?

This could be an upstream update of black:

```
  ===== 5190 passed, 6 skipped, 1 xfailed, 78 warnings in 1163.11s (0:19:23) =====
  py311: commands[1]> python -m black -l 80 . --check
  would reformat /home/runner/work/Axelrod/Axelrod/axelrod/classifier.py
  would reformat /home/runner/work/Axelrod/Axelrod/axelrod/makes_use_of.py
  would reformat /home/runner/work/Axelrod/Axelrod/axelrod/strategies/memorytwo.py
  would reformat /home/runner/work/Axelrod/Axelrod/axelrod/tests/strategies/test_gambler.py
```

The CI installed `black==25.1.0` could you check if that's different to the version you're running?

dongwonmoon:
Thank you. Version was different.
It passed the test!

axelrod/strategies/momentum.py
```diff
@@ -0,0 +1,63 @@
...
+        alpha=0.9914655399877477,
+        threshold=0.9676595613724907,
+    ) -> None:
+        super().__init__()
+        self.alpha = alpha
+        self.threshold = threshold
+        self.momentum = 1.0
+
+    def __repr__(self):
+        return f"Momentum: {self.alpha}, {self.threshold}"
```
Nikoleta-v3: Include `{self.momentum}` in the name as well for completeness.

Nikoleta-v3 (commented):
Thank you, @dongwonmoon, again for your `PR`. One small comment: since the player depends on `alpha`, `threshold`, and `momentum`, all three values can be included in the name.

Everything else looks great!

Nikoleta-v3:
Your `def test_repr(self):` is failing because the name was changed 😄

dongwonmoon:
Thank you. I applied that.
Please check and let me know if there is anything else I need to do

Nikoleta-v3 (approved):

PR #1471, opened 2025-02-28
Add EvolvedAttention: A transformer-based neural network strategy for Prisoner's Dilemma

moderouin:
## Description
This PR introduces the `EvolvedAttention` strategy, a novel approach that uses a transformer neural network with self-attention mechanisms to make decisions. The strategy analyzes game history through attention patterns to determine optimal moves.

## Features
- Transformer architecture with self-attention (24 layers, 8 attention heads)
- Memory depth of 200 moves with reverse-chronological processing
- GPU acceleration when available (with CPU fallback)
- Pre-trained model weights from evolutionary self-play

## Technical Implementation
- Game states encoded as tokens (CC, CD, DC, DD plus special CLS/PAD tokens)
- Position embeddings to maintain sequence information
- Decision boundary using sigmoid activation (< 0.5 → Cooperate, ≥ 0.5 → Defect)

## Performance Considerations
The neural network is relatively complex but runs efficiently on modern hardware. The strategy balances analytical depth with reasonable computational requirements.

### DIFF
--- a/axelrod/data/all_classifiers.yml
+++ b/axelrod/data/all_classifiers.yml
@@ -441,7 +441,7 @@ Evolved ANN 5 Noise 05:
   stochastic: false
 EvolvedAttention:
   inspects_source: false
-  long_run_time: false
+  long_run_time: True
   makes_use_of: !!set {}
   manipulates_source: false
   manipulates_state: false
--- a/axelrod/data/all_classifiers.yml
+++ b/axelrod/data/all_classifiers.yml
@@ -1,4 +1,4 @@
-$\phi$:
+$\\phi$:
   inspects_source: false
   long_run_time: false
   makes_use_of: !!set {}
@@ -6,7 +6,7 @@ $\phi$:
   manipulates_state: false
   memory_depth: .inf
   stochastic: false
-$\pi$:
+$\\pi$:
   inspects_source: false
   long_run_time: false
   makes_use_of: !!set {}
@@ -438,6 +438,14 @@ Evolved ANN 5 Noise 05:
   manipulates_source: false
   manipulates_state: false
   memory_depth: .inf
+  stochastic: false
+EvolvedAttention:
+  inspects_source: false
+  long_run_time: false
+  makes_use_of: !!set {}
+  manipulates_source: false
+  manipulates_state: false
+  memory_depth: 200
   stochastic: false
 Evolved FSM 16:
   inspects_source: false
--- a/axelrod/strategies/attention.py
+++ b/axelrod/strategies/attention.py
@@ -352,7 +352,7 @@ class EvolvedAttention(Player):
         self,
     ) -> None:
         super().__init__()
-        self.model = None
+        self.model: Optional[PlayerModel] = None
 
     def load_model(self) -> None:
         """Load the model weights."""
@@ -366,6 +366,8 @@ class EvolvedAttention(Player):
         """Actual strategy definition that determines player's action."""
         # Load the model if not already loaded
         self.load_model()
+        assert self.model is not None, "Model must be loaded before playing."
+
         # Compute features
         features = compute_features(self, opponent).unsqueeze(0).to(DEVICES)
 
--- a/axelrod/strategies/attention.py
+++ b/axelrod/strategies/attention.py
@@ -18,8 +18,6 @@ PAD_TOKEN = 1
 
 DEVICES = torch.device("cpu")
 
-model_weights = load_attention_model_weights()
-
 
 class GameState(IntEnum):
     CooperateDefect = 2
@@ -354,13 +352,20 @@ class EvolvedAttention(Player):
         self,
     ) -> None:
         super().__init__()
-        self.model = PlayerModel(PlayerConfig())
-        self.model.load_state_dict(model_weights)
-        self.model.to(DEVICES)
-        self.model.eval()
+        self.model = None
+
+    def load_model(self) -> None:
+        """Load the model weights."""
+        if self.model is None:
+            self.model = PlayerModel(PlayerConfig())
+            self.model.load_state_dict(load_attention_model_weights())
+            self.model.to(DEVICES)
+            self.model.eval()
 
     def strategy(self, opponent: Player) -> Action:
         """Actual strategy definition that determines player's action."""
+        # Load the model if not already loaded
+        self.load_model()
         # Compute features
         features = compute_features(self, opponent).unsqueeze(0).to(DEVICES)
 
--- a/axelrod/strategies/attention.py
+++ b/axelrod/strategies/attention.py
@@ -344,7 +344,7 @@ class EvolvedAttention(Player):
     classifier = {
         "memory_depth": MEMORY_LENGTH,
         "stochastic": False,
-        "long_run_time": False,
+        "long_run_time": True,
         "inspects_source": False,
         "manipulates_source": False,
         "manipulates_state": False,
--- a/axelrod/strategies/attention.py
+++ b/axelrod/strategies/attention.py
@@ -1,0 +1,373 @@
+import copy
+from enum import IntEnum
+from typing import Optional, Tuple
+
+import torch
+from torch import nn
+
+from axelrod.action import Action
+from axelrod.load_data_ import load_attention_model_weights
+from axelrod.player import Player
+
+C, D = Action.C, Action.D
+
+MEMORY_LENGTH = 200
+
+CLS_TOKEN = 0
+PAD_TOKEN = 1
+
+DEVICES = torch.device("cpu")
+
+model_weights = load_attention_model_weights()
+
+
+class GameState(IntEnum):
+    CooperateDefect = 2
+    DefectCooperate = 3
+    CooperateCooperate = 4
+    DefectDefect = 5
+
+
+def actions_to_game_state(
+    player_action: Action, opponent_action: Action
+) -> GameState:
+    action_mapping = {
+        (C, D): GameState.CooperateDefect,
+        (D, C): GameState.DefectCooperate,
+        (C, C): GameState.CooperateCooperate,
+        (D, D): GameState.DefectDefect,
+    }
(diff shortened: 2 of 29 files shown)


marcharper:
Hi @moderouin, thanks for your contribution! Have you run the tests locally? I'm wondering if the new tests are very slow or if the issue is with Github's CI.

moderouin:
The test runs within 20 minutes on my PC with Ubuntu 22.04.5 LTS and also seems to work well on Windows and macOS. However, I’m able to reproduce the CI bug within a Docker container using `ubuntu:latest`; the tests never seem to end. It appears to be an issue with `*.rts`. I’m currently trying to fix this, but if you have any insights, that would be great! @marcharper

moderouin:
It seems like the error was caused by the forking of process on Linux. I use spawn instead to avoid any deadlocks @marcharper

marcharper:
How does the strategy perform? Have you run any tournaments?

moderouin:
When running a tournament with 10 repetitions against all the strategies not in long_run_time, it ranked first in median score, with a score close to that of EvolvedLookerUp2_2_2.

marcharper:
Can you tell us more about how you trained it?

moderouin:
This strategy was trained by performing multiple rounds of tournaments against all strategies and the current network of EvolvedAttention. After each tournament, the strategy learns to reproduce the moves of the best-performing strategy from the last tournament.

moderouin:
I’m also currently working on a second version with the same architecture, but incorporating an actor-critic approach with policy and value heads on top of the base network and training it using PPO.

moderouin:
Is there any adjustment I should make to this strategy? @marcharper

marcharper:
No, I just haven't had a chance to review thoroughly, and we require two maintainer review. It's not surprising that it does so well if you trained against all the other strategies (it might be overfit), but that's not a blocker to including it. For your future trainings, try using just the short runtime strategies, which has worked fine for the other ML strategies and saves a lot of computation time.

axelrod/data/all_classifiers.yml
```diff
@@ -439,6 +439,14 @@ Evolved ANN 5 Noise 05:
   manipulates_state: false
   memory_depth: .inf
   stochastic: false
+EvolvedAttention:
+  inspects_source: false
+  long_run_time: false
```
marcharper: Probably should be true. We don't have a formal definition of long run time, but based on the size of the parameters I'm guessing this one is relatively slow compared to most of the strategies in the library.

axelrod/strategies/attention.py
```diff
@@ -0,0 +1,373 @@
...
+
+    Names:
+    - EvolvedAttention: EvolvedAttention by Marc-Olivier Derouin
+    """
+
+    name = "EvolvedAttention"
+    classifier = {
+        "memory_depth": MEMORY_LENGTH,
+        "stochastic": False,
+        "long_run_time": False,
```
marcharper: Same here

axelrod/strategies/attention.py
```diff
@@ -0,0 +1,373 @@
...
+    player_history = player.history[-MEMORY_LENGTH:]
+    player_history = player_history[::-1]
+    opponent_history = opponent.history[-MEMORY_LENGTH:]
+    opponent_history = opponent_history[::-1]
+
+    feature_size = MEMORY_LENGTH + 1 if right_pad else len(player_history) + 1
+
+    game_history = torch.full((feature_size,), PAD_TOKEN, dtype=torch.int)
+    game_history[0] = CLS_TOKEN
+    for index, (action_player, action_opponent) in enumerate(
```
marcharper, line 54: This seems inefficient, since it's called every round and loops through the entire history, so it's going to be repeating a lot of calculations, roughly O(N^2) in the length of the match. IIUC you could cache the calculations (in the Player class) and then just append the latest calculation each time to make this linear.

axelrod/tests/integration/test_matches.py
```diff
@@ -29,7 +29,7 @@ class TestMatchOutcomes(unittest.TestCase):
         ),
         turns=integers(min_value=1, max_value=20),
     )
-    @settings(max_examples=5)
+    @settings(max_examples=5, deadline=None)
```
marcharper: If these tests are slow perhaps it's better to lower max examples. @drvinceknight wdyt?

axelrod/tournament.py
```diff
@@ -3,7 +3,9 @@
 import os
 import warnings
 from collections import defaultdict
-from multiprocessing import Process, Queue, cpu_count
+from multiprocessing import Process, Queue, cpu_count, set_start_method
+
+set_start_method("spawn", force=True)
```
marcharper, line 10: Can you add a comment why this is needed?

tox.ini
```diff
@@ -28,6 +28,7 @@ deps =
     isort
     black
     numpy==1.26.4
```
marcharper, line 30: @drvinceknight is there a reason we've fixed the versions here?

marcharper (changes requested):
Looks good overall, some minor comments. PTAL and thanks for the contribution!

marcharper:
@jsafyan PTAL, you're more of a transformer expert than I.

axelrod/strategies/attention.py
moderouin: Yes, this is ineffective, but I wasn’t sure if we wanted to store the features in the player class. This means we will also need to override the implementation of update_history, since we need to update the features tensor.

Is that what we want?

Considering that the processing time of compute_features is not the bottleneck of this strategy and that there are other operations with O(N²) complexity, this seems acceptable.

marcharper: It might be a bottleneck, if someone tries to run a tournament with a large match length, so if there's a linear alternative that's preferable. Or is there something else here quadratic in the match length?

It's fine to store something locally, many strategies do that. I don't think you'd have to overwrite `update_history`, just update the features tensor in the strategy function at the start of each round.

moderouin: After looking further into the repo, I understand why I don't need to overwrite update_history.

So even though someone might want to test a large match length, this strategy still checks the last 200 steps. As a result, the runtime will remain the same, even if we run a tournament with a large match length.

I need to insert the last game state at index 1 within the tensor, so I have to shift every element within [1:]. This means that even though I store the previous features, I still have an O(n) operation at each round.

Anyway, the overall strategy has a complexity of roughly O(n^3) in the length of the match, since there are n attention computations, each of which is O(n^2) (attention mechanism).

marcharper: Ok, let's mark the strategy as `long_run_time` (see the other comments) and leave this as is.

moderouin:
@marcharper Do you have an idea why this test fail in the last check? Seems like it's not related to my strategy?

tox.ini
drvinceknight: It was done here to avoid issues with 2.0: https://github.com/Axelrod-Python/Axelrod/pull/1446

axelrod/tests/integration/test_matches.py
drvinceknight: Yeah, even if we go down to 2 that's not a bad idea.

moderouin:
Is there anything else I should change ? @drvinceknight @marcharper

drvinceknight:
> Is there anything else I should change ?

Would you be able to add a few more tests (against known sequences of play or strategies) here axelrod/tests/strategies/test_attention.py please?

I see you have the test again two cooperations but it would be good to have more wide ranging ones (even if due to the nature of this strategy it serves more as a regression test).

moderouin:
Let me know what you think @drvinceknight

drvinceknight (approved):
This looks good to me.

drvinceknight:
@marcharper I think this is just waiting on you to ok.

marcharper:
The entire test suite is substantially slower now. Is there something we can do to mitigate this, such as loading the weights more lazily?

drvinceknight:
> The entire test suite is substantially slower now. Is there something we can do to mitigate this, such as loading the weights more lazily?

drvinceknight:
(Sorry I clicked a button by mistake closing the PR!)

drvinceknight:
> The entire test suite is substantially slower now. Is there something we can do to mitigate this, such as loading the weights more lazily?

I hadn't realised this. How mush slower is it? If @marcharper's good suggestion doesn't work what about marking some tests as "slow" (pytest has tooling for this: https://pytest-with-eric.com/pytest-best-practices/pytest-markers/)?

moderouin:
> The entire test suite is substantially slower now. Is there something we can do to mitigate this, such as loading the weights more lazily?

Looking into this

moderouin:
Lazy loading of the model weights has reduced the total test run time by roughly one hour. Let me know what you think @marcharper

drvinceknight:
Unless I'm not remembering correctly before this PR the test run would be about 20 minutes, and your latest commit is at 60 minutes.

Have I got that right? (not sure who this question is for)

moderouin:
> Unless I'm not remembering correctly before this PR the test run would be about 20 minutes, and your latest commit is at 60 minutes.
> 
> Have I got that right? (not sure who this question is for)

That's correct. `test_meta.py` is taking a really long time to run. I'll try to reduce the runtime while still preserving the quality of the tests

marcharper:
@drvinceknight I'm ok with this going in for now, we should think about how to speed up the test suite though, maybe by ensuring that fewer of the long run time strategies end up in test tournaments or meta strategies.

drvinceknight:
> That's correct. `test_meta.py` is taking a really long time to run. I'll try to reduce the runtime while still preserving the quality of the tests

Thanks I think you've done an awesome job, appreciate you working to get the tests under control :)

> @drvinceknight I'm ok with this going in for now, we should think about how to speed up the test suite though, maybe by ensuring that fewer of the long run time strategies end up in test tournaments or meta strategies.

100%!

PR #1474, opened 2025-06-13
Modify the pyprojetc.toml (possibly incorrectly)

drvinceknight:
## Background

Hatch uses hatchling as the underlying build system.

With older versions of hatchling the metadata is out of date with the expected metadata for pypi (something to do with the license file -- I'm not entirely sure what).

The way to build the dist with hatch/hatchling:

```
hatch build
```

You can check the Metadata version with:

```
tar -xzf dist/axelrod-*.tar.gz
cat axelrod-*/PKG-INFO | grep "Metadata-Version"
```

THIS SHOULD give you:

```
Metadata-Version: 2.4
```

Once that's done, assuming the auth is set up correctly you publish with

```
hatch publish
```

## Proposed actions

While I was fixing all this I was changing the pyproject.toml file.

We should open an issue to possibly put the license file back in and also document the release process (essentially putting some of the information I'm writing here).

For other projects of mine I have setup a publish worklow so that github actions takes care of this. We could look in to doing that as well.


marcharper:
Why does the license file have to be removed?

drvinceknight:
Good question! Just to be clear — we’re not removing or changing the license itself. It’s still declared as MIT in pyproject.toml and the file remains in the repo.

This was mainly a packaging quirk between hatch, hatchling, and pyproject.toml. The way we referenced the license file wasn’t compatible, and after a fair bit of trial and error (which I could only debug in the final step of uploading to PyPI), I landed on this fix.

I suggest we revisit this for the next release and try adding the file path back — we should open an issue to track it (or perhaps turn this PR in to that issue).

drvinceknight:
Are we ok to merge this?

PR #1477, opened 2025-09-14
Added an AdaptiveCooperator strategy. 

cjgwang:
Was unable to run `python rebuild_classifier_table.py` due to network issues, but the rest should be correct. I created adaptive_cooperator.py, test_adaptive_cooperator.py, made changes to the doctest, references, and _strategy.py.

### DIFF
--- a/axelrod/strategies/adaptive_cooperator.py
+++ b/axelrod/strategies/adaptive_cooperator.py
@@ -1,0 +1,63 @@
+from axelrod.action import Action
+from axelrod.player import Player
+
+C, D = Action.C, Action.D
+
+class AdaptiveCooperator(Player):
+    """
+    This is an adaptive strategy where the player starts
+    by cooperating, but switches to Tit-For-Tat if the
+    opponent's score exceed's its own score. Then, it
+    switches to Defector if the opponent's score is twice
+    its own score.
+    """
+
+    name = 'Adaptive Cooperator'
+    classifier = {
+        'memory': 1,
+        'stochastic': False,
+        'inspects_source': False,
+        'manipulates_source': False,
+        'manipulates_state': False
+    }
+
+    def __init__(self):
+        super().__init__()
+        self.is_tft = False
+        self.is_defect = False
+        self.current_score = 0
+        self.opponent_score = 0
+
+    def _score_last_round(self, opponent):
+        game = self.match_attributes["game"]
+        last_round = (self.history[-1], opponent.history[-1])
+        scores = game.score(last_round)
+        self.current_score += scores[0]
+        self.opponent_score += scores[1]
+
+    def strategy(self, opponent):
+        turn = len(self.history) + 1
+        if turn > 1:
+            self._score_last_round(opponent)
+
+        if self.opponent_score > self.current_score * 2:
+            self.is_defect = True
+        elif self.opponent_score > self.current_score:
+            self.is_tft = True
+
+        #response to strategies
+        if self.is_defect:
+            return D
+        elif self.is_tft:
+            if not opponent.history or len(self.history) == 0:
+                return C
+            else:
+                return opponent.history[-1]
+        else: #cooperate
+            return C
+
+    def reset(self):
+        super().__init__()
+        self.is_defect = False
+        self.is_tft = False
+
--- a/axelrod/strategies/adaptive_cooperator.py
+++ b/axelrod/strategies/adaptive_cooperator.py
@@ -14,7 +14,7 @@ class AdaptiveCooperator(Player):
 
     name = 'Adaptive Cooperator'
     classifier = {
-        'memory': 1,
+        "memory_depth": 1,
         'stochastic': False,
         'inspects_source': False,
         'manipulates_source': False,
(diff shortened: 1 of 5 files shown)


drvinceknight:
So the build is failing because:

```
  ____ ERROR collecting axelrod/tests/strategies/test_adaptive_cooperator.py _____
  ImportError while importing test module '/Users/runner/work/Axelrod/Axelrod/axelrod/tests/strategies/test_adaptive_cooperator.py'.
  Hint: make sure your test modules/packages have valid Python names.
  Traceback:
  /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/importlib/__init__.py:126: in import_module
      return _bootstrap._gcd_import(name[level:], package, level)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  axelrod/tests/strategies/test_adaptive_cooperator.py:2: in <module>
      from test_player import TestPlayer
  E   ModuleNotFoundError: No module named 'test_player'
  ```

I see that what you have written [follows the documentation perfectly]() but that looks to be out of date (this is a documentation bug). If you take a look at a similar test file for example [here](https://github.com/Axelrod-Python/Axelrod/blob/dev/axelrod/tests/strategies/test_adaptor.py#L5C1-L5C36) you'll see that in fact the import needs to be:

```
```

If you fix that and push it hopefully that fixes the error.

**Comment by cjgwang** [comment]
Just fixed the TestPlayer class and pushed again - checks still failed. I think this has to do with the classifier, since `test_adaptive_coordinator.py` was unable to initialise.

File: axelrod/tests/strategies/test_adaptive_cooperator.py [hunk 1]
```diff
@@ -0,0 +1,28 @@
+import axelrod
+from .test_player import TestPlayer
+
+C, D = axelrod.Action.C, axelrod.Action.D
+
+class TestAdaptiveCooperator(TestPlayer):
+    name = 'Adaptive Cooperator'
+    player = axelrod.AdaptiveCooperator
+    expected_classifier = {
+        'memory': 1,
```
drvinceknight:
```suggestion
       "memory_depth": 1,
```

axelrod/strategies/adaptive_cooperator.py
```diff
@@ -0,0 +1,63 @@
...
+    This is an adaptive strategy where the player starts
+    by cooperating, but switches to Tit-For-Tat if the
+    opponent's score exceed's its own score. Then, it
+    switches to Defector if the opponent's score is twice
+    its own score.
+    """
+
+    name = 'Adaptive Cooperator'
+    classifier = {
+        'memory': 1,
```
drvinceknight:
```suggestion
        "memory_depth": 1,
```

drvinceknight (commented):
So it looks like you used an incorrect classification key.

cjgwang:
I tried again and it doesn't seem to work. It looks like the test is failing because of test coverage, but when I added some more tests, it still fails!

drvinceknight:
You're right, it's coverage but there are also some tests that are failing.

**Coverage**

From the log:

```
  axelrod/strategies/adaptive.py                                      27      0   100%
  axelrod/strategies/adaptive_cooperator.py                 37      1    97%   53
  axelrod/strategies/adaptor.py                                       30      0   100%
  axelrod/strategies/alternator.py                                    12      0   100%
```

Line 53 is not being hit when running the tests:

```
       elif self.is_tft:
            if not opponent.history or len(self.history) == 0:
                return C
```

That implies that we don't have a test which checks this specific set of conditions.

**Test failures**

There are some remaining tests that are not passing. 

One of them is for example:

```
 ___________ TestAdaptiveCooperator.test_reset_history_and_attributes ___________
  
  self = <axelrod.tests.strategies.test_adaptive_cooperator.TestAdaptiveCooperator testMethod=test_reset_history_and_attributes>
  
      def test_reset_history_and_attributes(self):
          """Make sure resetting works correctly."""
          for opponent in [
              axl.Defector(),
              axl.Random(),
              axl.Alternator(),
              axl.Cooperator(),
          ]:
              player = self.player()
              clone = player.clone()
              match = axl.Match((player, opponent), turns=10, seed=111)
              match.play()
              player.reset()
  >           self.assertEqual(player, clone)
  E           AssertionError: Adaptive Cooperator != Adaptive Cooperator
```

I believe this is because you have not reset all the attributes in the rest method:

```
  def reset(self):
        super().__init__()
        self.is_defect = False
        self.is_tft = False
```

should be:

```
  def reset(self):
        self.is_tft = False
        self.is_defect = False
        self.current_score = 0
        self.opponent_score = 0
```

**although** I think that you could remove the method completely and it will use the parent class `reset` which will call the init method itself.

There are a few other failures that are in the doctests but also some more in the unit tests that I'd need to look at a bit more closely.

You **might** find running the tests locally helpful so that you can iterate this a bit more quickly. There is some information on doing that here: https://axelrod.readthedocs.io/en/stable/how-to/contributing/running_tests.html

For example to just run the test file for your strategy you could run:

```
python -m pytest axelrod/tests/strategies/test_adaptive_cooperator.py
```

PR #1492, opened 2026-06-27
Fix dead prisoners-dilemma.com links with Wayback Machine archives

wali-reheman:
www.prisoners-dilemma.com has new ownership and no longer hosts IPD strategy content. Replaced the two broken HTTP links with Wayback Machine archives — one in the HARD_TFT docs and one in the bibliography entry.


drvinceknight:
The failure isn't due to your contribution but to an upstream change in numpy:

```

  E       - Joss-Ann Tit For Tat: (np.float64(0.0), np.float64(0.0))
  E       + Joss-Ann Tit For Tat: (0.0, 0.0)
  ```

